PR #65, opened 2017-02-14
will not merge, testing out buddybuild

npark-mozilla:
adding sample shell script and espresso test to see what happens, will not merge


pocmo:
@npark-mozilla What's the state of this? Can we close it? For buddybuild to start a build It should be enough to push a branch to Github.

Issue #37, opened 2017-02-07
Update Settings iconography on Main view to match 


antlam:
We used a placeholder from the Android Lib (I think?) in Kona, we should update that to match ours to help with cross-product consistency here.

antlam:
<img width="386" alt="screen shot 2017-02-16 at 6 52 59 pm" src="https://cloud.githubusercontent.com/assets/1994820/23050768/2dd73196-f479-11e6-8b2f-2ab051702a47.png">

@pocmo what do you think about this? Using the overflow would allow us to offer options like these:

- Settings
- About
- Help
- Your Rights

... in the typical UI of an overflow menu.

pocmo:
This is great!

pocmo:
![](https://cloud.githubusercontent.com/assets/89638/23062421/eded2eb8-f505-11e6-8253-cef901a504b5.gif)

This is how the current version looks like.

PR #83, opened 2017-02-17
HomeFragment: Transform settings button into overflow menu. (#37).

pocmo:


pocmo:
I thought I added a styles.xml line to make it show up above the anchor. Anyways, this is only a temporary solution until we can build the actual menu and then we might need PopupWindow and the same code like in Fennec.

Issue #76, opened 2017-02-17
Don't load blocklists on the UI thread


ahunt:
We currently load the blocklists on the UI thread. That's bad.

We only ever access the lists (i.e. do URL matching) on a background thread, i.e. webkit calls shouldInterceptRequest() off the UI thread. There's no need at all to have this data be available for UI thread calls. We could easily load the data the first time shouldInterceptRequest() is called.

I think it would be even better to try to preload the data as early as possible, e.g. during app startup - that means once we try to load a URL we already have all data available and loaded. (I have local patches which do this. We are still able to load during shouldInterceptRequest() as a backup, but we try to load at startup.)

pocmo:
> I think it would be even better to try to preload the data as early as possible, e.g. during app startup - that means once we try to load a URL we already have all data available and loaded.

I agree! There's a high chance that the user just visits a specific URL and then leaves Focus again. So we shouldn't make the actual load of this website slower but instead load the lists while the user might still be typing etc.

PR #88, opened 2017-02-17
Implement blocklist preloading (#76)

ahunt:
Previously, blocklists would be loaded on the UI thread (!), every time
the webview was shown. Instead, we should keep a static reference to the
lists and only load them once per session.

Loading the lists at app startup is helpful, since we should then have all
necessary data in memory by the time the user finishes typing a URL. Even
if we ever open directly to the webview, the UI will no longer be blocked.
(The loading/processing time for blocklists seems to be insignificant, but
 we should still do this for those times when reading from disk might be slow.)

### DIFF
--- a/app/src/gecko/java/org/mozilla/focus/web/WebViewProvider.java
+++ b/app/src/gecko/java/org/mozilla/focus/web/WebViewProvider.java
@@ -23,6 +23,10 @@ import org.mozilla.gecko.GeckoView;
  * WebViewProvider implementation for creating a Gecko based implementation of IWebView.
  */
 public class WebViewProvider {
+    public static void preload(final Context context) {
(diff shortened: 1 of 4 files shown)


app/src/gecko/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -23,6 +23,10 @@
  * WebViewProvider implementation for creating a Gecko based implementation of IWebView.
  */
 public class WebViewProvider {
+    public static void preload(final Context context) {
+        // Nothing: there's no Gecko preloading (yet?).
```
pocmo: We could startup Gecko here. That's what I wanted to try and for that I needed an early call to the provider too. I might look into this in a new issue. :)

Issue #35, opened 2017-02-01
Play store links crash: IllegalArgumentException: Malformed URI supplied


pocmo:
When opening a market:// URL linking to the Play store the app crashes:

    System.err  W  java.lang.IllegalArgumentException: Malformed URI supplied
                W      at org.mozilla.focus.webkit.matcher.EntityList.isWhiteListed(EntityList.java:38)
                W      at org.mozilla.focus.webkit.matcher.UrlMatcher.matches(UrlMatcher.java:85)
                W      at org.mozilla.focus.webkit.TrackingProtectionWebViewClient.shouldInterceptRequest(TrackingProtectionWebViewClient.java:30)
                W      at android.webkit.WebViewClient.shouldInterceptRequest(WebViewClient.java:185)
                W      at com.android.webview.chromium.WebViewContentsClientAdapter.shouldInterceptRequest(WebViewContentsClientAdapter.java:321)
                W      at org.chromium.android_webview.AwContents$BackgroundThreadClientImpl.shouldInterceptRequest(AwContents.java:527)
                W      at org.chromium.android_webview.AwContentsBackgroundThreadClient.shouldInterceptRequestFromNative(AwContentsBackgroundThreadClient.java:38)

STR:
 * Go to mozilla.org
 * Scroll down to the Firefox teaser
 * Click on the "Get it on Google Play" button

ahunt:
I'll focus on not-crashing in this bug, #32 will take care of actually opening an external app.

PR #107, opened 2017-02-22
Don't process main frame requests (#35)

ahunt:
This also upgrades to the non-deprecated version of shouldInterceptRequest.
That means we're now firmly locked to Android 5+.


pocmo:
I think that's fine. In the UI layer I am also using some Android 5+ features. In the security section of the Android docs they mention that WebView has some security flaws in Android 4.4 and lower - so we probably shouldn't go lower as we are loading random resources from the web.

pocmo:
Is the blocking of main frame requests the reason why I had problems loading Google services?

ahunt:
@pocmo Do you have specific sites that didn't work? I would've expected all google sites to break, but current builds seem to work for me.

pocmo:
I only remember that several images wouldn't load on google.com (like the daily doodle). But this seems to work now.

PR #111, opened 2017-02-22
Add support for google override list

ahunt:
We have an override list for google properties, we need to handle that too.

Given the complexity of blocklist processing code, I think it would be worth writing some python
scripts that are run at build-time (similar to what iOS does), then we could simply read two json files (definitive blocklist, and entitylist) at runtime. The blocklist is also more complex than what's needed for blocking (i.e. we discard data when reading it into memory), so the blocklist parsing would be even simpler than ever before.

That probably isn't needed for v1 though, the existing code works well enough for that.

### DIFF
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -2,3 +2,4 @@
 
 src/webkit/res/raw/blocklist.json
 src/webkit/res/raw/entitylist.json
+src/webkit/res/raw/google_mapping.json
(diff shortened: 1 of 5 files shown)


pocmo:
Approved, but there are build errors that seem to be related to this change!

ahunt:
Yeah, using a different blocklist location requires the blocklist script to ensure that the folder actually exists before copying stuff there - I've fixed that now. (The old location already had the topdomains.txt, hence the folder is created by git - the new location has no files in git.)

PR #124, opened 2017-02-28
Ahunt/113 back

ahunt:
Fix the back button behaviour


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -178,4 +179,12 @@ public boolean onMenuItemClick(MenuItem item) {
...
         return false;
     }
+
+    public boolean canGoBack() {
+        return webView.canGoBack();
+    }
+
+    public void goBack() {
+        webView.goBack();
+    }
```
pocmo: Another option would be to morph this into a more generic onBackPressed() that returns true if consumed. This would allow us to use this in other fragments too later.

pocmo: ... or not. For the overflow menu I need canGoForward() :)

Issue #120, opened 2017-02-24
App name should be put inside "Open With" string


petercpg:
String `menu_open_with` does not include the app name to open with, which violates the grammar in Chinese. We usually use 使用 {app} 開啟, literally 'Use {app} to Open'.

The string itself should be changed to something like "Open With %s".

ahunt:
FWIW we don't actually do anything with "Open With" right now (i.e. we have a menu item which actually only shows "open with" and does nothing and has no app now).

We'll be adding the actual functionality in Bug #26, but I'll try to fix the string now - that said, I don't know if this string will actually be used in future.

pocmo:
Yeah, this string should go away soon. I'm working on the final menu (#52) and as part of that I'll talk to antlam to figure out how the browser switching should work (#26).

pocmo:
#26 will introduce two strings:
* "Open with %1$s" - For opening the current URL in another browser or app
* "Open with…" - If there are multiple other options. Selecting this item will show a screen with all available options to choose from.

Mock from #26:
![screen shot 2017-03-02 at 12 01 18 pm](https://cloud.githubusercontent.com/assets/1994820/23524558/05ce44b0-ff40-11e6-8d5f-0dac8333fcfb.png)

Issue #52, opened 2017-02-09
Use overflow button to show menu of navigational options


antlam:
If we make the FAB erase, lets default back to the overflow button in the action bar for the following:

- Back (?)
- Forward
- Refresh
- Share
- Open with
- Settings

pocmo:
After refactoring the FAB button (#43) I created an overflow menu with the options mentioned above. This is just using the default plain text menu style. As soon as we have some mocks ready I'll update the menu accordingly.

antlam:
<img width="842" alt="screen shot 2017-02-16 at 2 27 56 pm" src="https://cloud.githubusercontent.com/assets/1994820/23044232/28086c00-f454-11e6-973e-d604d306baa4.png">

Attaching visual of "overflow button" here. Will post the menu contents next.

antlam:
<img width="390" alt="screen shot 2017-02-16 at 3 57 07 pm" src="https://cloud.githubusercontent.com/assets/1994820/23046794/9b911fbc-f460-11e6-8647-f1c3a42e9219.png">

Pressed list item color #E0E0E0

antlam:
Something to think about @bbinto ... are we OK with Firefox, Chrome, etc being tucked under the "Open with..." menu item with the rest of the other apps that could handle links?

or, we could special case ourselves and pull out "Firefox (Private browsing)" into an additional menu list item.

pocmo:
> Something to think about @bbinto ... are we OK with Firefox, Chrome, etc being tucked under the "Open with..." menu item with the rest of the other apps that could handle links?

For roughly two weeks after Kona I used the prototype as my default browser. And every additional click I had to do in order to get to my real browser (when needing cookies or passwords) was super annoying. This was ultimately the reason why I switched back to Nightly. It would be nice if the "Open with.." at least shows the last/most used browser as an icon (like our Quick share in Fennec!) for direct access.

antlam:
If it's not an immense amount of Engineering, I think adding a "last used" icon on the right of the Share and Open with.. option sounds like a great idea. But as a user, I would have concerns around privacy seeing this type of "smart"/"adaptive" behavior.

antlam:
<img width="379" alt="screen shot 2017-02-17 at 10 52 47 am" src="https://cloud.githubusercontent.com/assets/1994820/23078862/40b65a18-f4ff-11e6-90db-88f4233535a2.png">

@pocmo I forget, why did we need an "Open with..." button again?

[...]

Issue #26, opened 2017-01-26
Allow switching to a full featured browser


pocmo:
Currently we just have a share action but no way to switch to an actual browser.

pocmo:
@antlam We need to define how this should look like.

On new Android versions you can get the list of browser anymore if the user has a default browser set: We only get the default browser. However if Focus actually *is* the default browser then this doesn't help us.

However we can look which browsers from a list of known browsers are installed and then offer to open in one of those browsers. I would love to see something like our quick sharing feature in Fennec: First offer all browsers, then later allow quick access to the most used one. And do we want to promote Firefox here somehow? E.g. offer to install it if it's not yet.

antlam:
We should make sure we're switching to Private Browsing in Firefox. That's a change we made in iOS after some user feedback.

I feel like the UI for this should go in our Overflow menu... like the latest mock in #52 ... thoughts?

(After the -> and 'refresh' buttons) perhaps something like:

- Share...
- Open in Firefox (Private Browsing)
- Open in _Default browser_ 
- Settings

Where the "Default browser" option will only show up if it's not Firefox or Focus? And the "Firefox (Private Browsing)" link will lead to the GPS install page if not installed?

bbinto:
"We should make sure we're switching to Private Browsing in Firefox. That's a change we made in iOS after some user feedback."
--> Can you point me to that feedback, I thought this was something some of the team decided. 

I wanted to post here: https://github.com/mozilla-mobile/focus-android/issues/52 but thought it might fit better here:

How do you change the default browser though if it's set to Firefox or Focus, would there be no "Open with ..." option?

I think we can/should be proud of our cross-promotion and give Firefox a more prominent placeholder in the DOT menu. (that's the whole point of app diversification :))

- We might not need the Back button (if the device back button works, doesn't currently work)
- Forward
- Refresh
- Share

a) Default browser == Firefox
     Open with Firefox 
    "Open with ..." (I think we need this, otherwise the user is stuck with Firefox, unless they reset app Settings on their phone)
b) Default browser == Focus
    Open with ... (to overwrite)
c) Not Firefox/Focus as Default
    Open with Firefox ( if Firefox is not installed, link to Google Play Store)
    Open with _Default browser_

- Settings

For first version, we can/should keep it simple, add telemetry to help us understand how often users actually pop-out.

antlam:
[...]

PR #126, opened 2017-02-28
Add format placeholder to "open with" #120

ahunt:
We don't use this yet, but to avoid wasting localisers' time we should
just add the placeholder now.

(I don't know if we really want to land this - opinions welcome.)


pocmo:
Let's wait until #52 and #26 to avoid adding another unused string. I'm not convinced that the proposed menu covers all use cases yet.

PR #130, opened 2017-03-01
Implement overflow menu. (#52)

pocmo:
(work in progress)


pocmo:
This first version can land - After that it will go through another round of polishing with antlam.

pocmo:
> (I wonder whether people are going to ask for a back button :p )

Yeah, me too. I guess antlam is using this as a test. :)

PR #139, opened 2017-03-01
Ahunt/32 external

ahunt:
Allow opening external links

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/IntentUtils.java
+++ b/app/src/main/java/org/mozilla/focus/utils/IntentUtils.java
@@ -1,0 +1,154 @@
+/* -*- Mode: Java; c-basic-offset: 4; tab-width: 20; indent-tabs-mode: nil; -*-
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package org.mozilla.focus.utils;
+
+import android.app.Activity;
+import android.content.DialogInterface;
+import android.content.Intent;
+import android.content.pm.PackageManager;
+import android.content.pm.ResolveInfo;
+import android.net.Uri;
+import android.support.annotation.StringRes;
+import android.support.v7.app.AlertDialog;
+
+import org.mozilla.focus.R;
+
+import java.net.URISyntaxException;
+import java.util.List;
+
+public class IntentUtils {
+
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/focus/utils/IntentUtils.java
```diff
@@ -0,0 +1,154 @@
...
+
+        builder.setNegativeButton(R.string.external_app_prompt_cancel, new DialogInterface.OnClickListener() {
+            @Override
+            public void onClick(final DialogInterface dialog, final int which) {
+                dialog.dismiss();
+            }
+        });
+
+        // TODO: remove this, or enable it - depending on how we decide to handle the multiple-app/>1
+        // case in future.
```
pocmo: Should we file an issue for that?

app/src/main/res/values/strings.xml
```diff
@@ -93,4 +93,14 @@
     <string name="firstrun_description">Automatically block online trackers while you browse. Then tap to erase visited pages, searches, cookies and passwords from your device.</string>
     <string name="firstrun_button">Ok, got it!</string>
     <string name="firstrun_footer">To be clear, sites and service you use can still know where you\'ve been.</string>
+
+    <!-- This string is shown when the user has clicked a link that needs to be opened in a different app.
+         Argument 2 is the external app name, argument 1 is our name (i.e. Firefox Focus, or Firefox Klar). -->
+    <string name="external_app_prompt">This link will open in %2$s. Are you sure you want to exit %1$s?</string>
```
pocmo: It's a bit confusing that in the original string argument 2 comes before argument 1. But with the comment it shouldn't be a problem.

app/src/main/res/values/strings.xml
```diff
@@ -93,4 +93,14 @@
...
+    <string name="external_app_prompt_no_app_title">No app to handle link</string>
+    <string name="external_app_prompt_no_app">None of the apps on your device are able to open this link. Do you want to exit %1$s and open %2$s to find a suitable app?</string>
+    <string name="external_app_prompt_ok">Ok</string>
+    <string name="external_app_prompt_cancel">Cancel</string>
```
pocmo: I also added a 'Cancel' string for a dialog in one of my patches that I'm currently landing. We should make sure to unify them and just have one string before exporting them again.

app/src/main/java/org/mozilla/focus/utils/IntentUtils.java
ahunt: Filed:
https://github.com/mozilla-mobile/focus-android/issues/223

Issue #23, opened 2017-01-26
Add 'Your Rights' screen


pocmo:

pocmo:
@antlam: I can't find a "Your Rights" screen in Focus for iOS anymore (Only About/Help). Has this been removed?

pocmo:
There's still code to go from the "About" page (See #21) to a "Your Rights" screen. But I do not see it in the app anymore.

pocmo:
Okay, found it. The about screen is scrollable... :)

![your_rights](https://cloud.githubusercontent.com/assets/89638/23565499/1c4eea1e-004e-11e7-9aca-f10ae2227043.png)

antlam:
Let's make sure we keep the text styles, and padding the same across all 3 pages here. The "About" page is the styling we should follow.

On iOS we have an issue to make that all consistent too.

PR #157, opened 2017-03-05
Add 'Your Rights' screen

ltedone:


pocmo:
Hi Luigi. Thank you for this patch. I'll review it today.

* Can you add the issue number to the commit message? Like
> Add 'Your Rights' screen (#23)

* The resources you added - are those copied from Focus for iOS?

app/src/main/AndroidManifest.xml
```diff
@@ -48,7 +48,8 @@
 
         <activity android:name=".activity.HelpActivity" />
 
-        <activity android:name=".activity.RightsActivity" />
+        <activity android:name=".activity.RightsActivity"
+            android:label="Your Rights"/>
```
pocmo: Please add a string for the label to strings.xml - Even though the actual website might not be translated.

app/src/main/assets/style.css
```diff
@@ -0,0 +1,9 @@
+body, html {
+    background: #221F1F;
+    color: #E1E5EA;
+    font: -apple-system-body;
```
pocmo: I guess this doesn't have any affect on Android. As far as I understand it this will use the font of the system to render the web content. Is there something like that on Android? Otherwise we could try to use Roboto or just not specify any font at all.

app/src/main/java/org/mozilla/focus/activity/RightsActivity.java
```diff
@@ -1,6 +1,34 @@
...
+import org.mozilla.focus.R;
+
+public class RightsActivity extends AppCompatActivity {
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        setContentView(R.layout.activity_rights);
+        WebView mWebView = null;
+        mWebView = (WebView) findViewById(R.id.webview);
```
pocmo: It would be nice if we could use WebViewProvider here - so that we use the same rendering engine depending on build type. But that's something we could do in a follow-up issue.

app/src/main/res/layout/activity_rights.xml
```diff
@@ -0,0 +1,19 @@
...
+    android:layout_height="match_parent">
+
+    <android.support.v7.widget.Toolbar
+        android:id="@+id/toolbar"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:background="#ff363b40"
+        android:elevation="4dp"/>
+
+    <WebView  xmlns:android="http://schemas.android.com/apk/res/android"
```
pocmo: nit: namespace not needed here

app/src/main/res/layout/activity_rights.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:orientation="vertical" android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <android.support.v7.widget.Toolbar
+        android:id="@+id/toolbar"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:background="#ff363b40"
```
pocmo: Please reference this color from colors.xml

pocmo:
Can you attach a screenshot here so that @antlam (UX) can review it too? :)

pocmo:
From now on buddybuild should build pull requests from your fork too.

app/src/main/java/org/mozilla/focus/activity/RightsActivity.java
ltedone: I would prefer to manage it in a follow-up issue

ltedone:
Hi Sebastian, thanks for your suggestions. Yes, I used the resources from Focus on iOS.
This is the screenshot:
screenshot_20170306-224952

antlam:
Looking good! 

For styles, I think we could tidy things up a bit. 

We're using Roboto Medium, #FFFFFF, 18px for headers so we should use that here for the "Your Rights" header. For the body, we're using Roboto Regular, #FFFFFF, 14px 

The vertical spacing between should be 24px. Left and right padding can be 24px too.

@ltedone Can I see a screenshot of these changes? :)

ltedone:
@antlam this is the screenshot as you asked:

schermata 2017-03-07 alle 22 44 12

ltedone:
@antlam this is the screenshot you requested:
screenshot_20170307-231527

Is it fine for you?

antlam:
Looking good @ltedone ! Just wanted to double check the font family here. 

Is it Roboto Medium and Roboto Regular respectively?

ltedone:
@antlam yes, I added them to the repository and used them into the stylesheet

app/src/main/java/org/mozilla/focus/activity/RightsActivity.java
ahunt: I've been thinking about this a little - it would be good to share the same activity code for all of about/help/your-rights, and also some SUMO pages ("Set default browser"/"Learn more" for telemetry). Using the WebViewProvider has a potential bug/feature where image blocking being enabled makes SUMO pages a bit uglier - I filed #182 to make it easier to discuss that.

antlam:
screen shot 2017-03-08 at 8 36 57 am

Hey @ltedone I've mocked it up quickly so please ignore the copy for the time being.

Looking at your screenshot, it looks like the padding is larger than 24 and the font seems to be bold instead of Medium? I could be wrong though. 

Let me know!

ltedone:
Hi @antlam, thanks for your suggestions. I fixed the issues and this is the new screenshot:

screenshot_20170308-231914

antlam:
Looks better! thanks!

pocmo:
The code is looking good. I'm only wondering whether we can reference the fonts from the system instead of putting them in the APK. We are supporting Android 5+ devices - so in theory the font should be available everywhere.

ahunt:
I have one question re: localisation - IIRC assets/ files can't be translated - do we need that for the your rights screen (I'm pretty sure we will need that for about/help). We could provide locale-specific html by using the raw/ and raw-XX/ directories, but I don't know if our translations process supports that - do we need to look at including (html formatted?) text in strings.xml for this?

(That's something I've been looking into as part of the error pages bug, where we'd also need html formatted text.)

pocmo:
I merged this version. I'll file a new issue for the fonts.

Issue #36, opened 2017-02-07
Create First Run screen for new users


antlam:
Matching what we have on iOS would make the most sense here. Not sure if we 100% need the same animation with the wave graphic but I just wanted to file this first in case I forgot.

antlam:
Idea from @bbinto 

Allow opt-out in the onboarding screen for Adjust?

bbinto:
Let's not consider this for now (opt-out in onboarding) and replicate what we have for iOS.

ahunt:
I've got the code in place to show the first run, which leaves just the design (I've used the focus wordmark as a placeholder).

One thing I'm wondering: if the user installs focus, and opens a link (without having run focus before that point), do we want to:
A. open the link immediately (and show the first-run experience when focus is opened normally and not via a link)
B. show the first-run experience (and open the link once "OK, GOT IT" is pressed).

Either is possible, I've implemented (B) but it's easy to change.

antlam:
We've gone back and forth about this before on Fennec but I think for Focus, we should just show the First Run experience since it's relatively light-weight.

I vote B.

bbinto:
+1 for B

pocmo:
In Fennec we do (A) btw. :)

antlam:
Haha I know, and that has caused a bunch of edge cases that we have to deal with individually. This approach also seems less fitting for Focus given how light-weight our First Run experience is right now.

ahunt:
@antlam: do you have any mocks for the appearance?

(Here's a screenshot of what I've been using so far for testing, I just used a random background, etc.)
![firstrun2](https://cloud.githubusercontent.com/assets/1546915/23485430/3ec8d6aa-fe90-11e6-8206-9762d6ace863.png)

antlam:
Thanks @ahunt ! here's the mock. It uses the same gradient colors and setting as the toolbar when "inactive.

![screen shot 2017-03-02 at 12 21 37 pm](https://cloud.githubusercontent.com/assets/1994820/23525311/deecb810-ff42-11e6-825b-a4aeba6140d8.png)

antlam:
![screen shot 2017-03-02 at 12 29 44 pm](https://cloud.githubusercontent.com/assets/1994820/23525625/f57a0c62-ff43-11e6-9e51-79645e59c46c.png)

Specs!

ahunt:
And an updated screenshot:
[...]

PR #158, opened 2017-03-06
Add first run screen (#36)

ahunt:


app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    Button okButton;
+
+    @Override
+    public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {
+        final View view = inflater.inflate(R.layout.fragment_firstrun, container, false);
+
+        okButton = (Button) view.findViewById(R.id.firstrun_exitbutton);
+        okButton.setOnClickListener(this);
```
pocmo: Seems like we do not use okButton anywhere else. Could be just:
> view.findViewById(R.id.firstrun_exitbutton).setOnClickListener(this);

app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private static final String FIRSTRUN_PREF = "firstrun_shown";
+
+    public static FirstrunFragment create() {
+        return new FirstrunFragment();
+    }
+
+    public static boolean shouldShowFirstrun(final Context context) {
+        return !PreferenceManager.getDefaultSharedPreferences(context).getBoolean(FIRSTRUN_PREF, false);
+    }
```
pocmo: I recently introduced the `Settings` class for such helpers. Maybe you want to move it there.

app/src/main/res/values/strings.xml
```diff
@@ -78,4 +78,9 @@
 
     <!-- Content description (not visible, for screen readers etc.): Reload current website -->
     <string name="content_description_reload">Reload website</string>
+
+    <string name="firstrun_title">Browse like no one’s watching.</string>
+    <string name="firstrun_description">Automatically block online trackers while you browse. Then tap to erase visited pages, searches, cookies and passwords from your device.</string>
+    <string name="firstrun_button">OK, GOT IT!</string>
```
pocmo: You do not need to write this in all caps. You can use `android:textAllCaps` on the `Button`. However I think that's already the default on at least the latest Android versions. The advantage of letting Android do it is that it will adapt to whatever makes sense in a language.

ahunt: Yup, that's the default - it ALLCAPS buttons (at least in English)!

PR #160, opened 2017-03-06
Add functionality to switch to Firefox or default browser. (#26)

pocmo:
This is part one, adding functionality for "Switch to Firefox" and "Switch to [default browser/app]". I'll work on "Open With.." which will show a screen with all options to select from in a follow-up patch (Unfortunately we can't re-use the native Intent chooser for that).


app/src/main/java/org/mozilla/focus/utils/Browsers.java
```diff
@@ -0,0 +1,166 @@
...
+        GHOSTERY_BROWSER("com.ghostery.android.ghostery");
+
+        public final String packageName;
+
+        KnownBrowser(String packageName) {
+            this.packageName = packageName;
+        }
+    }
+
+    private Map<String, ActivityInfo> browsers;
```
ahunt: nit: these two could probably be final?

PR #161, opened 2017-03-06
Add first batch of error pages

ahunt:


app/src/main/res/values/strings.xml
```diff
@@ -93,4 +94,57 @@
...
+    <string name="errorpage_refresh">Try Again</string>
+
+    <string name="error_connectionfailure_title">Unable to connect</string>
+    <string name="error_connectionfailure_message">
+    <![CDATA[
+      <ul>
+        <li>The site could be temporarily unavailable or too busy. Try again in a few moments.</li>
+        <li>If you are unable to load any pages, check your mobile device’s data or Wi-Fi connection.</li>
+      </ul>
+    ]]>
```
pocmo: We need to test whether this works with our translation export/import. If not - maybe we could split that into separate strings without HTML?

app/src/main/res/values/strings.xml
```diff
@@ -93,4 +94,57 @@
...
+        <li>Check the address for typing errors such as
+          <strong>ww</strong>.example.com instead of
+          <strong>www</strong>.example.com</li>
+        <li>If you are unable to load any pages, check your device’s data or Wi-Fi connection.</li>
+      </ul>
+    ]]></string>
+
+
+    <string name="error_malformedURI_title">The address isn’t valid</string>
+    <string name="error_malformedURI_message">.... web addresses....</string>
```
pocmo: This string looks weird. Is this the correct one? If yes then this definitely needs a comment for translators. :)

app/src/webkit/java/org/mozilla/focus/webkit/ErrorPage.java
```diff
@@ -0,0 +1,177 @@
...
+        substitutionMap.put("%messageLong%", resources.getString(errorResourceIDs.second, desiredURL));
+
+        substitutionMap.put("%css%", cssString);
+
+        final String errorPage = loadResourceFile(webView.getContext(), R.raw.errorpage, substitutionMap);
+
+        // We could load the raw html file directly into the webview using a file:///android_res/
+        // URI - however we'd then need to do some JS hacking to do our String substitutions. Moreover
+        // we'd have to deal with the mixed-content issues detailed above in that case.
+        webView.loadDataWithBaseURL(desiredURL, errorPage, "text/html", "UTF8", desiredURL);
```
pocmo: Shouldn't this be "UTF-8"?

app/src/main/res/raw/errorpage_style.css
```diff
@@ -0,0 +1,229 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+html,
+body {
+  margin: 0;
+  padding: 0;
+  height: 100%;
+  --moz-vertical-spacing: 10px;
```
pocmo: There are a bunch of --moz-* modifiers in here that probably do not do anything in our case?

pocmo:
Argh. Looks like all the HTML gets escaped when exporting/importing strings:

```
msgctxt "error_connectionfailure_message"
msgid ""
"&lt;ul&gt; &lt;li&gt;The site could be temporarily unavailable or too "
"busy. Try again in a few moments.&lt;/li&gt; &lt;li&gt;If you are unable "
"to load any pages, check your mobile device’s data or Wi-Fi "
"connection.&lt;/li&gt; &lt;/ul&gt;"
msgstr ""
```

```
<string name="error_sslhandshake_message">&lt;ul&gt; &lt;li&gt;This could be a problem with the server’s configuration, or it could be someone trying to impersonate the server.&lt;/li&gt; &lt;li&gt;If you have connected to this server successfully in the past, the error may be temporary, and you can try again later.&lt;/li&gt; &lt;/ul&gt;</string>
```

Do you think it would be feasible to split strings and HTML?

app/src/main/res/raw/errorpage_style.css
ahunt: (IIUC) These are actually css variables - AFAICS prefixes have a single dash (-moz-something), and two dashes (--something) are variables. The use of --moz is confusing, but it seems easiest to not modify what the original page is using?

ahunt:
It turns out Android can actually deal with escaped html (i.e. the escaped text gets de-escaped when you getString()) - but I'm guessing it's not easy to translate if it's all escaped? I'm still thinking about how we could split html and text, the main issue is that errors vary, and some include <strong > text within each list item which makes it harder to split the text from html.

pocmo:
I assume the lists are easy to split and then re-construct before replacing?

For the strong formatting I'd say: Leave it out and file a follow-up. This shouldn't block the release. Or maybe try using palceholders and then add the <strong> tags from code. But this needs a very good comment for translators so that they do not mix up the order.

ahunt:
We have a total of 4 complicated strings:
- 2 of them have two <li> items, we could just have string_1 and string_2 for each of those.
- 2 of them have two <li> items, with some bolded text - we could split those, and give up on bolding.

Maybe we should just come up with simpler error strings, so that every error only has a one-line error text?

ahunt:
I've actually seen some html markup in other projects in pontoon ( e.g. https://pontoon.mozilla.org/de/amo/LC_MESSAGES/djangojs.po/ ), so I wonder what happens if we import our encoded html into pontoon? (If encoded doesn't work, maybe we can just modify android2po to give expected output?)

app/src/main/res/values/strings.xml
```diff
@@ -105,11 +105,70 @@
...
+          <strong>www</strong>.example.com</li>
+        <li>If you are unable to load any pages, check your device’s data or Wi-Fi connection.</li>
+      </ul>
+    ]]></string>
+
+
+    <string name="error_malformedURI_title">The address isn’t valid</string>
+    <string name="error_malformedURI_message"><![CDATA[
+<ul>
+  <li>Web addresses are usually written like  <strong>http://www.example.com/</strong></li>
```
flodolo: Unneeded double space before <strong>

app/src/main/res/values/strings.xml
```diff
@@ -105,11 +105,70 @@
...
+</ul>
+    ]]></string>
+
+    <string name="error_redirectLoop_title">The page isn’t redirecting properly</string>
+    <string name="error_redirectLoop_message"><![CDATA[<ul><li>This problem can sometimes be caused by disabling or refusing to accept cookies.</li></ul>]]></string>
+
+    <string name="error_unsupportedprotocol_title">The address wasn’t understood</string>
+    <string name="error_unsupportedprotocol_message"><![CDATA[<ul><li>You might need to install other software to open this address.</li></ul>]]></string>
+
+    <string name="error_sslhandshake_title">Secure Connection Failed</string>
```
flodolo: Other titles don't seem to use title case.

app/src/main/res/values/strings.xml
```diff
@@ -105,11 +105,70 @@
...
+    <string name="error_malformedURI_title">The address isn’t valid</string>
+    <string name="error_malformedURI_message"><![CDATA[
+<ul>
+  <li>Web addresses are usually written like  <strong>http://www.example.com/</strong></li>
+  <li>Make sure that you’re using forward slashes (i.e. <strong>/</strong>).</li>
+</ul>
+    ]]></string>
+
+    <string name="error_redirectLoop_title">The page isn’t redirecting properly</string>
+    <string name="error_redirectLoop_message"><![CDATA[<ul><li>This problem can sometimes be caused by disabling or refusing to accept cookies.</li></ul>]]></string>
```
flodolo: Maybe worth showing these on multiple lines, structuring the code like the others? In case, same for `error_unsupportedprotocol_message`

app/src/main/res/values/strings.xml
```diff
@@ -105,11 +105,70 @@
...
+    <string name="error_sslhandshake_message"><![CDATA[
+      <ul>
+        <li>This could be a problem with the server’s configuration, or it could be
+someone trying to impersonate the server.</li>
+        <li>If you have connected to this server successfully in the past, the error may
+be temporary, and you can try again later.</li>
+      </ul>
+     ]]></string>
+
+    <string name="error_generic_title">Oops.</string>
```
flodolo: A title with a period?

Issue #133, opened 2017-03-01
Toolbar jumps below status bar


pocmo:
Could be related to #132. When scrolling (and maybe zooming) it can happen that the toolbar jumps down and you can see the website behind the status bar.

![statusbar_website](https://cloud.githubusercontent.com/assets/89638/23469232/5b54c668-fea3-11e6-9c81-3ae61f1971a4.png)

ahunt:
I'm not seeing this anymore (in local builds) with my fix for #132, so I'll close this as a dup for now.

Issue #132, opened 2017-03-01
Website header is cut off


pocmo:
I'm confused why I didn't see this before but the top part of a website is hidden behind the toolbar. It should be revealed by scrolling but somehow this doesn't work.

Maybe related to #133.

Poopooracoocoo:
@pocmo This happens to me when headers are at the bottom. I think the toolbar of Focus pushes everything down.

PR #164, opened 2017-03-06
Use app:layout_behaviour to ensure webview isn't partially hidden (#132)

ahunt:


pocmo:
Nice! This was much easier than I thought. I remember removing the behavior at some point because I had problems with the FAB not disappearing. But I rewrote that part and now the behavior doesn't seem to be a problem anymore (just tested your patch). Yay! And yay that it fixes #133 too. 🥇

Issue #150, opened 2017-03-02
Erase button still clickable when hidden


pocmo:
When scrolling a website and hiding the 'erase' button then you can still click the invisible button and we end the browser session.

ahunt:
(FWIW #132, which some CoordinatorLayout related issues, doesn't fix this)

pocmo:
Some animation listeners to set it to VISIBLE/HIDE might be enough here.

PR #166, opened 2017-03-06
Ensure FAB is GONE when we hide it to avoid phantom clicks (#150)

ahunt:
We're still able to click on the FAB even when it's been animated to
zero-size. We need to make it truly gone to avoid having it intercept
clicks.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/widget/FloatingActionButtonBehavior.java
+++ b/app/src/main/java/org/mozilla/focus/widget/FloatingActionButtonBehavior.java
@@ -74,7 +74,7 @@ public class FloatingActionButtonBehavior extends CoordinatorLayout.Behavior<Flo
         animate(button, true);
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/widget/FloatingActionButtonBehavior.java
```diff
@@ -74,9 +74,15 @@ private void hideButton() {
...
     }
 
-    private void animate(View child, final boolean hide) {
+    private void animate(final View child, final boolean hide) {
         animating = true;
 
+        if (!hide) {
+            // Ensure the child will be visible before starting animation: if it's hidden, we've also
+            // set it to View.GONE, so we need to restore that now, _before_ the animation starts.
+            child.setVisibility(View.VISIBLE);
```
pocmo: Instead of doing it here you can override onAnimationStart() in the AnimatorListenerAdapter below.

Issue #173, opened 2017-03-07
Image-blocking pref requires BrowserFragment restart to take effect


ahunt:
STR:
1. Open website e.g. bbc.com (images are present)
2. Open menu->settings, enable "Block Images"
3. Press back to return to browser
4. Open menu->refresh
Expected: images disappear
Actual: images still visible (images continue being visible while you browse in this session)

If you exit the browser (and return to the homescreen), and reopen the same website, in place of step (3), then image blocking works.

We probably need to add a registerOnSharedPreferenceChangeListener() call to WebViewProvider, similar to what we do in UrlMatcher.

pocmo:
Entering a new URL (and therefore replacing the browser fragment) should work too. But yeah, we might want to add a preference listener for the already existing WebView. :)

ahunt:
It looks like cached images are still rendered. I'm guessing that's something we don't need to really worry about, but it does mean that refreshing an existing page wouldn't change anything if you start blocking images (but navigating to other pages without cached images will result in those images not being shown). I've already written the code, so I guess it's better to land that so that anyone who's testing the image blocking setting doesn't get confused by the presence of "unexpected" images.

pocmo:
@antlam Do you think it makes sense to explicitly mention that this blocks images loaded over the network? Maybe as a description below "Block Images"?

antlam:
Not sure I'm following along but that's how I would expect it to work. Why doesn't it work like that?
Shouldn't the images not load anyways after the user "Erases"?

@bbinto what're your thoughts on including this in V1? Should we still try and do that?

pocmo:
We decided to skip image blocking in v1: We only added it because it seemed simple and now we discovered edge cases. I'll remove/hide the setting and file a new bug for re-adding it (and fixing/deciding on this here) - marked as "nice to have".

antlam:
Note to self: 
- Subtitle to describe that there is a need to "restart"/"reload" the webpage the user is currently on in order for this to take effect
- Followed by a simple dialog to confirm the action (only applicable if the pref is not flipped on via the Home screen)

PR #174, opened 2017-03-07
Update webview image blocking when pref changes (#173)

ahunt:


app/src/webkit/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -67,6 +70,26 @@ private static void configureSettings(Context context, WebSettings settings) {
         settings.setAllowFileAccess(false);
 
         settings.setBlockNetworkImage(appSettings.shouldBlockImages());
+
+        appSettings.subscribeToBlockImagesPreference(new Settings.BooleanPrefListener() {
+            final WeakReference<WebView> webViewWeakReference = new WeakReference<>(webView);
```
pocmo: Even though you are using a WeakReference here there's still a implicit reference to the final WebView and Context objects from the parent class.

But even with the WeakReference working this might create a bunch of listeners that don't do anything and do not get cleaned up. Maybe this can move inside WebkitView: When the view is attached it registers a listener, updates its setting on changes and unregisters if it is being detached.

app/src/webkit/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -67,6 +70,26 @@ private static void configureSettings(Context context, WebSettings settings) {
...
+                final WebView webView = webViewWeakReference.get();
+                if (webView != null) {
+                    webView.getSettings().setBlockNetworkImage(blockImages);
+
+                    if (blockImages) {
+                        // Images that are already cached will still be rendered even after enabling
+                        // this setting. To avoid confusing anyone who is actively testing this setting
+                        // it's probably best to clear the cache to make sure absolutely no images
+                        // are shown. (The cache will be cleared when erase()'ing anyway.)
+                        webView.clearCache(true);
```
pocmo: This might prevent confusion but if someone is actually disabling this to save bandwidth then it wouldn't be appropriate to reload everything that's not an image, I guess? Let's ask Product/UX.

ahunt: Existing pages aren't reloaded until you explicitly reload, so this isn't too terrible - but yeah, I'm also not convinced either way.

ahunt:
Closing since we're disabling this pref for now, might come back in #188

Issue #55, opened 2017-02-10
Set user agent


pocmo:
**Firefox for Android**

    Mozilla/5.0 (Android 7.1.1; Mobile; rv:54.0) Gecko/54.0 Firefox/54.0

**Firefox for iOS**

    Mozilla/5.0 (iPhone; CPU iPhone OS 8_3 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) FxiOS/1.0 Mobile/12F69 Safari/600.1.4

**Focus for iOS**

    Mozilla/5.0 (iPhone; CPU Iphone OS 10_2 like Mac OS X) AppleWebKit/602.3.12 (KHTML, like Gecko) Focus/3.1 Mobile/14C89

**Default Android WebView**

    Mozilla/5.0 (Linux; Android 7.1.1; Pixel XL Build/NOF26V; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/56.0.2924.87 Mobile Safari/537.36

ahunt:
Firefox for Android replaces Mobile with Tablet for tablets - I wonder if we'd want to do something similar (it uses Configuration.SCREENLAYOUT_SIZE_LARGE and larger to set tablet mode).

pocmo:
Related: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent/Firefox

@st3fan @thebnich Is there anything we should take care of (especially in regards to tracking)? It looks like the iOS user agent has the same structure like Firefox for iOS?

ahunt:
(AFAICT only firefox on Android uses the Tablet/Mobile strings within the platform part of the UA. Everyone else uses Mobile only for mobile devices, later in the UA. I don't know which route we'd want to go here...)

st3fan:
@ahunt Can we look at some other Android WebKit based browsers to compare? How about UC Browser, Chrome, etc. Is there a common pattern there?

miketaylr:
Here's [Chrome's format](https://developer.chrome.com/multidevice/user-agent):

`Mozilla/5.0 (Linux; <Android Version>; <Build Tag etc.>) AppleWebKit/<WebKit Rev> (KHTML, like Gecko) Chrome/<Chrome Rev> Mobile Safari/<WebKit Rev>`

[...]

PR #175, opened 2017-03-07
Add custom UA (#55)

ahunt:
This produces strings like:
Mozilla/5.0 (Linux; Android 6.0.1) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Focus/1.0  Chrome/56.0.2924.87 Mobile Safari/537.36
General format:
Mozilla/5.0 (Linux; Android <Android Version>) AppleWebKit/<WebKit Rev> (KHTML, like Gecko) Version/4.0 Focus/<Focus Rev>  Chrome/<Chrome Rev> Mobile Safari/<WebKit Rev>

### DIFF
--- a/app/src/webkit/java/org/mozilla/focus/web/WebViewProvider.java
+++ b/app/src/webkit/java/org/mozilla/focus/web/WebViewProvider.java
@@ -7,9 +7,11 @@ package org.mozilla.focus.web;
 
 import android.annotation.SuppressLint;
 import android.content.Context;
-import android.content.SharedPreferences;
+import android.content.pm.PackageManager;
 import android.graphics.Bitmap;
-import android.preference.PreferenceManager;
+import android.os.Build;
+import android.support.annotation.VisibleForTesting;
+import android.text.TextUtils;
 import android.util.AttributeSet;
 import android.view.View;
 import android.webkit.CookieManager;
@@ -20,6 +22,8 @@ import android.webkit.WebView;
 import android.webkit.WebViewDatabase;
 
 import org.mozilla.focus.BuildConfig;
+import org.mozilla.focus.R;
+
 import org.mozilla.focus.utils.Settings;
 import org.mozilla.focus.webkit.NestedWebView;
 import org.mozilla.focus.webkit.TrackingProtectionWebViewClient;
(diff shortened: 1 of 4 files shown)


app/src/webkit/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -67,6 +68,48 @@ private static void configureSettings(Context context, WebSettings settings) {
...
+        try {
+            appVersion = context.getPackageManager().getPackageInfo(context.getPackageName(), 0).versionName;
+        } catch (PackageManager.NameNotFoundException e) {
+            // This should be impossible - we should always be able to get information about ourselves:
+            throw new IllegalStateException("Unable find package details for Focus", e);
+        }
+
+        for (int i = 0; i < tokens.length; i++) {
+            if (tokens[i].startsWith("Chrome")) {
+                tokens[i] = "Focus/" + appVersion + "  " + tokens[i];
```
pocmo: I looked at the iOS Klar build and they use "Klar" instead of Focus. We could reuse R.string.launcher_name but let's just create another string with translatable=false set.

app/src/webkit/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -67,6 +68,48 @@ private static void configureSettings(Context context, WebSettings settings) {
...
+        // so we skip that too.
+        uaBuilder.append(" (Linux; Android ").append(Build.VERSION.RELEASE).append(") ");
+
+        // Use the default WebView agent string here for everything after the platform, but insert
+        // Focus in front of Chrome.
+        // E.g. a default webview UA string might be:
+        // Mozilla/5.0 (Linux; Android 7.1.1; Pixel XL Build/NOF26V; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/56.0.2924.87 Mobile Safari/537.36
+        // And we reuse everything from AppleWebKit onwards, except for adding Focus.
+        final String existing = settings.getUserAgentString();
+        int start = existing.indexOf("AppleWebKit");
```
pocmo: Is it guaranteed that this string will always be present? Maybe we should parse for tokens like "Something/Version (Optional stuff)".

app/src/webkit/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -67,6 +68,48 @@ private static void configureSettings(Context context, WebSettings settings) {
...
+        final String appVersion;
+        try {
+            appVersion = context.getPackageManager().getPackageInfo(context.getPackageName(), 0).versionName;
+        } catch (PackageManager.NameNotFoundException e) {
+            // This should be impossible - we should always be able to get information about ourselves:
+            throw new IllegalStateException("Unable find package details for Focus", e);
+        }
+
+        for (int i = 0; i < tokens.length; i++) {
+            if (tokens[i].startsWith("Chrome")) {
```
pocmo: Is "Chrome" also present in Android versions where WebView is not using Chrome directly? What happens on devices without Google services (even though we currently do not plan to ship to them ourselves)?

ahunt: AFAICT Chrome supplies webview only only on Android 7 and newer. Both the Android 6 device, and Android 5 emulator I'm testing with, include Chrome in the UA string, so I'm guessing this is safe (the Android 5 emulator doesn't have Chrome installed at all) - but I've got no idea about other devices.

app/src/webkit/java/org/mozilla/focus/web/WebViewProvider.java
ahunt: Yeah - I was considering just searching for the first ")" which should usually give the same results.

app/src/webkit/java/org/mozilla/focus/web/WebViewProvider.java
ahunt: Maybe I should add a fallback: if we don't detect Chrome, we can add Focus/1.0 to the end of the UA?

app/src/klar/res/values/strings.xml
```diff
@@ -5,4 +5,6 @@
 <resources>
     <string name="app_name" translatable="false">Firefox Klar</string>
     <string name="launcher_name" translatable="false">Klar</string>
+    <string name="ua_name" translatable="false">Klar</string>
```
pocmo: Shouldn't this be "useragent_appname" to override the other string?

Issue #149, opened 2017-03-02
Link to default browser SUMO page


pocmo:
In #72 we added a setting that links to Android's "default apps" page for setting Focus as default browser (Android N+). On older versions we need to link to a SUMO page. Right now we just hide the setting in this case.

ahunt:
I wonder if we should get a new focus-specific and/or generic version of this page - it mentions firefox multiple times:
https://support.mozilla.org/kb/make-firefox-default-browser-android?utm_source=inproduct&amp;utm_medium=settings&amp;utm_campaign=mobileandroid

pocmo:
@bbinto @antlam What do you think? Who could help us create a page if needed?

antlam:
I think we just talked about this and @bbinto mentioned she would reach out to Joni or Roland? Is that correct?

bbinto:
+ @jonisavage, could we add Focus to this page as well for launch?

pocmo:
@ahunt Here are the links: https://docs.google.com/a/mozilla.com/spreadsheets/d/1utt7lgWClxyl6LdnkLFEtFZH7wf-nBg5MTbFfuZLgf4/edit?disco=AAAAA-Tevfw&usp=comment_email_document

Issue #182, opened 2017-03-08
Consider using WebViewProvider for About/Help/Your Rights (/ SUMO pages)


ahunt:
Sebastian suggested using the WebViewProvider (in place of the system WebView) for the Your Rights screen.

We actually have multiple such screens (about/help/your rights will be hosted within the app, we also show SUMO pages for "set Default Browser", and also a "learn more" link for telemetry). I think it would be useful to have one Activity implementation, which is able to show all the different pages (i.e. we just need to change the URL and title for every page/screen).

Using WebViewProvider means those screens also inherit Focus's content blocking settings: that's not an issue for the bundled pages (I don't think any of those actually include images). However the SUMO pages do have images, so if Image Blocking is enabled, the SUMO pages looks a bit uglier.

I don't know whether that should be considered a bug or a feature. (Twiddling the web-settings just for this scenario seems like a bad idea, because we don't want to risk accidentally breaking those settings for the actual browser webview.) If it's a bug, it's probably easiest to just use the system WebView for all support pages, if it's a feature then the WebViewProvider is the way to go!

pocmo:
I was mostly interested in using whatever engine is configured for this build variant (WebKit vs. Gecko). But you are right we shouldn't do the setup we do for blocking content. I think we can just ignore this for v1: We are going to use WebView - and we can look at Gecko once we switch. We already regressed the Gecko build anyways.

Having a single activity for all screens that are actually showing a website makes sense. Although we might need to translate some of them ("Your Rights" is always English afaik). Let's file a new issue for that. I'm going to remove the triage label from this one.

ahunt:
One further issue which Sebastian brought up: even if we use a plain WebView, we're still sharing data with the (webkit variant) BrowserFragment - i.e. cache, cookies, webstorage are all shared. Ideally we'd have separate data storage to make InfoActivity and BrowserFragment independent - but WebView doesn't appear to offer any way of doing that.

PR #183, opened 2017-03-08
Ahunt/149 defaultsumo

ahunt:
We might change the URL, I'll leave #149 until that's decided. I'd like to land this code now though so we can reuse InfoActivity for other SUMO pages. (The API might change in future bugs, I'm not too sure yet.)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -50,6 +50,9 @@
 
         <activity android:name=".activity.RightsActivity" />
 
+        <activity android:name=".activity.InfoActivity"
+                  android:theme="@style/SettingsTheme"/>
+
         <service
             android:name=".customtabs.CustomTabsService"
             android:exported="true">
--- a/app/src/main/java/org/mozilla/focus/activity/InfoActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/InfoActivity.java
@@ -1,0 +1,69 @@
+/* -*- Mode: Java; c-basic-offset: 4; tab-width: 4; indent-tabs-mode: nil; -*-
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.activity;
+
+import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
+import android.support.annotation.Nullable;
+import android.support.v7.app.AppCompatActivity;
+import android.support.v7.widget.Toolbar;
+import android.view.View;
+import android.webkit.WebView;
+import android.webkit.WebViewClient;
+
+import org.mozilla.focus.R;
+
+/**
+ * A generic activity that supports showing additional information in a WebView. This is useful
+ * for showing any web based content, including About/Help/Rights, and also SUMO pages.
+ */
+public class InfoActivity extends AppCompatActivity {
+    private static final String EXTRA_URL = "extra_url";
+    private static final String EXTRA_TITLE = "extra_title";
+
+    public static final Intent getIntentFor(final Context context, final String url, final String title) {
+        final Intent intent = new Intent(context, InfoActivity.class);
+
+        intent.putExtra(EXTRA_URL, url);
+        intent.putExtra(EXTRA_TITLE, title);
+
+        return intent;
+    }
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        setContentView(R.layout.activity_info);
+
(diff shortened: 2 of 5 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -50,6 +50,8 @@
 
         <activity android:name=".activity.RightsActivity" />
 
+        <activity android:name=".activity.InfoActivity" />
```
ahunt: We probably need to add a theme here, right now I'm seeing a purpley statusbar which is ugly, and I think setting a theme should fix that.

pocmo:
SupportUtils is missing in this patch.

ahunt:
Heh, I ended up not actually using SupportUtils (I'll need that for other sumo page) - but didn't remove the include... I'm going to rebuild this to use plain old WebView anyway, based on comments in #182.

app/src/main/java/org/mozilla/focus/activity/InfoActivity.java
```diff
@@ -0,0 +1,69 @@
...
+        setContentView(R.layout.activity_info);
+
+        final String url = getIntent().getStringExtra(EXTRA_URL);
+        final String title = getIntent().getStringExtra(EXTRA_TITLE);
+
+        final WebView webView = (WebView) findViewById(R.id.webview);
+        // By default WebView will try to open links and redirects in an external browser (and sumo
+        // links usually use a redirect to get to the desired target page). Somehow, setting a default
+        // WebViewClient() is enough to disable that behaviour.
+        webView.setWebViewClient(new WebViewClient());
```
pocmo: I wonder whether we should go through WebSettings and there's nothing enabled by default that could lead to problems.

Do WebViews share a cache? E.g. is everything we browse to here available to the WebView in BrowserFragment? Should we clear the cache etc. in onDestroy()?

app/src/main/res/layout/activity_info.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
```
pocmo: nit: license header

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -20,23 +20,5 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
-            if (preference instanceof PreferenceGroup) {
-                setupPreferences((PreferenceGroup) preference);
-            }
-
-            if (preference instanceof DefaultBrowserPreference
-                    && !((DefaultBrowserPreference) preference).shouldBeVisible()) {
-                group.removePreference(preference);
-                i--;
-            }
-        }
```
pocmo: I guess removing this part is a mistake?

app/src/main/java/org/mozilla/focus/activity/InfoActivity.java
ahunt: As far as I can tell, there's only one webview cache per (Android) application, which all WebViews then reuse. I don't actually see any way of overriding this behaviour - which means we might have to share data between BrowserFragment and here. Clearing the cache/history/etc will then affect an existing BrowserFragment, which we probably don't want either.

But: as far as I can tell: history is specific to a WebView (so we don't need to care about that), actually shared data is restricted to the cache (which we can clear without huge harm to the current session, except for increased data usage), cookies (we could remove/restore sumo cookies), WebStorage (same as cookies), maybe more?

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
ahunt: Do we still need this? AFAICT we now show the DefaultBrowserPreference on all Android versions, so we don't need to manually remove it?

app/src/main/java/org/mozilla/focus/widget/DefaultBrowserPreference.java
```diff
@@ -23,13 +25,17 @@ public DefaultBrowserPreference(Context context, AttributeSet attrs) {
...
     @Override
     protected void onClick() {
-        Intent intent = new Intent(Settings.ACTION_MANAGE_DEFAULT_APPS_SETTINGS);
-        getContext().startActivity(intent);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            Intent intent = new Intent(Settings.ACTION_MANAGE_DEFAULT_APPS_SETTINGS);
+            getContext().startActivity(intent);
+        } else {
+            // This link apparently doesn't need the usual SUMO processing: it's not app specific, but platform specific
+            final String url = "https://support.mozilla.org/kb/make-firefox-default-browser-android?utm_source=inproduct&amp;utm_medium=settings&amp;utm_campaign=mobileandroid";
```
pocmo: What about the utm_ flags. Do we need to change them so that we can see users are coming from Focus? (Let's move this question to the issue)

Issue #146, opened 2017-03-02
Remove overflow menu from settings


pocmo:
Instead move help etc. into the list of settings as an item.

pocmo:
@antlam I wonder if it's weird that on the start page you have settings AND help etc. And then when you go to settings you have those options again. Should we remove them from this overflow menu too (but then we only have settings in there) or only from the browser toolbar menu?

antlam:
Yes! Sorry that must've been unclear. I was thinking we would have the following states for the Overflow button.

Home:
- Settings
- About
- Help
- Your rights

Browsing (toolbar):
- Settings
- plus everything else...

Settings:
-Remove

@pocmo can we try that ^?

ahunt:
Just to double check: we don't want to link to about/help/rights from settings at all? (IIRC Sebastian was suggesting listing about/help/rights as settings items, like we do in firefox?)

antlam:
We want to link to them. But have them as items in the list rather than shortcuts in the overflow button.

ahunt:
![abouthelprights](https://cloud.githubusercontent.com/assets/1546915/23656673/f23ac2f8-033a-11e7-84f1-894b34f01b4a.png)

I'm wondering if we want better strings? E.g. "About Focus" (and possibly descriptions below the title?)

antlam:
Yeah, we could say "About Firefox Focus". But we would need to localize those strings.

I don't think we need descriptions below.

@bbinto thoughts? Keep in mind "Send anonymous usage data - Learn more" will be change to be about the Core ping later. So this grouping still kind of makes sense to me ("Mozilla").

bbinto:
@antlam, did you check with Michelle H. for wording?  

In general, I lean towards how we do things in Firefox Android. There it says "About Firefox", so I think "About Firefox Focus" sounds right.

antlam:
@bbinto just reached out - stay tuned!

pocmo:
String freeze is on Friday. We need a decision here.

mheubusch:
I like the idea of About Firefox Focus - I don't know that it adds much beyond what a user would assume, but it does reiterate the name and the association between the app and Firefox.  I think also that because this is a link and not a setting, there is no descriptive content needed other than that title. The description should be reserved for settings you perform in the app.

ahunt:
Are we doing this for just the item in Settings, or also for the main app menu?

antlam:
[...]

Issue #89, opened 2017-02-17
Do something with "learn more" link in Preferences -> Telemetry


ahunt:
We currently show "learn more" under the "Send anonymous usage data" preference. I'm guessing we'll either want to show something when pressing that and/or remove the "learn more" text.

st3fan:
@ahunt on iOS this links to https://support.mozilla.org/1/mobile/3.0/iOS/en-US/usage-data - the 3.0 and iOS parts are the app version and platform. We use this function to generate these SUMO links:

https://github.com/mozilla-mobile/focus/blob/master/Blockzilla/SupportUtils.swift#L14

ahunt:
On Android, tapping anywhere on the preference changes that preference. That makes me wonder if we should move "Learn more" into a separate compact item - we already do that for a "Learn More" link in Fennec's Advanced Preferences:
![learnmore_cropped](https://cloud.githubusercontent.com/assets/1546915/23624855/7b0bca62-02a7-11e7-8753-133318334602.png)

ahunt:
@antlam 
Does this look like an acceptable option? (Ignore the icon for now, that's just a placeholder)
![learnmore_proto1](https://cloud.githubusercontent.com/assets/1546915/23704246/3e4dfdee-0404-11e7-8d86-6e8f14d9e4a9.png)

antlam:
Talked to @ahunt on Slack a bit. 

Looks like we can just underline the "Learn more" text and keep it where it was before but have it link out to a webpage if the user presses on that link versus just flipping the toggle.

ahunt:
Screenshots of this in action:
![landed_learnmore](https://cloud.githubusercontent.com/assets/1546915/23849832/56c70744-07dd-11e7-8664-933d2ed28b91.png)

And what's shown when you click on that (I've scrolled down a little):
![landed_learnmore_content](https://cloud.githubusercontent.com/assets/1546915/23849856/6633228a-07dd-11e7-825c-61527f61d049.png)

antlam:
I think the title isn't necessary. 

But then it makes me think... why aren't we just using a Custom Tab? I briefly talked about this with @pocmo so mostly leaving a comment here in that regard so i dont forget :)

ahunt:
I'm adding custom tabs usage for all info screens in #199, therefore closing this bug for now!

PR #185, opened 2017-03-08
Move about/help/rights into preferences from preferences-menu (#146)

ahunt:
We might still change the strings / categorisation as part of #146, but landing this initial move will help avoid conflicts when making the "learn more" link work (#89).


app/src/main/java/org/mozilla/focus/activity/SettingsActivity.java
```diff
@@ -22,32 +22,4 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
 
         getSupportActionBar().setDisplayHomeAsUpEnabled(true);
     }
-
-    @Override
-    public boolean onCreateOptionsMenu(Menu menu) {
-        getMenuInflater().inflate(R.menu.menu_settings, menu);
```
pocmo: Did you remove menu_settings.xml too?

Issue #163, opened 2017-03-06
some translated locales still displays in English ('id', 'nl')


npark-mozilla:
After setting the locale to 'id' (Bahasa Indonesia I believe), the Focus app still displays menus in English.  I also tried in Bahasa Melayu, but shows in English as well.  Tried both on simulator and Experia Z3C with marshmallow.  (Tried both manually and using screengrab)

![browserviewmenu_1488815180187](https://cloud.githubusercontent.com/assets/6930774/23617332/cb92f8fe-025a-11e7-9e26-5146c1219151.png)

npark-mozilla:
Just found 'nl' also shows in english as well.

tomer:
Hebrew is missing as well. It is 100% translated on Pontoon, and the generated screenshots show the strings in English. (But featuring an Hebrew keyboard layout…) 

Locale code `he` everywhere but `iw` on Java/Android.

ahunt:
I'd filed #186 to fix the locale mappings (which would fix Hebrew, i.e. iw/he), but I then realised that Indonesian (id/ih) is also one of the locales needing a mapping, i.e. my bug is a dupe.

Outdated mappings doesn't explain Bahasa Melayu though, that might need more investigation

ahunt:
> Just found 'nl' also shows some locales in english as well. it looks like perhaps not all locales that are marked as 100% in https://l10n.mozilla-community.org/webstatus/?product=focus-android are fully localized.

I think for that we just need to do a new string import values-nl/strings.xml in master only has 24-ish strings, whereas that dashboard states there are 37 strings.

pocmo:
@npark-mozilla Is there any locale left that doesn't work (and has strings imported) or can this be closed?

pocmo:
Looking at the latest screenshots 'nl' and 'id' show up correctly now.

PR #187, opened 2017-03-08
Add outdated locale mapping (#163)

ahunt:
This should hopefully make Hebrew, Indonesian, and Yiddish all work.

(I've tested hebrew to work locally - previously only english fallback strings were visible. We'll
need a new string import for default builds to actually reflect these changes.)


pocmo:
Wow, that's weird. Might be worth filing this upstream too:
https://github.com/miracle2k/android2po/issues

PR #229, opened 2017-03-14
Search Engines

pocmo:
This is not feature complete yet. Those patches add the plugins and a JSON configuration, as well as a simple parser so that we can show the actual search engines for the current locale in settings (and finalize the UI).

Missing pieces:
* Parse and use the URL pattern from the search plugin
* Set and read the default engine from shared preferences
* Scripts for automating the import of new configurations


app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
```diff
@@ -5,27 +5,151 @@
...
+            // No configuration for the current locale found. Let's use the default configuration.
+            return configuration.getJSONArray("default");
+        } catch (JSONException e) {
+            // Assertion error because this shouldn't happen: We check whether a key exists before
+            // reading it. An error here would mean the JSON file is corrupt.
+            throw new AssertionError("Reading search configuration failed", e);
+        }
+    }
+
+    public synchronized List<SearchEngine> getSearchEngines() {
```
ahunt: Hmm - since we're already waiting for loading to finish - do we need to be synchronised? (I guess there's no real difference either way though, so this is probably unimportant.)

pocmo:
> lgtm (although I can't verify whether a lot of the other language search engine definitions are actually correct :p )

Yeah, me neither. This is a 1:1 import from Focus for iOS which imports them from Fennec and modifies them slightly. I still need to find a person who can tell us whether we can just use the iOS ones or if we need to customize things.

app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
pocmo: We need to synchronize so that the main thread sees the data loaded by the background thread. In non synchronized code the main thread could see a cached version of `searchEngines` that is still null [[1]](https://www.securecoding.cert.org/confluence/display/java/Concurrency,+Visibility,+and+Memory).

PR #235, opened 2017-03-14
Update external-app dialog styling

ahunt:
Trying to set the style using:
<item name="alertDialogStyle">@style/DialogStyle</item>
in AppTheme, results in:
android.content.res.Resources$NotFoundException: Resource ID #0x0

We're passing in the correct activity, so I've got no idea why this is
happening. Explicitly setting the style when creating the dialog works
so lets just stick with that for now. (Extending AlertDialog.AppCompat
results in the theming being ignored. Even an empty style that
extends Theme.AppCompat.Dialog.Alert results in crashes unless we set
it in code, so it seems something is wrong with only the xml codepath.)

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/IntentUtils.java
+++ b/app/src/main/java/org/mozilla/focus/utils/IntentUtils.java
@@ -114,7 +114,7 @@ public class IntentUtils {
     // end up needing more or a variable number we can change this, but java varargs are a bit messy
     // so let's try to avoid that seeing as it's not needed right now.
     private static void showConfirmationDialog(final Activity activity, final Intent targetIntent, final String title, final @StringRes int messageResource, final CharSequence param) {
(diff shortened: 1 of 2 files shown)


pocmo:
I wonder why the style is needed. The dialog in SearchEnginePreference has the accent color without me setting a theme.

ahunt:
> I wonder why the style is needed. The dialog in SearchEnginePreference has the accent color without me setting a theme.

For the buttons, it seems colours work: they use colorAccent, however colorAccent for the main activity is more purple than blue (and antlam wanted blue buttons). The title is always white by default, but we want blue for the link so that needs to be set regardless.

(We might not want to make the title blue anymore since we're no longer showing a link in the dialog, I'll hold off on landing until that's clarified.)

ahunt:
(With the new strings we only need blue text in the buttons, and not for the title - hence I've updated the patch.)

Issue #228, opened 2017-03-14
Create a way to induce specific web connection errors


npark-mozilla:
With commit https://github.com/mozilla-mobile/focus-android/commit/243073f149b9cc9be366fa24763bba814b579e08, error messages show up when connection errors are present.  We need a way to induce each type of error if we need to check the localization of the page.

pocmo:
@ahunt: Would it be possible to create a custom protocol that shows the error pages? A bit like about:neterror in Fennec.

ahunt:
Yeah - a custom protocol might be useful for certain cases anyway.

Another option might be to find the WebView, get it's WebViewClient, and call onReceivedError() with the desired error code. That's hacky and specific to the webview edition of focus though.

ahunt:
I've got a first draft. WebView does weird things when you try to use about: pages, so I'm using about2:neterror as the provisional URL.

Do we want any specific syntax? For now I'm using about2:neterror?wkerror=N, where N corresponds to the equivalent webkit error code as defined in:
https://developer.android.com/reference/android/webkit/WebViewClient.html

(Note: we don't yet support all those errorcodes, our list is at the top of ErrorPage.java)

pocmo:
> Do we want any specific syntax? For now I'm using about2:neterror?wkerror=N, where N corresponds to the equivalent webkit error code as defined in:
https://developer.android.com/reference/android/webkit/WebViewClient.html

As long as we are only using this internal and are not exposing this to users I don't think it's important. How about
- focus:error?wkerror=-12
- or just: error:-12

npark-mozilla:
i was gonna say about2:wkerror=N but I'm fine with anything

ahunt:
@npark-mozilla It's landed with "error:<CODE>" syntax (that's simplest to implement, so least risky too)

The list of supported codes is here:
https://github.com/mozilla-mobile/focus-android/blob/b561c6e8c001ca2e53d219e3efbaa01c9a49f753/app/src/webkit/java/org/mozilla/focus/webkit/ErrorPage.java#L23

npark-mozilla:
I see the customized error msgs now at will, nice!

PR #237, opened 2017-03-15
Implement error: page for error testing (#228)

ahunt:
If you make an error in the URL syntax, you'll end up on the
BAD_URL error page.

### DIFF
--- a/app/src/webkit/java/org/mozilla/focus/webkit/TrackingProtectionWebViewClient.java
+++ b/app/src/webkit/java/org/mozilla/focus/webkit/TrackingProtectionWebViewClient.java
@@ -9,7 +9,6 @@ import android.graphics.Bitmap;
 import android.net.http.SslError;
 import android.os.AsyncTask;
 import android.support.annotation.WorkerThread;
-import android.webkit.HttpAuthHandler;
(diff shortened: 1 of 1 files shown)


app/src/webkit/java/org/mozilla/focus/webkit/TrackingProtectionWebViewClient.java
```diff
@@ -116,6 +115,32 @@ public void onReceivedError(final WebView webView, int errorCode,
         // we just check whether the failing URL is the current URL, which is enough to detect an error
         // in the main frame.
 
+        // WebView swallows odd pages and only sends an error (i.e. it doesn't go through the usual
+        // shouldOverrideUrlLoading), so we need to handle special pages here:
+        // about: urls are even more odd: webview doesn't tell us _anything_, hence the use of
+        // a different prefix:
+        if (failingUrl.startsWith("error:")) {
+            // format: error:<error_code>
+            final int errorCodePosition = "error:".length();
```
pocmo: I guess we could hardcode "6" here? Or even nicer: Move "error" to a public static final field and use that in both cases.

PR #238, opened 2017-03-15
Add About view

ahunt:


app/src/main/java/org/mozilla/focus/utils/SupportUtils.java
```diff
@@ -36,4 +36,9 @@ public static String getSumoURLForTopic(final Context context, final String topi
 
         return "https://support.mozilla.org/1/mobile/" + appVersion + "/" + osTarget + "/" + langTag + "/" + escapedTopic;
     }
+
+    public static String getManifestoURL() {
+        final String langTag = Locales.getLanguageTag(Locale.getDefault());
+        return "https://www.mozilla.org/" + langTag + "/about/manifesto/";
```
pocmo: I wonder if there are locales this site does not support? If I use a made up locale then I see a 404 page but I can't find a valid locale that doesn't work.

app/src/main/java/org/mozilla/focus/activity/RightsActivity.java
```diff
@@ -1,34 +0,0 @@
...
-import android.webkit.WebView;
-
-import org.mozilla.focus.R;
-
-public class RightsActivity extends AppCompatActivity {
-    @Override
-    protected void onCreate(@Nullable Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
-
-        setContentView(R.layout.activity_rights);
```
pocmo: Can we remove the layout file too?

app/src/main/res/layout/activity_info.xml
```diff
@@ -11,7 +11,7 @@
         android:id="@+id/toolbar"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:background="@color/colorTitleBar"
```
pocmo: Can we remove the color from colors.xml?

PR #240, opened 2017-03-15
Disable SECURE mode in debug builds

ahunt:


app/src/main/java/org/mozilla/focus/utils/Settings.java
```diff
@@ -36,6 +37,12 @@ public boolean shouldShowFirstrun() {
     }
 
     public boolean shouldUseSecureMode() {
+        // Always allow screenshots in debug builds - it's really hard to get UX feedback
+        // without screenshots.
+        if (BuildConfig.DEBUG) {
```
pocmo: Currently our beta builds are debuggable too. Instead you might want to use BUILD_TYPE here.

Issue #218, opened 2017-03-13
Reference Roboto fonts from system in web content


pocmo:
We added the "Your Rights" screen in #23 (Pull request: #157). For that we added the Roboto fonts to the assets folder. However we do not want to ship them with the app. Instead we should reference them from the system. We are building Focus for API 21+. All versions should include those fonts already.

PR #241, opened 2017-03-15
Use bundled roboto fonts (#218)

ahunt:
It's hard to tell that this is working: the easiest way to verify
is the "Your rights" html content title in "Your rights". Its weight
increases when we use sans-serif-medium vs. no setting (and webview
seems to default to roboto regardless, so the remaining text doesn't
change).


ahunt:
Some people have mentioned Samsung devices running Android 4 breaking when you use sans-serif, I don't have any more modern (Android 5+) Samsungs to test whether that still applies to the Android versions we support.

Issue #44, opened 2017-02-07
Add basic support for Android tablets


antlam:
UX should work together with Eng here to figure out how to keep this simple and moving forward. We will likely be able to fork over a lot of the UX but there will still need to be some massaging I think.

pocmo:
Back, forward, Refresh -> Toolbar. FAB and menu stay.

antlam:
Details to follow:

Quick iPad screenshot for reference.

![simulator screen shot nov 30 2016 11 38 35 am](https://cloud.githubusercontent.com/assets/1994820/23567221/29c20a62-000a-11e7-8c0b-9726f4fd8266.png)

pocmo:
Reminder: The bottom sheet I added in #26 might need to be adapter for tablets too (at least the width).

ahunt:
Similarly, I think the first run screen might need adapting.

antlam:
![screen shot 2017-03-08 at 12 15 01 pm](https://cloud.githubusercontent.com/assets/1994820/23722111/fa8875fe-03f8-11e7-8468-ccb6652bb59d.png)

Had some time so I started on a WIP landscape mode to gauge the spacing etc.

pocmo:
![tablet_preview](https://cloud.githubusercontent.com/assets/89638/23964339/72b25fd4-09b4-11e7-9858-302451226875.png)

PR #249, opened 2017-03-15
Add toolbar layout for tablets. (#44)

pocmo:


pocmo:
Yeah, we are doing it correctly here :)

PR #252, opened 2017-03-15
Update translated strings from L10N repository.

pocmo:
A fresh import.


pocmo:
Seems like we need to fix #253 first.

pocmo:
Okay, now with CDATA :)

Issue #253, opened 2017-03-15
CDATA escaping missing when importing strings.


pocmo:
I just did a string import (See #252) and saw that the CDATA escaping is missing from the strings.xml files. Because of that we are loosing the HTML tags / formatting when displaying those strings.

pocmo:
@ahunt: I wonder if we could just wrap an imported string in CDATA whenever there's something in there that looks like a tag.

ahunt:
I think I forgot to push my branch, I'm trying to recover it now...

ahunt:
nvm: this isn't working:
https://github.com/mozilla-mobile/focus-android/commit/da1293679f0727c80b1b0fc38d112d525dfe2974

We need to expand our detection for at least the about page which doesn't have <ul>

ahunt:
Looking at the latest import, I noticed even <ul> isn't being detected - which is really odd. (I'd been testing on some fake data locally), so this might need more work:
https://github.com/mozilla-mobile/focus-android/pull/252/commits/28e24734b3c8198f01781c0492d38b93265d7c61

PR #254, opened 2017-03-15
Improve string import html detection (#253)

ahunt:
Also use the CDATA util that etree provides.


ahunt:
This works on my system, but given that the previous version failed completely in the following commit I don't know if this commit will actually help on all systems, or only mine:
https://github.com/mozilla-mobile/focus-android/pull/252/commits/28e24734b3c8198f01781c0492d38b93265d7c61

PR #256, opened 2017-03-15
Show "About <APPNAME>" in Settings (#146)

ahunt:
setupPreferences() first appeared in:
2717d17a345eccafada5de841a943de339aa2f6e
and was removed again in:
1710608811ebebbbfcfca0ebfe741352e21f4b65

### DIFF
--- a/app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
+++ b/app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
@@ -9,14 +9,12 @@ import android.content.Intent;
 import android.os.Bundle;
 import android.preference.Preference;
 import android.preference.PreferenceFragment;
-import android.preference.PreferenceGroup;
 import android.preference.PreferenceScreen;
 import android.support.annotation.Nullable;
 
 import org.mozilla.focus.R;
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -24,6 +24,23 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
+
+    private void setupPreferences(PreferenceGroup group) {
+        for (int i = 0; i < group.getPreferenceCount(); i++) {
+            final Preference preference = group.getPreference(i);
+
+            if (preference instanceof PreferenceGroup) {
+                setupPreferences((PreferenceGroup) preference);
+            }
+
+            if (preference.getTitle().equals(getString(R.string.preference_about))) {
```
pocmo: I'd have a better feeling if we'd use a key to find the preference here.

pocmo: Another option would be to write a quick AboutPreference that extends Preference and sets the title in the constructor. That way we wouldn't need to traverse the preferences here.

PR #258, opened 2017-03-16
Handle data: URIs correctly

ahunt:


app/src/webkit/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -80,6 +81,10 @@ public WebkitView(Context context, AttributeSet attrs) {
 
             setWebViewClient(client);
             setWebChromeClient(createWebChromeClient());
+
+            if (BuildConfig.DEBUG) {
```
ahunt: (I'm guessing we do want this for all debug builds, since people on beta might have the need/experience to debug web content issues that we've caused.)

PR #264, opened 2017-03-16
Use custom tabs for external links

ahunt:
This doesn't necessarily avoid data sharing, since we might end up reusing focus's custom tabs.
But it does let us use custom tabs, from any browser, for external pages.


pocmo:
The more I think about this I come to the conclusion that we shouldn't land this in v1:
* It feels weird and unexpected to jump to a different/third-party browser to show some of the pages, even though it looks like we are still inside our app visually.
* It makes some things easier but introduces other issues (e.g. with Focus being a custom tab provider too)
* Things like #283 require that we have control about how content is displayed - and we shouldn't have multiple different ways of showing content inside the app.

pocmo:
Closing this request - We talked on Slack and decided to not merge this for v1.

Issue #250, opened 2017-03-15
Sometimes "loading does not finish" when displaying error pages


pocmo:
Sometimes when showing an error page the loading indicator is still visible and the toolbar color doesn't change. Refreshing the page usually fixes that.

![errorpage_loading](https://cloud.githubusercontent.com/assets/89638/23965178/74a50424-09b7-11e7-8539-b317beef1927.png)

ahunt:
I don't know if there's a clean way for us to fix this: webview is doing the following:

1. pageStarted() when trying to load the page
2. onErrorReceived() (where we load the error page)
3. onPageFinished()
4. onPageStarted() (we start the progress bar again)
5. Only sometimes (about 50/50), onPageFinished() again

We could probably write some hacky code to ignore onPageStarted() when called directly after onErrorReceived(), I'm not sure how reliable that would be.

(All callbacks supply the same URL, since we're pretending to load the original URL in onErrorReceived()).

Trying to call WebView.stopLoading() before loading the error page doesn't seem to help either.

PR #274, opened 2017-03-17
Hack around occasionally missing onPageFinished() for error pages (#250)

ahunt:
This isn't very pretty, but it's the best method I've though of so far.


pocmo:
I'm a bit afraid that this workaround will bite us in some edge cases. If we ever have an error that doesn't call onPageStarted() then this will break the UX of the next request.

When filing the bug I was thinking about maybe redirecting to error:XXX and just not updating the URL toolbar for error pages (For WebView we are on the error page, but we keep the actual URL in the toolbar because it is more convenient for the user). Although there could be problems with that too.

What do you think?

ahunt:
I've been thinking about redirecting: the problem with using an error page URL is that the reload button won't work. We can change the error-page reload button to simply open the broken URL (instead of doing a reload()), so that's maybe not a huge issue (the in-app reload button will probably need extra logic though, since it uses WebView.reload() IIRC).

Another solution might be to set a "pending" error in onErrorReceived(), we can then load that in onPageFinished(). It works quite well, except for an occasional white flash while reloading (but that was already happening before testing that approach).

ahunt:
> Another solution might be to set a "pending" error in onErrorReceived(), we can then load that in onPageFinished(). It works quite well, except for an occasional white flash while reloading (but that was already happening before testing that approach).

This approach doesn't work for no-connection type errors (e.g. if I disable wifi) - in that case I still get a stuck progress bar, even if I delay error page loading. The original patch seems to still be the most robust method I can find.

I think we're pretty  much guaranteed to get onPageStarted() since we're loading new page data, but I'll try to test a bit to see what happens if we deliberately break onPageStarted().

Issue #170, opened 2017-03-07
Transition from first-run screen to HomeFragment/BrowserFragment


ahunt:
A better transition from first-run to either the main screen (or direct to the Browser if the first-run is shown when opening Focus via a VIEW intent) could be nice.

I'm guessing this isn't super important, but this is the first thing a user sees when using the app, so it's still quite nice to have?

antlam:
On iOS, I believe we went with a 200ms linear fade just so its not so jarring.

Try that! @ahunt

ahunt:
My phone doesn't seem to want to let me make great videos:
![fade2](https://cloud.githubusercontent.com/assets/1546915/24035742/11b41670-0af7-11e7-87e1-1ecf75ce8c8c.gif)

antlam:
@ahunt I don't have my Android device with me right now so I'll have to give feedback later or you could try sending a .MOV file to me instead? QuickTime has a pretty easy way to record too. 

File > New Screen Recording and then click the Record button

antlam:
@ahunt could you just push me a personal buddy build to try this out? I'm updating to master but I cant trigger this anymore

PR #275, opened 2017-03-17
Add exit fade to first-run (#170)

ahunt:
I wonder if we should move the 200ms definition into xml, but that doesn't seem to be someting we'll need to reconfigure.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.java
@@ -9,10 +9,11 @@ import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.support.annotation.Nullable;
 import android.support.v4.app.Fragment;
+import android.transition.Transition;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.java
```diff
@@ -23,7 +24,13 @@
     public static final String FIRSTRUN_PREF = "firstrun_shown";
 
     public static FirstrunFragment create() {
-        return new FirstrunFragment();
+        final Fade exitFade = new Fade();
+        exitFade.setDuration(200);
```
pocmo: Why not an XML animation? It's often easier to tweak and add additional effects. :)

ahunt:
I'm not quite sure how to request a new review (I've moved the transition into xml, but not 100% sure I'm doing things sensibly), therefore I'm doing a manual rereview request ping: @pocmo !

ahunt:
(Waiting on final UX review before pushing.)

PR #290, opened 2017-03-17
Refactor and fix URL handling

ahunt:
Enabling external intent URL processing for URLs entered by hand exposed some issues in the external intent handling, which I've now fixed.

These patches also move browser specific webviewclient code into it's own class, and removes non-tracking-protection relevant code from the TrackingProtectionWebViewClient.


app/src/webkit/java/org/mozilla/focus/web/FocusWebViewClient.java
```diff
@@ -0,0 +1,133 @@
+package org.mozilla.focus.web;
```
pocmo: nit: License header.

PR #293, opened 2017-03-19
First set of UI Tests

npark-mozilla:
a set of UI tests that does basic sanity checking

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -89,6 +89,11 @@ dependencies {
 
     testCompile 'junit:junit:4.12'
     testCompile "org.robolectric:robolectric:3.1.4"
+    androidTestCompile 'com.android.support.test.espresso:espresso-idling-resource:2.2.2'
+    androidTestCompile 'com.android.support.test.uiautomator:uiautomator-v18:2.1.2'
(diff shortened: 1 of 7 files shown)


npark-mozilla:
A couple more tests will be added once the URL refactoring work is done

app/build.gradle
```diff
@@ -78,6 +78,12 @@ repositories {
     }
 }
 
+configurations.all {
+    resolutionStrategy {
+        force 'com.android.support:support-annotations:25.1.1'
```
pocmo: Can you help me understand this line? It looks like you are enforcing a specific version support-annotations for all projects and dependencies. But later you are excluding support-annotations from other dependencies. Isn't only one needed (Excluding sounds like a nice strategy)?

app/src/androidTest/java/org/mozilla/focus/activity/OpenwithDialogTest.java
```diff
@@ -0,0 +1,119 @@
+package org.mozilla.focus.activity;
```
pocmo: nice: License header here and in the other new files.

PR #302, opened 2017-03-20
add 3 more UI tests

npark-mozilla:
- That checks for bad url prefix
- Checks the partial content of About Us/ Your Rights
- Checks the visit of http:// https:// and turns off blockers


app/src/androidTest/java/org/mozilla/focus/activity/PageVisitTest.java
```diff
@@ -0,0 +1,127 @@
+package org.mozilla.focus.activity;
```
pocmo: nit: License header

npark-mozilla:
@pocmo yup will do, I ran into a issue where in API 23 and 24, the resource ids are not present- they were only showing on API 25.  I guess the way to define resources ids are not backwards compatible?  
Also I think I should refactor the code so I don't have to define locators on every script- will clean up and ask for your review.

npark-mozilla:
@pocmo, could you take a look please? I at least removed the repetition of UiLocator declarations. Thanks!

PR #308, opened 2017-03-21
Don't crash if webView == null

ahunt:
I'm not too what's the "correct" solution here. There's no point in checking the back state of the webview unless it's actually in use. The nullchecks are ugly, but since it's legitimate for a Fragment to not be attached/visible (but still alive and in a backstack), we should really handle that case too.


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -258,18 +258,24 @@ private void updateToolbarButtonStates() {
     }
 
     public String getUrl() {
```
pocmo: Let's add @Nullable here. This might not be expected. I guess we could keep a reference to the url ourselves. We should always have some because we pass one in when we create the Fragment (even if there's no WebView yet).

pocmo:
Please wait with merging this one. I need to restructure some code to fix #313 (We need to forward lifecycle methods from the fragment to the WebView).

pocmo:
As part of #314 I added null checks too. I guess we'd only need the first commit here.

PR #329, opened 2017-03-22
Remove obsolete onDetach()

ahunt:
Since we're now destroying the webView in onDestroy, there's
no webView to remove callbacks from. We can still remove the callback
in onDestroy just to be safe, but hopefully webView doesn't send
any more callbacks after we've told it to destroy itself.

This avoids crashing when exiting the browser using the back button.


pocmo:
Ah, this sounds like #331 :)

Issue #301, opened 2017-03-20
 URL spoofing on location change/stop


thebnich:
https://people-mozilla.org/~bnicholson/test/ios-security-test.html

[See here](https://github.com/mozilla-mobile/focus/commit/ed7c9a8) for the iOS fix.

ahunt:
I've got a provisional patch which seems to do a similar thing. I'm going to spend a bit more time to figure out if there are any possible workarounds for my fix, but it seems like it's a robust fix.

PR #336, opened 2017-03-22
Only update URL if URL was actually fetched (#301)

ahunt:
I'm going to leave the bug open while I think about it more - but I'd be keen to get feedback on any obvious workarounds / oversights in this approach.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -10,11 +10,9 @@ import android.graphics.drawable.TransitionDrawable;
 import android.net.Uri;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
-import android.support.v4.app.Fragment;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
-import android.webkit.WebView;
 import android.widget.ProgressBar;
 import android.widget.TextView;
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -66,12 +67,29 @@ public String getInitialUrl() {
...
 
+    private void updateURL(final String urlString) {
+        try {
+            URI uri = new URI(urlString);
+            final String userInfo = uri.getUserInfo();
+            if (userInfo != null) {
+                // Strip the userInfo to minimise spoofing ability. This only affects what's shown
+                // during browsing, this information isn't used when we start editing the URL:
+                uri = new URI(uri.getScheme(), null, uri.getHost(), uri.getPort(), uri.getPath(), uri.getQuery(), uri.getFragment());
+            }
```
pocmo: May that's something for UrlUtils?

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -66,12 +67,29 @@ public String getInitialUrl() {
...
+
+            urlView.setText(uri.toString());
+        } catch (URISyntaxException e) {
+            // WebView should always have supplied a valid URL
+            throw new IllegalStateException("WebView is expected to always supply a valid URL");
```
pocmo: I wonder if we should only crash on beta/debug and return the original URL on release.

app/src/webkit/java/org/mozilla/focus/web/FocusWebViewClient.java
```diff
@@ -59,8 +94,17 @@ public void onPageStarted(WebView view, String url, Bitmap favicon) {
...
+    public void onPageFinished(WebView view, final String url) {
         if (callback != null) {
+            final String trueURL;
+            // We still receive onPageFinished if loading is interrupted, in which case we might
+            // still be on the previous page. WebView's URL indiciates the "true" page URL.
+            if (view.getUrl().equals(url)) {
+                trueURL = url;
+            } else {
+                trueURL = null;
+            }
```
pocmo: I'm confused. It looks like we do not even use trueURL?

PR #341, opened 2017-03-23
Import strings from L10N repository.

ahunt:


ahunt:
This doesn't build, I'll need to investigate that first.

ahunt:
The issue is caused by "%\@" being included in about_content in values-ur/strings.xml, which confuses AAPT's string replacement checker.

This appears as "%@" in the input .po file.

pocmo:
I think we saw that before when someone just reused the iOS string without  looking. Just fix the string manually in strings.xml (if you know how, for some languages that might be hard). Then we can land this and ping delphine/flod to get the source string fixed.

ahunt:
flod fixed the string in pontoon, I've updated the import and everything works again!

Issue #321, opened 2017-03-22
Domain autocompletion is case-sensitive


thebnich:
Type "G". I'd expect "Google.com" to be suggested, but nothing happens (I see the same behavior on Firefox Android).

ahunt:
I've filed an equivalent fennec bug (that code isn't a 1:1 copy though):
https://bugzilla.mozilla.org/show_bug.cgi?id=1349797

PR #345, opened 2017-03-24
Make domain autocomplete case insensitive (#321)

ahunt:
We could make all our later String comparisons case insensitive, but
that requires more substantial modifications to the Autocomplete
EditText. This solution avoids EditText modifications.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/autocomplete/UrlAutoCompleteFilter.java
+++ b/app/src/main/java/org/mozilla/focus/autocomplete/UrlAutoCompleteFilter.java
@@ -12,6 +12,7 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/autocomplete/UrlAutoCompleteFilter.java
```diff
@@ -20,21 +20,34 @@ public UrlAutoCompleteFilter(Context context) {
...
+
     @Override
-    public void onFilter(String searchText, InlineAutocompleteEditText view) {
+    public void onFilter(final String rawSearchText, InlineAutocompleteEditText view) {
         if (domains == null || view == null) {
             return;
         }
 
+        // Search terms are all lowercase already, we just need to lowercase the search text
+        final String searchText = rawSearchText.toLowerCase();
```
pocmo: We should specify a locale here. As our domain list is ASCII only it should be good enough to always use Locale.US here.

Issue #320, opened 2017-03-22
Search text not updated after autocompletion


thebnich:
STR:
1. Type "g", see g[oogle.com].
2. Tap anywhere in the text field to accept the "google.com" completion.

This should change the search text to "search for google.com", but it still says "search for g".

PR #346, opened 2017-03-24
Fixed Search text not updated after autocompletion (#320)

sabbir044:
There was no listener after completing after user commit the
autocompletion suggestion, added a listener and update searchview based
on that.


sabbir044:
Seems there is better solution.Going to remove the pull request :).

ahunt:
Sorry, I never noticed that you were also working on this : (. Feel free to assign bugs to yourself and/or comment on them so we don't collide, and thanks for also looking into it : )!

PR #347, opened 2017-03-24
Use WebFragment for InfoView

ahunt:
(I'm not sure if we should bother with the webview hiding part since it's extra logic, but it does make the experience nicer on my slow devices.)

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/InfoFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/InfoFragment.java
@@ -17,6 +17,7 @@ import org.mozilla.focus.web.IWebView;
 
 public class InfoFragment extends WebFragment {
     private ProgressBar progressView;
+    private View webView;
 
     private static final String ARGUMENT_URL = "url";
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/focus/fragment/InfoFragment.java
```diff
@@ -0,0 +1,92 @@
+package org.mozilla.focus.fragment;
```
pocmo: nit: license header

app/src/main/java/org/mozilla/focus/fragment/InfoFragment.java
```diff
@@ -0,0 +1,92 @@
...
+    @NonNull
+    @Override
+    public View inflateLayout(LayoutInflater inflater, @Nullable ViewGroup container,
+                              @Nullable Bundle savedInstanceState) {
+        final View view = inflater.inflate(R.layout.fragment_info, container, false);
+        progressView = (ProgressBar) view.findViewById(R.id.progress);
+        webView = view.findViewById(R.id.webview);
+
+        final String url = getInitialUrl();
+        if (!(url.startsWith("http://") || url.startsWith("https://"))) {
```
pocmo: nit: might be something for UrlUtils.

PR #354, opened 2017-03-27
Some lint fixes

ahunt:
A bunch of lint fixes. There's still a lot more to do, but these ones helped discover one missing accessibility string, so it seems to have been a worthwhile use of time.


app/src/main/res/layout-sw600dp/toolbar.xml
```diff
@@ -18,7 +18,6 @@
         android:layout_marginTop="8dp"
         android:background="?android:attr/selectableItemBackgroundBorderless"
         android:padding="4dp"
-        android:contentDescription="@string/menu_back"
```
pocmo: Let's add a string here. Even though if we do not do an export for v1. We definitely want to have it in the next time. :)

app/src/main/res/layout/toolbar.xml
```diff
@@ -33,7 +33,6 @@
         android:gravity="center_vertical"
         android:lines="1"
         android:maxLines="1"
-        android:singleLine="true"
```
pocmo: Removing this breaks the fading edge. Don't ask me why.

ahunt:
I've readded the string (delphine suggested just adding it back anyway, and I guess worst case that means we'll have an english only back string if translations aren't prepared in time), and readded the singleLine (with ignore=deprecated)!

app/src/main/res/values/strings.xml
```diff
@@ -117,6 +117,9 @@
     <!-- Content description (not visible, for screen readers etc.): Reload current website -->
     <string name="content_description_reload">Reload website</string>
 
+    <!-- Content description (not visible, for screen readers etc.): Navigate backw (browsing history) -->
```
pocmo: typo: backw

PR #362, opened 2017-03-29
Handle http: by itself (and misc. minor optimisation)

ahunt:


app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -71,4 +83,12 @@ public static String stripUserInfo(String url) {
             }
         }
     }
+
+    public static boolean focusSupportURLProtocol(final String url) {
+        return (!url.startsWith("http:")) &&
+                (!url.startsWith("https:")) &&
+                (!url.startsWith("file:")) &&
```
pocmo: But I thought we do not support `file`?

ahunt: We do for Your Rights (which uses file://android_assets/etc), but that's only because we're now using the same WebViewProvider. Anything else should be disabled (and will show an error).

PR #365, opened 2017-03-30
Cleanup in onCreate and onPause() && isFinishing()

ahunt:


ahunt:
This does clear all data with "Don't keep activities". We also don't retain the currently opened site, which we'd presumably also need to add support for anyway (in which case we can performCleanup() if there's no restore data available?).

ahunt:
@pocmo re-review request for the last part (I've also filed #391 to make state restoration work properly, since it only restores the _first_ URL right now).

Issue #368, opened 2017-03-30
Configure android:fullBackupContent


ahunt:
See:
https://developer.android.com/guide/topics/data/autobackup.html

(We also get a lint warning because we don't configure this.)

The only thing we'd want backed up is the app preferences, stored in:
shared_preferences/APPID_preferences.xml

(Browsing data could potentially be backed up if a user force-kills the app, meaning browsing data isn't cleaned up/remains in the app directory.)

PR #369, opened 2017-03-30
Disable backup (#368)

ahunt:
This is the only data that users might want to persist. No browsing
data should be persisted on disk, but we want to be as safe as possible
so explicitly not backing up anything else seems to be wise.


pocmo:
I think I'd just disable backups and avoid upload anything. Telemetry is going to use some shared preferences too and I don't think we would like to restore that on a new install/device. :)

ahunt:
Yeah that seems simpler. We could just backup the default shared prefs, but those contain the app name and I'm not sure if there's a simple way of referring to that in the backup list (e.g. org.mozilla.focus.debug.prefs etc).

Issue #375, opened 2017-03-30
First-run screen expanding beyond viewable screen area


pocmo:
I might have caused that with my tablet polishing patches.

AaronMT:
Nexus 6 (7.1.1) running 4e3131 

![device-2017-03-30-110307](https://cloud.githubusercontent.com/assets/102331/24510936/86b9ae44-1538-11e7-8be2-e13169f9d498.png)

PR #376, opened 2017-03-30
Redirect match_parent dimen via another item to make it work (#375)

ahunt:
This fixes first run to actually have match_parent.


ahunt:
I wonder if we should just split the layouts instead, but this is still simpler it seems.

(We could keep the core part of the layout in firstrun_content, and then <include> that from firstrun_fragment in both normal and sw600dp version?)

PR #389, opened 2017-03-31
URL spoofing followups

ahunt:
I don't know if this covers the recent breakage, but at least we'll update the URL when loading finishes which is better than never updating the URL.


app/src/test/java/org/mozilla/focus/utils/UrlUtilsTest.java
```diff
@@ -3,7 +3,7 @@
 import junit.framework.Assert;
 
 import org.junit.Test;
-
+g
```
pocmo: Oops :)

Issue #391, opened 2017-03-31
State restoration


ahunt:
It looks like we partly already save MainActivity instance state: if you don't setup fragments in MainActivity.onCreate() when savedInstancedState != null, then BrowserFragment is restored - however it opens up to the first page that you opened. 

(When debugging, I noticed that UrlInputFragment gets opened, onCommit() gets called with the original input string, and BrowserFragment reopens.)

ahunt:
See the following commit/message for more:
https://github.com/mozilla-mobile/focus-android/pull/365/commits/d861b740ebf3cf1114d4c8f6ede49a391233e4d2

PR #393, opened 2017-03-31
Restore webview state when restarting activity (#391)

ahunt:
I'll try testing this a bit more before formally requesting review, but feedback is welcome!

### DIFF
--- a/app/src/main/java/org/mozilla/focus/web/IWebView.java
+++ b/app/src/main/java/org/mozilla/focus/web/IWebView.java
@@ -5,6 +5,8 @@
 
 package org.mozilla.focus.web;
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/web/IWebView.java
```diff
@@ -42,4 +44,8 @@
     boolean canGoForward();
 
     boolean canGoBack();
-}
+
+    void restoreWebviewState(Bundle savedInstanceState);
+
+    void onSaveInstanceState(Bundle outState);
```
pocmo: You forgot GeckoWebView. :)

PR #399, opened 2017-04-03
Fixed #338 and #398

### LINKED ISSUE #338: Firefox Focus should be able to detect installed Firefox Beta
Hi there.
I'm testing Firefox Focus on Android via Buddy Builds. My main task is checking if focus working with RTL languages correctly?
I saw focus have an option in it's menu, named Open with Firefox. When i touched that option it went to Google Play to show the page of Firefox in Google Play.
But i installed Firefox Beta via Google Play before that.
I think focus should be understand if i installed Firefox beta instead of Firefox? If a version of Firefox Beta is on Android now, that option should be changed to Open With Firefox Beta or it should be add another menu option named open with Firefox Beta in addition of Open With Firefox.
It's true that i can use open with... option and see all of browsers on my Android phone, then select one of them(Included Firefox Beta). But i think if Focus be able to show Open with Firefox Beta in main menu, it will be better. Because one of missions of focus is Adverting for Firefox and should be able to show All Firefox versions in it's menu.

Thanks.
Amir Farsi.
Member of Persian L10n Team.

### LINKED ISSUE #398: Open with Firefox does not open Firefox in playstore when not installed
When Firefox is not installed, if I press on open with firefox, it starts the playstore but shows "Item not found" in screen. It should go to the firefox page to install firefox.

sabbir044:
Fixed #338-Firefox Focus should be able to detect installed Firefox Beta
Now focus will able to detect the following version of firefox and show
in menu if installed.
1. Firefox
2. Firefox-Beta
3. Firefox-Aurora
4. Firefox-nightly

Fixed #398
open firefox install page in playstore if firefox is not
installed.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -331,13 +331,13 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
 
                 final Browsers browsers = new Browsers(getContext(), webView.getUrl());
 
-                if (browsers.isInstalled(Browsers.KnownBrowser.FIREFOX)) {
+                if (browsers.hasFirefoxBrandedBrowserInstalled()) {
                     final Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(webView.getUrl()));
-                    intent.setPackage(Browsers.KnownBrowser.FIREFOX.packageName);
+                    intent.setPackage(browsers.getFirefoxBrandedBrowser().packageName);
                     startActivity(intent);
                 } else {
                     final Intent intent = new Intent(Intent.ACTION_VIEW,
-                            Uri.parse("market://details?id=" + Browsers.KnownBrowser.FIREFOX));
+                            Uri.parse("market://details?id=" + Browsers.KnownBrowser.FIREFOX.packageName));
                     startActivity(intent);
                 }
 
--- a/app/src/main/java/org/mozilla/focus/menu/BrowserMenuAdapter.java
+++ b/app/src/main/java/org/mozilla/focus/menu/BrowserMenuAdapter.java
@@ -52,8 +52,14 @@ public class BrowserMenuAdapter extends RecyclerView.Adapter<BrowserMenuViewHold
 
         items.add(new MenuItem(R.id.share, resources.getString(R.string.menu_share)));
 
-        items.add(new MenuItem(R.id.open_firefox, resources.getString(
-                R.string.menu_open_with_default_browser, "Firefox")));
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -286,7 +286,57 @@ public void onClick(View view) {
...
                     final Intent intent = new Intent(Intent.ACTION_VIEW,
-                            Uri.parse("market://details?id=" + Browsers.KnownBrowser.FIREFOX));
+                            Uri.parse("market://details?id=" + Browsers.KnownBrowser.FIREFOX.packageName));
+                    startActivity(intent);
+                }
+
+                break;
+            }
+
+            case R.id.open_firefox_beta: {
```
pocmo: This here and in `BrowserMenuAdapter` is a lot of duplicated code. Can we hide this functionality in the `Browsers` class? We might just need a `ActivityInfo getFirefoxBrandedBrowser()` method which we can use to determine whether we want to show a menu item and also for opening the browser. Almost like the default browser item, but just for firefox branded browsers.

sabbir044: Okay, got it.

sabbir044:
Pushed the change request.

app/src/main/java/org/mozilla/focus/menu/BrowserMenuAdapter.java
```diff
@@ -52,8 +52,14 @@ private void initializeMenu(String url) {
 
         items.add(new MenuItem(R.id.share, resources.getString(R.string.menu_share)));
 
-        items.add(new MenuItem(R.id.open_firefox, resources.getString(
-                R.string.menu_open_with_default_browser, "Firefox")));
+        if(browsers.hasFirefoxBrandedBrowserInstalled()) {
```
pocmo: nit: space after if.

app/src/main/java/org/mozilla/focus/utils/Browsers.java
```diff
@@ -75,6 +79,20 @@ public Browsers(Context context, String url) {
 
         this.browsers = browsers;
         this.defaultBrowser = findDefault(packageManager, url);
+        this.firefoxBrandedBrowser = findFirefoxBrandedBrowser();
+    }
+
+    private ActivityInfo findFirefoxBrandedBrowser() {
+        if(browsers.containsKey(KnownBrowser.FIREFOX.packageName)) {
```
pocmo: nit: space after "if" here and for the "else if" cases.

pocmo:
Thanks!

PR #407, opened 2017-04-04
Ahunt/405

ahunt:
This reverts the original "fix", and also adds a check to make sure we never hang. I'm not 100% convinced by my anti-hang-detection, feedback welcome : )


app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
```diff
@@ -34,6 +34,12 @@
 
     private List<SearchEngine> searchEngines;
 
+    /**
+     * A flag indicating that data has been loaded, or is loading. This lets us detect if data
+     * has been requested without a preceeding init().
+     */
+    private volatile boolean loadHasBeenTriggered = false;
```
pocmo: This is only read or written from synchronized blocks - so `volatile` isn't really needed. Inside synchronized blocks you'll never see stalled data.

app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
```diff
@@ -157,6 +164,10 @@ public synchronized SearchEngine getDefaultSearchEngine(Context context) {
     }
 
     public void awaitLoadingSearchEnginesLocked() {
+        if (!loadHasBeenTriggered) {
+            throw new IllegalStateException("Attempting to retrieve search engines without a corresponding init()");
```
pocmo: We could trigger a load here - but I like your approach more. Throwing will uncover logical errors we made.

PR #410, opened 2017-04-05
Resize BrowserFragment when keyboard appears

ahunt:


app/src/main/java/org/mozilla/focus/utils/KeyboardUtil.java
```diff
@@ -0,0 +1,103 @@
...
+import android.view.ViewTreeObserver;
+import android.view.inputmethod.InputMethodManager;
+
+/**
+ * Created by mikepenz on 14.03.15.
+ * This class implements a hack to change the layout padding on bottom if the keyboard is shown
+ * to allow long lists with editTextViews
+ * Basic idea for this solution found here: http://stackoverflow.com/a/9108219/325479
+ */
+public class KeyboardUtil {
```
pocmo: This class contains a bunch of things we do not use. I'd only pick the bits we need and fold it into BrowserFragment.

app/src/main/java/org/mozilla/focus/utils/KeyboardUtil.java
```diff
@@ -0,0 +1,103 @@
...
+    private View decorView;
+    private View contentView;
+
+    public KeyboardUtil(Activity act, View contentView) {
+        this.decorView = act.getWindow().getDecorView();
+        this.contentView = contentView;
+
+        //only required on newer android versions. it was working on API level 19
+        if (Build.VERSION.SDK_INT >= 19) {
+            decorView.getViewTreeObserver().addOnGlobalLayoutListener(onGlobalLayoutListener);
```
pocmo: We register a global layout listener here but we never unregister. Because of that we might keep references to the views and activity.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -116,6 +117,8 @@ public View inflateLayout(LayoutInflater inflater, @Nullable ViewGroup container
         menuView = view.findViewById(R.id.menu);
         menuView.setOnClickListener(this);
 
+        new KeyboardUtil(getActivity(), view);
```
pocmo: This is a super weird API. We create an object that isn't even needed. It looks like this should have been a static helper method.

PR #434, opened 2017-04-11
Add disconnect list integration tests

ahunt:


ahunt:
I decided to go for the option of "Maybe it's okay for the test to fail locally if the adjust ID is not set.". I'm guessing most people don't run the full test suite locally anyway - maybe I should look into a "workaround" of setting an appropriate adjust env variable for local builds if needed?

PR #435, opened 2017-04-11
Reenable proguard (with ACRA workaround)

ahunt:


ahunt:
(Note: I've enabled SDK integration for this branch, i.e. I've been able to test against the same configuration that was causing failures on master)

PR #438, opened 2017-04-12
Infer fixup (mostly resource leakage)

ahunt:


app/src/main/java/org/mozilla/focus/autocomplete/UrlAutoCompleteFilter.java
```diff
@@ -77,6 +77,12 @@ private void loadUrls(Context context) {
                 } catch (IOException e) {
                     // No autocomplete for you!
                     return null;
+                } finally {
+                    try {
+                        reader.close();
+                    } catch (IOException e) {
+                        // There's really nothing we can do now.
+                    }
```
pocmo: Reader implements Closable. So you should be able to use the helper in IOUtils (here and in the other cases). I wonder if our API level is high enough to actually use try-with-resources? That would be sweet.

PR #451, opened 2017-04-12
Migrate to try-with-resource

ahunt:
I forgot to do this before landing the infer fixes. try-with-resource seems to make our code a bit cleaner!


ahunt:
(I did file an equivalent bug for Fennec, but we can't upgrade until minSdk=19/Android 4.4)

Issue #322, opened 2017-03-22
Visited links not cleared after erase


thebnich:
STR:
1. Do a search for "foo".
2. Tap any link.
3. Hit erase.
4. Repeat the search for "foo".

The link is purple. We should be clearing all user data after each erase.

ahunt:
Huh, I've noticed that restarting the app does succesfully clear the link colouring data. (Maybe they're kept in memory, maybe something else is happening.)

ahunt:
Even if I manually wipe _all_ on-disk data while focus is running (except for shared_prefs), the link colouring still happens (until I restart the app).

(adb shell, run-as org.mozilla.focus.debug, rm -rf everything except shared_prefs)

It's possible that WebView's destroy() fixes this, however we currently only do that when the WebFragment gets destroyed (and I don't know if we can force fragment destruction).

ahunt:
destroy() doesn't help either. (It gets run if you press back to exit from the browser, instead of using the erase button.)

(Unassigning for now - I'm not too sure what other options we can try, short of force-restarting the app when erase'ing().)

ahunt:
I've done some more digging (TL;DRL version is: no progress : / ):

The only java way in to the visited links storage seems to be via WebChromeClient.getVisitedHistory(), which is called every time a new WebView is created.

However that only adds to the visited links list, which is stored in C++ land:
https://chromium.googlesource.com/chromium/src.git/+/d0ef9df6be5983f6df7e4e050bbad4eb5030e7a2/android_webview/browser/aw_browser_context.cc#140
and actual storage:
https://github.com/adobe/chromium/blob/master/chrome/browser/visitedlink/visitedlink_master.cc

We'd somehow need to call VisitedLinkMaster::DeleteAllURLs(), and I haven't found any paths that let us trigger that from Java.

The clearHistory() call ends up somewhere completely different:
https://github.com/cvsuser-chromium/chromium/blob/master/content/browser/android/content_view_core_impl.cc#L1194
which calls into the navigation history for a given view - i.e. doesn't help with clearing the in memory visited links list.

ahunt:
As mentioned on IRC:
- visited_links is stored in chromium in C++, it's part of the BrowserContext
- BrowserContext is a singleton, stored as a C++ only global variable
- The only way of resetting visited_links at this time is to kill the entire process (there's no way to unload native code in Android it seems, and WebView doesn't offer any unloading mechanism)

[...]

PR #471, opened 2017-04-13
Use JS to clear :visited link's CSS styling (#322)

ahunt:
We really should get some performance numbers for this.


app/src/webkit/java/org/mozilla/focus/web/FocusWebViewClient.java
```diff
@@ -43,6 +43,49 @@ public void setCallback(IWebView.Callback callback) {
...
+            // moving in the other direction we'd need to remember to process a given index
+            // again which is more complicated).
+            "  for (i = nRules - 1; i >= 0; i--) {" +
+            "    cssRule = stylesheet.cssRules[i];" +
+            // Depending on style type, there might be no selector
+            "    if (cssRule.selectorText && cssRule.selectorText.includes(':visited')) {" +
+            "      stylesheet.deleteRule(i);" +
+            "    }" +
+            "  }" +
+            "}";
```
pocmo: When executing this in the global scope of the website we should try to avoid potentially breaking the site's JS.

How about wrapping this in:

```
(function() {
 ...
})();
```

This will generate an anonymous function and execute it immediately. In addition to that we should use `let` to avoid overriding global variables.

app/src/webkit/java/org/mozilla/focus/web/FocusWebViewClient.java
```diff
@@ -43,6 +43,49 @@ public void setCallback(IWebView.Callback callback) {
...
+                // a <link>'d css stylesheet is loaded:
+                "links = document.getElementsByTagName('link');" +
+                "for (i = 0; i < links.length; i++) {" +
+                "  link = links[i];" +
+                "  if (link.rel == 'stylesheet') {" +
+                "    link.addEventListener('load', cleanupVisited, false);" +
+                "  }" +
+                "}"
+                , null);
+
```
pocmo: Same here: Can we wrap this to avoid overriding other scripts?

Issue #478, opened 2017-04-19
Block favicon.ico


ahunt:
As suggested by @pocmo on IRC: webview always requests favicon.ico (I wonder if it doesn't do that when a page defines its own icon in html?), that icon isn't actually used anywhere, so we might as well block it.

We need to make sure nothing odd breaks by doing this. Hopefully sites aren't using /favicon.ico in page content, in which case we can just block any URL consisting of $current_domain/favicon.ico.

ahunt:
To reply to myself: if the site defines their own `<link rel="shortcut icon"`, then webview loads that instead of $current_domain/favicon.ico. I suspect that's the case for most websites - if that link points to something ending in favicon.ico, I guess we could still easily block it - but otherwise we don't really have any way of determining whether a given resource is an unneeded favicon.

PR #480, opened 2017-04-19
Block unneeded /favicon.ico (#478)

ahunt:


pocmo:
WebViewDataTest might need to be adapted. I guess it will fail without this request now.

ahunt:
Update the test - buddybuild seems happy (I specifically enable UI tests for my branch), hopefully this can land!

PR #481, opened 2017-04-19
Migrate UrlMatcher.matches() to use Uri

ahunt:
We used to convert the Uri to a String, only to parse it as URL
in order to call getHost() on it. We can instead just pass the original
Uri around.

(This is a historical remainder from when we used to support the older
 shoudlInterceptRequest() version, which would only give us the resource
 URL string, hence UrlMatcher was designed around handling a URL string.)


pocmo:
> ahunt wants to merge 1 commit into ahunt/478-favicon from ahunt/matcherURI 

Is this correct? I guess you want to merge into master?

ahunt:
> Is this correct? I guess you want to merge into master?

Hmm. I did want to merge into master, but I was basing this patch off the patches in my favicon branch (since they both touch the URL matching path, and would otherwise conflict). I don't think I did the right thing with the github UI though.

ahunt:
I think I've managed to fix things - in future I'll probably just wait until PR 1 has landed, or have 1 PR with all patches that might depend on each other.

PR #482, opened 2017-04-19
[BLOCKED] SearchEngineManager: make variables volatile

ahunt:
Both of these are accessed across threads (i.e they will be written
to from a background thread during loadFromDisk(), and we're reading
them from a different thread in awaitLoadingSearchEnginesLocked()).
In order to avoid a race condition we should make these volatile.

(Inspired by findbugs complaining about lack of synchronisation for these values)


pocmo:
This seems to be a false positive. All reads of `loadHasBeenTriggered` and `searchEngines` are synchronized. awaitLoadingSearchEnginesLocked() is only called from synchronized methods. For some reason awaitLoadingSearchEnginesLocked() is public though (but doesn't need to be). Does findbugs still complain if it's private?

PR #493, opened 2017-04-20
Hide FAB when keyboard is showing (#485)

ahunt:
I don't know if this is over-the-top architecture wise. We could probably just findViewById in the ResizeableKeyboardLayout.


pocmo:
> I don't know if this is over-the-top architecture wise. We could probably just findViewById in the ResizeableKeyboardLayout.

Yeah. I think for us it's okay to build customized components that are not re-usable. :)

PR #508, opened 2017-04-21
Attempt at adding code coverage information in Github

npark-mozilla:


npark-mozilla:
@pocmo so for BB, how can i add extra gradle command parameter to execute coverall?  I need to do something like: ./gradlew --info clean assembleDebug lint testFocusWebkitDebugUnitTest jacocoTestFocusWebkitDebugUnitTestReport coveralls

npark-mozilla:
Let's see how this goes in master branch

Issue #506, opened 2017-04-21
Add "search" action into text selection menu of third-party apps


pocmo:
**Theme**: _Default Browser & Android Advantage_

**User Story**

> I want to quickly start a private Focus search from my homescreen

**Feature**

> Include action in text selection menu.

**Example**

![1 d4zzzplbtk5cen9qn0-cba](https://cloud.githubusercontent.com/assets/89638/25283853/78360ff0-26b5-11e7-9396-141559bffe64.gif)

**Technical details**
https://medium.com/google-developers/custom-text-selection-actions-with-action-process-text-191f792d2999

**Tasks**
- [x] Implement text selection action receiver (ACTION_PROCESS_TEXT)
- [x] Trigger search from action
- [x] Update copy: "Search" / "Private Search" / "Focus Search" / ... ?
- [x] Add telemetry probe

antlam:
If it's just adding a "SEARCH" item to the context menu/action bar here that triggers Firefox Focus, then we can just jump right into loading a URL in Firefox Focus after they press "SEARCH"

pocmo:
In Firefox we add the product name: "Firefox search", "Nightly search", ... Should we do this here too? Focus search? Or maybe even Private search?

antlam:
Maybe! but I worry that would seem out of guidelines for these context menu/action bars... 

From my understanding, this SEARCH action is usually done by the default browser. So if Firefox Focus is my default, it seems unnecessary to call that out again in the menu. 

Thoughts?

pocmo:
The action will show up always - independently from whether we are the default browser or not. For example in my case I see search entries for all my Firefoxes and Focus.

pocmo:
> Thoughts?

Users might want to be sure that the action is launching their privacy browser that doesn't keep any history etc. - Launching the search in a different browser might mean that this search will be connected with your profile.

antlam:
> The action will show up always

Ah, my mistaken. Then this makes more sense if we label it explicitly.

brampitoyo:
Some copy ideas other than the ones mentioned above:

* Search with Firefox Focus (too long?)
* Search with Focus

Otherwise, I like “Private Search”. Even though it’s less unique because it doesn’t contain our name, it communicates exactly what user expects: “When I tap on this, my search will not be recorded into history”.

antlam:
@pocmo and I were also talking about "Focus Search" because of the system conventions (Google Search, Yahoo Search, etc)... 

[...]

PR #509, opened 2017-04-24
Add text selection action for launching private search (from third-party apps). (#506)

pocmo:
(Waiting for UX feedback regarding string to use)


pocmo:
I'll keep track of the string in #506.

PR #518, opened 2017-04-26
Findbugs

ahunt:
It looks like we don't run findbugs in automation. Let's make sure we do : )


ahunt:
> As long as this doesn't run for every local build in such a way that it slows down developing... :)

This only runs if you do a full "gradle build" - it doesn't affect normal builds in AS!

Issue #522, opened 2017-04-26
Crash when clicking erase notification on Android 6


ahunt:
STR:
1. Open Focus on Android 6
2. Start browsing
3. Tap the erase notification

Expected: data is erased
Actual: focus crashes

java.lang.RuntimeException: Unable to start service org.mozilla.focus.notification.BrowsingNotificationService@facb2d5 with Intent { act=erase cmp=org.mozilla.focus.debug/org.mozilla.focus.notification.BrowsingNotificationService }: android.util.AndroidRuntimeException: Calling startActivity() from outside of an Activity  context requires the FLAG_ACTIVITY_NEW_TASK flag. Is this really what you want?
                                                                             at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:3027)
                                                                             at android.app.ActivityThread.-wrap17(ActivityThread.java)
                                                                             at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1442)
                                                                             at android.os.Handler.dispatchMessage(Handler.java:102)
                                                                             at android.os.Looper.loop(Looper.java:148)
                                                                             at android.app.ActivityThread.main(ActivityThread.java:5417)
                                                                             at java.lang.reflect.Method.invoke(Native Method)
                                                                             at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
                                                                             at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
                                                                          Caused by: android.util.AndroidRuntimeException: Calling startActivity() from outside of an Activity  context requires the FLAG_ACTIVITY_NEW_TASK flag. Is this really what you want?
                                                                             at android.app.ContextImpl.startActivity(ContextImpl.java:672)
                                                                             at android.app.ContextImpl.startActivity(ContextImpl.java:659)
                                                                             at android.content.ContextWrapper.startActivity(ContextWrapper.java:331)
                                                                             at org.mozilla.focus.notification.BrowsingNotificationService.onStartCommand(BrowsingNotificationService.java:102)
                                                                             at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:3010)
                                                                             at android.app.ActivityThread.-wrap17(ActivityThread.java) 
                                                                             at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1442) 
                                                                             at android.os.Handler.dispatchMessage(Handler.java:102) 
[...]

PR #523, opened 2017-04-26
Don't crash when erasing via notification on Android 6 (#522)

ahunt:
I haven't tested Android 5, or 7.0 yet.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/notification/BrowsingNotificationService.java
+++ b/app/src/main/java/org/mozilla/focus/notification/BrowsingNotificationService.java
@@ -97,6 +97,10 @@ public class BrowsingNotificationService extends Service {
                     activityIntent.setFlags(Intent.FLAG_ACTIVITY_NO_ANIMATION);
                 }
 
+                // This doesn't seem to be needed on Android 7.1. It is needed on Android 6, or otherwise
+                // we crash with "AndroidRuntimeException: Calling startActivity() from outside of an Activity context requires the FLAG_ACTIVITY_NEW_TASK flag."
+                activityIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/notification/BrowsingNotificationService.java
```diff
@@ -97,6 +97,10 @@ public int onStartCommand(Intent intent, int flags, int startId) {
                     activityIntent.setFlags(Intent.FLAG_ACTIVITY_NO_ANIMATION);
                 }
 
+                // This doesn't seem to be needed on Android 7.1. It is needed on Android 6, or otherwise
+                // we crash with "AndroidRuntimeException: Calling startActivity() from outside of an Activity  context requires the FLAG_ACTIVITY_NEW_TASK flag."
```
pocmo: nit: two spaces after "Activity" and a lonely quotation mark at the end

app/src/main/java/org/mozilla/focus/notification/BrowsingNotificationService.java
```diff
@@ -97,6 +97,10 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
+                activityIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
```
pocmo: Technically the flag isn't needed as our activity is singleTask anyways. So maybe the check is now less strict on Android 7+.

app/src/main/java/org/mozilla/focus/notification/BrowsingNotificationService.java
ahunt: Heh, the double spaces came from Android and I never noticed while copying and pasting : )

PR #536, opened 2017-04-28
Actually run findbugs in buddybuild

ahunt:
Issue 1: buddybuild only seems to run assemble, and tests. We therefore need to call findbugs ourselves.

Issue 2: buddybuild removes our flavour disabling variantFilter, so if you run 'gradle findbugs' it will try to build focus/klarGeckoRelease. That doesn't build because we only supply AdjustHelper for focusWebkitRelease/klarWebkitRelease (in addition to debug/beta builds). It seemed simplest to just make focusGeckoRelease buildable, however the gradle code for that is a bit annoying (gradle doesn't detect focusRelease/klarRelease, so we have to add that support ourselves).

This all seems quite complicated, alternative suggestions are welcome : ).


ahunt:
Note also: this branch was based off a master that had findbugs issues (see #533 ), after rebasing it will *hopefully* build correctly (I've just triggered an update from fixed master).

PR #544, opened 2017-05-01
Only use FLAG_ACTIVITY_NEW_TASK for Android <= 6 (#522)

ahunt:
7.0 is the first Android version that seems to no longer need this,
as verified by testing on devices and emulators.


pocmo:
Overall it doesn't hurt to be safe and always add the flag? :)

ahunt:
Yeah, no harm in the flag - I guess we can always remove it if/when we don't need to suport Android < 7.

PR #545, opened 2017-05-02
Disable findbugs in buddybuild for now

ahunt:
Running gradle breaks the build, but only in buddybuild with the
sdk integration enabled. It seems buddybuild injects its SDK calls
into FocusApplication.java, however gradle doesn't actually know
about that dependency during postbuild. I have no idea how the main
build accesses it, this will need more investigation before reenabling.


ahunt:
I'd recommend landing this ASAP to unblock building master : )

(I didn't see the issues because we only enable the SDK on master and not on other CI builds, I've installed the SDK onto ahunt/findbugs-bb2 where I'm looking into how we can fix this.)

pocmo:
Yeah, this SDK magic is quite annoying at times. Back when I enabled it I was looking for a standalone version that I could integrate myself (like you can for iOS), but I couldn't find one.

Issue #504, opened 2017-04-21
Create localized domain completion lists


pocmo:
**User Story**

> I want to quickly and easily browse to the most popular websites, and not have to type them in manually.

**Feature**

> A better (localized) domain completion list. It is very US centric now.

**See**

* http://gecko.readthedocs.io/en/latest/mobile/android/fennec/defaultdomains.html

**Details**

> We could start by adding domain completion lists for major locales

- [x] Define scope: What locales/languages/countries do we want to support?
- [x] Create lists (Based on Alexa Top 500? Filter inappropriate content)
- [x] Use different lists based on configuration (like we do for search engines)
- [x] Merge multiple lists on Android 7+ (`LocaleList.getDefault()`)

bbinto:
Here is the list of our top tier countries, I chose Top 50 (only) as it needed a lot of manual sanitization, and I believe we will cover most of the user's favourite websites. 

I don't think we should replace the existing top 500 global list with the top 50 for each locale, so I was wondering if we should merge them with the locale-specific ones, and remove duplicates? Should we do this in the code or should I provide the updated lists as Excel sheets?

https://docs.google.com/a/mozilla.com/spreadsheets/d/1A39cpy87arrLr8LBd_PL-FKRrgazMoGdKjqEf5b5iOo/edit?usp=sharing

pocmo:
We can keep the lists separate and do the merging in code. :)

pocmo:
It's now in the latest Beta builds.

We prefer the user's locale over the global list and if the user is using Android 7+ then we support multilocale too.

E.g. my phone is set to: en_US, de_DE - When typing we look first in the "us" list, then "de" list then global list.

PR #562, opened 2017-05-03
Use localized domain completion lists. (#504)

pocmo:


ahunt:
It looks like findbugs found some issues : )

Issue #500, opened 2017-04-21
Support "Custom Tabs" UI customizations


pocmo:
**Theme:** _Default Browser & Android Advantage_

**User Story**

> While using shopping apps or reading shopping email promotions, I want to have peace of mind to not be tracked for any link I click.

**Details**

> Discuss UI for custom tabs, can we leverage current 3 dot menu from Focus chrome?

**Technical details**

* https://developer.chrome.com/multidevice/android/customtabs

**Tasks**

- [x] Define scope: What UI customization features do we want to support (in v1)? How do we implement them without completely giving up the Focus branding?
- [x] Customization: Support toolbar color
- [x] Customization: Support action button
- [x] Customization: Support custom menu items
- [x] Hide erase button (but erase when going back to previous app via up/back button)
- [x] Show "X" up button
- [x] Add telemetry probes
- [x] "Maybe we could extend the FloatingActionButton class to permanently hide the button in some states."
- [x] Tests

**Not in scope**
- Pre-warming
- "May open URL"
- Up button customization
- Custom in/out animations
- Bottom toolbar
- "Up button" customization

ahunt:
Another UX question: *if* we add a bottom toolbar, do we want to make it disappear, when scrolling, similar to the FAB? (Chrome usually does that, although it was disabled for some time in Chrome Dev.) I'm guessing that's what app devs will expect, so we probably should do this.

pocmo:
I'm pretty certain that we are not going to support the bottom toolbar. It's not used much and we already decided to not implement it in the first iteration in Fennec.

ahunt:
Further TODOs:

- [ ] Import SafeIntent (or find alternative), and use that elsewhere in Focus too
- [ ] Check close icon size (as supplied by external app)
- [ ] And close icon padding
- [ ] Consider long-press to copy URL (similar to chrome - what does Firefox do?)
- [ ] Tests for intent handling

ahunt:
I'm still wondering what telemetry probes we want, here's what I've though of sofar:

- [ ] Custom tab opened (vs normal URL opened)
- [ ] Custom tab specific features used, i.e. included in the intent (action button / menu item / bottom toolbar etc)
- [ ] Custom tab feature selected by user (exited using back button vs up button vs using CT menu item/actionbutton/etc)

ahunt:
[...]

PR #565, opened 2017-05-05
Custom Tabs - part 1 (exit/erase/config/urlbar colour)

ahunt:
I still need to dig into the (fragment) lifecycle to make sure we're not doing anything stupid / breaking switching between CT and normal browsing.


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -136,14 +141,85 @@ public View inflateLayout(LayoutInflater inflater, @Nullable ViewGroup container
...
+
+        // Unfortunately there's no simpler way to have the FAB only in normal-browser mode.
+        // - ViewStub: requires splitting attributes for the FAB between the ViewStub, and actual FAB layout file.
+        //             Moreover, the layout behaviour just doesn't work unless you set it programatically.
+        // - View.GONE: doesn't work because the layout-behaviour makes the FAB visible again when scrolling.
+        // - Adding at runtime: works, but then we need to use a separate layout file (and you need
+        //   to set some attributes programatically, same as ViewStub).
+        final View erase = view.findViewById(R.id.erase);
+        final ViewGroup eraseContainer = (ViewGroup) erase.getParent();
+        eraseContainer.removeView(erase);
```
pocmo: Maybe we could extend the FloatingActionButton class to permanently hide the button in some states.

app/src/main/res/layout/fragment_browser.xml
```diff
@@ -57,6 +57,7 @@
 
         <android.support.design.widget.FloatingActionButton
             android:id="@+id/erase"
+            android:visibility="gone"
```
pocmo: Does this need to be "gone" by default now?

ahunt:
I've added "Maybe we could extend the FloatingActionButton class to permanently hide the button in some states." as an item to the custom tab bug ( #500 )

PR #566, opened 2017-05-05
fix small UI test issues

npark-mozilla:


npark-mozilla:
Quick fix for UI test failure in BB due to locator change / sumo issue.  The crash I see on API 22 and 24 seems to be bb related - I can't repro it on both sim/device of API 22, and from the logcat they provide, it seems that the device is slow to kill the test process upon test completion, and it actually kills focus when it's in middle of the next test.

Issue #579, opened 2017-05-11
Use different app names for telemetry


pocmo:
* Focus
* Klar

PR #581, opened 2017-05-11
Use different app name (Focus/Klar) for telemetry. (#579)

pocmo:


app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
```diff
@@ -34,7 +34,8 @@
 import org.mozilla.telemetry.storage.TelemetryStorage;
 
 public final class TelemetryWrapper {
-    private static final String TELEMETRY_APP_NAME = "Focus";
+    private static final String TELEMETRY_APP_NAME_FOCUS = "Focus";
```
ahunt: Would we want to reuse the UA name (R.string.useragent_appname) here?

pocmo: We have a dependency on the server side processing here - So I guess it's safer to keep them separate.

Issue #498, opened 2017-04-21
Make number of blocked trackers visible


pocmo:
**Theme:** _General Improvement_

**User Story**

> As a user I want to see how many trackers have been blocked on a page, so I feel safer knowing what the product does

**Feature**

* Shows the number of ads/trackers blocked
* By session, page, total

**Tasks**
- [x] Define scope (What do we count exactly? Do we count per session / webpage?)
- [x] UX mocks / flow
- [x] Count trackers and persist count
- [x] Add UI for displaying counts to user

ahunt:
As an extension on the scope part, what do we count as a tracker?
- single URL
- single full domain (foobar.imatracker.com)
- single blocked/base domain (anything from imatracker.com)
- single owner (anything from TrackerCorp including trackercorp.com, trackercorp.ca, etc)

The lower you go in the above list, the harder the implementation gets. The last option would require significant rework of our blocklist code (and I would argue is infeasible to do well since e.g. google domains are split across ads/tracking/social lists, so we'd have to reunify that - which would be messy to implement).

bbinto:
![](https://github.trello.services/images/mini-trello-icon.png) [Shows the number of ads/trackers blocked](https://trello.com/c/VaSzbZbJ/15-shows-the-number-of-ads-trackers-blocked)

PR #582, opened 2017-05-11
Count blocked trackers and show in menu. (#498)

pocmo:


app/src/main/res/layout/menu_blocking_switch.xml
```diff
@@ -2,17 +2,56 @@
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<Switch xmlns:android="http://schemas.android.com/apk/res/android"
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
ahunt: The current menu item lets you click anywhere (ie. on the text instead of the switch), and it still toggles. IIUC the new version will only switch content blocking if you click on the switch, is that desired (I haven't tested it yet, so I'm not sure if the menu will just hide, without the content blocking setting changing, or if the menu will just stay open in this scenario?).

pocmo: The menu will stay open. There are some ideas floating around to make the other part clickable and show some more information about what trackers have been blocked or something like that.

PR #585, opened 2017-05-11
Fix some low priority findbugs issues

ahunt:
These warnings are increasingly non-useful, I wonder if we should just not bother with findbugs' low priority warnings.


pocmo:
> These warnings are increasingly non-useful, I wonder if we should just not bother with findbugs' low priority warnings.

Yeah, those annotations add a lot of noise and we do not really gain much.

ahunt:
Yeah, I think I'll scrap the pure annotation patches in this PR (the unused variables / confusing method name still seem worthwhile so I'll land those).

PR #611, opened 2017-05-18
Custom Tabs: the rest (hopefully)

ahunt:


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -222,7 +222,7 @@ private void showBrowserScreen(String url) {
         if (intent.getBooleanExtra(EXTRA_TEXT_SELECTION, false)) {
             TelemetryWrapper.textSelectionIntentEvent();
         } else if (BrowsingSession.getInstance().isCustomTab()) {
-            TelemetryWrapper.customTabIntentEvent();
+            // Telemetry is sent while parsing the custom tab config
```
pocmo: Why? It's a bit unusual to have a parser sent UI events?

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -576,8 +579,10 @@ public void onClick(View view) {
             }
 
             case R.id.customtab_close: {
-                erase();
```
pocmo: No erase anymore?

ahunt: I filed #616 to clarify what we want to do here, since I can't find any documented decisions either way. It seemed weird to have different behaviour for (system) back button vs close button, but that's probably for UX to decide!

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
ahunt: I've rewritten the telemetry to be sent separately (we still need to parse unsupported features during parseCustomTabIntent, but they're store din a list until whenever we actually send telemetry.) Rereview requested for the last 2 patches (one adds telemetry for supported features, one adds the unsupported features) : )!

PR #613, opened 2017-05-19
Npark mozilla/image test

npark-mozilla:
Create a UI Test that taps images on webpage


app/src/androidTest/java/org/mozilla/focus/activity/ImageSelectTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+        // Erase browsing session
+        TestHelper.floatingEraseButton.perform(click());
+        TestHelper.erasedMsg.waitForExists(waitingTime);
+        Assert.assertTrue(TestHelper.erasedMsg.exists());
+        Assert.assertTrue(TestHelper.urlBar.exists());
+        TestHelper.waitForIdle();
+    }
+
+    private String readTestAsset(String filename) throws IOException {
```
ahunt: Is this intentionally duplicated (readTestAsset/readStreamIntoString) - can we just use the TestHelper version instead (they'd need to be made public I guess)?

npark-mozilla:
@ahunt oh yes this is work in progress, sebastian's code could only handle text files, so I took it and changed it to handle binaries as well.  trying to write one more test that copies image URLs

npark-mozilla:
@ahunt it says "FindBugs rule violations were found. See the report at: file:///tmp/sandbox/workspace/app/build/reports/findbugs/findbugs.html" as a build error. how can I see this report so I can fix my build?

PR #650, opened 2017-05-25
Fix crashes and inconsistencies in share/open with code

ahunt:


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -612,10 +597,13 @@ private void updateToolbarButtonStates() {
         stopButton.setVisibility(isLoading ? View.VISIBLE : View.GONE);
     }
 
-    @Nullable
+    @NonNull
```
pocmo: Here you define @NonNull but above in onClick() you do a null check? :)

PR #653, opened 2017-05-25
List of 50 most visited brazil sites according to Alexa

danilodorgam:
I noticed that you did not have domains .br


pocmo:
Thank your for your pull request! 👍 

Unfortunately I won't be able to merge it: We are following a strict process for sanitizing and filtering lists of domains that we ship. However I pinged our product team and they will look into creating a list for Brazil that we can integrate and ship with our release.

pocmo:
A list has been merged in #659 :)

Issue #656, opened 2017-05-26
Pre-process strings.xml to replace app name?


pocmo:
I'm currently replacing some strings with hard-coded app name (Focus / Firefox Focus) to use placeholders instead. What's annoying is that I now have to find views using the strings and assign them manually after replacing the placeholder with R.string.app_name or R.string.launcher_name. This is actually nicer in Fennec where we fill an entity with the app name and then use that entity in strings.xml. Unfortunately I'm not able to create an entity that is different for each build configuration with just pure gradle. Maybe some pre-processing step would be helpful here.

We have a similar problem with shortcuts.xml - We need to hard-code the application ID in the XML file - but the application id is different for each build configuration. Here I replace {applicationId} via gradle manually.

PR #657, opened 2017-05-26
Stealth

pocmo:


app/src/main/java/org/mozilla/focus/firstrun/FirstrunPagerAdapter.java
```diff
@@ -45,6 +46,13 @@ private View getView(int position, ViewPager pager) {
             finishView.setOnClickListener(listener);
         }
 
+        // Special casing page 2 where we need to replace the app name. That's quite shitty and annoying to do
```
pocmo: Filed #656 to find a better solution for things like this.

pocmo:
@ahunt I merged this without review to get the strings in for the export today. Let me know if there's something in here that we should change.

PR #664, opened 2017-05-26
Strict mode - disable for beta, print logs in dev

ahunt:


app/src/main/java/org/mozilla/focus/FocusApplication.java
```diff
@@ -31,17 +31,21 @@ public void onCreate() {
...
+        // Android/WebView sometimes commit strict mode violations, see e.g.
+        // https://github.com/mozilla-mobile/focus-android/issues/660
+        if (AppConstants.isReleaseBuild() || AppConstants.isBetaBuild()) {
             return;
         }
 
         StrictMode.setThreadPolicy(new StrictMode.ThreadPolicy.Builder()
                 .detectAll()
+                .penaltyLog()
                 .penaltyDeath()
```
pocmo: penaltyLog() and penalyDeath() at the same time doesn't really make sense, I think.

With just logging it's easy to miss things - I think I would prefer false positive crashes (in Dev) over missing penalties because they are only logged.
At least for ThreadPolicy we could use penaltyDialog() - maybe even on Beta.

How about this:
* Thread Policy
  * Beta: Dialog
  * Debug: Dialog
* VmPolicy
  * Beta: Log
  * Debug: Death

ahunt: I added the penaltyLog() mainly to enable logs (penaltyDeath() by itself doesn't print a stack trace, at least on the devices I've been using, but penaltyDeath+penaltyLog means the app dies and there's a visible trace to find what the violation was). (I'll update my patches with the new policies!)

PR #667, opened 2017-05-26
disable onboarding screen and create separate test

npark-mozilla:
separate html for separate tests that use mockwebserver
add OnBoardingTest

### DIFF
--- a/app/src/androidTest/assets/test.html
+++ b/app/src/androidTest/assets/test.html
@@ -9,8 +9,6 @@
 <h1>focus test page</h1>
 <!-- More random words that should not end up on disk -->
 <p>groovy rabbits</p>
-<img id="rabbitImage" src="rabbit.jpg" alt="Smiley face" width="251" height="201">
-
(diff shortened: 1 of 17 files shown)


app/src/androidTest/assets/test.html
```diff
@@ -9,8 +9,6 @@
 <h1>focus test page</h1>
 <!-- More random words that should not end up on disk -->
 <p>groovy rabbits</p>
-<img id="rabbitImage" src="rabbit.jpg" alt="Smiley face" width="251" height="201">
```
pocmo: Hey, why did you remove that? It's used in my test to make sure the image does not remain on disk.

npark-mozilla: Maybe I'll change the html so the image is loaded after the script, and see whether that solves the issue. I'm not sure why the script loading was keep being interrupted with the image.  I don't think mockwebserver is designed to 'host' a normal website with many components, but I could be using it incorrectly :/

PR #671, opened 2017-05-29
Tracker counter: Show "-" instead of 0 if content blocking is disable…

pocmo:
…d. (#699)


app/src/main/java/org/mozilla/focus/menu/BlockingItemViewHolder.java
```diff
@@ -49,6 +53,15 @@ public void run() {
         });
     }
 
+    private void disableTrackingCount(final TextView view) {
+        ThreadUtils.postToMainThread(new Runnable() {
+            @Override
+            public void run() {
+                view.setText("-");
```
ahunt: I wonder if we'd ever want to show something different in certain locales - but I guess we'll get feedback if that's needed!

pocmo: Good point. I'll ask flod whether he has any concerns.

PR #711, opened 2017-06-06
flake8 improvements for android2po

mastizada:
All warnings solved.

NOTE: added requirements.txt file with requirements from source project, but with newer version of lxml, please, check if everything works. If it will crash due to lxml, please, write output and information how to represent that error.


mastizada:
This one is ready

pocmo:
It might make sense to upstream those changes:

PR #730, opened 2017-06-08
check translation pre-build in buddybuild

### LINKED ISSUE #701: Build time check for placeholders in strings.
See #698 - Sometimes we get crashes because a placeholder is not added or wrong in a translation of a string.

It would be nice if we had a script (in tools/ folder) that loops over all strings in our source file (values/strings.xml) and makes sure that placeholders in the source string show up correctly in the translated string as well.

* Source: no placeholders -> No check -> OK
* Source: %1$s -> Translation contains %1$s too. -> Otherwise fail build.
* Source %1$s, %2$s -> Translation contains %1$s and %2$s -> Otherwise fail build.
* Source:  %1$s -> String does not exist in Translation. -> OK, might not be translated yet.

Android does support a lot of format strings, but so far we are only using %x$s:
https://developer.android.com/reference/java/util/Formatter.html

This script could run on buddybuild as a pre-build script:
http://docs.buddybuild.com/docs/custom-prebuild-and-postbuild-steps

mastizada:
fix #701

### DIFF
--- a/buddybuild_prebuild.sh
+++ b/buddybuild_prebuild.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+# Pre-build script for buddybuild
+# http://docs.buddybuild.com/docs/custom-prebuild-and-postbuild-steps#section-pre-build
+
+python tools/l10n/check_translations.py
--- a/tools/l10n/check_translations.py
+++ b/tools/l10n/check_translations.py
@@ -1,0 +1,65 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#  This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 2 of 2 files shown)


mastizada:
@pocmo We need to install python requirements before tests. Should I add requirements.txt in somewhere where we will add all requirements for python files (in this case the ones that we use for buddybuild)

pocmo:
Looking at the code I realized that my description in the bug wasn't super clear. Sorry for that.

You can ignore the po files (and the l10n repository) and work with the strings.xml files only.

Use values/strings.xml as source. For every placeholder in a string in this file make sure it's also in every translated strings.xml file in values-XXXXX/strings.xml. If a string doesn't exist in the translated file then this is fine. However if it exists it needs to have exactly the same placeholders somewhere in the string.

Without needing to read the po files - Do we still need an additional dependency?

pocmo:
Thanks. This is great! 👍

PR #758, opened 2017-06-12
[BLOCKED] Build with and target SDK 26.

pocmo:
Even if this builds I do not want to merge this until after the release - to not introduce additional risk of breaking things.


pocmo:
`Robolectric does not support API level 26.`

Some tests need to either be pinned to 25. Or we wait until a Robolectric version that supports 26 is released.

Issue #774, opened 2017-06-14
Focus app unexpectedly closes if history is cleared twice from the notification tray


MihaiNinu:
MihaiNinu commented 29 minutes ago
Device: Android only

STR:

Tap a site address (e.g. qq.com since it takes a long time to load )
while site is loading, drag to open android bar
Tap on Erase focus history
Actual: Focus Crashes
Note: Please take in consideration that you might need to repeat the STR to trigger the crash

For further details please check : https://pastebin.mozilla.org/9024585

AaronMT:
Log is set to expire after a day 
[9024585.txt](https://github.com/mozilla-mobile/focus-android/files/1075141/9024585.txt)

pocmo:
Opened a PR with a fix for this. Can't test this thoroughly while in transit. :)

mcomella:
After looking at the code, I discovered a simpler STR:
1. Start focus
1. load website
1. use notification to clear data
1. load a new website
1. use notification to clear data

Expected: data is clear, on homepage
Actual: app is closed

mcomella:
I filed a related issue (#779) which is **not** fixed by Sebastian's patch.

PR #777, opened 2017-06-14
Correctly track visibility state of MainActivity. (#774)

pocmo:
#774.

When clicking the notification to erase the browsing history our BrowsingNotificationService
needs to decide:

* Is the app in the background? Then close it so that we start fresh the next time it is started.
* Is it in the foreground? Then notify it and play the erase animation.

The previous version tracked forground/background state by looking at onResume() and onPause().
This is not correct in all cases: An activity can be visible but not in the foreground (e.g.
another translucent screen is on top). In this case we still want to play the animation and
not (visibly) close the app.

With this patch we now track the state from onStart() and onStop(). Those callbacks are exactly
for this: They are called once an activity becomes visible to the user.

Additionally with this patch we will track the start of a browsing session as a "foreground"
event. This is necessary because since the service could have been removed since the last
foreground event (it's only "ongoing" *during* a session). However we are *always* in the
foreground when starting a browsing session.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/notification/BrowsingNotificationService.java
+++ b/app/src/main/java/org/mozilla/focus/notification/BrowsingNotificationService.java
@@ -70,6 +70,15 @@ public class BrowsingNotificationService extends Service {
     public int onStartCommand(Intent intent, int flags, int startId) {
         switch (intent.getAction()) {
             case ACTION_START:
+                // When we use ACTION_STOP to stop the browsing session, we also stop this Service (to conserve
+                // resources) so we lose track of the MainActivity's foreground/background state. If we start the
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/notification/BrowsingNotificationService.java
```diff
@@ -70,6 +70,7 @@ public static void stop(Context context) {
     public int onStartCommand(Intent intent, int flags, int startId) {
         switch (intent.getAction()) {
             case ACTION_START:
+                foreground = true;
```
mcomella: Add a comment explaining why this is necessary (because we stop the Service with ACTION_STOP so it loses the Activities' foreground state and we can assume when ACTION_START is called, the Activity is in the foreground).

Issue #682, opened 2017-05-31
Design icon to show blocking is OFF


brampitoyo:
The current icon – shield with ‘x’: https://cl.ly/3d1E1W1Q1o2y – isn’t yet final.

<img src="https://cloud.githubusercontent.com/assets/51007/26614814/3bdc1324-460e-11e7-9de8-b24b56b93e6c.png" width="360" />

We should align this redesigned icon with the rest of Photon iconography.

liuche:
Android intern is coming next week! This looks like a great first bug :) Would the icons be finalized by then, or does this require more UX time (to be consistent with Photon design)?

brampitoyo:
@antlam Do you think we can design a Photon-style 🛡 icon next week? We may be able to take what we’ve done in this bug and reuse it for the rest of Photon desktop and mobile.

pocmo:
We moved this after v1 to not rush it - so let's not rush it because of the internship again :)

Apart from the icon I wonder whether "having another button that opens the menu" is the best we can do here. Especially with other potentially buttons showing up. From my point of view there's a disconnect between the things we show in the menu (tracker counter, switch to disable blocking) and the icon in the menu: What does it actually mean? How can the user "explore" it - there's no direct action or detailed information to trigger?

brampitoyo:
<img width="600" alt="shield-big" src="https://cloud.githubusercontent.com/assets/51007/26772719/f561057c-4a12-11e7-88ef-0e240e4c3f8c.png">

<img width="84" alt="shield-small" src="https://cloud.githubusercontent.com/assets/51007/26772745/1b40f7fc-4a13-11e7-9e8f-763928ec7c82.png">

A few Photon-style shield icons. Thoughts, @antlam & @amilalhazwani?

brampitoyo:
Icons seen in context of the browser chrome:

<img width="376" alt="screen shot 2017-06-06 at 5 54 48 pm" src="https://cloud.githubusercontent.com/assets/51007/26817189/542a0fb2-4ae1-11e7-9277-1f45c3a71d7f.png">

brampitoyo:
@pocmo wrote:

> From my point of view there's a disconnect between the things we show in the menu (tracker counter, switch to disable blocking) and the icon in the menu: What does it actually mean? How can the user "explore" it - there's no direct action or detailed information to trigger?

[...]

PR #782, opened 2017-06-15
Design icon to show blocking is OFF (#682)

ekager:


ekager:
I added the new image files and added a block badging visibility method, but as we discussed in person, left finding the view with the rest!

ekager:
@pocmo There is an svg for this icon, but it was appearing distorted on certain API levels for unknown reasons... So I switched to the pngs. I can go back and investigate the svg issue more if it will keep the app small!

pocmo:
Okay. In this case let's land the code as-is (to get it out and tested) and let's file a new issue for updating the icon with a SVG?

In the past we had luck with other web-based SVG to Drawable XML converters that were smarter than Android Studio. There's also a list of unsupported SVG features - maybe this can help our designers to create a compatible version of the icon:
https://developer.android.com/studio/write/vector-asset-studio.html#PSD

app/src/main/res/layout/toolbar.xml
```diff
@@ -67,16 +67,35 @@
...
-        android:layout_marginTop="8dp"
-        android:background="?android:attr/selectableItemBackgroundBorderless"
-        android:padding="4dp"
-        android:contentDescription="@string/content_description_menu"
-        android:src="@drawable/ic_menu" />
+    <FrameLayout
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:paddingTop="4dp"
+        android:paddingEnd="4dp">
```
pocmo: This padding moved the menu button slightly to the left. It's almost not noticeable. :)

Maybe you can move the badge view with "margin" into position instead?

app/src/main/res/layout/toolbar.xml
```diff
@@ -67,16 +67,35 @@
...
+                android:layout_height="18dp"
+                android:layout_gravity="top|right|center"
+                android:src="@drawable/ic_tracking_protection_16_disabled"
+                android:tint="#FFFFFF"
+                android:background="@drawable/gold_badge_background"
+                android:visibility="gone" />
+
+    </FrameLayout>
+
+
```
pocmo: nit: two whitespaces

ekager:
@pocmo believe I made the requested changes! I tried to use two different online svg to drawable xml convertors, but they were still distorted when I tested on API 22 devices.

ekager:
SVG drawable is no longer distorted! I changed the color of the circle background as requested in the thread as well.

pocmo:
Adding [blocked] label just because it's not in the v1.1 list. Let's try to land this after the release.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -846,4 +852,9 @@ public boolean isBlockingEnabled() {
         final IWebView webView = getWebView();
         return webView == null || webView.isBlockingEnabled();
     }
+
+    // In the future, if more badging icons are needed, this should be abstracted
+    public void updateBlockingBadging(boolean enabled){
```
sputnikci: \[Checkstyle\] ERROR: '\{' is not preceded with whitespace\.

ekager:
So would this be ready to land in the next release @pocmo ?

pocmo:
Let's merge it in! :)

PR #786, opened 2017-06-15
Closes #783: Allow pages to load about:blank.

### LINKED ISSUE #783: Yelp redirect to app store loop
There are many ways to reproduce this behavior on the yelp website. The quickest is to do the following

* Do not have the yelp app installed
* Visit https://m.yelp.com
* tap the open in app button

Expected: warning that the play store is going to open

Actual: warning that the play store will open and a repeating attempt to open an external link that will eventually cause the OS to kill Focus. It is impossible to switch apps or try and kill Focus until this loop has run for a while and starts slowing down the phone.

Video of the behavior https://www.youtube.com/watch?v=4oUDw1oI2Lo

mcomella:
#783.

### DIFF
--- a/app/src/webkit/java/org/mozilla/focus/webkit/FocusWebViewClient.java
+++ b/app/src/webkit/java/org/mozilla/focus/webkit/FocusWebViewClient.java
@@ -186,6 +186,13 @@ import java.util.Map;
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


mcomella:
Merging before review so I can make a potential 1.0-RC4 release build - @liuche is currently reviewing.

Issue #784, opened 2017-06-15
Urls with trailing whitespace should load url, not search


liuche:
When entering a url, whitespace should be trimmed and the resulting string should be either loaded as a url (if it's a url) or searched for (if it's a search term), the way it is on the other browsers.

STR:
Enter url with a trailing space.

Expected:
Url is loaded.

Actual:
Url is searched for on search engine, leading to search results page, not url.

ekager:
I added a leading/trailing space trim before it checks whether the string is a url. Am I oversimplifying?

liuche:
Yep! That's the right fix.

ekager:
Okay! Added trimming in isUrl and normalizeUrl methods in UrlUtils.

liuche:
Closed by #789 .

PR #789, opened 2017-06-16
Urls with trailing whitespace should load url, not search (#784)

ekager:


liuche:
Doesn't look like we have any fires on Focus, so I'm going to merge some of the simple blocked-on-release requests!

liuche:
Closes #784.

PR #811, opened 2017-06-23
Add Custom Tabs status bar colour

jonalmeida:
This was cherry-picked from the [fennec patch][1] and goes hand-in-hand with the toolbar colour changes as well.

Left is before, right is after.

out

[1]: https://hg.mozilla.org/mozilla-central/file/1c529127c7a4/mobile/android/base/java/org/mozilla/gecko/customtabs/CustomTabsActivity.java

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -17,6 +17,7 @@ import android.graphics.Color;
 import android.graphics.drawable.Drawable;
 import android.graphics.drawable.TransitionDrawable;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
@@ -26,6 +27,8 @@ import android.support.v4.content.ContextCompat;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.Window;
+import android.view.WindowManager;
 import android.webkit.CookieManager;
 import android.webkit.URLUtil;
 import android.widget.ImageButton;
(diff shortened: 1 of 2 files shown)


pocmo:
I think we want to explicity keep the Focus color here.

@antlam @brampitoyo Opinions ?

brampitoyo:
Actually, in addition to @pocmo’s idea re: status bar colour gradient, I wonder if it would also be possible to have the Focus gradient be on the URL bar?

This helps consistency and brand identity (“When it’s a purple gradient, you know it’s Focus. Pages are going to load faster/without ads, and your history will be erased afterwards”).

brampitoyo:
I just checked the Custom Tabs implementation, and it sounded like each individual app (not us) have control over the colour of the URL bar.

I still think that keeping our colour gradient on the status bar is the right thing to do, with one caveat.

In app, we change our toolbar gradient from dark (desaturated) to light (saturated) as the website finishes loading. Currently, on Custom Tabs, the status bar gradient changes. But we shouldn’t do this anymore.

The URL bar colour is set by the app and never changes, so our status bar gradient will look off if it changes all the time. We should always maintain the same light/saturated gradient, whatever the site loading status may be.

jonalmeida:
> I just checked the Custom Tabs implementation, and it sounded like each individual app (not us) have control over the colour of the URL bar.

The way it works is that the app gives us (the browser) the colour that they want for the URL bar so that it matches their UX and styles, making it look less like a browser and more like it's part of the app. We have control in whether we should be applying those colour changes or not.

> I still think that keeping our colour gradient on the status bar is the right thing to do, with one caveat.
> In app, we change our toolbar gradient from dark (desaturated) to light (saturated) as the website finishes loading. Currently, on Custom Tabs, the status bar gradient changes. But we shouldn’t do this any more.

You could consider taking the colour the app provides and making a gradient from it so that you're still listening to the colour the app is asking for and also maintaining your brand identity. Although, I think that should be a different issue from this one.

> The URL bar colour is set by the app and never changes, so our status bar gradient will look off if it changes all the time. We should always maintain the same light/saturated gradient, whatever the site loading status may be.

My patch is the bit of incomplete implementation that does this. It's part of the [demo samples provided][1] as well as what's being done in the fennec custom tabs implementation (linked above).


[1]: https://github.com/GoogleChrome/custom-tabs-provider/blob/master/app/src/main/java/com/example/android/customtabsbrowser/CustomTabController.java#L292

pocmo:
Let's move this discussion over to an issue. I'll close this PR.

PR #840, opened 2017-06-30
Update GeckoView and fix up the provider a bit

snorp:


app/build.gradle
```diff
@@ -88,14 +88,6 @@ android {
...
     }
 
-    variantFilter { variant ->
-        def flavors = variant.flavors*.name
-        // We only need a gecko debug build for now.
-        if (flavors.contains("gecko") && (variant.buildType.name != "debug" || flavors.contains("klar"))) {
-            setIgnore(true)
-        }
-    }
-
```
pocmo: Did you need to do non-debug gecko builds? I would like to keep this so that we do not build all configurations (release, beta, ..) with gecko every time on buddybuild.

app/src/gecko/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -128,6 +126,10 @@ private ContentListener createContentListener() {
                 @Override
                 public void onTitleChange(GeckoView geckoView, String s) {
                 }
+
+                @Override
+                public void onFullScreen(GeckoView geckoView, boolean fullScreen) {
+                }
```
pocmo: Nice! I'm about to add fullscreen support to Focus.

app/build.gradle
snorp: Yeah, I wanted to at least get beta. I changed it to just include that with the debug flavor.

Issue #776, opened 2017-06-14
Refrain blank searches


AaronMT:
Currently we let you do a blank search via no input in the address bar but by hitting the Go/Arrow on the keyboard.

=> Tap the address bar 
=> Hit Go

We should decide wether we grey out the keyboard button until a character exists in the field.

npark-mozilla:
Actually, the fix recognizes empty spaces as valid input, instead of ignoring it. The behavior in Chrome (and in Firefox for Android) is that it ignores the input unless there is a non-space char.  Can we add a fix for it?

ekager:
@npark-mozilla Thanks! I think I fixed this by adding a trim.

pocmo:
Merged. :)

PR #854, opened 2017-07-04
Refrain blank searches (#776)

ekager:


app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.java
```diff
@@ -390,19 +390,21 @@ private void dismiss() {
 
     @Override
     public void onCommit() {
-        ViewUtils.hideKeyboard(urlView);
+        if (!urlView.getText().toString().isEmpty()) {
```
pocmo: nit: Let's avoid calling getText() and toString() multiple times by assigning it to some variable (final String input = ..)

Issue #860, opened 2017-07-05
Floating action button missaligned after exiting full screen mode


pocmo:
STR:
* Go to ted.com
* Start a video
* Press full screen button on video player
* Exit full screen

Expected: When scrolling the FAB is still at the same position
Actual: The FAB is partially positioned off screen

![fab-misplaced](https://user-images.githubusercontent.com/89638/27857918-b4284782-6173-11e7-8fba-ba1d55ba04e0.png)

pocmo:
This has been a side effect of us not leaving immersive mode correctly. Fixed in ee9b7e99943170eb31d79abdeade3cdf765d602f.

PR #865, opened 2017-07-05
Floating action button missaligned after exiting full screen mode #860

ekager:


pocmo:
I fixed issue #872 -> We didn't exit immersive mode correctly. And it looks like this fixed this issue too. The misalignment might have happened because we were still drawing in full screen mode  below the navigation bar.

I think initially we didn't include the floating action button into the ResizableKeyboardLayout so that the button doesn't move up when showing the keyboard (and probably overlaying the text field, especially in landscape mode).

ekager:
Okay, I'll close this PR then!

Issue #14, opened 2016-12-18
Support file downloads


pocmo:
**Theme**: _Default Browser_

**User Story**

> I discovered content of a page that I want to revisit later (save to my device)

**Feature**

> Support for downloads (file downloads, PDF, execs)

**Background**

Currently just nothing happens when clicking a link that should start a download.

This should be fairly easy to implement using WebView (I guess), but could be more complicated with Gecko. Downloads are not part of the GeckoView MVP. However we might be able to work around that with some custom JavaScript.

Maybe we just want to intercept the download and let Android's download manager perform the task (privacy concerns?).

**Tasks**
- [x] Implement [DownloadListener](https://developer.android.com/reference/android/webkit/WebView.html#setDownloadListener(android.webkit.DownloadListener))
- [x] UX mocks / flow
- [x] Show dialog to let the user confirm the download (see UI mock)
- [x] Show snackbar once download is finished
- [x] Request runtime permission to write to external storage
- [x] Forward Download to Android's Download Manager
- [x] Add telemetry probe

**UI mock**

<img width="1532" alt="screen shot 2017-07-06 at 8 43 59 pm" src="https://user-images.githubusercontent.com/51007/27905544-f7023f36-628b-11e7-9331-7fbaf11afdd2.png">

pocmo:
@antlam I implemented downloads and they will ask for the storage permission (if needed). However I assume we want to show some kind of UI before starting the download?

We know the following things at the time of the download and could display it:
* URL from which we are downloading
* File name (This is either defined by the server or generated by us)
* MIME type (Optional)
* Size of file (Optional)

antlam:
Gonna leave this one to @brampitoyo !

pocmo:
Btw.: Downloads are already implemented (without UI, except that we may ask for a permission) - but not enabled in builds as we decided to not ship it in v1.

danilodorgam:
[Request runtime permission to write to external storage](https://developer.android.com/training/permissions/requesting.html)

Follows documentation about the request.

If no one implements this function, I can try to do this later.

pocmo:
> If no one implements this function, I can try to do this later.

That's already implemented!

brampitoyo:
First take on the download dialogue:

![download dialog - mdpi](https://cloud.githubusercontent.com/assets/51007/26570449/6629004a-455c-11e7-9668-aa480c17700d.png)

[...]

PR #885, opened 2017-07-10
Support file downloads (#14)

ekager:


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -501,7 +580,64 @@ private void queueDownload(Download download) {
...
+                            snackbar.show();
+                        }
+                    } else {
+                        cursor.close();
+                    }
+                }
+            }
+        };
+
+        getActivity().registerReceiver(downloadBroadcastReceiver, filter);
```
pocmo: You register the receiver but you never unregister it (leak). The receiver can be independent from queuing a download. So we could create it once in onCreate() and then register/unregister in onResume() and onPause().

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -501,7 +580,64 @@ private void queueDownload(Download download) {
...
+                                        @Override
+                                        public void onClick(View view) {
+                                            if (mimeType != null) {
+                                                Intent openFileIntent = new Intent(Intent.ACTION_VIEW);
+                                                openFileIntent.setDataAndType(uriFile, mimeType);
+                                                openFileIntent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
+                                                try {
+                                                    startActivity(openFileIntent);
+                                                } catch (ActivityNotFoundException e) {
+                                                    // Ignore if no activity was found.
```
pocmo: I wonder whether we should show an error toast/dialog here. Otherwise just nothing happens when the user clicks the "Open" action. Another option would be to query available apps to open this intent before hand and only show the "Open" action if there's an activity to handle this intent. Let's ask the UX team - but we would need to land a string (if needed) beginning of this week.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -501,7 +580,64 @@ private void queueDownload(Download download) {
...
+                    DownloadManager.Query query = new DownloadManager.Query();
+                    query.setFilterById(mDownloadReference);
+                    Cursor cursor = manager.query(query);
+                    if (cursor.moveToFirst()) {
+                        int columnIndex = cursor.getColumnIndex(DownloadManager.COLUMN_STATUS);
+                        if (DownloadManager.STATUS_SUCCESSFUL == cursor.getInt(columnIndex)) {
+                            String uriString = cursor.getString(cursor.getColumnIndex(DownloadManager.COLUMN_LOCAL_URI));
+                            cursor.close();
+
+                            if (uriString.substring(0, 7).matches("file://")) {
```
pocmo: nit: startsWith()

app/src/main/AndroidManifest.xml
```diff
@@ -16,6 +16,17 @@
         android:supportsRtl="true"
         android:theme="@style/AppTheme"
         android:name=".FocusApplication">
+
+        <provider
+            android:name="android.support.v4.content.FileProvider"
```
pocmo: Wow, it's annoying that we have to create a FileProvider for this and can't just share the file:/// uri to the already public file. :(

app/src/main/res/xml/provider_paths.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<paths xmlns:android="http://schemas.android.com/apk/res/android">
+    <external-path
+        name="Download"
+        path="Download/" />
```
pocmo: Is it possible to define some kind of wildcard path here?

- Currently the app crashes when saving an image (long press image -> Save image) because it's not in the download path
- There might be devices where the download directory is named differently. In code we are using the Environment.DIRECTORY_* constants.

app/src/main/AndroidManifest.xml
ekager: I know :( It made it much more complex. Security change on Nougat was :

> Passing file:// URIs outside the package domain may leave the receiver with an unaccessible path. Therefore, attempts to pass a file:// URI trigger a FileUriExposedException. The recommended way to share the content of a private file is using the FileProvider.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
ekager: I had a discussion with Bram and we decided to take away the "Open" action if no activity is found to handle the intent.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -493,15 +534,72 @@ private void queueDownload(Download download) {
...
+        manager = (DownloadManager) context.getSystemService(Context.DOWNLOAD_SERVICE);
+
+        final IntentFilter filter = new IntentFilter(DownloadManager.ACTION_DOWNLOAD_COMPLETE);
+
+        getActivity().registerReceiver(downloadBroadcastReceiver, filter);
+
+        mDownloadReference = manager.enqueue(request);
+    }
+
+    public void displaySnackBar(long referenceId) {
```
pocmo: Could we move all this snackbar code into it's own receiver class file? I assume it's pretty independent from this fragment except that it needs to get a reference to a view to show the snackbar.

So the rough idea would be to create an instance of this in onCreateView (That's where we have a view) - then register in onResume() and unregister in onPause() (See below).

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -493,15 +534,72 @@ private void queueDownload(Download download) {
...
+                                        }
+                                    }
+                                });
+                        if (activitiesFoundForIntent) {
+                            snackbar.setActionTextColor(ContextCompat.getColor(getContext(), R.color.download_dialog_text));
+                        } else {
+                            snackbar.setAction("", null);
+                        }
+                        snackbar.show();
+                        getActivity().unregisterReceiver(downloadBroadcastReceiver);
```
pocmo: We still need to get here in order to successfully unregister (e.g. we do not if there's an exception or the query is empty). If we do not then this receiver (and via reference this fragment) will exist indefinitely in memory. Register/Unregister in onResume()/onPause() is safer!

app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
```diff
@@ -0,0 +1,99 @@
...
+            final DownloadManager.Query query = new DownloadManager.Query();
+            query.setFilterById(downloadReference);
+            try (Cursor cursor = downloadManager.query(query)) {
+                if (cursor.moveToFirst()) {
+                    int columnIndex = cursor.getColumnIndex(DownloadManager.COLUMN_STATUS);
+                    if (DownloadManager.STATUS_SUCCESSFUL == cursor.getInt(columnIndex)) {
+                        String uriString = cursor.getString(cursor.getColumnIndex(DownloadManager.COLUMN_LOCAL_URI));
+
+                        if (uriString.startsWith("file://")) {
+                            uriString = uriString.substring(7);
```
mcomella: super-nit: "file://" should be a constant and we should reference that constant when we get the substring, i.e.:
```java
uriString = uriString.substring(FILE_SCHEME.length)
```

app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
```diff
@@ -0,0 +1,99 @@
...
+                        }
+                        final String localUri = uriString;
```
mcomella: super-nit: slightly more readable to use a ternary:
```java
final String localUri = uriString.startsWith(FILE_SCHEME) ? uriString.substring(FILE_SCHEME.length) : uriString;
```

Since it's only one statement.

app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
```diff
@@ -0,0 +1,99 @@
...
+
+import org.mozilla.focus.BuildConfig;
+import org.mozilla.focus.R;
+
+import java.io.File;
+import java.util.HashMap;
+import java.util.List;
+
+public class DownloadBroadcastReceiver extends BroadcastReceiver {
+    private final View browserContainer;
```
mcomella: Views contain a reference to Context so this should be a weak reference.

In the current implementation, it's unlikely to cause issues (if the Activity is destroyed, so will the fragment so so too should this Receiver) but could cause issues if the implementation changes

app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
```diff
@@ -0,0 +1,99 @@
...
+                    if (DownloadManager.STATUS_SUCCESSFUL == cursor.getInt(columnIndex)) {
+                        String uriString = cursor.getString(cursor.getColumnIndex(DownloadManager.COLUMN_LOCAL_URI));
+
+                        if (uriString.startsWith("file://")) {
+                            uriString = uriString.substring(7);
+                        }
+                        final String localUri = uriString;
+
+                        final String fileExtension = MimeTypeMap.getFileExtensionFromUrl(localUri);
+                        final String fileName = URLUtil.guessFileName(localUri, null, null);
```
mcomella: Should the mimeType be passed in here?

app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
```diff
@@ -0,0 +1,99 @@
...
+                                    public void onClick(View view) {
+                                        if (activitiesFoundForIntent && mimeType != null) {
+                                            context.startActivity(openFileIntent);
+                                        }
+                                    }
+                                });
+                        if (activitiesFoundForIntent) {
+                            snackbar.setActionTextColor(ContextCompat.getColor(context, R.color.download_dialog_text));
+                        } else {
+                            snackbar.setAction("", null);
```
mcomella: nit: instead of setting the action to null here if there are no activities found, we should just never add the action in the first place

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -467,12 +480,37 @@ public void onRequestPermissionsResult(int requestCode, @NonNull String[] permis
...
 
+    void showDialog(Download download) {
+        final DialogFragment newFragment = DownloadDialogFragment.newInstance(download);
+        newFragment.setTargetFragment(BrowserFragment.this, 300);
+        newFragment.show(getFragmentManager(), "dialog");
+    }
+
+    @Override
+    public void onFinishDownloadDialog(Download download, boolean shouldDownload) {
+        if (shouldDownload){
```
mcomella: nit: `if (shouldDownload) {`

app/src/main/java/org/mozilla/focus/fragment/WebFragment.java
```diff
@@ -38,6 +38,7 @@
      */
     @Nullable
     public abstract String getInitialUrl();
+    public abstract void onCreateViewCalled();
```
mcomella: Add a comment explaining what this method is and why it's necessary (because `onCreateView` is final so others don't forget to call it).

app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
```diff
@@ -0,0 +1,99 @@
...
+        this.browserContainer = view;
+    }
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        final long mDownloadReference = intent.getLongExtra(DownloadManager.EXTRA_DOWNLOAD_ID, -1);
+        displaySnackBar(context, mDownloadReference);
+    }
+
+    public void displaySnackBar(final Context context, long downloadReference) {
```
mcomella: nit: it's confusing what the diff between `referenceIds` and `downloadReference` is. Perhaps change them (respectively to): `queuedDownloadReferences` and `completedDownloadReference`.

app/src/main/java/org/mozilla/focus/fragment/DownloadDialogFragment.java
```diff
@@ -0,0 +1,118 @@
...
+
+        return alert;
+    }
+
+    public interface DownloadDialogListener {
+        void onFinishDownloadDialog(Download download, boolean shouldDownload);
+    }
+
+    public void sendButtonClicked(Download download, boolean shouldDownload) {
+        final DownloadDialogListener listener = (DownloadDialogListener) getTargetFragment();
```
mcomella: super-nit: This is a little fragile. It's probably better practice to have the more conventional `setDownloadDialogListener`, check if it's null, if not call it, etc.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -409,7 +417,12 @@ public void onExitFullScreen() {
             public void onDownloadStart(Download download) {
                 if (PackageManager.PERMISSION_GRANTED == ContextCompat.checkSelfPermission(getContext(), Manifest.permission.WRITE_EXTERNAL_STORAGE)) {
                     // We do have the permission to write to the external storage. Proceed with the download.
-                    queueDownload(download);
+                    if (download.getDestinationDirectory().equals(Environment.DIRECTORY_DOWNLOADS)) {
```
ekager: This is my work-around to differentiate direct downloads that need this download dialog from long press on image downloads which have their own dialog.

pocmo:
> I suggested Emily add a list/set/map of queued downloads (addQueuedDownload) but it appears the back-end only supports one download (pendingDownload). This inconsistency is concerning (I wonder if there are bugs in the back-end implementation around multiple downloads?). I didn't comment about this in my review below

The backend supports unlimited number of parallel downloads. `pendingDownload` is only a temporary reference and only used when we have to show a runtime permission dialog. While we ask for the permission the user can leave the app and the activity can get destroyed. That's why there's additional code to save and restore `pendingDownload`. As soon as we have the permission this reference is not used anymore. See BrowserFragment.onDownloadStart() and onRequestPermissionsResult(). The naming I chose here isn't perfect and I think I should make this more clear in the code.

app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
```diff
@@ -0,0 +1,120 @@
...
+                    final String mimeType = MimeTypeMap.getSingleton().getMimeTypeFromExtension(fileExtension);
+
+                    final File file = new File(localUri);
+                    final Uri uriForFile = FileProvider.getUriForFile(context, BuildConfig.APPLICATION_ID + FILE_PROVIDER_EXTENSION, file);
+
+                    final Intent openFileIntent = createOpenFileIntent(uriForFile, mimeType);
+                    showSnackbarForFilename(openFileIntent, context, fileName);
+                }
+            }
+        }
```
ekager: Right!

app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.HashSet;
+import java.util.List;
+
+/**
+ * BroadcastReceiver for finished downloads
+ */
+public class DownloadBroadcastReceiver extends BroadcastReceiver {
+    private static final String FILE_SCHEME = "file://";
+    private static final String FILE_PROVIDER_EXTENSION = ".fileprovider";
+    private final View browserContainer;
```
mcomella: nit: Conventionally, constants (static final) should be separated from instance variables; i.e. add a newline above the view.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
mcomella: Please add a comment explaining ^ as we spoke about!

In general, if a reviewer asks a question about why something is the way it is, it may be worth adding a comment because future developers reading your code may also have the same questions!

app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
```diff
@@ -0,0 +1,120 @@
...
+        query.setFilterById(completedDownloadReference);
+        try (Cursor cursor = downloadManager.query(query)) {
+            if (cursor.moveToFirst()) {
+                int statusColumnIndex = cursor.getColumnIndex(DownloadManager.COLUMN_STATUS);
+                if (DownloadManager.STATUS_SUCCESSFUL == cursor.getInt(statusColumnIndex)) {
+                    String uriString = cursor.getString(cursor.getColumnIndex(DownloadManager.COLUMN_LOCAL_URI));
+
+                    final String localUri = uriString.startsWith(FILE_SCHEME) ? uriString.substring(FILE_SCHEME.length()) : uriString;
+                    final String fileExtension = MimeTypeMap.getFileExtensionFromUrl(localUri);
+                    final String fileName = URLUtil.guessFileName(localUri, null, null);
```
mcomella: Should we be passing in the mimeType here?

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -500,8 +541,9 @@ private void queueDownload(Download download) {
 
         request.allowScanningByMediaScanner();
 
-        final DownloadManager manager = (DownloadManager) context.getSystemService(Context.DOWNLOAD_SERVICE);
-        manager.enqueue(request);
+        long downloadReference = manager.enqueue(request);
+
+        DownloadBroadcastReceiver.addQueuedDownload(downloadReference);
```
mcomella: Is there any reason for this to be static?

In general, mixing static (these queued downlads) and member variables (our receiver instance) can create bugs since it's harder to create track of the lifecycles of your program state. For example, if it's all static (it lasts basically forever) or member (it lasts as long as the object does) but if they're mixed, it's easy to overlook.

In this specific case, we may overlook the system's lifecycles: would the system return a new downloadManager instance if the Activity (and thus this broadcast receiver) is recreated? If it's a new instance, do the download references within the new instance match the download references we had already stored statically? I'm not sure... it's safer to stick with member variables!

That being said, static *does* let us do more (if the Activity is recreated, we can avoid losing any pending download references as we would with member variables) but there's always a question of code complexity vs. functionality. imo, it's unlikely the Activity to be destroyed so I'm fine dropping download snackbars if it happens (especially because the system sends notifications too) to make the code much less complex.

app/src/main/java/org/mozilla/focus/fragment/DownloadDialogFragment.java
```diff
@@ -0,0 +1,118 @@
...
+        String fileName = URLUtil.guessFileName(
+                download.getUrl(), download.getContentDisposition(), download.getMimeType());
+
+        args.putString("fileName", fileName);
+        args.putParcelable("download", download);
+        frag.setArguments(args);
+        return frag;
+    }
+
+    public AlertDialog onCreateDialog(Bundle bundle) {
```
mcomella: super-nit: fwiw, this function could benefit from helper functions (it's more than a screen long!) but no need to change it now and potentially reintroduce bugs (though, I guess intellij can actually do this refactor for you so... :P)

app/src/main/java/org/mozilla/focus/fragment/DownloadDialogFragment.java
```diff
@@ -0,0 +1,118 @@
...
+
+        downloadDialogWarningMessage.setText(getSpannedTextFromHtml(R.string.download_dialog_warning, R.string.app_name));
+
+        final AlertDialog alert = builder.create();
+
+        downloadDialogCancelButton.setOnClickListener(
+                new View.OnClickListener() {
+                    @Override
+                    public void onClick(View v) {
+                        sendButtonClicked(pendingDownload, false);
```
mcomella: I wonder if we'd want a telemetry event for cancelled downloads too...

app/src/main/java/org/mozilla/focus/fragment/DownloadDialogFragment.java
```diff
@@ -0,0 +1,118 @@
...
+                        sendButtonClicked(pendingDownload, true);
+                        TelemetryWrapper.downloadDialogEvent();
+                        dismiss();
+                    }
+                });
+
+        return alert;
+    }
+
+    public interface DownloadDialogListener {
```
mcomella: nit: by convention, inner interfaces & classes should be at the bottom (or some cases, at the top!) of a class

app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
```diff
@@ -0,0 +1,105 @@
...
+            snackbar.setActionTextColor(ContextCompat.getColor(context, R.color.download_dialog_text));
+        }
+        snackbar.show();
+    }
+
+    private boolean isFocusDownload(long completedDownloadReference) {
+        return (queuedDownloadReferences.contains(completedDownloadReference));
+    }
+
+    private void removeFromHashSet(long completedDownloadReference) {
```
mcomella: nit: I don't think this warrants a separate helper method. `hashSet.remove(thingie)` is already pretty readable and to which hash set does "removeFromHashSet" refer to?

PR #905, opened 2017-07-12
Closes #895: Add a telemetry probe for "default browser"

### LINKED ISSUE #895: Add a telemetry probe for "default browser"
Looking at #888 it would be very interesting to know how many users (%) have set Focus as their default browser.

liuche:
This isn't my ideal way of adding this probe, as an event - it'd be better to have this as a Settings pref, but in this case, "is default browser" isn't truly an Android SharedPreference, so it doesn't get added in SettingsMeasurement. Since it's not uncommon to have non-SharedPreference prefs (on Fennec at least) maybe we can add a callback into the SettingsMeasurement.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
+++ b/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
@@ -16,6 +16,7 @@ import org.mozilla.focus.R;
 import org.mozilla.focus.search.SearchEngine;
 import org.mozilla.focus.search.SearchEngineManager;
 import org.mozilla.focus.utils.AppConstants;
+import org.mozilla.focus.utils.Browsers;
 import org.mozilla.telemetry.Telemetry;
 import org.mozilla.telemetry.TelemetryHolder;
 import org.mozilla.telemetry.config.TelemetryConfiguration;
@@ -43,6 +44,7 @@ public final class TelemetryWrapper {
 
     private static class Category {
         private static final String ACTION = "action";
+        private static final String PREF = "pref";
     }
 
     private static class Method {
@@ -81,6 +83,7 @@ public final class TelemetryWrapper {
         private static final String CUSTOM_TAB_CLOSE_BUTTON = "custom_tab_close_but";
         private static final String CUSTOM_TAB_ACTION_BUTTON = "custom_tab_action_bu";
         private static final String FIRSTRUN = "firstrun";
+        private static final String IS_DEFAULT = "is_default_browser";
     }
 
     private static class Value {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
```diff
@@ -208,6 +214,14 @@ public static void urlBarEvent(boolean isUrl) {
         }
     }
 
+    private static TelemetryEvent isDefaultBrowserEvent() {
+        final Context context = TelemetryHolder.get().getConfiguration().getContext();
+        final Browsers browsers = new Browsers(context, "http://www.mozilla.org");
+        final String isDefault = browsers.isDefaultBrowser(context) ? Value.TRUE : Value.FALSE;
+
+        return TelemetryEvent.create(Category.PREF, Method.BACKGROUND, Object.IS_DEFAULT, isDefault);
```
pocmo: My only concern here is: How do we analyze this on the server side? What we'd want to see in re:dash is a pie chart showing the % of using Focus as a default browser. Can we do this easily with an event?

If there wouldn't be a code freeze would we use something else? Like a field in the event ping for that? In this case I'd prefer to move this to the next iteration. What do you think?

pocmo:
> This isn't my ideal way of adding this probe, as an event - it'd be better to have this as a Settings pref, but in this case, "is default browser" isn't truly an Android SharedPreference, so it doesn't get added in SettingsMeasurement. Since it's not uncommon to have non-SharedPreference prefs (on Fennec at least) maybe we can add a callback into the SettingsMeasurement.

That's a great idea! Currently SettingsMeasurement just gets the default shared preferences from the system and this doesn't work in all cases. Adding something like an optional SharedPreferencesProvider to SettingsMeasurement sounds like a great idea. I wanted to fix another bug in the telemetry library today - I'll see if I could add this for you.

pocmo:
Oh, one thing I forgot. Whenever we add a new probe we update the telemetry doc too:
https://github.com/mozilla-mobile/focus-android/blob/master/docs/telemetry.md

pocmo:
In #911 I update the telemetry library that now allows you to define your own `SettingsProvider`. By default a `SharedPreferenceSettingsProvider` is used (reads just from default SharedPreferences).

If you add a made-up preference to `setPreferencesImportantForTelemetry(..)` in `TelemetryWrapper` and setup your own `SettingsProvider` implementation then you can inject the value for the default browser.

Sketch:
```java
        SettingsMeasurement.SettingsProvider provider = new SettingsMeasurement.SharedPreferenceSettingsProvider() {
            @Override
            public boolean containsKey(String key) {
                if (key.equals("....default..browser...")) {
                    return true;
                }

                return super.containsKey(key);
            }

            @Override
            public Object getValue(String key) {
                if (key.equals("....default..browser...")) {
                    return ...;
                }
                
                return super.getValue(key);
            }
        };
```

liuche:
Closing in favor of new approach in #916 using SettingsMeasurement instead of Events.

Issue #897, opened 2017-07-11
NullPointerException crash in telemetry library


pocmo:
Tracking here too so that we make sure we update the library in one of our releases:
https://github.com/mozilla-mobile/telemetry-android/issues/51

This our Top 2 crash.

Reports: 1,557
Impacted users: 39

PR #911, opened 2017-07-12
Update telemetry dependency to 1.1.0.

pocmo:
For fixing #897.

https://github.com/mozilla-mobile/telemetry-android/compare/v1.0...v1.1

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -148,7 +148,7 @@ dependencies {
         transitive = false
     }
 
-    compile 'org.mozilla.telemetry:telemetry:1.0.0'
+    compile 'org.mozilla.telemetry:telemetry:1.1.0'
 
     focusCompile 'com.adjust.sdk:adjust-android:4.11.4'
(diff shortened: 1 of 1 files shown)


mcomella:
I'm going to go review the changes in telemetry-android from v1.0..v1.1 right now – will post back when review is completed.

mcomella:
r+ on telemetry from v1.0..v1.1. Here are some notes though:
* Tag `v1.1` should be on the commit that changed version to v1.1
* Why is `setCollectionEnabled` independent of `setUploadEnabled`? It seems likely one will mistakenly be called without the other
* When a user disables telemetry collection, should we clear all pending pings?
* Added additional comments to this commit: https://github.com/mozilla-mobile/telemetry-android/commit/4e16343ca2a86e99a248e551d8cb04e4975a106e

mcomella:
Merging so we can get this in before code freeze but @pocmo should verify tomorrow that this is desired.

PR #913, opened 2017-07-12
Make UI Tests run on beta build

npark-mozilla:


npark-mozilla:
@pocmo um actually, I sent you a slack msg about how it's causing compile error and wanted your input on it - "this should be enough, but it fails build with (Error:Execution failed for task ‘:app:transformClassesAndResourcesWithProguardForFocusWebkitBetaAndroidTest’.
> Job failed, see logs for details)"

npark-mozilla:
Detail error msg can be seen here: https://dashboard.buddybuild.com/apps/584f67f6f3d6eb01000842d6/build/59666c5fdc94b90001029812#logs

npark-mozilla:
@pocmo actually, I think I figured out how to get around proguard error.  I created another rule file and excluded test related files.  (Although I'm not exactly sure what proguard does)   Could you take another look at the changes please?  Thanks!

app/build.gradle
```diff
@@ -8,6 +8,7 @@ apply plugin: 'checkstyle'
 android {
     compileSdkVersion 25
     buildToolsVersion "25.0.3"
+    testBuildType "beta"
```
pocmo: I saw that we can configure the build variant to run tests on in buddybuild. Do we even need this in the gradle file?

npark-mozilla: @pocmo so I had to specify which build variant the test can run, as it seems that tests can run on only one variant at a time.  without above, it will not apply dependencies from 'androidTestCompile' or 'testCompile' apparently. (https://stackoverflow.com/questions/32484985/android-studio-instrumentation-testing-build-variant)

pocmo:
I'll close this one. Let's find a different way to run the tests if strict mode is our only problem with the current setup.

PR #916, opened 2017-07-12
Closes #895: Add a telemetry probe for "default browser"

### LINKED ISSUE #895: Add a telemetry probe for "default browser"

liuche:
This is the new version of the telemetry probe using the SettingsMeasurement.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
+++ b/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
@@ -16,12 +16,14 @@ import org.mozilla.focus.R;
 import org.mozilla.focus.search.SearchEngine;
 import org.mozilla.focus.search.SearchEngineManager;
 import org.mozilla.focus.utils.AppConstants;
+import org.mozilla.focus.utils.Browsers;
 import org.mozilla.telemetry.Telemetry;
 import org.mozilla.telemetry.TelemetryHolder;
 import org.mozilla.telemetry.config.TelemetryConfiguration;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
```diff
@@ -172,6 +176,31 @@ public static void init(Context context) {
...
+                }
+
+                return super.containsKey(key);
+            }
+
+            @Override
+            public java.lang.Object getValue(String key) {
+                if (key.equals(prefKeyDefaultBrowser)) {
+                    final Context context = TelemetryHolder.get().getConfiguration().getContext();
+                    final Browsers browsers = new Browsers(context, "http://www.mozilla.org");
```
mcomella: nit: Since we re-use this `"http://www.mozilla.org"` incantation, we should probably define a constant, if not define a method that makes this use case more apparent.

liuche:
Changes from comments (that were eaten by GH): Added the url as TRADITIONAL_BROWSER_URL constant.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
liuche: Added this as a constant, Browsers.TRADITIONAL_BROWSER_URL with a comment.

liuche:
Data-review on Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1380179

Next: Making changes to the server-side data ping processing

PR #917, opened 2017-07-12
Closes #785: Separate gradient background view into two views

### LINKED ISSUE #785: Focus is missing URL bar when loading first page
I see this regularly on my older GS5 when I load the first page:
screenshot_2017-06-15-13-41-15

It shortly (5 seconds max?) transitions to show the URL bar and the web view. It's ugly but still functional. I don't see this on faster devices.

mcomella:
#785

Tested on:
- N5 (Android 6)
- GS5 (Android 6)

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -20,6 +20,7 @@ import android.net.Uri;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+import android.support.annotation.UiThread;
 import android.support.design.widget.AppBarLayout;
 import android.support.v4.app.Fragment;
 import android.support.v4.content.ContextCompat;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -774,8 +794,22 @@ public void setBlockingEnabled(boolean enabled) {
...
+        appBarLayout.setBackgroundResource(backgroundResource);
+        backgroundTransitionGroup = getUpdatedBackgroundTransitionGroup(); // must be called after resource update.
+    }
+
+    /**
+     * Returns an updated drawable group for the background views, based on their current view resources.
+     *
+     * For more details on these background views, see comments around their declaration in fragment_browser.xml.
+     */
+    private TransitionDrawableGroup getUpdatedBackgroundTransitionGroup() {
```
liuche: Nit: I might call this getCurrentBackgroundTransitionGroup - "updated" from what?

app/src/main/res/layout/fragment_browser.xml
```diff
@@ -13,10 +13,20 @@
         android:layout_width="match_parent"
         android:layout_height="match_parent">
 
+        <!-- The toolbar gradient background extends from under the system status bar to the app's Toolbar.
```
liuche: nit: through the app's Toolbar

liuche:
Looks like there are some conflicts, so I'll just let mcomella fix and merge on Thurs.

pocmo:
This looks broken on my device (Pixel XL, Android 8.0):

statusbar-broken

mcomella:
> This looks broken on my device (Pixel XL, Android 8.0):

As per Chenxia's comment above, I think what's happening is the returned status bar height is 0 on this device – I'm not sure why but Chenxia pointed out others had this problem in the StackOverflow post where I found the parts of my solution.

The reason the status bar background is visible at all (and does not have a height of 0) is because, I think, the status bar background view is taking the height of the progress bar, which was added to to it so that the two background views are continuous.

pocmo:
I think in some of the layouts we just naively assumed a fixed status bar height of 25?

I'm going to close this PR - just so that it's not in the list indefinitely. However it's still going to be linked from the issue to pick it up again at a later point.

Issue #899, opened 2017-07-11
NullPointerException: UrlUtils.isPermittedResourceProtocol()


pocmo:
```
java.lang.NullPointerException: 
  at org.mozilla.focus.utils.UrlUtils.isPermittedResourceProtocol(UrlUtils.java:90)
  at org.mozilla.focus.utils.UrlUtils.isSupportedProtocol(UrlUtils.java:97)
  at org.mozilla.focus.webkit.FocusWebViewClient.shouldOverrideUrlLoading(FocusWebViewClient.java:209)
  at org.mozilla.focus.webkit.WebkitView.loadUrl(WebkitView.java:132)
  at org.mozilla.focus.utils.IntentUtils.handleUnsupportedLink(IntentUtils.java:94)
  at org.mozilla.focus.utils.IntentUtils.handleExternalUri(IntentUtils.java:60)
  at org.mozilla.focus.fragment.BrowserFragment$2.handleExternalUrl(BrowserFragment.java:347)
  at org.mozilla.focus.webkit.FocusWebViewClient.shouldOverrideUrlLoading(FocusWebViewClient.java:211)
  at android.webkit.WebViewClient.shouldOverrideUrlLoading(WebViewClient.java:73)
  at com.android.webview.chromium.WebViewContentsClientAdapter.shouldOverrideUrlLoading(WebViewContentsClientAdapter.java:353)
  at org.chromium.android_webview.AwContentsClient.shouldIgnoreNavigation(AwContentsClient.java:168)
  at org.chromium.android_webview.AwContentsClientBridge.shouldOverrideUrlLoading(AwContentsClientBridge.java:257)
  at org.chromium.base.SystemMessageHandler.nativeDoRunLoopOnce(Native Method:0)
  at org.chromium.base.SystemMessageHandler.handleMessage(SystemMessageHandler.java:41)
  at android.os.Handler.dispatchMessage(Handler.java:102)
  at android.os.Looper.loop(Looper.java:154)
  at android.app.ActivityThread.main(ActivityThread.java:6310)
  at java.lang.reflect.Method.invoke(Native Method:0)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:872)
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:762)
```

Reports: 626
Impacted users: 444

PR #920, opened 2017-07-13
UrlUtils: Scheme passed to isPermittedResourceProtocol() and isSupportedProtocol() can be null. (#899)

pocmo:
In various places we pass the value from Uri.getScheme() to those methods. However getScheme() can return null and in those cases we crash. I haven't been able to actually find a way to trigger
this crash. But it seems to be rare looking at the number of crash reports we received.


pocmo:
> Alas, if only we had null-checking in the type system... ;)

Soon! :)

PR #935, opened 2017-07-14
update codecov setting to allow 1% drop max, with precision to 0 dec place

npark-mozilla:


npark-mozilla:
@pocmo it's the delta, so it means it could drop to 8% in worst case, according to the doc:https://github.com/codecov/support/wiki/Codecov-Yaml

pocmo:
@npark In this case let's get much lower. Even 1% would be a lot :D

PR #971, opened 2017-07-17
Execute codecov on travis instead of buddybuild.

pocmo:


pocmo:
Travis and Codecov are both enabled for our GitHub org. They somehow are able to figure things out with that. ;)

Issue #970, opened 2017-07-17
(BrowserFragment) IllegalStateException: Can not perform this action after onSaveInstanceState


pocmo:
```
java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState

4 | android​.support​.v4​.app | DialogFragment.java​ Line 143 DialogFragment.show
5 | org​.mozilla​.focus​.fragment | BrowserFragment.java ​Line ​510
6 | org​.mozilla​.focus​.fragment | BrowserFragment.java ​Line  501 BrowserFragment.onRequestPermissionsResult
```

![image](https://user-images.githubusercontent.com/89638/28280554-68aa106e-6b24-11e7-9092-4919ff74e638.png)

[View Full Crash Details](https://dashboard.buddybuild.com/apps/584f67f6f3d6eb01000842d6/crashreport/99edb87817d35403857716ea57cfd34c)

pocmo:
@ekager This is related to the downloads dialog. Do you have time to look into this?

pocmo:
https://issuetracker.google.com/issues/37067655

ekager:
Yes I can look into this today!

PR #973, opened 2017-07-18
Changed download dialog from Fragment to Dialog with callback

ekager:
Changed implementation in response to (#970)


pocmo:
The problem I see with using a dialog without a DialogFragment is that Android won't restore the dialog if the activity is getting destroyed while in the background (e.g. test with "don't keep activities" developer setting). Instead the dialog will disappear and Android will log a message that we leaked the previous view:

> android.view.WindowLeaked: Activity org.mozilla.focus.activity.MainActivity has leaked window DecorView@b3a9ac6[] that was originally added here [..] 

In addition I'm hesitant about merging this patch now that we have almost no time left to send this to our beta testers.

I've been playing today with showing the fragment from onResume() instead of onRequestPermissionsResult(). See PR #974. It seems to work. But I haven't tested it thoroughly yet. What do you think?

ekager:
In my research that solution came up, but I thought I would try to follow the form of WebContextMenu, but I guess the location of calls to show WebContextMenu are different. Fine with #974 if it works and can easily be merged! :)
		
		

		Get Outlook for iOS
	




On Tue, Jul 18, 2017 at 3:50 AM -0700, "Sebastian Kaspari" <notifications@github.com> wrote:












The problem I see with using a dialog without a DialogFragment is that Android won't restore the dialog if the activity is getting destroyed while in the background (e.g. test with "don't keep activities" developer setting). Instead the dialog will disappear and Android will log a message that we leaked the previous view:



android.view.WindowLeaked: Activity org.mozilla.focus.activity.MainActivity has leaked window DecorView@b3a9ac6[] that was originally added here [..]



In addition I'm hesitant about merging this patch now that we have almost no time left to send this to our beta testers.


I've been playing today with showing the fragment from onResume() instead of onRequestPermissionsResult(). See PR #974. It seems to work. But I haven't tested it thoroughly yet. What do you think?

PR #982, opened 2017-07-20
Show search term instead of search query URL - Closes #891

### LINKED ISSUE #891: Show search term instead of search query URL
In Fennec's awesomebar, if the user searched for something like "swiss chard recipes", they would get to a results page, with a long search query URL. 

Once the user presses the URL, the search term "swiss chard receipes" is displayed instead of the long search query URL. This helps with re-searching and correcting typos. 

@bbinto do we want something similar here? cc @pocmo

ekager:


app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -62,6 +64,74 @@ public static String createSearchUrl(Context context, String searchTerm) {
...
+            searchTermsOrUrl = url;
+        }
+
+        return searchTermsOrUrl;
+    }
+
+    /**
+     * @return If a url matches the form of a first search URL for the current default search engine.
+     */
+    private static boolean isDefaultSearchUrl(Context context, String urlSearch) {
```
sputnikci: \[PMD\] INFO: The method 'isDefaultSearchUrl' has a Cyclomatic Complexity of 6\.

ekager:
Currently this will check if the search was by the default search engine and if so, will show the search terms instead of the created search URL when the user presses on the URL bar. 

Limitation is that it will only work for the first search with default engine (and subsequent searches if they use this feature to edit or add new search terms because it creates a new search URL) but would not work if they try to change the search terms in the search engine search bar and then select the URL bar, as the search URL probably no longer matches our default search URL formatting. It would also not work if they navigated to a different search engine by themselves, searched, and then clicked the URL bar. Are these limitations blockers?

pocmo:
> Are these limitations blockers?

No, I don't think so. Fennec has this feature too and it has the same limitations afaik.

app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
pocmo: This can be fixed by splitting the functionality into more/helper methods. Maybe using TextUtils (see blow) is already enough.

app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -62,6 +64,74 @@ public static String createSearchUrl(Context context, String searchTerm) {
...
+        if (urlSearch == null || urlSearch.isEmpty()) {
```
pocmo: nit: TextUtils.isEmpty() checks for null and empty strings.

app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -62,6 +64,74 @@ public static String createSearchUrl(Context context, String searchTerm) {
...
+            return false;
+        }
+
+        final SearchEngine searchEngine = SearchEngineManager.getInstance()
+                .getDefaultSearchEngine(context);
+
+        String baseSearchUrl = searchEngine.getBaseSearchUrl();
+        if (baseSearchUrl == null || baseSearchUrl.isEmpty()) {
```
pocmo: nit: TextUtils.isEmpty()

app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -62,6 +64,74 @@ public static String createSearchUrl(Context context, String searchTerm) {
...
+     */
+    private static boolean isDefaultSearchUrl(Context context, String urlSearch) {
+        if (urlSearch == null || urlSearch.isEmpty()) {
+            return false;
+        }
+
+        final SearchEngine searchEngine = SearchEngineManager.getInstance()
+                .getDefaultSearchEngine(context);
+
+        String baseSearchUrl = searchEngine.getBaseSearchUrl();
```
pocmo: nit: final?

app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -62,6 +64,74 @@ public static String createSearchUrl(Context context, String searchTerm) {
...
+        if (baseSearchUrl == null || baseSearchUrl.isEmpty()) {
+            return false;
+        }
+        // %7B is a placeholder for search terms - if it is a default search engine url, the urls should match before the placeholder.
+        if (baseSearchUrl.contains("%7B")) {
+            int searchPlaceholderIndex = baseSearchUrl.lastIndexOf("%7B");
```
pocmo: nit: final

app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -62,6 +64,74 @@ public static String createSearchUrl(Context context, String searchTerm) {
...
+    }
+
+    /**
+     * @return The search parameters decoded from the search url or null if they could not be extracted.
+     */
+    private static String getSearchTermsFromUrl(Context context, String url) {
+        String searchParams = null;
+        Uri uri = Uri.parse(url);
+        try {
+            String queryKey = getSearchQueryKeyForSearchEngine(context);
```
pocmo: nit: final here and for Uri above

app/src/main/java/org/mozilla/focus/search/SearchEngineParser.java
```diff
@@ -99,6 +99,9 @@ private static void readUrl(XmlPullParser parser, SearchEngine searchEngine) thr
             if (tag.equals("Param")) {
                 final String name = parser.getAttributeValue(null, "name");
                 final String value = parser.getAttributeValue(null, "value");
+                if (value.equals("{searchTerms}")) {
+                    searchEngine.searchTermsParamName = name;
+                }
```
pocmo: I guess in theory we could get this value from a SearchEngine already by looping over the query parameters of the first resultsUris?

app/src/main/java/org/mozilla/focus/search/SearchEngineParser.java
```diff
@@ -99,6 +99,9 @@ private static void readUrl(XmlPullParser parser, SearchEngine searchEngine) thr
             if (tag.equals("Param")) {
                 final String name = parser.getAttributeValue(null, "name");
                 final String value = parser.getAttributeValue(null, "value");
+                if (value.equals("{searchTerms}")) {
```
pocmo: We should create a constant for this value.

ekager:
> One issue with this approach seems to be when the website redirects to a different URL. For example using Wikipedia I noticed that they redirect to the article page directly if the search term matches the article title (e.g. search for "Fennec"). 

Hmm... That's true, but if Wikipedia links you to the page you were looking for then you probably would be more likely to want to share the link than to edit your search, no?

> That's an interesting approach (Matching the URL and extracting the keywords). I guess I'd tried to store the last entered text in the BrowsingSession object. But your approach might be more flexible. I'll try this a bit. Did you look at how Fennec implements this? I didn't yet. :)

And I decided to extract from the URL because it would be one less data object to keep track of and I thought it would keep things less complex and more flexible. I also didn't look at how Fennec does this but I can go check it out!

pocmo:
Okay, here's what I think we should do: We take the patch you wrote in and ship it with the next version of Focus. Meanwhile I have been writing some code that is the foundation for supporting multiple sessions (aka multitasking / tabs) at the same time. I think with the updated session code there might be the opportunity to attach the search keywords to a session (and clear them once the user navigates away).

app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -62,6 +64,74 @@ public static String createSearchUrl(Context context, String searchTerm) {
         return searchEngine.buildSearchUrl(searchTerm);
     }
 
+    /**
+     * @return The search terms for the first search performed with the default search engine or the current URL.
+     */
+    public static String getSearchTermsOrUrl(Context context, String url) {
```
pocmo: Can you file a follow-up bug that we write some unit tests for those new helper methods? I'll merge the patch before that so that we get some testing time from QA already.

PR #986, opened 2017-07-20
RTL Language Bullets Display Error Fixes #788

### LINKED ISSUE #788: RTL Language Bullets Display Error
Focus is displaying bullets on left for Urdu Language. As Urdu is RTL Language so bullets should be placed at right side 
unnamed

ekager:
Manually added RTL to CDATA about_content string for fa and ur string files. Other RTL languages (Arabic + Hebrew) seem to be working without this fix.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/InfoFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/InfoFragment.java
@@ -37,6 +37,7 @@ public class InfoFragment extends WebFragment {
     @Override
     public View inflateLayout(LayoutInflater inflater, @Nullable ViewGroup container,
                               @Nullable Bundle savedInstanceState) {
+
         final View view = inflater.inflate(R.layout.fragment_info, container, false);
         progressView = (ProgressBar) view.findViewById(R.id.progress);
         webView = view.findViewById(R.id.webview);
@@ -49,6 +50,8 @@ public class InfoFragment extends WebFragment {
             // external pages, because they are both light-coloured, and significantly slower loading.
             webView.setVisibility(View.INVISIBLE);
         }
+
+        applyLocale();
 
         return view;
     }
--- a/app/src/webkit/java/org/mozilla/focus/webkit/FocusWebViewClient.java
+++ b/app/src/webkit/java/org/mozilla/focus/webkit/FocusWebViewClient.java
@@ -11,6 +11,8 @@ import android.graphics.Bitmap;
 import android.net.Uri;
 import android.net.http.SslError;
 import android.support.v4.util.ArrayMap;
+import android.support.v4.view.ViewCompat;
+import android.view.View;
 import android.webkit.SslErrorHandler;
 import android.webkit.WebResourceRequest;
 import android.webkit.WebResourceResponse;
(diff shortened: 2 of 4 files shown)


app/src/main/res/values-fa/strings.xml
```diff
@@ -20,7 +20,7 @@
...
...575;&#1576;&#8204;&#1607;&#1575; &#1578;&#1594;&#1740;&#1740;&#1585; &#1583;&#1607;&#1740;&#1583;)</li>"\n    "<li>&#1662;&#1575;&#1705; &#1705;&#1606;&#1740;&#1583; &#1578;&#1575; &#1575;&#1591;&#1604;&#1575;&#1593;&#1575;&#1578;&#1740; &#1605;&#1575;&#1606;&#1606;&#1583; &#1705;&#1608;&#1705;&#1740;&#8204;&#1607;&#1575; &#1608; &#1607;&#1605;&#1670;&#1606;&#1740;&#1606; &#1580;&#1587;&#1578;&#8204;&#1608;&#1580;&#1608;&#1607;&#1575; &#1608; &#1578;&#1575;&#1585;&#1740;&#1582;&#1670;&#1607; &#1605;&#1585;&#1608;&#1585; &#1581;&#1584;&#1601; &#1588;&#1608;&#1606;&#1583;</li>"\n  "</ul>\n</p>\n<p xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">%1$s &#1578;&#1608;&#1587;&#1617;&#1591; &#1605;&#1608;&#1586;&#1740;&#1604;&#1575; &#1605;&#1606;&#1578;&#1588;&#1585; &#1588;&#1583;&#1607; &#1575;&#1587;&#1578;. &#1605;&#1571;&#1605;&#1608;&#1585;&#1740;&#1578; &#1605;&#1575; &#1578;&#1575; &#1585;&#1608;&#1575;&#1580; &#1740;&#1705; &#1575;&#1740;&#1606;&#1578;&#1585;&#1606;&#1578; &#1587;&#1575;&#1604;&#1605;&#1548; &#1576;&#1575;&#1586; &#1575;&#1587;&#1578;.<br/>\n<a href="%2$s">&#1576;&#1740;&#1588;&#1578;&#1585; &#1576;&#1583;&#1575;&#1606;&#1740;&#1583;</a></div></p>]]></string>
```
pocmo: Making this change in the string files is problematic: They are automatically imported from the L10N repo and translators use a bunch of different tools to translate them. Even if we convince those translators to update the strings then this problem might come back as soon as we add other RTL languages.

Initially I made this content RTL compatible by adding `dir=auto` to the body tag here:
https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/res/raw/about.html#L10

I wonder if we could instead insert ltr/rtl based on the current configuration (like we insert other strings into this HTML content). It seems like "auto" doesn't work for all our supported locales. :(

ekager: Okay that makes sense, I'll look into dynamically inserting ltr/rtl into about.html !

pocmo: I think we are using ViewCompat.getLayoutDirection() somewhere in the code already.

ekager:
I did work on this today and found that the main problem was InfoFragment was not being updated with the new configuration when the language was changed within the app, (It would always log as "en") so I added an applyLocale method. Additionally, we need to dynamically add the tag for the language direction in the html file as “auto” doesn’t work on its own, even after updating the configuration.

app/src/main/java/org/mozilla/focus/fragment/InfoFragment.java
```diff
@@ -50,10 +57,24 @@ public View inflateLayout(LayoutInflater inflater, @Nullable ViewGroup container
             webView.setVisibility(View.INVISIBLE);
         }
 
+        applyLocale();
```
pocmo: This fragment already extends from LocaleAwareFragment. I wonder if we should move this into LocaleAwareFragment - I guess all fragments would need to do this?

app/src/webkit/java/org/mozilla/focus/webkit/FocusWebViewClient.java
```diff
@@ -304,6 +306,18 @@ private void loadAbout(final WebView webView) {
         final String wordmark = HtmlLoader.loadPngAsDataURI(webView.getContext(), R.drawable.wordmark);
         substitutionMap.put("%wordmark%", wordmark);
 
+        String direction = "auto";
```
pocmo: nit: If you do not set "`auto`" here then you can define `direction` as `final` and set `auto` from an '`else`' branch in your code below.

app/src/main/java/org/mozilla/focus/fragment/InfoFragment.java
ekager: I moved the applyLocale method code into WebFragment, but I don't think any other fragments that currently extend WebFragment or LocaleAwareFragment would need the same update. The method is abstract within LocaleAwareFragment since different fragments updates differently when the locale is changed.

Issue #651, opened 2017-05-25
TransactionTooLargeException in Android code


pocmo:
Buddybuild detected a crash

[View Full Crash Details](https://dashboard.buddybuild.com/apps/584f67f6f3d6eb01000842d6/crashreport/953a29ee4a3aa8b79c2e09d64bb5920f)

ahunt:
I don't know if/how we triggered this, all the threads are in Android framework code.

ahunt:
These crashes happened on a pixel. One other pixel user has complained about the app hanging and eventually crashing, also on a Pixel, when using the menu. We're creating intents when inflating the menu, so I wonder if those reports are related to this crash.

See also:
https://developer.android.com/reference/android/os/TransactionTooLargeException.html

pocmo:
We saw some new crashes. All happening on a single Google Pixel XL running Android 7.1.2.

pocmo:
This happened on Joe's device and we were able to track it down. It's not actually a visible crash. This happens in the background when the app is killed by swiping it away in "recent apps".

Some further details:
* https://www.reddit.com/r/androiddev/comments/53690f/transactiontoolargeexception_crashes_on_nougat/
* http://blog.sqisland.com/2016/09/transactiontoolargeexception-crashes-nougat.html
* https://stackoverflow.com/questions/39098590/android-os-transactiontoolargeexception-on-nougat

From the postings it looks like the data saved in onSaveInstanceState might be too large. As we are not really permanently saving data and as the user is killing the app in this case (with no visible crash) we can look at this after the v1 release.

liuche:
Looks like the other v1.1 issues are all being investigated or handled, so I'll investigate this issue and add any more information I learn.

pocmo:
Okay, I have an idea how to fix this. :)

PR #1006, opened 2017-07-25
Keep WebView state in memory when view gets destroyed. (#651)

pocmo:
The issue in #651 is happening when WebView's state is too large to be saved in the state bundle that Android will take care of. Instead we are keeping the state in memory and only store a UUID in the state bundle.


app/src/webkit/java/org/mozilla/focus/webkit/WebkitView.java
```diff
@@ -78,10 +82,18 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
     }
 
     @Override
-    public void restoreWebviewState(Bundle savedInstanceState) {
-        // We need to have a different method name because restoreState() returns
-        // a WebBackForwardList, and we can't overload with different return types:
-        final WebBackForwardList backForwardList = restoreState(savedInstanceState);
+    public void restoreWebViewState(Bundle inBundle) {
```
sputnikci: \[PMD\] INFO: The method 'restoreWebViewState' has a Cyclomatic Complexity of 5\.

PR #1013, opened 2017-07-25
Added correct telemetry for Download Dialog

ekager:
Realized I only added positive telemetry and we're not recording download dialog _cancel_ events. I also noticed the method should probably be "click" instead of "download" but if it's too difficult to change this because it's already "out in the wild" and would change telemetry reports, I can take that change out.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/DownloadDialogFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/DownloadDialogFragment.java
@@ -78,7 +78,7 @@ public class DownloadDialogFragment extends DialogFragment {
                     @Override
                     public void onClick(View v) {
                         sendDownloadDialogButtonClicked(pendingDownload, shouldDownload);
-                        TelemetryWrapper.downloadDialogDownloadEvent();
+                        TelemetryWrapper.downloadDialogDownloadEvent(shouldDownload);
                         dismiss();
                     }
                 });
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/fragment/DownloadDialogFragment.java
```diff
@@ -78,7 +78,11 @@ private void setButtonOnClickListener(Button button, final Download pendingDownl
                     @Override
                     public void onClick(View v) {
                         sendDownloadDialogButtonClicked(pendingDownload, shouldDownload);
-                        TelemetryWrapper.downloadDialogDownloadEvent();
+                        if (shouldDownload) {
+                            TelemetryWrapper.downloadDialogDownloadEvent();
+                        } else {
+                            TelemetryWrapper.downloadDialogCancelEvent();
+                        }
```
pocmo: Another option here would be to pass "shouldDownload" to TelemetryWrapper.downloadDialogDownloadEvent() and decide what event to build/send in there. Just for making our main code simpler I'd try to reduce telemetry to one line as much as possible. :)

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java
```diff
@@ -249,11 +248,11 @@ public static void customTabsIntentEvent(final List<String> options) {
     }
 
     public static void downloadDialogDownloadEvent() {
-        TelemetryEvent.create(Category.ACTION, Method.DOWNLOAD, Object.DOWNLOAD_DIALOG, Value.DOWNLOAD).queue();
+        TelemetryEvent.create(Category.ACTION, Method.CLICK, Object.DOWNLOAD_DIALOG, Value.DOWNLOAD).queue();
```
pocmo: I'm okay with this change but I'd ask @bbinto to make sure we are not messing with her analytics.

Please also update the wiki after landing those probes:
https://github.com/mozilla-mobile/focus-android/wiki/Telemetry

PR #1018, opened 2017-07-26
Home Screen shortcut, use Photon colors

antlam:
Keeping the same shape, going with purple 80 and finessing the drop shadow to better match our app icon.


antlam:
ic_homescreen_shape

PR #1035, opened 2017-07-28
Decode URI special characters before creating File - closes #1011

### LINKED ISSUE #1011: Opening downloaded APK from snackbar shows error: "There was a problem parsing the package."
When downloading an APK (usually an update I get from buddybuild) and trying to install the APK from the Snackbar ("open" action) an error dialog pops up saying: "There was a problem parsing the package.". When opening the APK from the download manager notification everything is fine though.

ekager:
In reference to (#1011), this includes a check to see if the file exists before showing the snackbar and I added Uri decoding to take care of special properties that should only appear in the Uri and not when creating the File or displaying the file name.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
+++ b/app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
@@ -63,11 +63,10 @@ public class DownloadBroadcastReceiver extends BroadcastReceiver {
                     final String localUri = uriString.startsWith(FILE_SCHEME) ? uriString.substring(FILE_SCHEME.length()) : uriString;
                     final String fileExtension = MimeTypeMap.getFileExtensionFromUrl(localUri);
                     final String mimeType = MimeTypeMap.getSingleton().getMimeTypeFromExtension(fileExtension);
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/broadcastreceiver/DownloadBroadcastReceiver.java
```diff
@@ -63,13 +63,14 @@ private void displaySnackbar(final Context context, long completedDownloadRefere
...
-
-                    final File file = new File(localUri);
-                    final Uri uriForFile = FileProvider.getUriForFile(context, BuildConfig.APPLICATION_ID + FILE_PROVIDER_EXTENSION, file);
-
-                    final Intent openFileIntent = IntentUtils.createOpenFileIntent(uriForFile, mimeType);
-                    showSnackbarForFilename(openFileIntent, context, fileName);
+                    final String fileName = URLUtil.guessFileName(Uri.decode(localUri), null, mimeType);
+
+                    final File file = new File(Uri.decode(localUri));
+                    if (file.exists()) {
```
pocmo: I'd either remove this check - now that we expect this file to always exist.

Or: Throw an exception (like IllegalStateException) so that we see a crash and know that there are still issues.

Issue #863, opened 2017-07-05
Build with Android O SDK (Android 8.0 / API 26)


pocmo:
This is at least needed to support home screen shortcuts on Android O (#833).

I tried using the new SDK in June (#758). Building worked fine but running our Robolectric tests failed.

* [x] Update SDK and other dependencies like support library in build.gradle files
* [x] Use new Shortcut API to implement #833
* [x] Use startForegroundService() to show permanent notification (this depends on multitasking refactoring)
* [x] Create a notification channel for our on-going notification
* [x] Fix other behavior changes or deprecated code: https://developer.android.com/preview/behavior-changes.html
* [x] Fix automated tests that depend on a different SDK version
* [x] Replace "placeholder code" that doesn't work with newest build tools anymore
* [x] Finalize notification channel strings (See comment with screenshots below)

ekager:
Would we also have to refactor now-deprecated methods? If this isn't too complicated I think it would be great to be able to include homescreen shortcuts (#833) for Android O+ with the ShortcutManager. 👍

pocmo:
If it's straight-forward to do. Otherwise I'd be okay to do it later. I'm more worried about some behavior changes in the SDK that break some of the existing features (like the ongoing notification). I'll try to get patches up for that soon! :)

pocmo:
@antlam @mheubusch While Focus is active we show an on-going notification. Now on Android 8+ we need to create a so called "notification channel" for all notifications. Whenever we display a notification we need to assign it to a channel. A user can configure how (or if) notifications from a specific channel show up. Here's a pretty good article about the feature:
https://www.androidcentral.com/notification-channels

For our on-going notification we need to create a channel with a "name" and an (optional) description. The Android guidelines say:
* name: The recommended maximum length is 40 characters; the value may be truncated if it is too long.
* description: The recommended maximum length is 300 characters; the value may be truncated if it is too long.

I have absolutely no clue how we should name this thing.

However here are some screenshots from Android's UI showing the strings:

Showing all notification channels of the app. "Session" is how I named the channel in my build:
![overview](https://user-images.githubusercontent.com/89638/29579878-7571b628-8774-11e7-96f8-3a55df304828.png)

This is the screen shown when tapping on the channel to configure it. "Text text" is the description of the channel I used in my build:
![channel](https://user-images.githubusercontent.com/89638/29579925-9473b684-8774-11e7-86cf-8c1884628689.png)

And just in case, this is the notification we show:
[...]

PR #1036, opened 2017-07-28
Add to Homescreen 

### LINKED ISSUE #833: Add to home screen
**User Story**

> I want to easily go back to websites I regularly visit in Focus (without ads)

> As a Focus user, I want to easily frequent my favorite website so that I don't have to type in the full-length, same, URL each time

**Tasks**

- [x] UX design / mocks
- [x] ~~Implement additional UI~~ (Not part of the first iteration)
- [x] Add functionality for adding pages to the home screen
- [x] Get a good icon representing the website
- [x] Generate a fallback icon
- [x] Add Telemetry probes
- [x] Add support for new shortcut API in Android O (Requires SDK 26 -> #863)
- [x] Add dialog to update the title of the shortcut
- [x] Finalize strings

ekager:
I have the Android O code, but I will create a new PR for it and we could block it until it's ready for merge (#863). I will also update the telemetry wiki after merge.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -21,6 +21,7 @@ import org.mozilla.focus.fragment.HomeFragment;
 import org.mozilla.focus.fragment.UrlInputFragment;
 import org.mozilla.focus.locale.LocaleAwareAppCompatActivity;
 import org.mozilla.focus.notification.BrowsingNotificationService;
+import org.mozilla.focus.shortcut.HomeScreen;
 import org.mozilla.focus.telemetry.TelemetryWrapper;
 import org.mozilla.focus.utils.SafeIntent;
 import org.mozilla.focus.utils.Settings;
@@ -250,6 +251,8 @@ public class MainActivity extends LocaleAwareAppCompatActivity {
 
         if (intent.getBooleanExtra(EXTRA_TEXT_SELECTION, false)) {
             TelemetryWrapper.textSelectionIntentEvent();
+        } else if (intent.hasExtra(HomeScreen.ADD_TO_HOMESCREEN_TAG)) {
+            TelemetryWrapper.openHomescreenShortcutEvent();
         } else if (BrowsingSession.getInstance().isCustomTab()) {
             TelemetryWrapper.customTabsIntentEvent(BrowsingSession.getInstance().getCustomTabConfig().getOptionsList());
         } else {
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -14,7 +14,6 @@ import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.pm.ActivityInfo;
 import android.content.pm.PackageManager;
-import android.graphics.Bitmap;
 import android.graphics.Color;
 import android.graphics.drawable.Drawable;
 import android.graphics.drawable.TransitionDrawable;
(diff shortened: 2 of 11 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -802,16 +822,16 @@ public void onClick(View view) {
...
             case R.id.add_to_homescreen:
-                final IWebView webView = getWebView();
-                if (webView == null) {
-                    return;
-                }
-
-                final Bitmap icon = webView.getIcon();
-                final String url = webView.getUrl();
-                final String title = webView.getTitle();
+                if (!isLoading()) {
```
pocmo: Why are we not showing this dialog if we are still loading? Shouldn't we hide or disable the menu item in this case too?

app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.java
```diff
@@ -20,45 +19,24 @@
...
      * on top of a generic launcher icon shape that we provide.
      */
-    /* package */ static Bitmap generateLauncherIcon(Context context, Bitmap rawIcon, String url) {
+    public static Bitmap generateLauncherIcon(Context context, String url) {
         final BitmapFactory.Options options = new BitmapFactory.Options();
         options.inMutable = true;
 
         final Bitmap shape = BitmapFactory.decodeResource(context.getResources(), R.drawable.ic_homescreen_shape, options);
 
         final Bitmap innerIcon;
```
pocmo: nit: technically this isn't an **inner** icon anymore.

app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.java
```diff
@@ -20,45 +19,24 @@
...
 
-        if (rawIcon == null || rawIcon.getWidth() < MINIMUM_INNER_ICON_SIZE || rawIcon.getHeight() < MINIMUM_INNER_ICON_SIZE) {
-            innerIcon = generateInnerIcon(context, url);
-        } else {
-            final int targetWidth = Math.min(rawIcon.getWidth() * 2, INNER_ICON_SIZE);
-            final int targetHeight = Math.min(rawIcon.getHeight() * 2, INNER_ICON_SIZE);
-
-            innerIcon = Bitmap.createScaledBitmap(rawIcon, targetWidth, targetHeight, true);
-        }
+        innerIcon = generateInnerIcon(context, url);
```
pocmo: nit: This could move to the declaration now.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
ekager: Yes changed this to not show the menu item instead!

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -141,6 +142,9 @@ protected void onNewIntent(Intent unsafeIntent) {
             // We can't update our fragment right now because we need to wait until the activity is
             // resumed. So just remember this URL and load it in onResumeFragments().
             pendingUrl = intent.getDataString();
+            if (intent.hasExtra(HomeScreen.ADD_TO_HOMESCREEN_TAG)) {
+                TelemetryWrapper.openHomescreenShortcutEvent();
+            }
```
pocmo: There's a problem here: This telemetry probe is only fired when the app is already running (onNewIntent). And additionally we send the telemetry probe for the VIEW intent.

We could move this into show showBrowserScreen() [where all incoming URLs go through] as an additional else branch:
https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/activity/MainActivity.java#L230

Btw. you can test telemetry probes locally (and I should really document this somewhere) by setting telemetryEnabled = true [1] and replacing the upload client with a DebugLogClient [2]. The next time you leave the app this should print the telemetry ping to the log instead of uploading it.

[1] https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java#L149
[2] https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.java#L171

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
ekager: @antlam changed this, and now we will grey out the menu item for add to homescreen if the user opens the menu while the page is loading.

PR #1037, opened 2017-07-28
Add to Homescreen for Android O

ekager:
Blocked until (#863)


pocmo:
@ekager After #1194 is merged we can finally get this in too. :)

Issue #894, opened 2017-07-11
Filter out Focus in the BottomSheet when "Open with..." is clicked


mTwTm:
It makes no sense to open with Focus again so this will help simplify/clarify the list. Furthermore this may also simplify the flow since opening the same url again can be a source of bugs in the future which is unnecessary. We shall take care of the case when Focus is the only browser available, though.

ioana-chiorean:
Just saw this - and I agree
Curious fact is that i see it only on 2/3 devices 

<img src="https://user-images.githubusercontent.com/1391451/28212504-61390faa-68aa-11e7-9b96-927d445cad60.jpg" width="550" height="300" />

mitchellkager:
look into On my fork, I offered a quick and dirty solution to this problem. Can't think of any way to prevent AppCompatDialogFragment from providing same package in getArguments(), so ignoring may be the best action. Tangentially: what behaviour should occur when there are no applications other than Focus available?

One note is that my solution compares to see if the packageName of each AppInfo is equal to the current application's packageName: if you are using, for example, Gecko, you will still be prompted to open the app in Focus. Could potentially instead filter based on the existence of the expression "org.mozilla.focus" within the AppInfo's packageName.

pocmo:
The filtering should probably already happen in Browsers.java so that we do not show this menu item if there's no other browser.

mitchellkager:
Ended up migrating logic to Browsers.java - fit cleaner there anyway. Now, need to look for a place where one can check to see if any browsers are available, choosing not to display the "Open in" menu option if none are.

pocmo:
This is what happens here:
https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/menu/BrowserMenuAdapter.java#L83

PR #1045, opened 2017-07-31
Filter out Focus in the BottomSheet when "Open with..." is clicked #894

mitchellkager:
Apply filter so that current app (Mozilla Focus or Gecko) is not included from Browsers' 'resolveBrowsers' method. This prevents the current app from ever having opportunity to be included in 'Open with...' menu. Additionally, provided very simple update to logic in 'hasMultipleThirdPartyBrowsers' - had to adjust to check for just two other browsers (as opposed to 3) since current browser is no longer contributing to list size.

Worth noting that this will still include Focus when using Gecko (and vice versa). Did not necessarily consider this a bug. In order to change this, I would adjust filter within resolveBrowsers to exclude both browsers based on commonality in package names (org.mozilla.focus is present in both). This is a sufficiently unique key which I feel would always appropriately filter out browsers, should it be decided that the sibling app not be included in browser list.


pocmo:
Overall this change looks good to me! However the commits are a bit all over the place. Can you squash your changes into a single commit and push it? Thank you for your contribution! 👍 

Some hints:
* https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git
* https://stackoverflow.com/questions/6934752/combining-multiple-commits-before-pushing-in-git

PR #1046, opened 2017-07-31
Allow opening of shared URLs - closes #889

### LINKED ISSUE #889: Allow opening of shared URLs
Currently Focus only handles Intents to **open** a specific URL. Additionally Focus could also handle Intents when an URL is **shared**. That's what Fennec does too (Shows up as "Add to Firefox").

I saw this SUMO request.
https://support.mozilla.org/en-US/questions/1167043

Technically the SHARE intent is not required to contain an URL. It could be just normal text. In this case we could trigger a search?

ekager:
Will update the Telemetry Wiki when approved and merged!

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -9,6 +9,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
 import android.support.v4.app.FragmentManager;
+import android.text.TextUtils;
 import android.util.AttributeSet;
 import android.view.View;
 import android.view.WindowManager;
@@ -23,6 +24,7 @@ import org.mozilla.focus.notification.BrowsingNotificationService;
 import org.mozilla.focus.telemetry.TelemetryWrapper;
 import org.mozilla.focus.utils.SafeIntent;
 import org.mozilla.focus.utils.Settings;
+import org.mozilla.focus.utils.UrlUtils;
 import org.mozilla.focus.web.BrowsingSession;
 import org.mozilla.focus.web.IWebView;
 import org.mozilla.focus.web.WebViewProvider;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -143,6 +145,23 @@ protected void onNewIntent(Intent unsafeIntent) {
             pendingUrl = intent.getDataString();
         }
 
+        if (Intent.ACTION_SEND.equals(intent.getAction())) {
```
pocmo: onNewIntent() is only called if the app is already running. Try killing Focus and then sharing to Focus. You will need to add a second entry point from onCreate(). See VIEW handling.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -143,6 +145,23 @@ protected void onNewIntent(Intent unsafeIntent) {
...
+            String dataString = intent.getStringExtra(Intent.EXTRA_TEXT);
+            if (!TextUtils.isEmpty(dataString)) {
+                boolean isUrl = UrlUtils.isUrl(dataString);
```
pocmo: nit: final

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -143,6 +145,23 @@ protected void onNewIntent(Intent unsafeIntent) {
             pendingUrl = intent.getDataString();
         }
 
+        if (Intent.ACTION_SEND.equals(intent.getAction())) {
+            String dataString = intent.getStringExtra(Intent.EXTRA_TEXT);
```
pocmo: nit: final

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -143,6 +145,23 @@ protected void onNewIntent(Intent unsafeIntent) {
...
+                boolean isUrl = UrlUtils.isUrl(dataString);
+                TelemetryWrapper.shareIntentEvent(isUrl);
+                if (isUrl) {
+                    // We can't update our fragment right now because we need to wait until the activity is
+                    // resumed. So just remember this URL and load it in onResumeFragments().
+                    pendingUrl = dataString;
+                } else {
+                    // We can't update our fragment right now because we need to wait until the activity is
+                    // resumed. So just remember this URL and load it in onResumeFragments().
+                    pendingUrl = UrlUtils.createSearchUrl(this, dataString);
```
pocmo: nit: Maybe short syntax is more compact / readable here:

pendingUrl = isUrl ? .. : ..;

Issue #870, opened 2017-07-06
Add notification action for 'Erase + Start from Home View'


antlam:
From @bbinto (and I'm paraphrasing obvs) 

> What if we had another notification action that allowed the user to 'Erase' their current browsing session **and** 'Open' Focus at the same time? Sometimes I want to just fire off a new search with Focus but I'm in another app.

antlam:
@pocmo how hard would this be?

@mheubusch this is going to be a bit more complicated than 'Open'. Do you have thoughts on this?

pocmo:
Stop. No new features in 1.1 (I removed the label and added the triage label).

I wonder if this could be something for a shortcut action?

Anyhow: I think that's doable. And after #832 this will be "just" switching to the app and erase. Immediately being able to type a new URL / search will be the default for all.

bbinto:
![](https://github.trello.services/images/mini-trello-icon.png) [Improve notification actions](https://trello.com/c/5vOsYjHR/58-improve-notification-actions)

pocmo:
@antlam @bbermes Maybe this is also a good thing for adding as a shortcut?

antlam:
Shortcuts from the Home screen? Would it be something like "Erase" and "Erase and start a new search"? that could work!... 🤔

bbinto:
@pocmo yes, in addition to the existing shortcut (Erase), we could add the "Erase & start new search". However, would this still make sense though after we do #832?

mheubusch:
@antlam what do you think of 
Firefox Focus
Erase history and open app
(although not sure how this will work on my Samsung - there is an "OPEN" that opens the app with the last site visited appearing in the browser)

mheubusch:
![screenshot_2017-07-27-17-46-59](https://user-images.githubusercontent.com/20115914/28696328-8965f358-72fa-11e7-8287-19de2b104f05.png)

pocmo:
I'll add the string to the repo (and export) today.

@antlam We also need to decide what icon we want to show in the notification on older devices. The firefox like for open - or the trash can? or something else?

antlam:
@pocmo is it ok if we dont supply an icon?

pocmo:
@antlam The API for adding an action to a notification requires supplying an icon. On my Z3C it works if I do not specifiy a valid icon resource id (see screenshot below). However it's not mentioned in the docs that this is supported. So this could be fragile. We would need to test on a variety of devices and (older) Android versions to be sure.

[...]

PR #1050, opened 2017-08-01
Add "Erase and open" notification action. (#870)

pocmo:


pocmo:
Yeah, the flash is annoying. I actually thought about delaying the animation. But I'll ping antlam to see whether he wants the page to show up (and animate without flash) or if we should pretend the session never existed. I think you have a good point about not showing the previous content at all.

pocmo:
Mr. UX says: Show start screen immediately and snackbar. I'll update the patch!

pocmo:
@ekager I updated the patch. :)

PR #1051, opened 2017-08-01
Filter out Focus in the BottomSheet when "Open with..." is clicked #894

mitchellkager:
Apply filter so that current app (Mozilla Focus or Gecko) is not included from Browsers' 'resolveBrowsers' method. This prevents the current app from ever having opportunity to be included in 'Open with...' menu. Additionally, provided very simple update to logic in 'hasMultipleThirdPartyBrowsers' - had to adjust to check for just two other browsers (as opposed to 3) since current browser is no longer contributing to list size.

Worth noting that this will still include Focus when using Gecko (and vice versa). Did not necessarily consider this a bug. In order to change this, I would adjust filter within resolveBrowsers to exclude both browsers based on commonality in package names (org.mozilla.focus is present in both). This is a sufficiently unique key which I feel would always appropriately filter out browsers, should it be decided that the sibling app not be included in browser list.

Created a new pull request as I am now working with a new (much better organized) repository.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/Browsers.java
+++ b/app/src/main/java/org/mozilla/focus/utils/Browsers.java
@@ -76,7 +76,7 @@ public class Browsers {
 
         final Uri uri = Uri.parse(url);
 
-        final Map<String, ActivityInfo> browsers = resolveBrowsers(packageManager, uri);
+        final Map<String, ActivityInfo> browsers = resolveBrowsers(context, packageManager, uri);
 
         // If there's a default browser set then modern Android systems won't return other browsers
         // anymore when using queryIntentActivities(). That's annoying and our only option is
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/utils/Browsers.java
```diff
@@ -71,8 +71,11 @@
     // installed then higher priority one will be stored here
     private ActivityInfo firefoxBrandedBrowser;
 
+    private Context context;
+
     public Browsers(Context context, @NonNull String url) {
         final PackageManager packageManager = context.getPackageManager();
+        this.context = context;
```
pocmo: Sorry, I missed this before. Please just pass the context as an argument to resolveBrowsers() instead of keeping a reference to it. The reason for that is that the Context we pass in is often an Activity object and this Browser object here might live longer than the Activity. If we keep a reference to the Activity here then we are keeping the (destroyed) Activity in memory.

pocmo: Ah, I think your previous patch actually did that :)
https://github.com/mozilla-mobile/focus-android/pull/1045/files

FYI: You can use "git commit --amend" to update your existing commit and then (force) push the branch to your repository to update this PR.

Issue #1065, opened 2017-08-02
When google is set to search default, search term is not displayed outside U.S.


npark-mozilla:
Found this while verifying #891 

I'm in Canada, so it goes to google.ca site (not sure whether this matters)

STR:
- Set the default search to Google
- Enter search term, press enter to go to the search result webpage
- tap URL bar

Expected:
- Show highlighted search term
Actual:
- highlights actual long URL

ekager:
google.ca isn't in our ca search resource folder which I use to check if it's a default search. google.ca is automatically used for you when you search from the focus homescreen? https://github.com/mozilla-mobile/focus-android/tree/master/app/src/main/assets/search/ca

npark-mozilla:
Oh sorry, let me clarify, I chose google as the search engine in Settings, and since I'm in Canada, it uses google.ca and not google.com.

pocmo:
Duplicate of #891

PR #1070, opened 2017-08-02
Modified show search terms or URL (#1065)

ekager:
While we are probably changing the implementation in the future, this is more flexible than before and should work for different GeoIP pickups. I also moved getRepresentativeSnippet to UrlUtils.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
+++ b/app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
@@ -18,6 +18,7 @@ import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URLDecoder;
+import java.util.List;
 
 public class UrlUtils {
     public static String normalize(@NonNull String input) {
@@ -67,11 +68,12 @@ public class UrlUtils {
     /**
      * @return The search terms for the first search performed with the default search engine or the current URL.
      */
-    public static String getSearchTermsOrUrl(Context context, String url) {
+    public static String getSearchTermsOrUrl(String url) {
         String searchTermsOrUrl = null;
+        final SearchEngine foundSearchEngine = UrlUtils.isDefaultSearchUrl(url);
 
-        if (UrlUtils.isDefaultSearchUrl(context, url)) {
-            searchTermsOrUrl = UrlUtils.getSearchTermsFromUrl(context, url);
+        if (foundSearchEngine != null) {
+            searchTermsOrUrl = UrlUtils.getSearchTermsFromUrl(url, foundSearchEngine);
         }
 
         if (TextUtils.isEmpty(searchTermsOrUrl)) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -82,38 +84,32 @@ public static String getSearchTermsOrUrl(Context context, String url) {
     }
 
     /**
-     * @return If a url matches the form of a first search URL for the current default search engine.
+     * @return The detected search engine or null if one was not matched with the URL
      */
-    private static boolean isDefaultSearchUrl(Context context, String urlSearch) {
+    private static SearchEngine isDefaultSearchUrl(String urlSearch) {
```
pocmo: nit: I expect methods that start with "is" to always return a boolean. Maybe we should rename this method now.

app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -82,38 +84,32 @@ public static String getSearchTermsOrUrl(Context context, String url) {
...
-            final int searchPlaceholderIndex = baseSearchUrl.lastIndexOf("%7B");
-            return urlSearch.startsWith(baseSearchUrl.substring(0, searchPlaceholderIndex));
-        } else {
-            return false;
+        final String representativeSnippet = UrlUtils.getRepresentativeSnippet(urlSearch);
+        final List<SearchEngine> searchEngines = SearchEngineManager.getInstance().getSearchEngines();
+        for (SearchEngine searchEngine : searchEngines) {
+            if (representativeSnippet.contains(searchEngine.getName().toLowerCase())) {
+                return searchEngine;
+            }
```
pocmo: I wonder why we iterate over all search engines now?

Also I'm a bit afraid that checking whether the name of the search engine is in the URL is a bit fragile. For example the name for the German Wikipedia search engine is "Wikipedia (de)".

app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -82,38 +84,32 @@ public static String getSearchTermsOrUrl(Context context, String url) {
...
      */
-    private static String getSearchTermsFromUrl(Context context, String url) {
+    private static String getSearchTermsFromUrl(String url, SearchEngine searchEngine) {
         String searchParams = null;
         final Uri uri = Uri.parse(url);
         try {
-            final String queryKey = getSearchQueryKeyForSearchEngine(context);
-            if (queryKey != null) {
+            final String queryKey = searchEngine.getSearchTermsParamName();
+            if (queryKey != null && url.contains(queryKey)) {
```
pocmo: Afaik the param name for Google is just "q". It's likely that there's a URL that contains google.com and a "q". You already have an `Uri` object here, you might be able to ask it if there's an actual query parameter with the name "q". But nevertheless it could happen that there are URLs with google.com and a "q" parameter that are not actually search result pages. :(

app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -82,38 +84,32 @@ public static String getSearchTermsOrUrl(Context context, String url) {
...
+            }
         }
+        return null;
     }
 
     /**
      * @return The search parameters decoded from the search url or null if they could not be extracted.
      */
-    private static String getSearchTermsFromUrl(Context context, String url) {
+    private static String getSearchTermsFromUrl(String url, SearchEngine searchEngine) {
```
pocmo: We'd now always show the search terms even if I haven't started a search but when I'm on a search page, right? I think that's okay too, but we should confirm with UX that this behavior is okay.

pocmo:
Closing this PR - I guess we are switching to your new approach eventually?

PR #1088, opened 2017-08-04
add search term display check on the URL bar

npark-mozilla:


npark-mozilla:
will enable the google check once #1065 is fixed. Or, since the fix might be available soon, I should wait merging until then.

npark-mozilla:
waiting until next release

pocmo:
@npark-mozilla This should be unblocked now.

PR #1093, opened 2017-08-04
Created variables to keep track of search instead of parsing URLs (#1…

ekager:
…065)
Different implementation that instead of parsing URLs and search engine keys, keeps track of the search terms and whether the action was a search in static variables. This should work for different geoIPs.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -75,11 +75,18 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
     private static int REQUEST_CODE_STORAGE_PERMISSION = 101;
     private static final int ANIMATION_DURATION = 300;
     private static final String ARGUMENT_URL = "url";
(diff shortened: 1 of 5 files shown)


ekager:
I think we should disable the feature for now if everyone's okay with that, I am much prouder of this way to do this, and we should keep behavior consistent not let it be working for only some users. We could polish it and then release and advertise it for the next release along with multitasking!

pocmo:
Alright. I talk to barbara and we can also discuss this in the Focus meeting tomorrow.

pocmo:
Can you prepare a patch to disable this for this release?

ekager:
Yes, working on it now!

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -172,7 +187,7 @@ public View inflateLayout(LayoutInflater inflater, @Nullable ViewGroup container
 
         urlView = (TextView) view.findViewById(R.id.display_url);
         updateURL(getInitialUrl());
-
+
```
sputnikci: \[Checkstyle\] ERROR: Trailing whitespace

PR #1112, opened 2017-08-09
remove the search term display check

npark-mozilla:


npark-mozilla:
This fixes the current test failures in master

PR #1133, opened 2017-08-11
Fix default browser switch

### LINKED ISSUE #1125: Default browser setting off even though Focus is my default
On my Android O device i've set Focus as my default browser but the switch in the settings is off.

joydeepsaha05:
Fixes #1125 

I've tested this on Android M and N.


joydeepsaha05:
@pocmo The problem in the Browsers.resolveBrowsers() method, the Focus package name isn't added to the list of browsers and when this list is queried for the Focus package name, it obviously doesn't find it.

I've made the requested changes (changed the Browsers.findDefault() method to get the default browser directly).

pocmo:
We had to make a Focus 1.2 RC 3 build to fix this issue immediately for the upcoming version (a2cca5aefe3fc2482d6fdad71627d33ce72eb06f). Therefore I unfortunately hadn't the time to help push this patch over the finish line. Sorry!

PR #1183, opened 2017-08-21
Import translations from L10N repository.

pocmo:


pocmo:
Two errors in this import:

> [..]
> 48 Checking for placeholders in pl
> 49  * [values-pl/strings.xml] missing placeholder in translation, key: firstrun_shortcut_text
> [..]
> 66 Checking for placeholders in sk
> 67  * [values-sk/strings.xml] missing placeholder in translation, key: firstrun_shortcut_text
> [..]

flodolo:
@piotrdrag and @jurajcigan (and @delphine)
Looks like we have a problem with a string in pl and sk, and removing one instance of the placeholder causes errors.

@pocmo 
Is this a warning or generates errors?

jurajcigan:
I have edited the Slovak string. It should be without errors now.

piotrdrag:
The sentence would feel much more natural without the second variable, but I added it for now.

pocmo:
Thanks! I re-imported the strings and now the build passes.

> Is this a warning or generates errors?

This is from our self-written script to check that we do not land strings that do not contain the same placeholders as the template string. We didn't think of this special case (the same placeholder twice) and Android probably has no problem if we omit one of them in a translation. If this is something we should allow then let's file an issue and we can fix that!

flodolo:
> If this is something we should allow then let's file an issue and we can fix that!

That seems to make sense.

pocmo:
Okay, I filed #1186.

pocmo:
@piotrdrag @jurajcigan From now on how often placeholders appear in a translated string is not important anymore (as long as they show up at least once). You can now update your strings if you want to. :)

Issue #125, opened 2017-02-28
Add "Paste & Go" context menu to URL entry


ahunt:
Similar to what fennec does: a "paste & go" context menu item would be nice to have, for URL entry both on the main screen, and when browsing. (This is probably primarily useful while testing, I'm not sure if there's much value for typical users.)

We should check fennec telemetry (if available) to see whether "paste & go" is regularly used there.

pocmo:
For testing you could also add an additional run configuration to Android Studio that starts the app with a VIEW intent.

pocmo:
For testing I added in Android Studio an additional Run Configuration with Launch Flags: "-a android.intent.action.VIEW -d https://www.mozilla.org"

Sdaswani:
Sorry for the newb question, but how can I access the 'Paste & Go' menu is Firefox for Android?

pocmo:
We moved the string freeze and release date. I landed a "Paste & Go" string (R.string.context_menu_paste_go) as part of PR #361. So this feature could land in v1.

Sdaswani:
Dropped my assignment in case someone else wants to pick it up for v1.

ahunt:
I've got patches which add this to the home screen. That leaves the actual browser screen. We could show the same context menu (which is what firefox does), or we could try using Chrome's behaviour (long-press expands the URL bar into editing mode AND triggers the text copy/paste/etc menu - that doesn't have a paste & go option though).

bbinto:
![](https://github.trello.services/images/mini-trello-icon.png) [Add "Paste & Go" context menu to URL entry](https://trello.com/c/Dsb93LHs/46-add-paste-go-context-menu-to-url-entry)

ekager:
Hi @ahunt ! Were you working on this specific issue for the non-testing features? If so could you push all of the code to a branch?
I started to work on the issue before seeing this thread, as I was working off of this description/specs https://bugzilla.mozilla.org/show_bug.cgi?id=1310265

pocmo:
There's a pull request with the code open. It's waiting for testing and reviewing after v1 has been released.

antlam:
@bbinto did we forget to include this in V2? if not, feel free to remove the 'needs triage' label :)

pocmo:
@antlam If we re-structure the UI so that the keyboard is always up by default then we would not need this, I think? You could just long-press the input and get an overflow toolbar with paste etc.

antlam:
[...]

PR #1184, opened 2017-08-22
Implement Paste and Go (#125)

### LINKED ISSUE #1118: Show clipboard URL as user starts typing
_Added by @bbinto_

**Why**
- Provide users with a convenient way to quickly access the URL they just copied to the clipboard earlier

**What**
-  We detect that the string in the clipboard is a URL.
-  If true, then show “Link you copied: [url.com]” as soon as you type something into the URL bar.
- Tap on the URL shown to go to the URL.

**Acceptance Criteria (how do I know when I’m done?)**
- Basic flow as described above can be performed
- Probes included to show how often users 
   - come into the app, having a URL copied to their clipboard
   - use the feature, i.e. Tap on the URL

--
We have this in iOS right now and it seems to be a great way to provide value given the context of Focus. It's also a quick way to help users move links into Focus. But we're still learning and iterating on this feature. 

Filing this to see if we also want to bring this over to Android. @bbinto thoughts?

Latest bug here for iOS: https://github.com/mozilla-mobile/focus-ios/issues/269

codebu5ter:
Closes #1118 and #125

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.java
@@ -54,8 +54,6 @@ public class UrlInputFragment extends LocaleAwareFragment implements View.OnClic
     private static final String ARGUMENT_HEIGHT = "height";
     private static final String ARGUMENT_OVERLAY = "translucent";
 
-    private String pasteString = "";
-
     private static final String ANIMATION_BROWSER_SCREEN = "browser_screen";
 
     private static final int ANIMATION_DURATION = 200;
@@ -126,6 +124,8 @@ public class UrlInputFragment extends LocaleAwareFragment implements View.OnClic
     private View toolbarBackgroundView;
     private View menuView;
 
+    private String pasteString;
+
     private @Nullable PopupMenu displayedPopupMenu;
 
     private volatile boolean isAnimating;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.java
```diff
@@ -138,6 +145,17 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
...
 
+        searchViewClip =  (TextView) view.findViewById(R.id.search_hint_clip);
+        searchViewClip.setOnClickListener(this);
+
+        ClipboardManager clipboard = (ClipboardManager) getContext().getSystemService(CLIPBOARD_SERVICE);
+        ClipData clip = clipboard.getPrimaryClip();
+
+        if (clip != null && clip.getItemCount() > 0) {
+            ClipData.Item item = clip.getItemAt(0);
+            pasteString = item.getText().toString();
```
xfix: I think `item.getText()` can return `null` if paste doesn't contain text (not sure however)?

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.java
```diff
@@ -50,6 +54,8 @@
     private static final String ARGUMENT_HEIGHT = "height";
     private static final String ARGUMENT_OVERLAY = "translucent";
 
+    private String pasteString = "";
```
xfix: Should this be a global variable? Its value is initialized when view is created, but the paste can feasibly change since then.

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.java
codebu5ter: @xfix I've changed this to coerceToText()

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.java
codebu5ter: @xfix You are right. This was an oversight on my end. I've now moved the initialization to just before the onCreateView

PR #1188, opened 2017-08-22
Show Search Terms or URL with Sessions (#1065)

ekager:


app/src/main/java/org/mozilla/focus/session/Session.java
```diff
@@ -128,6 +130,22 @@ public void setSearch(boolean isSearch) {
...
     }
 
+    public void setSearchTerms(String searchTerms) {
+        this.searchTerms = searchTerms;
+    }
+
+    public String getSearchTerms() {
+        return searchTerms;
+    }
+
```
sputnikci: \[Checkstyle\] ERROR: Trailing whitespace

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.java
```diff
@@ -462,14 +490,40 @@ public void onCommit() {
...
+                if (!isUrl) {
+                    session.setSearch(true);
+                    session.setSearchTerms(input.trim());
+                } else {
+                    session.setSearch(false);
+                }
+            } else {
+                if (!isUrl) {
+                    isSearchNewSession = true;
+                    searchTermsNewSession = input.trim();
```
pocmo: Instead of keeping references in the UrlInputFragment instance we could pass those values to openUrl(), I guess?

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.java
```diff
@@ -182,8 +185,33 @@ public boolean onPreDraw() {
...
-            urlView.setText(session.getUrl().getValue());
+            final String searchURL = session.getSearchUrl();
+            final String currentURL = session.getUrl().getValue();
+            final String searchTerms = session.getSearchTerms();
+            if (!TextUtils.isEmpty(currentURL)) {
+                if (session.isSearch()) {
+                    if ((!TextUtils.isEmpty(searchTerms))) {
+                        urlView.setText(searchTerms);
+                        session.setSearch(false);
+                        session.setSearchUrl(currentURL);
```
ekager: I set this to false so when the user navigates away from their search by clicking on another link, the search terms will not keep showing up. But if they just click away and click back on the UrlInputFragment, then they will show up because the URL matches the search URL.

pocmo:
Thank you. This looks good. :)

Issue #1059, opened 2017-08-01
Support keyboards with automatic incognito mode


buttercookie42:
i.e. add [`IME_FLAG_NO_PERSONALIZED_LEARNING`](https://developer.android.com/reference/android/support/v13/view/inputmethod/EditorInfoCompat.html#IME_FLAG_NO_PERSONALIZED_LEARNING) to `outAttrs.imeOptions` in `onCreateInputConnection`.

At the moment Gboard is the only keyboard that I know of that supports this, but hopefully others might follow. See also [the corresponding bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1266683#c22) for Firefox's private mode.

buttercookie42:
Since somebody mentioned this somewhere on IRC some time ago, on "GeckoFocus" everything handled by `GeckoInputConnection` should get this for free thanks to [bug 1266683](https://bugzilla.mozilla.org/show_bug.cgi?id=1266683), but things like the URL bar or the Webview need to be handled separately.

PR #1192, opened 2017-08-23
Use the keyboard's incognito mode if available. (#1059)

pocmo:


app/src/main/java/org/mozilla/focus/utils/ViewUtils.java
```diff
@@ -26,6 +26,13 @@
 
 public class ViewUtils {
     /**
+     * Flag of imeOptions: used to request that the IME does not update any personalized data such
+     * as typing history and personalized language model based on what the user typed on this text
+     * editing object.
+     */
+    public static final int IME_FLAG_NO_PERSONALIZED_LEARNING = 0x01000000;
```
xfix: This constant is directly available in API 26, it may be a good idea to use it after #1194 gets merged.

xfix: This constant is available in API 26, it may be a good idea to use it after #1194 is merged.

PR #1210, opened 2017-08-26
Consider upgrading to Java 8 language level

jonalmeida:
This should help reduce some unnecessary code for writing listeners (for example) with method references.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -17,6 +17,10 @@ android {
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
 
         multiDexEnabled true
+
+        jackOptions {
+            enabled true
+        }
     }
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -17,6 +17,10 @@ android {
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
 
         multiDexEnabled true
+
+        jackOptions {
+            enabled true
```
pocmo: The jack toolchain is deprecated:
https://android-developers.googleblog.com/2017/03/future-of-java-8-language-feature.html

jonalmeida: Oh no.. closing then. Will wait for AS 3.0 to be stable.

PR #1224, opened 2017-08-28
Placeholder count mismatch as warning, unused placeholders exception (#1186)

### LINKED ISSUE #1186: check_translations.py: Allow omitting duplicate placeholders
See #1183.

We have a string that contains the same placeholder twice, e.g. "Hello %1$s World %1$s". Some translations only needed this placeholder once, e.g. "Hallo %1$s Welt". Our check_translations.py script failed the build because the translated string does not include all placeholders. We should check that all placeholders are in the translated string, but we should not enforce how often the placeholder is used.

mastizada:
fix #1186

### DIFF
--- a/tools/l10n/check_translations.py
+++ b/tools/l10n/check_translations.py
@@ -21,8 +21,23 @@ def etree_to_dict(tree):
     return d
 
 
-def raise_exception(lang, code):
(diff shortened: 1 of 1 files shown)


pocmo:
Thank you @mastizada! :)

Issue #960, opened 2017-07-17
Multitasking within Focus, switching between sites


bbinto:
### Tasks

* [x] Finalize UX spec
* [x] Land strings so that L10N team can translate them
* [x] Refactor app architecture to support multiple sessions
* [x] Implement new Floating Action Button UI for opening and switching between tabs
* [x] Add option to open a new tab to the context menu
* [x] Add telemetry probes for all actions
* [x] Verify whether we need additional telemetry (Number of tabs per session?)
* [x] Testing: Do we still send all telemetry correctly?
* [x] Testing: Custom tabs and other edge cases (Share to Focus, First run etc.)

### Follow-up bugs

* #1230 - In new tab, trackers blocked setting is not remembered upon revisiting.
* #1231 - New tab shows more than first subdomain

pocmo:
See also: #751 - Support for tabs.

antlam:
There's a lot to do for 2.0 already. I think this is going to be one of the larger UX efforts (to get right at that) and I might suggest pushing this to after.

bbinto:
We agreed to move this to the next major update (3.x)

andreicristianpetcu:
Please stick with tabs for the multitasking aspect. Please do not invent a new UX concept just for this. Tabs are awesome. Thank you!

antlam:
Thanks @andreicristianpetcu 

Could you share some more helpful insight into what you most often use tabs for (on mobile) exactly?

TruboXL:
While not exactly a use case for tabs, I have been viewing certain websites that requires pop-ups.

Without tabs, websites just redirect to the ads (instead of opening up another tab/window) and it is sometimes impossible to advance to the next page of the intended website when you Go Back.

andreicristianpetcu:
@antlam I am a heavy reddit user and most of my browsing is on Reddit or starts from there. I usually start several tabs in parallel from various bookmarks. I have named bookmarks for [refi](https://www.reddit.com/r/firefox/new/) (Reddit Firefox), [remo](https://www.reddit.com/r/mozilla/new/) Reddit Mozilla or [rena](https://www.reddit.com/r/worldnews+europe+news+politics+live) (Reddit News All). Since Reddit loads slow, I open several tabs and wait for them to load.

I either open them by typing the name of the bookmark in the URL or I just click on the Activity Stream pinned items if it is on the list. If I find something interesting I open a new tab with that Reddit post.

[...]

PR #1229, opened 2017-08-29
Add simple test that opens and closes tabs. (#960)

pocmo:


app/src/androidTest/java/org/mozilla/focus/activity/MultitaskingTest.java
```diff
@@ -0,0 +1,174 @@
...
+                    .perform(click());
+
+            onView(withText(R.string.tabs_tray_action_erase))
+                    .perform(click());
+
+            assertFalse(SessionManager.getInstance().hasSession());
+        }
+
+
+        SystemClock.sleep(5000);
```
npark-mozilla: do we need this sleep() call?

pocmo:
It will open in the same state as you left it.

But yeah, there are a bunch of additional tests we could/should write. This is just a start. :)

PR #1243, opened 2017-08-30
avoid double typing on MultitaskingTest

npark-mozilla:


npark-mozilla:
I think this fixes it?

Issue #808, opened 2017-06-22
Allow users to set custom search engines


vext01:
Hi,

I'm a new user of firefox focus. I love it, but I would like to use my search engine of choice: startpage.com.

Perhaps we could:

 * Add startpage.com to the list of default search engines.
 * And allow users to add their own search engines.

What do you think? Thanks

jpggithub:
Privacy and Google (or Yahoo,...) search engines are paradixal, no?

afarsi:
We want at least, www.startpage.com become one of included search engines in Focus. Focus should support secure and open source search engines.
Focus shoudl focus on secure and open source search engines.

pocmo:
Closing this as a duplicate of #961. In the next sprint we are exploring options to add/remove custom search engines.

PR #1264, opened 2017-09-02
Add Bing search

### LINKED ISSUE #1263: Add Bing search option
Bing search is missing from the default search options (en-US locale).

Someone else requested support for Bing search at https://support.mozilla.org/en-US/questions/1169564

wesdawg:
Fix #1263
* Create bing.xml default search engine asset based on [Fennec's](https://hg.mozilla.org/mozilla-central/file/tip/mobile/locales/searchplugins/bing.xml)
* Add bing to `en-US` search config


pocmo:
Search plugin XML files are not edited by hand. We import them using a script:
https://github.com/mozilla-mobile/focus-android/blob/master/tools/search/scrape_plugins.py

Bing is removed from the list on purpose. See #808 about adding the ability to allow custom search engines.

Issue #1154, opened 2017-08-16
email client subject field not populated


njeyaakili:
When using the "share" capability to an email client, the "Subject:" field isn't populated with the page title. This does work properly with Firefox with both tested mail clients.

Tested: 

browsers:
Focus - 1.2 - doesn't populate
Firefox - 54.0.1 - populates

email clients:
Gmail - 7.7.16.164685024 
K-9 mail - 5.207

bbinto:
UX: XS Eng: M

PR #1270, opened 2017-09-04
Add page title to the share

rmisev:
Fixes: https://github.com/mozilla-mobile/focus-android/issues/1154

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -797,9 +797,19 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
             }
 
             case R.id.share: {
+                final String url = getUrl();
                 final Intent shareIntent = new Intent(Intent.ACTION_SEND);
                 shareIntent.setType("text/plain");
-                shareIntent.putExtra(Intent.EXTRA_TEXT, getUrl());
+                shareIntent.putExtra(Intent.EXTRA_TEXT, url);
+                // Use title from webView if it's content matches the url
+                final IWebView webView = getWebView();
+                if (webView != null) {
+                    final String contentUrl = webView.getUrl();
+                    if (contentUrl != null && contentUrl.equals(url)) {
+                        final String contentTitle = webView.getTitle();
+                        shareIntent.putExtra(Intent.EXTRA_SUBJECT, contentTitle);
+                    }
+                }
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -797,9 +797,19 @@ public void onClick(View view) {
...
+                final String url = getUrl();
                 final Intent shareIntent = new Intent(Intent.ACTION_SEND);
                 shareIntent.setType("text/plain");
-                shareIntent.putExtra(Intent.EXTRA_TEXT, getUrl());
+                shareIntent.putExtra(Intent.EXTRA_TEXT, url);
+                // Use title from webView if it's content matches the url
+                final IWebView webView = getWebView();
+                if (webView != null) {
+                    final String contentUrl = webView.getUrl();
+                    if (contentUrl != null && contentUrl.equals(url)) {
```
pocmo: Why can't we always set the title? Why the URL check?

rmisev: I added this check because an URL and a title are shared from different sources: the URL from the URL bar (`urlView`) and a title from the WebView.

Sometimes WebView can display a page loaded from a different location than specified in the URL bar. I think the page title should not be shared then.

Examples of such cases:
1. An error page when a site is unreachable.
2. Press "Share..." while new page isn't loaded yet, then the `webView.getTitle()` returns a title of previously loaded page.

What do you think?

pocmo: > Sometimes WebView can display a page loaded from a different location than specified in the URL bar. I think the page title should not be shared then.

Do you know any situations like that? This would be a major bug.

pocmo: > An error page when a site is unreachable.

We'd show the error page's title?

> Press "Share..." while new page isn't loaded yet, then the webView.getTitle() returns a title of previously loaded page.

This sounds like a bug we should fix!

rmisev: > We'd show the error page's title? I think that might be fine.

My tests show, what this title has text "%pageTitle%", a bit annoying.

> This sounds like a bug we should fix!

Hmm, I don't think so. Say the WebView is in the state of connecting to a slow new site (specified in the URL bar). While it's in this state, I think it isn't a bug what the WebView displays ant returns a title of a currently loaded page.

PR #1293, opened 2017-09-07
Unify homescreen shortcut adding using support library

### LINKED ISSUE #1291: Creating a shortcut on the device's home screen works directly with legacy api
[HomeScreen](https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/shortcut/HomeScreen.java) class has different shortcut adding implementations for android API < 26 and 26+. Version for Android prior to Nougat operates with deprecated intent directly. But [ShortcutManagerCompat](https://developer.android.com/reference/android/support/v4/content/pm/ShortcutManagerCompat.html) support library implementation already exists. So it would be good to use this class from AppCompat. Though home screen will have to be opened manually on Android up to 7.

Bringoff:
Fixes #1291

### DIFF
--- a/app/src/main/java/org/mozilla/focus/session/SessionManager.java
+++ b/app/src/main/java/org/mozilla/focus/session/SessionManager.java
@@ -190,6 +190,14 @@ public class SessionManager {
         addSession(session);
     }
 
+    private void createSession(Context context, Source source, SafeIntent intent, String url, boolean blockingEnabled) {
+        final Session session = CustomTabConfig.isCustomTabIntent(intent)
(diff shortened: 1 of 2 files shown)


pocmo:
Something went wrong here. There are now multiple unrelated commits in this PR. Can you *rebase* this so that this PR only contains your commit (on top of master)?

Bringoff:
Something went wrong with rebasing, git considered all commits as mine. I'd prefer to create new PR because I can't modify source branch for PR.

Issue #1295, opened 2017-09-08
Add confirmation before clearing browsing history due to back button


spocko:
Using the back button to return to the start screen causes the browsing history to be erased. Sometimes I do this accidentally, and since the history was cleared, I can't go "forward" to get back where I was. 

To address this, I suggest adding the option to have a confirmation before clearing the history when triggered via the back button.

layely:
It happens to me also to miss the back button and touch the erase button.

@pocmo maybe I could handle this (????).

pocmo:
Flagging this for our bug triage meeting.

Personally I'm not a fan of a confirmation - it interrupts the flow and forces the user to make a decision every time.

@spocko To understand your problem better: Is it that you accidentally hit 'back' or are you navigating back in your web history and then at some point drop out of Focus accidentally (-> #1195)

spocko:
This is very similar to issue #1195, but my concern is not about exiting the app accidentally, it's about clearing the history accidentally. In the current version of the app, I don't think it's really possible to exit the app accidentally, because the back button always takes you back to the start screen before exiting. 

My scenario is that I'm intentionally using the back button to navigate back in my web history, but accidentally hit back too many times and therefore arrive back at the start screen. If I do that, then my history is automatically cleared, and I have then lost the ability to go forward to get back to the url that I started with.

I understand your point about a confirmation, so maybe that is not the best solution. More generally I'm looking for some way to avoid unintentionally clearing the history when using the back button.

A simple solution might be to clear history when exiting the app (i.e. one more press of the back button) instead of clearing it when returning to the start screen. Then if a user accidentally navigates back to the start screen, their history would still be available, so they could go forward to get back where they were. If they want to clear history and exit, they just need to hit the back button one more time. The home screen could even have some buttons added to it, for example "Clear and Exit", or a forward button to resume browsing using the current session history. This would definitely be better than a modal confirmation popup.

layely:
@pocmo 
On my Honor phone, i have the **overview button** on the right side (down the floating erase button). 


![screenshot_2017-09-08-17-50-47](https://user-images.githubusercontent.com/19215803/30224692-3cbecab6-94bf-11e7-9eb6-b7c05a83303d.png)

While using focus, i would jump to another app (like reading my whatsapp messages) and get back quickly with the **overview button**. But sometimes i accidentally hit the **floating erase button** instead.

spocko:
@layely 
[...]

PR #1297, opened 2017-09-08
Add confirmation before clearing browsing history due to back button. (#1295)

layely:


layely:
Will be really sorry for the damages made, it's just my very first pull request to an open source project.

Issue #1304, opened 2017-09-10
[2.0 RC1] address bar regression


phunkular:
on 2.0, when I visit page via search and tap the address bar, I only get the search terms appear - I can no longer view the full URL and edit it. If I visit a page via URL first instead then I can tap, highlight and then modify the full URL.

Eg. Visit the Firefox focus github by firstly searching for it and then by visiting GitHub directly. When you search for GitHub you will not be able to modify the URL to see the /releases page. However if you go to GitHub.com , search Firefox focus, you will then have full URL available and can browse /releases from there,

Thanks

pocmo:
Back-ported to Cobalt milestone for 2.0-RC2 build.

PR #1305, opened 2017-09-10
Address bar regression (#1304)

ekager:
Takes care of case where user clicks link before URL bar and we need to setSearch for current session to false.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/session/SessionCallbackProxy.java
+++ b/app/src/main/java/org/mozilla/focus/session/SessionCallbackProxy.java
@@ -56,6 +56,10 @@ public class SessionCallbackProxy implements IWebView.Callback {
     @Override
     public void onURLChanged(String url) {
         session.setUrl(url);
+        if (session.isSearch()) {
+            session.setSearch(false);
+            session.setSearchUrl(url);
+        }
     }
 
 


app/src/main/java/org/mozilla/focus/session/SessionCallbackProxy.java
```diff
@@ -56,6 +56,10 @@ public void onProgress(int progress) {
     @Override
     public void onURLChanged(String url) {
         session.setUrl(url);
+        if (session.isSearch()) {
+            session.setSearch(false);
+            session.setSearchUrl(url);
+        }
```
pocmo: Wouldn't this break in cases where the search page redirects to a different URL (e.g. Wikipedia)?

pocmo: nit: I think I saw this snippet a couple of times in our code base now. Maybe it's time for a session.clearSearchIfNeeded() method or something like that. :)

ekager: Ugh, yes it would probably break for redirects. Any ideas? Should we scrap the URL matching altogether?

pocmo: Oh, do we still do URL matching? I somehow missed that :D

ekager: Not regex matching, just literally comparing two URLs. That was the solution I implemented so that we could keep track and compare the search URL with current URL to see if it was a search (after clicking away from the URL bar). But I guess I forgot to account for if the user clicks away without clicking the URL bar...  >.<
https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.java#L187

pocmo: Ah, the WebView might be able to tell us whether a new URL came from a user interaction. I guess we could just keep the search terms (and display them) until there's a new request triggered by a user interaction (like a click on a link) - and completely ignore the URL? I'll write a patch and you can review it if you have time for it. :)

PR #1314, opened 2017-09-12
Keep search terms until a new page is loaded triggered by a user gesture. (#1304)

pocmo:
@ekager Does this make sense?

With that we keep displaying the search term until the user has loaded a new page after clicking something etc. It looks like with that we can remove the code in the UrlInputFragment. We would loose the ability to navigate back to a search page and then display the search terms again - but I guess that's fine...

### DIFF
--- a/app/src/webkit/java/org/mozilla/focus/webkit/FocusWebViewClient.java
+++ b/app/src/webkit/java/org/mozilla/focus/webkit/FocusWebViewClient.java
@@ -5,34 +5,22 @@
 package org.mozilla.focus.webkit;
 
 import android.content.Context;
-import android.content.pm.PackageInfo;
-import android.content.pm.PackageManager;
-import android.content.res.Resources;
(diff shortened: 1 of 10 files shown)


ekager:
Yeah I think this makes sense!

app/src/webkit/java/org/mozilla/focus/webkit/FocusWebViewClient.java
```diff
@@ -119,7 +104,7 @@ public void onLoadResource(WebView view, String url) {
     }
 
     @Override
-    public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceRequest request) {
+    public WebResourceResponse shouldInterceptRequest(WebView view, final WebResourceRequest request) {
```
sputnikci: \[PMD\] INFO: The method 'shouldInterceptRequest' has a Cyclomatic Complexity of 6\.

PR #1316, opened 2017-09-13
Fix typo error in comment

djKooks:
Fixed simple typo error `localised` in comment. Please look on.


pocmo:
This is not a typo as such - just the difference between British and American English. Afaik this was written by @ahunt... so I guess this was done on purpose. :)

djKooks:
@pocmo Oh, sorry. Thanks for your comment! I was just wondering that Activity class name is started with `Localized` but comment is written as `localised`.

PR #1355, opened 2017-09-21
Should check for isClickable(), not exists()

npark-mozilla:


npark-mozilla:
Sorry about this oversight.

PR #1356, opened 2017-09-21
Change Floating Erase Button to UiDevice type from ViewInteraction of espresso

npark-mozilla:
Best way to avoid timing error on clicking erase button is to wait for it.

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/activity/AdBlockingTest.java
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/AdBlockingTest.java
@@ -136,6 +136,6 @@ public class AdBlockingTest {
         assertTrue(blockAdTrackerValue.getText().equals("ON"));
         TestHelper.navigateUp.click();
         TestHelper.webView.waitForExists(waitingTime);
-        TestHelper.floatingEraseButton.perform(click());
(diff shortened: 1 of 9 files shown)


pocmo:
Why?

The failure we saw came from a  UIObject already. Note that there are two erase buttons used in ScreengrabTest. One is the ViewInteraction and the other one is a UIObject. I think we saw the other one fail on taskcluster. With a higher wait time this seems to work now. So far I feel like espresso interactions are more reliable than uiautomator. :)

npark-mozilla:
Oh, I assumed what we saw from TC was same as this error in master: https://dashboard.buddybuild.com/apps/584f67f6f3d6eb01000842d6/build/59c41b570698550001093d83#uitests  In this case, the failure comes from not being able to find the erase button. Okay, let's investigate this a bit more.  It seems that on master this is consistently failing here.

npark-mozilla:
Looks like two previous master runs have the ViewInteraction failures.

npark-mozilla:
@pocmo hmm, now I'm not sure... this failure only happens on Nexus 9 (tablet) (on master branch), and I cannot run the test locally as the screencapture call throws exception.  Could you take a quick look?  It's consistently failing so something must be off I think.

PR #1357, opened 2017-09-22
Npark mozilla/screengrabfix

npark-mozilla:


app/src/androidTest/java/org/mozilla/focus/activity/screenshots/ScreenGrabTest.java
```diff
@@ -95,6 +95,8 @@ protected void beforeActivityLaunched() {
                         .setBody(TestHelper.readTestAsset("rabbit.jpg")));
                 webServer.enqueue(new MockResponse()
                         .setBody(TestHelper.readTestAsset("download.jpg")));
```
npark-mozilla: Below is needed to load the site when opening a new tab in line 368. (not necessary, but cleaner)

app/src/androidTest/java/org/mozilla/focus/activity/screenshots/ScreenGrabTest.java
```diff
@@ -260,12 +262,16 @@ private void takeAddToHomeScreenScreenshot() throws UiObjectNotFoundException {
         TestHelper.AddtoHSCancelBtn.waitForExists(waitingTime);
         Screengrab.screenshot("AddtoHSDialog");
         TestHelper.pressBackKey();
-        TestHelper.pressBackKey();
+
```
npark-mozilla: Video shows that the simulator is slow enough that the keyboard isn't pulled when the dialog is opened. If Back key is pressed twice before the keyboard is visible, app will exit to the initial page.

app/src/androidTest/java/org/mozilla/focus/activity/screenshots/ScreenGrabTest.java
```diff
@@ -363,7 +369,9 @@ private void takeScreenshotOfContextMenu(UiDevice device) throws UiObjectNotFoun
         multiTabBtn.click();
         eraseHistoryBtn.waitForExists(waitingTime);
         Screengrab.screenshot("Multi_Tab_Menu");
-        eraseHistoryBtn.click();
```
npark-mozilla: the definition of eraseHistoryBtn is incorrect, at least on tablet mode. Uiautomator will pick whatever element that matches first, and there are more than one element that is android.widget.TextView class with index of 2. (namely, the url bar) If we prefer to erase it this way, let's assign an identifier to the erase history button.  As a workaround, I deleted the tabs from the notification bar.

npark-mozilla:
@pocmo I put some workaround to address two separate issues, but feel free to take over this PR.
So the two issues were:
- Sim too slow to pull keyboard when Add shortcut dialog is opened
- eraseHistoryBtn is incorrect

PR #1366, opened 2017-09-22
Add Korean translation for setting menu

djKooks:
I updated some Korean strings which are not being translated in settings menu.
Please look on. Thanks.


pocmo:
Thank you for your help!

Unfortunately I can't merge this PR. Focus is getting translated on Pontoon here:
https://pontoon.mozilla.org/projects/focus-for-android/

We import translated strings from there automatically and therefore can't edit them in our repository. You can register on Pontoon and start helping to translate Focus and other Mozilla products. You can get in touch with the L10N teams here:
https://wiki.mozilla.org/L10n:Teams

djKooks:
@pocmo Got it. Thanks!

umertanvir:
@pocmo  let me check what I can do for that i am translating using poontoon but I guess that's not an issue from translation side but still I will recheck

PR #1374, opened 2017-09-22
remove assert to continue test

npark-mozilla:


npark-mozilla:
Test failed from screengrab, but it fixes the intermittent in AddtoHSTest

Issue #1375, opened 2017-09-22
NullPointerException: SessionsSheetFragment.onAnimationEnd()


pocmo:
```
java.lang.NullPointerException: 
 
  at org.mozilla.focus.session.ui.SessionsSheetFragment$3.onAnimationEnd (SessionsSheetFragment.java:105)
 
  at android.view.RenderNodeAnimator.onFinished (RenderNodeAnimator.java:381)
 
  at android.animation.RevealAnimator.onFinished (RevealAnimator.java:43)
 
  at android.view.RenderNodeAnimator.callOnFinished (RenderNodeAnimator.java:488)
 
  at android.os.MessageQueue.nativePollOnce (Native Method)
 
  at android.os.MessageQueue.next (MessageQueue.java:323)
 
  at android.os.Looper.loop (Looper.java:136)
 
  at android.app.ActivityThread.main (ActivityThread.java:6123)
 
  at java.lang.reflect.Method.invoke (Native Method)
 
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:867)
 
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:757)
```

pocmo:
Duplicate of #1385.

PR #1391, opened 2017-09-25
Fix crash while double tapping a tab inside tablist (#1385)

### LINKED ISSUE #1385: Crash while double tapping a tab inside tablist.
Firefox focus always crash when I double tap any tab inside tablist.

codebu5ter:
Fixes #1385

### DIFF
--- a/app/src/main/java/org/mozilla/focus/session/ui/SessionsSheetFragment.java
+++ b/app/src/main/java/org/mozilla/focus/session/ui/SessionsSheetFragment.java
@@ -17,6 +17,7 @@ import android.view.ViewGroup;
 import android.view.animation.AccelerateInterpolator;
 
 import org.mozilla.focus.R;
+import org.mozilla.focus.activity.MainActivity;
 import org.mozilla.focus.locale.LocaleAwareFragment;
 import org.mozilla.focus.session.SessionManager;
 import org.mozilla.focus.telemetry.TelemetryWrapper;
@@ -102,10 +103,13 @@ public class SessionsSheetFragment extends LocaleAwareFragment implements View.O
         animator.addListener(new AnimatorListenerAdapter() {
             @Override
             public void onAnimationEnd(Animator animation) {
-                getActivity().getSupportFragmentManager()
-                        .beginTransaction()
-                        .remove(SessionsSheetFragment.this)
-                        .commit();
+                final MainActivity activity = (MainActivity) getActivity();
+                if (activity != null) {
+                    activity.getSupportFragmentManager()
+                            .beginTransaction()
+                            .remove(SessionsSheetFragment.this)
+                            .commit();
+                }
             }
         });
 


app/src/main/java/org/mozilla/focus/session/ui/SessionsSheetFragment.java
```diff
@@ -102,10 +102,12 @@ public void onAnimationEnd(Animator animation) {
         animator.addListener(new AnimatorListenerAdapter() {
             @Override
             public void onAnimationEnd(Animator animation) {
-                getActivity().getSupportFragmentManager()
-                        .beginTransaction()
-                        .remove(SessionsSheetFragment.this)
-                        .commit();
+                if (getActivity() != null) {
+                    getActivity().getSupportFragmentManager()
```
pocmo: Even if the first call returns a non null activity, the second call can still return null.

codebu5ter: @pocmo The first call returns a non null activity, and hence the code under the 'if' statement is executed, but the second call returns a null, and hence nothing is done, preventing the crash.
This seems to be working fine as per my testing.

pocmo: If the second call returns null then this would be a NullPointerException again.

Don't you want to do something like this:

```java
final Activity activity = getActivity();
if (activity != null) {
    activity.getSupportFragmentManager() ....
    ...
}
```

app/src/main/java/org/mozilla/focus/session/ui/SessionsSheetFragment.java
```diff
@@ -98,14 +99,17 @@ public void onAnimationEnd(Animator animation) {
 
     /* package */ Animator animateAndDismiss() {
         final Animator animator = playAnimation(true);
+        final MainActivity activity = (MainActivity)getActivity();
```
sputnikci: \[Checkstyle\] ERROR: 'typecast' is not followed by whitespace\.

sputnikci: \[Checkstyle\] ERROR: 'typecast' is not followed by whitespace\.

pocmo: I wonder why you are doing this here and not in onAnimationEnd()? With this we are keeping a reference to the activity the whole time unnecessarily.

Also the cast to MainActivity does not seem like something we need to do.

codebu5ter: @pocmo Shifted the activity reference to inside onAnimationEnd().
Also, without the typecasting, getSupportFragmentManager() is unable to get resolved, because it requires a reference to an AppCompatAcitivty. I can either typecast to an AppCompatAcitivty or MainActivity. I've chosen the latter since it extends AppCompatAcitivty.

pocmo:
Looks good now! Can you squash those two commits into a single one? Then it's ready to merge. :)

codebu5ter:
@pocmo Done! :)

codebu5ter:
@pocmo  I’m not sure, but this may also solve #1375

Issue #1373, opened 2017-09-22
IllegalArgumentException: null URL (SessionManager.createSession)


pocmo:
```
Caused by: java.lang.IllegalArgumentException: 
 
  at org.mozilla.focus.architecture.NonNullLiveData.setValue (NonNullLiveData.java:45)
 
  at org.mozilla.focus.architecture.NonNullMutableLiveData.setValue (NonNullMutableLiveData.java:25)
 
  at org.mozilla.focus.architecture.NonNullLiveData.<init> (NonNullLiveData.java:12)
 
  at org.mozilla.focus.architecture.NonNullMutableLiveData.<init> (NonNullMutableLiveData.java:9)
 
  at org.mozilla.focus.session.Session.<init> (Session.java:39)
 
  at org.mozilla.focus.session.SessionManager.createSession (SessionManager.java:176)
 
  at org.mozilla.focus.session.SessionManager.createSessionFromIntent (SessionManager.java:78)
 
  at org.mozilla.focus.session.SessionManager.handleIntent (SessionManager.java:60)
 
  at org.mozilla.focus.activity.MainActivity.onCreate (MainActivity.java:65)
 
  at android.app.Activity.performCreate (Activity.java:6705)
 
  at android.app.Instrumentation.callActivityOnCreate (Instrumentation.java:1119)
 
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2648)
```

PR #1399, opened 2017-09-27
IllegalArgumentException: null URL (SessionManager.createSession)

pocmo:
#1373

### DIFF
--- a/app/src/main/java/org/mozilla/focus/session/SessionManager.java
+++ b/app/src/main/java/org/mozilla/focus/session/SessionManager.java
@@ -78,24 +78,26 @@ public class SessionManager {
 
             if (intent.hasExtra(HomeScreen.ADD_TO_HOMESCREEN_TAG)) {
                 final boolean blockingEnabled = intent.getBooleanExtra(HomeScreen.BLOCKING_ENABLED, true);
-                createSession(context, Source.HOME_SCREEN;, intent, intent.getDataString(), blockingEnabled);
+                createSession(context, Source.HOME_SCREEN, intent, intent.getDataString(), blockingEnabled);
             } else {
                 createSession(context, Source.VIEW, intent, intent.getDataString());
             }
         } else if (Intent.ACTION_SEND.equals(action)) {
             final String dataString = intent.getStringExtra(Intent.EXTRA_TEXT);
-            if (!TextUtils.isEmpty(dataString)) {
-                final boolean isSearch = !UrlUtils.isUrl(dataString);
+            if (TextUtils.isEmpty(dataString)) {
+                return;
+            }
 
-                final String url = isSearch
-                        ? UrlUtils.createSearchUrl(context, dataString)
-                        : dataString;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/session/SessionManager.java
```diff
@@ -71,14 +71,11 @@ private void createSessionFromIntent(Context context, SafeIntent intent) {
         final String action = intent.getAction();
 
         if (Intent.ACTION_VIEW.equals(action)) {
-            final Source source;
             if (intent.hasExtra(HomeScreen.ADD_TO_HOMESCREEN_TAG)) {
-                source = Source.HOME_SCREEN;
-                boolean blockingEnabled = intent.getBooleanExtra(HomeScreen.BLOCKING_ENABLED, true);
-                createSession(context, source, intent, intent.getDataString(), blockingEnabled);
+                final boolean blockingEnabled = intent.getBooleanExtra(HomeScreen.BLOCKING_ENABLED, true);
+                createSession(context, Source.HOME_SCREEN;, intent, intent.getDataString(), blockingEnabled);
```
mcomella: nit: not great for bisection that we break the build in this commit (with the semicolon) but revert it in a later commit. That being said, if we create a no-ff merge commit, I assume it wouldn't go into branches when bisecting. Also, I don't know how often bisecting is actually used when we can just bisect Nightly builds.

app/src/main/java/org/mozilla/focus/session/SessionManager.java
```diff
@@ -71,6 +71,11 @@ private void createSessionFromIntent(Context context, SafeIntent intent) {
         final String action = intent.getAction();
 
         if (Intent.ACTION_VIEW.equals(action)) {
+            final String dataString = intent.getDataString();
```
mcomella: nit: this stored value can be used in the `createSession` calls below.

app/src/main/java/org/mozilla/focus/session/SessionManager.java
```diff
@@ -71,6 +71,11 @@ private void createSessionFromIntent(Context context, SafeIntent intent) {
...
+            if (TextUtils.isEmpty(dataString)) {
```
mcomella: This works but I wonder: do we want to catch the error closer to where it's used? e.g. we can create a session with a static method that can return null in situations like this. The upside is that we're less likely to regress if the code changes but there are big downsides:
1. it's harder to take action for specific failures in the static method both because you only get "null" as a return value and because it's far removed from where the data is initially introduced
1. The static method would require a lot of refactoring given there are multiple constructors

Then again, our current set-up is pretty good at finding regressions too (catch at the source, throw for illegal state, add a test).

Just thinking aloud...

mcomella: From the code, it looks like the behavior is: if we receive a null URI, we open the app and do nothing. Should we should a Toast for the users to let them know the Intent failed?

PR #1403, opened 2017-09-27
add link/image to README to point to F-Droid download page

eighthave:
Please include F-Droid as one of the options for getting Mozilla Klar.

Firefox Klar has been included in F-Droid since 2017-07-04, built directly from source using the official build system.  As with all apps included in f-droid.org, we also run our automatic verification process on it, you can see Klar diffs here:

* https://verification.f-droid.org/org.mozilla.klar_11.apk.diffoscope.html
* https://verification.f-droid.org/org.mozilla.klar_10.apk.diffoscope.html

It is also possible to include the Mozilla APK signature in f-droid.org provided that Klar builds reproducibly.  We test this by building from source on our own infrastructure, then copying the developer's APK signature into our APK.  If it verifies, we ship it.


pocmo:
Can you file an issue for that? This will need some more discussion. As long as those are not the APKs distributed by Mozilla it might be problematic to link to them - even though I personally like all the work you are doing!

rugk:
Can't you discuss it in this PR here?

They are not the APKs from Mozilla, but use the "official" (= without modifications) source code.

pocmo:
Closing the PR -> Discussion / prioritization happens in #1405. Will re-open once ready.

PR #1412, opened 2017-09-28
Kotlin android extensions

pocmo:
🚀

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/DisplayToolbar.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/DisplayToolbar.kt
@@ -8,29 +8,18 @@ import android.content.Context
 import android.support.design.widget.AppBarLayout
 import android.util.AttributeSet
 import android.util.TypedValue
-import android.view.View
-
-import org.mozilla.focus.R
-import org.mozilla.focus.widget.AnimatedProgressBar
+import kotlinx.android.synthetic.main.browser_display_toolbar.view.*
 
 /**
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/browser/DisplayToolbar.kt
```diff
@@ -43,7 +32,7 @@ class DisplayToolbar(context: Context, attrs: AttributeSet) : AppBarLayout(conte
 
         // If the toolbar is collapsed then we will move the progress view so that it's
         // still fully visible.
-        progressView.translationY = if (isCollapsed) collapsedProgressTranslationY else 0f
+        progress.translationY = if (isCollapsed) collapsedProgressTranslationY else 0f
```
mcomella: nit: Not kotlin related but I like postfixing views with `...View` because it gives an indication what these variables represent. e.g. `progress` could be `progressView` but it could also be a value (a percent?) representing how far something has progressed.

It'd be the same with `toolbarContent` (is it the data behind the content, or the actual view?).

Though I'm not convinced that all data needs to have `...Model`, or similar, after it so maybe I'm not entirely consistent.

pocmo: I like the "View" suffix a lot too! But it also feels weird to add this to the IDs. What I did instead is configuring Android Studio to display those extension variables in a different color. This way I immediately see it. But of course this only works locally.

Issue #929, opened 2017-07-14
Automate L10N import


pocmo:
_Added by @bbinto_

**Why/User Benefit**
- Improve productivity of Focus team

**What**
- Importing strings into builds of the app automatically

**Acceptance Criteria (how do I know when I’m done?)**

---
I want to automate the L10N import. Exporting strings can still be a manual process (because we want to review the changes). But importing the strings into builds of the app should be automated.

I talked to Stefan and Firefox for iOS seems to do something like this. I want to explore our options here. It shouldn't make it super hard for regular developers to get other locales in their local builds (like Fennec).

## Tasks

- [x] Make L10N scripts for import etc. independent from gradle (to run them in automation etc.)
- [x] Modify docker image to support importing of translations in automation (a2po dependencies)
- [ ] Import strings automatically when creating screenshots on taskcluster (see #980)
- [x] Keep a list of "release" locales in the repository and only import those for release builds
- [ ] (Maybe) remove translations from the repository, because they are imported at build time
- [ ] (Maybe) run check_translations.py in automation so that we can find issues with translated strings

Release builds should only contain the defined locales. All other builds can include all available locales.

pocmo:
Closing: This is half automated and the L10N team is working on making their tools support Android projects.

PR #1413, opened 2017-09-28
Move L10N tasks from gradle to standalone scripts. (#929)

pocmo:
This makes it easier to customize them and run them in automation.


tools/l10n/initialize.sh
```diff
@@ -0,0 +1,4 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
liuche: Is this file used anywhere?

app/build.gradle
```diff
@@ -422,75 +342,6 @@ clean.doLast {
...
-    } .collect {
-        return '"' + it + '"'
-    }
-
-    langs << "\"en-US\""
-
-    return langs
-}
-
-task verifyLocales << {
```
liuche: I didn't see this ported over, but is it because it's included (warnings at least) in ImportCommand?

tools/l10n/initialize.sh
pocmo: Ops. Initially I wanted to put the repository initialization here. But I moved it into the other scripts so that they work on their own. I'll remove this from the patch.

app/build.gradle
pocmo: Yeah, and going forward we explicitly do not want to import all locales in all cases. :)

Issue #1201, opened 2017-08-24
"What's New" in app


bbinto:
We want to inform our users of updates *inside* the app, rather than via the store update description.

UX to come up with a solution how to achieve that.

Requirements:
- needs to show the content of a What's New SUMO Page 
- shows after update of the app
- the presentation does not conflict with the simplicity of the app
- easily to be localized and to be updated if needed (without releasing another app update)

gonzalde:
We could just add a small notification in the search bar with a message below indicating there was an update which can be clicked on to show whats new with the update. I attached a simple mock-up.
[updates-focus.pdf](https://github.com/mozilla-mobile/focus-android/files/1249802/updates-focus.pdf)

gonzalde:
also should only be showm when starting a new search and not when browsing other pages

bbinto:
Eng: S
UX: L

antlam:
We should talk about this next week with the iOS issue (also Ebony, also L)

antlam:
Moving to Folly for now, CC @bbinto

lime124:
![ixd flow - what s new cards](https://user-images.githubusercontent.com/287322/30870002-8b45918e-a297-11e7-8318-843f0883e71b.png)
![ixd flow - what s new](https://user-images.githubusercontent.com/287322/30870003-8b4d4b40-a297-11e7-9c2b-2cb5e2bccf58.png)

Strings are TBD and visual spec is coming.LMK if you have any questions!

/cc @bbinto @pocmo

bbinto:
@lime124 what's the difference between the top and the bottom spec? Or is this for us to choose what we prefer?

I'm a bit nervous to embed so much localizable text inside the app while features/updates listed in there could potentially fall off the release after string freeze. I was under the impressing we'd be pulling/linking to the SUMO page (which would be up to date at launch & localized).

Thoughts?

pocmo:
Yeah, having those strings in the app can become a problem. Linking out to SUMO and giving the community indefinitely time to create those pages in languages they can maintain is definitely easier for us!

In Fennec we have this "home banner" concept which shows an icon and message at the bottom of the screen. Almost like a Snackbar[1] with a link. I wonder if that's a helpful concept here too. Visually your design is way nicer than a boring SUMO page though. :)

[1] https://material.io/guidelines/components/snackbars-toasts.html

lime124:
@bbinto the spec is both pages. there was too much to fit on one page.

[...]

PR #1421, opened 2017-09-29
Add Strings for Folly.

liuche:
Adding a single string from Issue #1201.


pocmo:
> From #1201 (comment), do we also need to add, "Next" and "Close" for the card UI?

The card UI will not get implemented after all :)

Issue #1376, opened 2017-09-22
IllegalAccessError: SessionManager.getSessionByUUID()


pocmo:
```
java.lang.IllegalAccessError: 
 
  at org.mozilla.focus.session.SessionManager.getSessionByUUID (SessionManager.java:136)
 
  at org.mozilla.focus.fragment.UrlInputFragment.onCreate (UrlInputFragment.java:129)
 
  at android.support.v4.app.Fragment.performCreate (Fragment.java:2329)
 
  at android.support.v4.app.FragmentManagerImpl.moveToState (FragmentManager.java:1377)
 
  at android.support.v4.app.FragmentManagerImpl.moveFragmentToExpectedState (FragmentManager.java:1740)
 
  at android.support.v4.app.FragmentManagerImpl.moveToState (FragmentManager.java:1809)
 
  at android.support.v4.app.FragmentManagerImpl.dispatchStateChange (FragmentManager.java:3217)
 
  at android.support.v4.app.FragmentManagerImpl.dispatchCreate (FragmentManager.java:3166)
 
  at android.support.v4.app.FragmentController.dispatchCreate (FragmentController.java:181)
 
  at android.support.v4.app.FragmentActivity.onCreate (FragmentActivity.java:312)
 
  at android.support.v7.app.AppCompatActivity.onCreate (AppCompatActivity.java:84)
 
  at org.mozilla.focus.locale.LocaleAwareAppCompatActivity.onCreate (LocaleAwareAppCompatActivity.java:38)
 
  at org.mozilla.focus.activity.MainActivity.onCreate (MainActivity.java:53)
 
  at android.app.Activity.performCreate (Activity.java:6021)
 
  at android.app.Instrumentation.callActivityOnCreate (Instrumentation.java:1105)
 
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:2298)
 
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:2405)
 
  at android.app.ActivityThread.access$800 (ActivityThread.java:155)
 
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1323)
 
  at android.os.Handler.dispatchMessage (Handler.java:102)
 
  at android.os.Looper.loop (Looper.java:135)
 
  at android.app.ActivityThread.main (ActivityThread.java:5376)
 
  at java.lang.reflect.Method.invoke (Native Method)
 
  at java.lang.reflect.Method.invoke (Method.java:372)
[...]

PR #1429, opened 2017-10-03
fix IllegalAccessError: SessionManager.getSessionByUUID() (#1376)

layely:
fix bug #1376


layely:
@pocmo sorry for the delay, was hanging on some other stuff theses days.
I think this is far more complicated that i thought. Plus it's really hard to reproduce the bug.

Issue #1148, opened 2017-08-16
String import/export: cak / trs / zam / mix  / meh is not a valid locale


pocmo:
Output from Android2Po when importing/exporting strings:

```
    [failed] cak is not a valid locale
    [failed] trs is not a valid locale
    [failed] zam is not a valid locale
```

pocmo:
* cak: Kaqchikel
  * http://www-01.sil.org/iso639-3/documentation.asp?id=cak
  * https://en.wikipedia.org/wiki/Kaqchikel_people
  * https://en.wikipedia.org/wiki/Kaqchikel_language
* trs: Chicahuaxtla Triqui 
  * http://www-01.sil.org/iso639-3/documentation.asp?id=trs
  * https://es.wikipedia.org/wiki/Triqui_de_Chicahuaxtla
* zam: Miahuatlán Zapotec
  * http://www-01.sil.org/iso639-3/documentation.asp?id=zam
  * https://en.wikipedia.org/wiki/Miahuatl%C3%A1n_Zapotec

pocmo:
Also see: https://bugzilla.mozilla.org/show_bug.cgi?id=1263437

mastizada:
Babel do not support these locales, but I think we can do some custom tweaks for these locales in android2po (instead of using `babel.Locale`). Also adding custom locales for android doc should be checked (for values-xx directory). Ideally, these locales should be added to `cldr` unicode database, but that is different issue.

pocmo:
That would be great. We have some additional Android runtime code to check whether a locale can be used (and to get the name of the language in other languages). We would need to check whether this works on any device with those locales.

pocmo:
We have the same issue with the new locale "mix": Mixtepec Mixtec - http://www.mexico.sil.org/language_culture/mixtec/mixtec-mix.

pocmo:
Android doesn't have a problem to build with cak / trs / zam / mix locales. Of course at runtime it's unable to display the name of the languages. We would need to add those manually.

 * @mastizada Do you have time to take care of patching android2po?
 * @delphine Can you help us to get the name of the language *in the language* for cak / trs / zam / mix?

mastizada:
Hi @pocmo , I think I will need a help this time :)

I already discussed that issue before, the Babel library do not support these locales and they are also absent in Unicode CLDR library (http://cldr.unicode.org/).

Looks like android2po uses this only for plural forms.

[...]

PR #1432, opened 2017-10-03
Import Ebony strings - missing eo, lt, pl, tr

liuche:
Ran the new import script, was unable to import some locales (#1148), but also ran into some problems with mismatched number of placeholders:

```
Checking for placeholders in translation files...
 * [values-eo/strings.xml] number of placeholders not matching, key: firstrun_shortcut_text
 * [values-lt/strings.xml] missing placeholder in translation, key: preference_default_browser2
 * [values-pl/strings.xml] number of placeholders not matching, key: firstrun_shortcut_text
 * [values-tr/strings.xml] number of placeholders not matching, key: your_rights_content5
```

Not sure what should be done about these, but maybe the translations haven't been updated fully? Will make a note in the wiki when I find out what actions should be taken.


pocmo:
```
* [values-eo/strings.xml] number of placeholders not matching, key: firstrun_shortcut_text
* [values-pl/strings.xml] number of placeholders not matching, key: firstrun_shortcut_text
* [values-tr/strings.xml] number of placeholders not matching, key: your_rights_content5
```

Those three are only warnings and not errors. And we should make this more obvious: I'm going to file a bug. Currently you can only see that by looking at the exit code (`echo $?`) if there are no other issues.

```
* [values-lt/strings.xml] missing placeholder in translation, key: preference_default_browser2
```

This is definitely an error. We can either contact Delphine directly or open a Bugzilla bug for the translators.

liuche:
Removed the offending string, and also filed a bug for the missing placeholder:
https://bugzilla.mozilla.org/show_bug.cgi?id=1407344

I also added a commit to fix the taskcluster linting issue (removing an unused import), but there are other failures now:
`Building 39% > :app:mergeKlarWebviewCoverageAssetsergeKlarWebviewCoverageAssets
Could not merge source set folders: 
Error: java.nio.file.NoSuchFileException: /opt/focus-android/app/build/intermediates/assets/klarWebview/coverage/search/default
...
Caused by: java.lang.RuntimeException: java.nio.file.NoSuchFileException: /opt/focus-android/app/build/intermediates/assets/klarWebview/coverage/search/default
`

Any thoughts, @pocmo? The Buddybuild now builds fine, and the build error seems to be the same one that I saw in #1452 earlier.

pocmo:
> I also added a commit to fix the taskcluster linting issue (removing an unused import)

I landed a patch for this today. If you rebase your branch then this one should be gone.

Looking at the other issue now...

pocmo:
This looks like a gradle bug / race condition to me. I'll force-merge this and then see how the taskcluster build for master turns out.

Issue #1231, opened 2017-08-29
New tab shows more than first subdomain


npark-mozilla:
According to the UX spec, the url should only display up to first subdomain, but opening a new tab on embedded url shows more subdomains past 1st level.

![screenshot_20170829-164738](https://user-images.githubusercontent.com/6930774/29843521-31f2688c-8cda-11e7-94be-3fcc48e0be5c.png)

pocmo:
So far the implementation only does:
* Remove the scheme (http://  https://)
* Removes common unspecific sub domains (m. www. mobile.) - Code borrowed from Fennec

I noticed two things:
* We only allow opening new tabs from the context menu. Using this I often opened a bunch of tabs  that are from the same "page". Meaning at least the domain (and sometimes parts of the path) are the same for a bunch of tabs. Making it hard or impossible to distinguish between them.
* Strictly following the spec I end up with pages that are reduced to the same "label":
** https://en.m.wikipedia.org/wiki/Firefox_Focus -> m.wikipedia.org/wiki
** https://en.m.wikipedia.org/wiki/Mozilla -> m.wikipedia.org/wiki

Picking a "label" from a URL is pretty hard. We had this discussion a bunch of times for Activity Stream and whatever algorithm we pick there are always cases where it doesn't work correctly. I understand that we need to somehow cut off the string if it's too long - but I'm not sure why we purposefully make it shorter if there's space to show more.

Unfortunately I have no other proposal. Sometimes the domain is important, sometimes the path and other times the page title is the most helpful.

@brampitoyo @antlam What do you think?

antlam:
What if we started with removing the scheme, and the common unspecific sub domains....

And then started truncating with ellipsis immediately but showed as many characters as we can from the end of the URL? That could be helpful if we're looking to differentiate URLs from the same domain.

Eg.
- wikipedia.org/wiki/.../Mozilla
- wikipedia.org/w.../red_pandas
- wikipedia.org...furry_wombats

brampitoyo:
@pocmo Hey. You’re right that we might have truncated way too much of the URL to be helpful, and to make good distinction between resources located on the same site.

I like @antlam’s suggestion, but want to expand on it a bit more.

Our intention isn’t only to truncate, but also help clarify what each URL might be about. This is a big, hairy problem!

> Sometimes the domain is important, sometimes the path and other times the page title is the most helpful.

[...]

PR #1437, opened 2017-10-04
SessionViewHolder: Call beautifyUrl() on raw URL. (#1231)

pocmo:
Calling this on the URL with a bunch of stuff already stripped doesn't really work well. :)

### DIFF
--- a/app/src/main/java/org/mozilla/focus/session/ui/SessionViewHolder.kt
+++ b/app/src/main/java/org/mozilla/focus/session/ui/SessionViewHolder.kt
@@ -16,7 +16,6 @@ import org.mozilla.focus.R
 import org.mozilla.focus.session.Session
 import org.mozilla.focus.session.SessionManager
 import org.mozilla.focus.telemetry.TelemetryWrapper
-import org.mozilla.focus.utils.UrlUtils
 import org.mozilla.focus.ext.beautifyUrl
 import java.lang.ref.WeakReference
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/session/ui/SessionViewHolder.kt
```diff
@@ -59,10 +58,7 @@ class SessionViewHolder internal constructor(private val fragment: SessionsSheet
     }
 
     private fun updateUrl(session: Session) {
-        val shortUrl = UrlUtils.stripCommonSubdomains(
-                UrlUtils.stripScheme(session.url.value))
-
-        textView.text = shortUrl.beautifyUrl()
+        textView.text = session.url.value.beautifyUrl()
```
mcomella: I wonder if we should make `beautifyUrl` a method on `Uri` instead:
1. it'd cause red flags when you pass an already stripped string into a Uri constructor
1. `BeautifyUrl` throws for error cases and then immediately converts to Uri (sounds like a Uri method :P).
1. It's more type-safe, by convention, than using String-encoded Urls

But the trade-off is we're not always passing around URLs as Uris so we could end up creating a lot of wasted objects.

pocmo: Yeah, that's what I was wondering too when I wrote this code initially. Back then I decided against adding the method to Uri because I didn't want to be forced to do the conversion everywhere I want to use the method. We are basically never using Uri objects and maybe it is just an implementation detail here (to extract parts of the URL easily). There's no reason why we couldn't remove the Uri conversion from beautifyUrl() and replace it with a different implementation. If it wouldn't require a lot of work I'd finally like to use our own URL class. This way we could finally remove all the thousand helper method and just attach them to our class. :)

mcomella: > If it wouldn't require a lot of work I'd finally like to use our own URL class. This way we could finally remove all the thousand helper method and just attach them to our class. :)

I really like this idea of making any urls we bring into the app type-safe by wrapping them in a URL type, rather than keeping them as Strings. With extension functions, this doesn't necessarily need to be our own URL class but then we maybe lose a little flexibility? I guess I haven't thought about it enough.

Issue #1009, opened 2017-07-25
Consider adding more "App Shortcuts" for API25+


ekager:
I realize we already have the Erase function here, but something to think about adding in a future release for users running newer versions of Android perhaps for the new "Restart"/"Open and Erase", or a quick search options? @antlam any thoughts?

bbinto:
@lime124 please let us know if you agree to have this added (I do :)).

PR #1442, opened 2017-10-05
Added shortcut to erase history and open browser - #1009

jotunacorn:
The previous shortcut only erased history. This new shortcuts enables the user
to open the browser at the same time.

I added it in a response to issue #1009. Maybe also add a quick-search shortcut?


pocmo:
Thank you for your patch! I'll flagged the issue for our next triage meeting so that we can discuss taking this in.

PR #1445, opened 2017-10-06
Fix #1148 - String import for unsupported locales

### LINKED ISSUE #1148: String import/export: cak / trs / zam / mix  / meh is not a valid locale

mastizada:
This is import only. Export is not supported.

Solved using list of missing locales and matched them with locales that has same plural rules with them (using info from pontoon).

### DIFF
--- a/tools/l10n/android2po/commands.py
+++ b/tools/l10n/android2po/commands.py
@@ -12,6 +12,7 @@ from babel.messages import pofile, Catalog
 from termcolor import colored
 
 import convert
+from patch import read_po
 from env import resolve_locale
 
 __all__ = ('CommandError', 'ExportCommand', 'ImportCommand', 'InitCommand',)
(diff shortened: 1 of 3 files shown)


mastizada:
@pocmo Also we can now set custom tags in po files instead of changing them everytime while merging with l10n files from pontoon.

tools/l10n/android2po/commands.py
```diff
@@ -12,6 +12,7 @@
 from termcolor import colored
 
 import convert
+from patch import read_po
```
pocmo: Having a patch file is a nice idea!

pocmo:
This works great.

I see some XHTML errors. But they are actually in the translations. I'll file separate issues for that.

mastizada:
Forgot to change message, after our discussion in IRC  I've also added export.

PR #1452, opened 2017-10-09
LocaleListPreference: Add language names for cak / trs / zam / mix / meh manually

pocmo:
#1148

### DIFF
--- a/app/src/main/java/org/mozilla/focus/widget/LocaleListPreference.java
+++ b/app/src/main/java/org/mozilla/focus/widget/LocaleListPreference.java
@@ -21,12 +21,30 @@ import java.nio.ByteBuffer;
 import java.text.Collator;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


mcomella:
Does this replace PR #1432?

liuche:
@mcomella it doesn't replace that issue because #1432 is for importing translated locales, and this PR is to manually add/update these missing locales.

However, this seems to be running into the same problem as #1432, with the values-lt/strings.xml having the same broken string - in both these cases, there's a string that is listed as causing problems (preferences_default_browser2) but I can't see any problems with the string itself, but it's breaking the Buddbuild build.

pocmo:
> That being said, I don't totally understand all these locale issues but this commit doesn't look like it totally fixes the issues in #1148 – is this just a part of a greater whole, Sebastian?

Yeah, this is one part of it. We already landed the changes needed to fix our import/export tool.

Issue #1434, opened 2017-10-04
check_translations.py: Make it more obvious if something is a warning vs. an error.


pocmo:
From the output it's often hard to see if something is an error or a warning. A warning is something we can investigate asynchronously by opening a bug and then continue. However errors break the build and therefore are more important and need to be fixed.

pocmo:
Currently we can only look at the "exit code" but this doesn't help us if there are multiple warnings/errors.

devophile:
Hi @pocmo This is my first time on open source. I would like to contribute to this project. Can you help me in getting the structure of the project... I mean the way things are working so that I can work on this? Or any sources you would like to refer me to understand this. Thank you. :smile:

pocmo:
@devophile Sure! This one is for developers who want to write some Python code. There's a script in our repository that runs whenever we import strings. It makes sure that all the placeholders we use in our strings.xml file are also used in the translated strings:
https://github.com/mozilla-mobile/focus-android/blob/master/tools/l10n/check_translations.py

Some checks are treated as error and others as warnings (errors return exit code 1):
https://github.com/mozilla-mobile/focus-android/blob/master/tools/l10n/check_translations.py#L75-L83

All those checks print a message but from the message you can't see if they are an error or a warning: https://github.com/mozilla-mobile/focus-android/blob/master/tools/l10n/check_translations.py#L24-L43

To solve this issue it might be enough to just print "warning: " or "error: " in front of the messages. :)

mastizada:
```
Checking for placeholders in translation files...
 * ERROR: [values-az/strings.xml] ...
 WARNING: [values-eo/strings.xml] ...
 WARNING: [values-pl/strings.xml] ...
```

arjunkrishnababu96:
can I work on it?

emilyeverett:
Hi, @pocmo, I see that this issue hasn't been assigned yet, and I'm interested in working on it. Mind if I take it on?

pocmo:
Fixed by 1296ce9d55856893e9850ec72edfe1c435f47b21

PR #1454, opened 2017-10-10
Mark warning messages with W instead of *

mastizada:
It was discussed in PR #1432

### DIFF
--- a/tools/l10n/check_translations.py
+++ b/tools/l10n/check_translations.py
@@ -23,21 +23,21 @@ def etree_to_dict(tree):
 
 def missing_target_exception(lang, code):
     # there is placeholder in source, but not in target.
-    print(" * [{lang}/strings.xml] missing placeholder in translation, key: {code}".format(
+    print(" * ERROR: [{lang}/strings.xml] missing placeholder in translation, key: {code}".format(
         lang=lang,
         code=code
     ))
 
 def missing_source_exception(lang, code):
     # there is placeholder in target, but not in source.
-    print(" * [{lang}/strings.xml] placeholder missing in source, key: {code}".format(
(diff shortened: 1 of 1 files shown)


mastizada:
Just saw issue #1434, I think it can be better to leave that for new contributor.

pocmo:
@mastizada Thanks! 👍 Let's wait some days and if no other contributor picks it up then I'll merge your patch. :)

tools/l10n/check_translations.py
```diff
@@ -37,7 +37,7 @@ def missing_source_exception(lang, code):
 
 def count_mismatch_warning(lang, code):
     # number of same placeholders used in source and target do not match
-    print(" * [{lang}/strings.xml] number of placeholders not matching, key: {code}".format(
+    print(" W [{lang}/strings.xml] number of placeholders not matching, key: {code}".format(
```
pocmo: I think we can be more verbose and actually scream "WARNING: " or "ERROR: " into the face of the script user :)

mastizada:
Example output:
```
Checking for placeholders in translation files...
 * ERROR: [values-az/strings.xml] ...
 WARNING: [values-eo/strings.xml] ...
 WARNING: [values-pl/strings.xml] ...
```

crud3:
@pocmo  I'd like to pick it up!

I'm not sure about how useful it is though, as someone already commented on #1434 and this PR here already exists with a correct solution and everything :)
 But if you think it's okay I'll just do it.

pocmo:
@crud3 Do you want to write Python code or are you interested in contributing in general (Android/Java)? We could try to find an other issue for you!

crud3:
@pocmo  If you have some tasks to delegate, I would be glad to help :) 
I'm interested in both. Right now I'm have a lot more experience writing Java code, but I'm trying to learn Python anyways and have written some more or less small scripts in the past

PR #1458, opened 2017-10-10
Improve docker image

pocmo:
There are two things I want to update:
* Install pip and all dependencies for android2po. This will allow us to run the import/export scripts on taskcluster. And I want to use pip for something else - so it's a good time to do this now. :)
* Previously we only ran "gradle assemble" to fetch dependencies and cache them in the docker image. However we now run a lot of more gradle commands and they all need to pull dependencies. To avoid that we run those commands when creating the image and then those dependencies will already be available.


pocmo:
Image build succeeded. Pushing image to docker hub now.

pocmo:
Push successful. Image available now.

PR #1461, opened 2017-10-10
Update support library to 26.1.0 and architecture components to 1.0.0-beta2.

pocmo:
Some of our custom lifecycle classes are not needed anymore as this functionality
has been implemented in the support library.


pocmo:
Seems like this broke one of our unit tests. Will have a look tomorrow.

Issue #1392, opened 2017-09-25
Previous links stay purple after clearing.


doublez13:
Tested on Android O on Pixel

For example...
1. Go to Wikipedia
2. Browse a link
3. Press the trash can to clear everything
4. Browse back to Wikipedia
5. Viewed link still appears purple

Pressing the trash can, then closing the app seems to actually clear the viewed links.

pocmo:
Thank you for filing this issue. Can you help me answer the following questions so that I can try to reproduce this issue?

* What device are you using?
* What Android version is installed on your device?
* What's the browser user agent if you go to this URL: http://www.whoishostingthis.com/tools/user-agent/

doublez13:
Google Pixel running Android 8.0.0

Mozilla/5.0 (Linux; Android 8.0.0) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Focus/2.0 Chrome/61.0.3163.98 Mobile Safari/537.36

doublez13:
Here's an exact test case that I've confirmed
1. Open Firefox Focus
2. Go to wiki.kernel.org
3. Tap the rt link (first link on the list)
4. Tap trash can
5. Go to wiki.kernel.org
6. rt link still appears purple

ekager:
#322 is where this issue was discussed previously. There doesn't seem to be much action in the [Chromium bug we filed. ](https://bugs.chromium.org/p/chromium/issues/detail?id=709627&can=2&start=0&num=100&q=&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified&groupby=&sort=) So I will try to work on the CSS injection bit because I don't think we're doing anything wrong in how we clear history, cache, cookies, etc.

pocmo:
@ekager Have you been able to reproduce this for Wikipedia? That's something I'm still unable to.

For wiki.kernel.org: They do not appear to use a style for links at all (default browser colors). That's not covered by our workaround and a known issue. This might be a duplicate of #990 (Fix in WebView required).

liuche:
Duping this to #1600

PR #1462, opened 2017-10-10
CSS workaround to clear visited link color (#1392)

ekager:
We were just removing any CSS for "visited", so I added a line to add our own visited CSS instead of letting it default. Limitations of this: If link colors are set for a significant reason, they will not be displayed. I tried to just set the visited field, but then it was displaying the visited links a different shade of blue, so I followed the [default link color from Mozilla](https://hg.mozilla.org/mozilla-central/file/3d57107fcf9f/modules/libpref/init/all.js#l212) to set the visited and link colors. With more complicated JavaScript, we could maybe set the visited color to match the original link color, but may be less flexible/more prone to errors. This fixes the STR example for this bug.

### DIFF
--- a/app/src/webview/java/org/mozilla/focus/webview/FocusWebViewClient.java
+++ b/app/src/webview/java/org/mozilla/focus/webview/FocusWebViewClient.java
@@ -46,34 +46,53 @@ import org.mozilla.focus.web.IWebView;
      */
     private static final String CLEAR_VISITED_CSS =
             "var nSheets = document.styleSheets.length;" +
-            "for (s=0; s < nSheets; s++) {" +
-            "  var stylesheet = document.styleSheets[s];" +
-            "  var nRules = stylesheet.cssRules ? stylesheet.cssRules.length : 0;" +
-            // rules need to be removed by index. That modifies the whole list - it's easiest
-            // to therefore process the list from the back, so that we don't need to care about
-            // indexes changing after deletion (all indexes before the removed item are unchanged,
-            // so by moving towards the start we'll always process all previously unprocessed items -
-            // moving in the other direction we'd need to remember to process a given index
-            // again which is more complicated).
-            "  for (i = nRules - 1; i >= 0; i--) {" +
-            "    var cssRule = stylesheet.cssRules[i];" +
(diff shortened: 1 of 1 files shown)


ekager:
Hmm I *think* it will go through every stylesheet associated with the loaded website and remove all ":visited" tags from the stylesheets, and then this will add a line to the end of every stylesheet to set the ":link" and ":visited" as the same blue color.

pocmo:
Oh, yeah, somehow I missed that you add a color for :link too.

I guess this should work. But wouldn't we break the design of websites with that?

ekager:
Yes. I'm worried if I try to extract the link color and set the visited color to those, it will get messy, but I can try to make it more robust.

ekager:
Okay I attempted to make this more robust by trying to do several things at a time. 

1. Remove :visited tags from all stylesheets (this was all we had before)
2. Attempt to find :link tags and a) add a new rule for :visited to match link CSS exactly and b) splice matching :visited tag into the other rule to piggyback. 
3. If we can't find a :link tag and we can't find an "a" tag at all, then we will inject the generic blue for links and visited. 

I tested it on a few sites and it seemed to work... Let me know if you have any other ideas!

pocmo:
Wow, that's pretty cool! 👍 I'll give it a try. :)

ekager:
Just checked for Reddit mobile site and it doesn't work... Not sure what else to include...

ekager:
I researched this more today. I'm not very experienced with CSS and realized that it goes much further than the cases I was looking for. With more complicated websites this will not find/change everything because of CSS classes and IDs and more specific selectors. I'm not convinced it would be possible with JavaScript injection to scrub all visited link styles reliably, or even to force color the links. If anyone is more experienced with JavaScript/CSS injection feel free to butt in!

pocmo:
Not catching all cases is okay - for the workaround. @ekager Do you still think your modification is something that makes it better or does it have negative effects in some cases?

ekager:
I *think* it is a slight improvement covering a few more cases of not too complex CSS. I didn't notice any negative effects, it just won't work in some cases (like Reddit...)

Issue #1260, opened 2017-09-01
[ux] UI audit of dialogs/ context menus in app


antlam:
Noticing that we're not actively keeping track of the UI here. So I'm filing this to keep things polished and in order. Strictly looking at some padding, color, and type consistencies here. Nothing major.

![screen shot 2017-08-30 at 11 09 35 am](https://user-images.githubusercontent.com/1994820/29982953-964e52b6-8f08-11e7-916c-9855d41aac3d.png)

Moved checklist to eng [meta]: #1659.

bbinto:
UX: S
Eng: S

antlam:
![screen shot 2017-09-08 at 10 59 20 am](https://user-images.githubusercontent.com/1994820/30224744-cc472812-9484-11e7-888e-ddfa26d48a14.png)

Top new, Bottom old. For comparison

antlam:
![spec_dialogs1](https://user-images.githubusercontent.com/1994820/30620468-61691760-9d5a-11e7-9ba5-400c67231760.png)

@pocmo should be the last of it!

layely:
@antlam @bbinto @lime124 is this something that i can work on ?

lime124:
hey @layely! Are you interested in contributing as a designer or an engineer? 

If it's the design side, there are designs posted for this issue but I can totally point you to other issues that need design work. 

If it's engineering, @pocmo would be a better person to ping for what things you should work on.

Thanks for the ping! :)

layely:
@lime124 Sure the engineer side is what I am interested to. Let's see if @pocmo has something to suggest doing here.

pocmo:
@layely Yes, help would be appreciated here. Currently this tracks the current milestone to be released with the next version of Focus. So there's some time pressure. Therefore I'd suggest to create small commits that fix only a specific thing and open separate pull request for them. This way we can merge them separately (and early) and someone can potentially take over later. If you try to fix all of that in a gigantic pull request then this might take too much time (with review).

bbinto:
awesome, thanks @layely

layely:
![2017-10-11-05-59-40](https://user-images.githubusercontent.com/19215803/31442366-bace7034-ae85-11e7-8caa-3ea8b67ea603.png)

![capture - copie](https://user-images.githubusercontent.com/19215803/31442397-d7f9b4ac-ae85-11e7-8e9c-30e6abf54119.PNG)

The first is my capture when I apply changes suggested by @antlam 
[...]

Issue #1358, opened 2017-09-22
Default to displaying "Share" in custom tabs


autonome:
This is a key feature in a very common task workflow, is quite frustrating that it's not there. The share menu item is present when opening the browser itself.

STR:

1. Open Reddit app
2. Open a link
3. Opens in Focus
4. Open the stacked dots menu

Expected: A "share" menu item is available.

Actual: There is no "share" menu item.

mcomella:
I can reproduce. Reddit opens a custom tab and it appears our custom tab implementation doesn't include the "Share" menu item.

@pocmo Was this intentional in any way?

autonome:
FWIW, I've seen it on other sites as well, not only Reddit.

pocmo:
> @pocmo Was this intentional in any way?

It depends. There's an Intent extra (CustomTabsIntent.EXTRA_DEFAULT_SHARE_MENU_ITEM) that tells us whether the third-party app wants to show a share item or not. So there could be multiple things happening here:

* Those apps do not want to show a share item (in this case: No bug - other browsers with custom tab support should behave the same)
* The apps want to show the share item - but there's a bug preventing it to show up
* The app does not define the extra; in which case we default to "false" and do not show the item. And in this case it would be interesting to know if we should default to "true" and what Chrome does.

autonome:
> Those apps do not want to show a share item

These are just web pages I'm visiting. As a user, I should be deciding
whether I want to share the URL, not anything else (eg, whether I have
native app that's registered as handler for that domain or not).

pocmo:
> These are just web pages I'm visiting. As a user, I should be deciding
> whether I want to share the URL, not anything else (eg, whether I have
> native app that's registered as handler for that domain or not).

I agree that a lot of the web related decisions Android has made give power to the developers of apps and take them away from the users (and their user agents). In custom tabs you have at least the option to switch to the full browser (via the menu). I do not think it would be worth fighting the custom tabs protocol here.

mcomella:
> in which case we default to "false" and do not show the item. And in this case it would be interesting to know if we should default to "true" and what Chrome does.

- I checked: Chrome does not display Share.
- [Here's the code where they default to `false`](https://cs.chromium.org/chromium/src/chrome/android/java/src/org/chromium/chrome/browser/customtabs/CustomTabIntentDataProvider.java?l=201&rcl=c5cbe85f310704b191dd6b6ad61e75c45c337fee). 
[...]

Issue #1179, opened 2017-08-21
Header and URL bar has seizure like flickers when scrolling horizontal images


clujin:
I'm using Android 7.1.2 with the latest Firefox Focus.

Steps to reproduce:
- Go to www.google.com
- Search something with news article that are placed horizontally like search for 'eclipse'.
- Horizontally scroll the news article bar left to right or right to left.

You see the header/notification bar at top flickering causing some degree of discomfort.

mcomella:
I can reproduce on an API 23 Nexus 5X emulator. Some cases when we change the URL bar colors:
1) Loading pages
2) When enabling/disabling tracking protection

To debug, I'd recommend throwing a breakpoint in where we change the colors and seeing where it's getting called from.

PR #1466, opened 2017-10-11
UI polishing: contextmenu (#1260)

layely:
Expected (UI/UX Team)
30620468-61691760-9d5a-11e7-9ba5-400c67231760 crop

Result:

### DIFF
--- a/app/src/main/res/layout/context_menu.xml
+++ b/app/src/main/res/layout/context_menu.xml
@@ -3,25 +3,31 @@
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
     android:orientation="vertical">
 
     <android.support.design.widget.NavigationView
+        android:theme="@style/AppTheme.ContextMenuItem"
         android:id="@+id/context_menu"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:layout_margin="0dp"
-        android:padding="0dp"
-        android:background="@android:color/transparent" />
+        android:paddingBottom="16dp"
+        android:background="@android:color/transparent"
+        app:itemTextAppearance="@style/ContextMenuTextAppearence"/>
 
     <TextView
         android:id="@+id/warning"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:background="#ff262626"
-        android:textColor="#ff8b8b8b"
+        android:textColor="#66ffffff"
         android:textSize="12sp"
-        android:padding="18dp" />
+        android:fontFamily="@string/font_roboto_regular"
+        android:paddingStart="24dp"
+        android:paddingEnd="24dp"
+        android:paddingTop="16dp"
+        android:paddingBottom="16dp" />
 
 </LinearLayout>
--- a/app/src/main/res/layout/context_menu_title.xml
+++ b/app/src/main/res/layout/context_menu_title.xml
@@ -10,6 +10,7 @@
     android:gravity="center_vertical"
     android:ellipsize="end"
     android:textSize="14sp"
-    android:padding="18dp"
-    android:textColor="#ff979797"
-    tools:text="http://www.mozilla.org" />
+    android:padding="24dp"
+    android:textColor="#ffb2b2b2"
+    tools:text="http://www.mozilla.org"
+    android:fontFamily="@string/font_roboto_regular" />
--- a/app/src/main/res/values/fonts.xml
+++ b/app/src/main/res/values/fonts.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<resources>
+    <string name="font_roboto_medium">sans-serif-medium</string>
+    <string name="font_roboto_regular">sans-serif</string>
+</resources>
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -39,4 +39,33 @@
     <style name="PopupMenu" parent="@style/Widget.AppCompat.Light.PopupMenu">
         <item name="overlapAnchor">true</item>
     </style>
+
+    <style name="AppTheme.ContextMenuItem">
+
+        <!-- All these attributes bellow need to be defined in a theme using android:theme
+            otherwise they will have no effect... -->
+        <item name="android:listPreferredItemHeightSmall">48dp</item>
+
+        <!-- Set the padding to the children (menu item) of the NavigationView
+             (instead of the NavigationView itself) allowing the hit targets
+              to span the full width of the view. For more info, see
+              https://github.com/mozilla-mobile/focus-android/pull/1466#discussion_r144679954.
+
+              Note: listPreferredItemPaddingStart and listPreferredItemPaddingEnd
+              don't work in this context because in the design_navigation_item.xml (layout for NavigationMenuItem)
+              only listPreferredItemPaddingLeft and listPreferredItemPaddingRight attribute
+              are used for paddingLeft and paddingRight properties.
+              Check design_navigation_item.xml source here :
+              https://android.googlesource.com/platform/frameworks/support.git/+/master/design/res/layout/design_navigation_item.xml
+            -->
+        <item name="listPreferredItemPaddingLeft">24dp</item>
+        <item name="listPreferredItemPaddingRight">24dp</item>
+    </style>
+
+    <style name="ContextMenuTextAppearence">
+        <item name="android:textSize">16sp</item>
+        <item name="android:fontFamily">@string/font_roboto_medium</item>
+        <item name="android:textStyle">normal</item>
+        <item name="android:textColor">#FFFFFF</item>
+    </style>
 </resources>


pocmo:
@mcomella @liuche Can anyone of you take this one?

app/src/main/res/layout/context_menu.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- This Source Code Form is subject to the terms of the Mozilla Public
+<?xml version="1.0" encoding="utf-8"?><!-- This Source Code Form is subject to the terms of the Mozilla Public
```
mcomella: nit: the file header should stay on the next line. It's unnecessary churn and, imo, it makes more sense on the next line. Here and in the next file.

app/src/main/res/values/fonts.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mcomella: nit: add license file header on the next line.

app/src/main/res/layout/context_menu.xml
```diff
@@ -12,16 +11,20 @@
...
+        android:paddingEnd="8dp"
+        android:paddingBottom="16dp" />
 
     <TextView
         android:id="@+id/warning"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:background="#ff262626"
+        android:paddingHorizontal="24dp"
+        android:paddingVertical="16dp"
```
mcomella: `paddingHorizontal/Vertical` are API 26+ and we support 21+ – can you change this to be start/end/top/bottom?

app/src/main/res/layout/context_menu_title.xml
```diff
@@ -1,15 +1,15 @@
...
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <TextView xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:maxLines="3"
-    android:gravity="center_vertical"
     android:ellipsize="end"
+    android:fontFamily="@string/font_roboto_regular"
+    android:gravity="center_vertical"
```
mcomella: nit: please set these lines to their original order! Reordering lines makes it more difficult to review your code and follow git history! You should have a really good reason for when you reorder lines!

app/src/main/res/layout/context_menu_title.xml
```diff
@@ -1,15 +1,15 @@
...
+    android:maxLines="3"
+    android:padding="24dp"
+    android:textColor="#ffb2b2b2"
```
mcomella: Where did you get this color from? I see an incorrect value in the mocks and I'm not sure this is the same color as the url text. I asked antlam for clarification in https://github.com/mozilla-mobile/focus-android/issues/1260#issuecomment-336595240 and I pinged him to respond soon!

app/src/main/res/layout/context_menu.xml
```diff
@@ -12,16 +11,20 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_margin="0dp"
-        android:padding="0dp"
-        android:background="@android:color/transparent" />
+        android:background="@android:color/transparent"
+        android:paddingTop="0dp"
+        android:paddingStart="8dp"
```
mcomella: Our hit targets look like this now:

screenshot_1507940598

I assume we want the hit targets to span the full width of the view: would it be possible to fix this? Perhaps we can add the padding to the NavigationView children instead.

app/src/main/res/layout/context_menu.xml
```diff
@@ -12,16 +11,20 @@
...
         android:id="@+id/warning"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:background="#ff262626"
+        android:paddingHorizontal="24dp"
+        android:paddingVertical="16dp"
         android:textColor="#ff8b8b8b"
-        android:textSize="12sp"
-        android:padding="18dp" />
+        android:textSize="12sp" />
```
mcomella: We should explicitly specify roboto regular here, right?

app/src/main/res/values/fonts.xml
```diff
@@ -0,0 +1,5 @@
...
+<resources>
+    <string name="font_roboto_medium">sans-serif-medium</string>
```
mcomella: I really like this! I always find it difficult to map UX specifications like `robotium_modium` to `san-serif-medium` – good work! :)

app/src/main/res/layout/context_menu.xml
layely: Nothing was specified in the mock here. But yeah I think it should be that.

app/src/main/res/layout/context_menu.xml
layely: Oh sorry that's the autoformatter.

app/src/main/res/layout/context_menu_title.xml
layely: Oops autoformatter again.

app/src/main/res/layout/context_menu_title.xml
layely: That was specified in the mock.

image

app/src/main/res/values/fonts.xml
layely: Thanks

app/src/main/res/layout/context_menu.xml
layely: Ok. Will be handle in the next commit

layely:
@mcomella the changes have been made.

app/src/main/res/layout/context_menu.xml
```diff
@@ -22,6 +26,10 @@
         android:background="#ff262626"
         android:textColor="#ff8b8b8b"
```
mcomella: Can you make this white with 40% opacity? via https://github.com/mozilla-mobile/focus-android/issues/1260#issuecomment-336601194

app/src/main/res/layout/context_menu.xml
```diff
@@ -8,11 +8,15 @@
     android:orientation="vertical">
 
     <android.support.design.widget.NavigationView
```
mcomella: How do we know the heights of these items are guaranteed to be 48dp? I searched the source and there appears to be a `listPreferredItemHeight`, a `*HeightSmall`, and a `*HeightLarge`. HeightSmall is what we want - how we know it'll get used?

Similarly, how do we know the text color, size, etc. will be what we want them to be? fwiw, there [appears to be some `itemText*` attributes](https://developer.android.com/reference/android/support/design/widget/NavigationView.html#attr_android.support.design:itemTextAppearance) - maybe we should use those?

layely: will be clearly set in my next commit

app/src/main/res/layout/context_menu.xml
```diff
@@ -8,11 +8,15 @@
...
+        android:theme="@style/ContextMenuItemStyle"
         android:id="@+id/context_menu"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_margin="0dp"
-        android:padding="0dp"
+        android:paddingTop="0dp"
```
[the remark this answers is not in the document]
layely: It's not necessary off course, so let's remove it.

app/src/main/res/layout/context_menu.xml
layely: Okay

app/src/main/res/values/styles.xml
```diff
@@ -39,4 +39,13 @@
     <style name="PopupMenu" parent="@style/Widget.AppCompat.Light.PopupMenu">
         <item name="overlapAnchor">true</item>
     </style>
+
+    <style name="ContextMenuItemStyle" parent="@style/AppTheme">
+        <!-- Set the padding to the children (menu item) of the NavigationView
```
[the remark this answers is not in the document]
layely: @mcomella  After some investigation i found that Items layout in **NavigationView** are defined in [design_navigation_item.xml](https://android.googlesource.com/platform/frameworks/support.git/+/master/design/res/layout/design_navigation_item.xml). 
```
<android.support.design.internal.NavigationMenuItemView
        xmlns:android="http://schemas.android.com/apk/res/android"
        android:layout_height="?attr/listPreferredItemHeightSmall"
        android:paddingLeft="?attr/listPreferredItemPaddingLeft"
        android:paddingRight="?attr/listPreferredItemPaddingRight"
        android:foreground="?attr/selectableItemBackground"
        android:focusable="true"/>
```

As you can see some properties are set using attributes that are specific to themes. So in order to use a custom value for a property, we need to override its related attribute.

> t'd be also great to comment that PaddingStart/End don't work in this context and a reason, if you know why (I think it's because only L/R is defined in AppCompat, whose theme we inherit from; only the system defines S/E).

I think it's just because ListPreferredItemPaddingStart and istPreferredItemPaddingEnd are not used in the source.

layely:
@mcomella commit resent

app/src/main/res/values/styles.xml
```diff
@@ -39,4 +39,31 @@
...
+              only listPreferredItemPaddingLeft and listPreferredItemPaddingRight attribute
+              are used for paddingLeft and paddingRight properties.
+              Check design_navigation_item.xml source here :
+              https://android.googlesource.com/platform/frameworks/support.git/+/master/design/res/layout/design_navigation_item.xml
+            -->
+        <item name="listPreferredItemPaddingLeft">24dp</item>
+        <item name="listPreferredItemPaddingRight">24dp</item>
+    </style>
+
+    <style name="ContextMenuTextApparence">m
```
mcomella: nit: -> `Appearance`

mcomella: No `m` at the end of the line.

app/src/main/res/layout/context_menu.xml
```diff
@@ -5,23 +5,29 @@
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
```
mcomella: nit: `xmlns:` attrs should be at the top of the tag, above all the other attrs.

app/src/main/res/values/styles.xml
```diff
@@ -39,4 +39,31 @@
     <style name="PopupMenu" parent="@style/Widget.AppCompat.Light.PopupMenu">
         <item name="overlapAnchor">true</item>
     </style>
+
+    <style name="AppTheme.ContextMenuItem">
+        <!-- All these attributes bellow need to be defined in a theme using android:theme
+            otherwise they will have no effect... -->
+        <item name="android:listPreferredItemHeightSmall">48dp</item>
+        <!-- Set the padding to the children (menu item) of the NavigationView
```
mcomella: nit: typically we have a line of whitespace above all comments

mcomella:
@layely By the way, do you know if there's any more fixes required to complete #1260? Do you want me to find you another issue to work on after this one?

app/src/main/res/layout/context_menu.xml
layely: fixed

app/src/main/res/values/styles.xml
layely: fixed

app/src/main/res/values/styles.xml
layely: fixed (forgive my english, it's because my official language is french kxkx).

layely: fixed

layely:
> This looks good! There are just a few more nits. If you want to save time, I could fix these up for you and land it (if so, let me know).

I did some changes again, but feel free to fix them this time.

> @layely By the way, do you know if there's any more fixes required to complete #1260?

Yes, now we have to polish the **home_screen dialog** and **the popup menu** which i think won't take much time.
I will send a commit about it very soon.

> Do you want me to find you another issue to work on after this one?

Yes please, it would be just great for me. Whether in this project or any other related to Java and JavaScript. I learn too much from you, Thank you.

mcomella:
> Yes, now we have to polish the home_screen dialog and the popup menu which i think won't take much time.
I will send a commit about it very soon.

Great – feel free to flag me as a reviewer!

> Yes please, let me get more involved, it would be just great for me. Whether in this project or any other related to Java and JavaScript. I learn too much from you, Thank you.

Glad we can help! :) I found two bugs that might be interesting:
- #1179: The header flickers when scrolling horizontally on Google Image results
- #1358: The "Share" item is missing when opening Focus from external apps

app/src/main/res/values/styles.xml
```diff
@@ -39,4 +39,33 @@
...
+              only listPreferredItemPaddingLeft and listPreferredItemPaddingRight attribute
+              are used for paddingLeft and paddingRight properties.
+              Check design_navigation_item.xml source here :
+              https://android.googlesource.com/platform/frameworks/support.git/+/master/design/res/layout/design_navigation_item.xml
+            -->
+        <item name="listPreferredItemPaddingLeft">24dp</item>
+        <item name="listPreferredItemPaddingRight">24dp</item>
+    </style>
+
+    <style name="ContextMenuTextAppearence">
```
mcomella: nit: this should be "Appearance" but I don't think it's worth the churn at this point.

mcomella:
Adding this to the milestone because we don't have an issue for this small amount of work.

Issue #1453, opened 2017-10-09
Export Granite Strings


Sdaswani:

liuche:
Checked in with UX and it sounds like there will be strings up this evening: strings for the feature as well as alternative strings for a fallback version if the scope of the mocks ends up being too big for the feature (and Brian will be on PTO, so we can't get more strings).

I'll reassign if I need someone else to export these strings for Wednesday.

liuche:
@pocmo If you don't get to this tomorrow during your work day I'll export the strings!

liuche:
PR opened to l10n, and pinged Delphine. It might be at the end of her day so I'll monitor and make changes if needed.

...also for the record, doing strings export without mocks is *extremely* painful, and requires a lot of cross-referencing and constant real-time discussion, so let's not give UX such little lead time next time between prioritization and string freeze.

pocmo:
> ...also for the record, doing strings export without mocks is extremely painful, and requires a lot of cross-referencing and constant real-time discussion, so let's not give UX such little lead time next time between prioritization and string freeze.

Absolutely agree. I'm more in more in favor of adding more time to translate things at the end of the sprint. If we run into issues actually implementing the feature we have no time to adapt anymore.

Sdaswani:
Localizing after the strings are set is an option, but it’s a bit risky as since we don’t have an SLA that strings will be localized in a certain time frame it could risk our release date, which other cross functional teams depend on. We used to indeed localize at the end of development at Lookout but the localization turn around was something we could plan for and it was pretty short (2-3 Days).

Sdaswani:
Also @antlam suggested we release features for English first and let the localizations follow but I’m worried about branching UI based on language 😂.

PR #1467, opened 2017-10-12
Strings for "Add search engine" feature.

liuche:
Added strings for #1453 for Granite.

The additional changes beyond what's included in the strings doc are:
- Adding an error string (per Brian Jones)
- Removing the clickable substring for manually adding a search engine
- Adding strings for manually adding search engine "MANUALLY ADD" and "CANCEL"
- Removed the "[]" around hint strings
- Added "Save" to save manually added search engines
- Added accessibility text

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -151,6 +151,55 @@
     <!-- Item in Preferences that shows the about page. Parameter 1 is the app name, i.e. Firefox Focus/Firefox Klar. -->
     <string name="preference_about">About %1$s</string>
 
+    <!-- Header for the list of installed search engines -->
+    <string name="preference_search_installed_search_engines">Installed search engines</string>
+
+    <!-- Action for restoring the default list of search engines -->
+    <string name="preference_search_restore">Restore default search engines</string>
+
+    <!-- Action for adding an additional search engine to the list -->
+    <string name="preference_search_add">Add another search engine</string>
+    <string name="preference_search_remove_title">Remove search engines</string>
+    <string name="preference_search_remove">Remove</string>
+
+    <!-- Clickable text in tutorial to open a dialog to manually add a search engine -->
+    <string name="tutorial_search_manually_add">Manually add</string>
+
+    <string name="tutorial_search_title">Add search engine</string>
+    <string name="tutorial_search_step1">1. Go to the search engine homepage.</string>
+    <string name="tutorial_search_step2">2. Tap into the search engine’s search field.</string>
+    <!-- An alternative string for Step 2 of the tutorial -->
+    <string name="tutorial_search_step2_alternative">2. Long press the search engine’s search field.</string>
+    <string name="tutorial_search_step3">3. Tap the magnifying glass icon above the keyboard to add the search engine.</string>
+    <!-- Clickable action text that shows when user long-presses on a search engine search field (this is similar to paste or copy when long-pressing a text field) -->
+    <string name="action_option_add_search_engine">Add search engine</string>
+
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -151,6 +151,55 @@
...
+    <!-- Action for restoring the default list of search engines -->
+    <string name="preference_search_restore">Restore default search engines</string>
+
+    <!-- Action for adding an additional search engine to the list -->
+    <string name="preference_search_add">Add another search engine</string>
+    <string name="preference_search_remove">Remove</string>
+
+    <!-- Clickable text in tutorial to open a dialog to manually add a search engine -->
+    <string name="tutorial_search_manually_add">MANUALLY ADD</string>
+    <string name="tutorial_search_cancel">CANCEL</string>
```
pocmo: I wouldn't capitalize the strings here (and further down). For buttons that's done by Android automatically (in languages where it makes sense!) and in other locations we can use the android:capitalize attribute to control it ourselves.

app/src/main/res/values/strings.xml
```diff
@@ -151,6 +151,55 @@
...
+    <string name="tutorial_search_title">Add search engine</string>
+    <string name="tutorial_search_step1">1. Go to the search engine homepage.</string>
+    <string name="tutorial_search_step2">2. Tap into the search engine\'s search field.</string>
+    <!-- An alternative string for Step 2 of the tutorial -->
+    <string name="tutorial_search_step2_alternative">2. Long press the search engine\'s search field.</string>
+    <string name="tutorial_search_step3">3. Tap the magnifying glass icon above the keyboard to add the search engine.</string>
+    <!-- Clickable action text that shows when user long-presses on a search engine search field (this is similar to paste or copy when long-pressing a text field) -->
+    <string name="action_option_add_search_engine">ADD SEARCH ENGINE</string>
+
+    <!-- An alternative string for Step 3 of the tutorial. Parameter 1 is &action_option_add_search_engine, which is the action text that the tutorial is referring to -->
```
pocmo: nit: &action_option_add_search_engine looks like you are referencing a Gecko string because of the & :)

app/src/main/res/values/strings.xml
pocmo: Ah, just saw that we also use this string in placeholders. In this case it might be a bit more complicated than I thought - not sure if we want to call uppercase() on the string before replacing the placeholder?

app/src/main/res/values/strings.xml
```diff
@@ -151,6 +151,55 @@
...
+
+    <string name="search_add_hint">Tap the magnifying glass icon above the keyboard to add this search engine.</string>
+
+    <string name="search_add_manually_name">Name to display</string>
+    <string name="search_add_manually_name_hint">Search engine name</string>
+    <string name="search_add_manually_string">Search string to use</string>
+    <string name="search_add_manually_save">Save</string>
+
+    <!-- Explanation of how to add search string, with instruction on how to modify the string if necessary. The string "%s" should not be changed or removed. -->
+    <string name="search_add_manually_string_hint">Paste or enter search string. If necessary, replace search term with: %s.</string>
```
pocmo: Here (and for the string below) I wonder whether Android will understand %s as a placeholder and whether we need to escape it somehow. I'd test this quickly in a layout.

liuche: I tested this in Focus quickly and it works -  it looks like as long as the %s isn't in the same string as other formatted strings (%1$s) it won't cause problems.

app/src/main/res/values/strings.xml
liuche: Sounds good, I'll capitalize these normally. I also removed Cancel because we already have that as a general string.

Issue #1436, opened 2017-10-04
NullPointerException: DownloadManager.enqueue


pocmo:
```
java.lang.NullPointerException: 
  at android.app.DownloadManager.enqueue (DownloadManager.java:1026)
  at org.mozilla.focus.fragment.BrowserFragment.queueDownload (BrowserFragment.java:659)
  at org.mozilla.focus.fragment.BrowserFragment.access$1300 (BrowserFragment.java:78)
  at org.mozilla.focus.fragment.BrowserFragment$8.onDownloadStart (BrowserFragment.java:476)
  at org.mozilla.focus.menu.WebContextMenu$2.onNavigationItemSelected (WebContextMenu.java:138)
  at android.support.design.widget.NavigationView$1.onMenuItemSelected (NavigationView.java:154)
  at android.support.v7.view.menu.MenuBuilder.dispatchMenuItemSelected (MenuBuilder.java:822)
  at android.support.v7.view.menu.MenuItemImpl.invoke (MenuItemImpl.java:171)
  at android.support.v7.view.menu.MenuBuilder.performItemAction (MenuBuilder.java:973)
  at android.support.design.internal.NavigationMenuPresenter$1.onClick (NavigationMenuPresenter.java:342)
```

pocmo:
b10f55f18c6ad959d43efec81c1b9c9c14d725d1

PR #1476, opened 2017-10-13
Added try catch blocks for downloads (#1436)

ekager:
Added some error handling for queuing downloads [copied from Firefox ](https://dxr.mozilla.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/BrowserApp.java#2175)
Not sure how to reproduce the original NPE error, but can continue to investigate. I thought if I tried to switch sessions while the download was queuing it would lead to a NPE because maybe the manager would be null, but I couldn't switch sessions fast enough. What do you think?


pocmo:
> Not sure how to reproduce the original NPE error, but can continue to investigate. 

Did you try stackoverflow etc.? Wouldn't be surprised if this is just something we have to expect from DownloadManager in certain Android versions. :(
You could also try "blame" on the Fennec code and see if the linked bug contains any helpful information.

PR #1478, opened 2017-10-13
Update check_translations.py

d6thsense:
Added "Error: and Warning: tags in print statements"


mcomella:
#1434.

Unfortunately, I have to run but hopefully @pocmo will get to this Monday.

d6thsense:
Thanks :)

PR #1480, opened 2017-10-16
Fix SwitchLocalTest Failure

npark-mozilla:


npark-mozilla:
moving it to Beforeclass seems to make it work - what do you think?

PR #1484, opened 2017-10-17
Update search engines preferences to use RadioButtons

liuche:
This is the first step to adding the Preference for manually adding a new search engine. (The button for manually adding search engines will be added below all the search engines.) Could also land this totally separately to get more testing/avoid bitrot.

screenshot_1508884140

Opening a PR right now just to make sure things are going the right direction, the Settings code was a little messy.

@lime124, I also added the icons which wasn't in your mock - let me know if that's desired (I can take it out easily), or if it makes the "add a manual search engine" messier (because we might not have an icon for those).


app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -76,6 +100,10 @@ public void onPause() {
         getPreferenceManager().getSharedPreferences().unregisterOnSharedPreferenceChangeListener(this);
     }
 
+    public void setTitleUpdater(TitleUpdater titleUpdater) {
+        this.titleUpdater = titleUpdater;
+    }
```
pocmo: `this.titleUpdater` will be null if the `Fragment` gets destroyed while the app is in the background and the next time you resume the app. Android will re-create the fragment but no one is calling `setTitleUpdater()` on it. Only the arguments `Bundle` will be preserved and passed to the new instance again.

As the `TitleUpdater` is in fact the activity of the fragment, you can remove this reference completely and just cast `getActivity()` to `TitleUpdater` where needed (Watch out, getActivity() could be null). If you want to be explicit about the activity having to implement this interface then you could override onAttach() and throw if it isn't an `instanceof TitleUpdater`.

app/src/main/java/org/mozilla/focus/search/SearchEngineAdapter.java
```diff
@@ -1,71 +0,0 @@
...
-import android.widget.TextView;
-
-import org.mozilla.focus.R;
-
-import java.util.List;
-
-/**
- * Adapter for displaying a list of search engines.
- */
-public class SearchEngineAdapter extends BaseAdapter {
```
pocmo: We might be able to remove more:
* SearchEnginePreference
* item_search_engine.xml

lime124:
@liuche hopefully this helps!

visual design

PR #1491, opened 2017-10-18
Address intermittent UITest faults in BB

npark-mozilla:
Seems like espresso is much faster than UiAutomator, exposing timing issues.

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/activity/AddtoHSTest.java
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/AddtoHSTest.java
@@ -243,8 +243,7 @@ public class AddtoHSTest {
         TestHelper.inlineAutocompleteEditText.setText("helloworld");
         TestHelper.hint.waitForExists(waitingTime);
         TestHelper.pressEnterKey();
-        TestHelper.progressBar.waitForExists(webPageLoadwaitingTime);
-        Assert.assertTrue(TestHelper.progressBar.waitUntilGone(webPageLoadwaitingTime));
+        TestHelper.progressBar.waitUntilGone(webPageLoadwaitingTime);
 
         openAddtoHSDialog();
 
--- a/app/src/androidTest/java/org/mozilla/focus/activity/SwitchLocaleTest.java
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/SwitchLocaleTest.java
@@ -10,8 +10,8 @@ import android.content.res.Configuration;
 import android.content.res.Resources;
 import android.preference.PreferenceManager;
 import android.support.test.InstrumentationRegistry;
+import android.support.test.filters.Suppress;
 import android.support.test.rule.ActivityTestRule;
-import android.support.test.runner.AndroidJUnit4;
 import android.support.test.uiautomator.UiObject;
 import android.support.test.uiautomator.UiObjectNotFoundException;
 import android.support.test.uiautomator.UiScrollable;
@@ -21,7 +21,6 @@ import org.junit.After;
 import org.junit.Assert;
 import org.junit.Rule;
 import org.junit.Test;
-import org.junit.runner.RunWith;
 
 import java.util.Locale;
 
@@ -33,7 +32,7 @@ import static org.mozilla.focus.activity.helpers.EspressoHelper.openSettings;
 import static org.mozilla.focus.fragment.FirstrunFragment.FIRSTRUN_PREF;
 
 // This test checks all the headings in the Settings menu are there
-@RunWith(AndroidJUnit4.class)
+@Suppress
 public class SwitchLocaleTest {
 
     @Rule
--- a/app/src/androidTest/java/org/mozilla/focus/activity/TestHelper.java
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/TestHelper.java
@@ -5,6 +5,7 @@
 
 package org.mozilla.focus.activity;
 
+import android.graphics.Rect;
 import android.os.RemoteException;
 import android.support.annotation.NonNull;
 import android.support.test.InstrumentationRegistry;
@@ -179,7 +180,7 @@ public final class TestHelper {
             .resourceId("org.mozilla.focus.debug:id/trackers_count"));
     static UiObject blockToggleSwitch = mDevice.findObject(new UiSelector()
             .resourceId("org.mozilla.focus.debug:id/blocking_switch"));
-    static UiObject menulist = mDevice.findObject(new UiSelector()
+    public static UiObject menulist = mDevice.findObject(new UiSelector()
             .resourceId("org.mozilla.focus.debug:id/list")
             .enabled(true));
     static String getMenuItemText(UiObject item) throws UiObjectNotFoundException {
@@ -232,6 +233,14 @@ public final class TestHelper {
     }
     static void openNotification() {
         mDevice.openNotification();
+    }
+
+    static void clearAllNotification() throws UiObjectNotFoundException {
+        UiObject clearButton = mDevice.findObject(new UiSelector()
+                .resourceId("com.android.systemui:id/dismiss_text")
+                .enabled(true));
+        clearButton.click();
+        clearButton.waitUntilGone(waitingTime);
     }
 
     static void swipeUpScreen () {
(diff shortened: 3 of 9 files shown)


app/src/androidTest/java/org/mozilla/focus/activity/TestHelper.java
```diff
@@ -312,4 +321,25 @@ static void waitForWebSiteTitleLoad() {
...
+                .description("Firefox Focus")
+                .enabled(true));
+        container.waitForExists(waitingTime);
+        mainLabel.waitForExists(waitingTime);
+
+        Rect containerBound = container.getBounds();
+        Rect labelBound = mainLabel.getBounds();
+
+
+        if (labelBound.bottom > containerBound.bottom/2) {
```
sputnikci: \[Checkstyle\] ERROR: '/' is not preceded with whitespace\.

sputnikci: \[Checkstyle\] ERROR: '/' is not followed by whitespace\.

npark-mozilla:
Not ready for review yet.

pocmo:
Please wait with that a bit. I have an almost done patch ready to split ScreengrabTest into multiple tests.... to avoid having a gigantic test where one failure means the whole thing stops. :)

npark-mozilla:
sure, that sounds great.  btw, I think now we addressed all of the intermittents - the tests are passing 3 times in a row in BB.  (I've disabled SwitchLocaleTest in this PR because the fix for that is on my other PR)

app/src/androidTest/java/org/mozilla/focus/activity/AddtoHSTest.java
```diff
@@ -243,8 +243,7 @@ public void SearchTermShortcutTest() throws InterruptedException, UiObjectNotFou
         TestHelper.inlineAutocompleteEditText.setText("helloworld");
         TestHelper.hint.waitForExists(waitingTime);
         TestHelper.pressEnterKey();
-        TestHelper.progressBar.waitForExists(webPageLoadwaitingTime);
-        Assert.assertTrue(TestHelper.progressBar.waitUntilGone(webPageLoadwaitingTime));
+        TestHelper.progressBar.waitUntilGone(webPageLoadwaitingTime);
```
pocmo: Can you explain why you removed the assertion? Without the assert this is basically just a sleep, isn't it?

app/src/androidTest/java/org/mozilla/focus/activity/SwitchLocaleTest.java
```diff
@@ -33,7 +32,7 @@
 import static org.mozilla.focus.fragment.FirstrunFragment.FIRSTRUN_PREF;
 
 // This test checks all the headings in the Settings menu are there
-@RunWith(AndroidJUnit4.class)
+@Suppress
```
pocmo: What's the goal here? Let's either fix the test or remove it. If we suppress it for a longer time then it will not work anymore eventually anyways.

app/src/androidTest/java/org/mozilla/focus/activity/TestHelper.java
```diff
@@ -234,6 +235,14 @@ static void openNotification() {
         mDevice.openNotification();
     }
 
+    static void clearAllNotification() throws UiObjectNotFoundException {
+        UiObject clearButton = mDevice.findObject(new UiSelector()
+                .resourceId("com.android.systemui:id/dismiss_text")
+                .enabled(true));
+        clearButton.click();
+        clearButton.waitUntilGone(waitingTime);
+    }
```
pocmo: I don't see this new method being used in this patch?

app/src/androidTest/java/org/mozilla/focus/activity/TestHelper.java
```diff
@@ -312,4 +321,25 @@ static void waitForWebSiteTitleLoad() {
...
         Assert.assertTrue("Website title loaded", titleMsg.waitForExists(waitingTime));
     }
+
+    // Checks whether keyboard is visible by looking at the coord of the logo w.r.t. container
+    public static boolean checkKeyboardPresenceInMainView() throws UiObjectNotFoundException {
+        UiObject container = TestHelper.mDevice.findObject(new UiSelector()
+                .resourceId("org.mozilla.focus.debug:id/container")
+                .enabled(true));
+        UiObject mainLabel = TestHelper.mDevice.findObject(new UiSelector()
+                .description("Firefox Focus")
```
pocmo: Let's start to use strings we get from the context object instead of plain English strings, for new code.

app/src/androidTest/java/org/mozilla/focus/activity/TestHelper.java
```diff
@@ -234,6 +235,14 @@ static void openNotification() {
         mDevice.openNotification();
     }
 
+    static void clearAllNotification() throws UiObjectNotFoundException {
+        UiObject clearButton = mDevice.findObject(new UiSelector()
```
pocmo: nit: Here and in a bunch of other cases this can be `final`.

app/src/androidTest/java/org/mozilla/focus/activity/ToggleBlockTest.java
```diff
@@ -73,10 +71,22 @@ public void tearDown() throws Exception {
...
-        onView(withId(R.id.url_edit))
-                .check(matches(isDisplayed()))
-                .check(matches(hasFocus()))
-                .perform(typeText("mozilla"), pressImeActionButton());
+        //
+        //onView(withId(R.id.url_edit))
+        //        .check(matches(isDisplayed()))
+        //        .check(matches(hasFocus()))
+        //        .perform(typeText("mozilla"), pressImeActionButton());
+        // Above types "mmozilla"
```
pocmo: We had this problem before and started to use replaceText() instead of typeText(). This seemed to work.

app/src/androidTest/java/org/mozilla/focus/activity/ToggleBlockTest.java
```diff
@@ -73,10 +71,22 @@ public void tearDown() throws Exception {
...
+        TestHelper.inlineAutocompleteEditText.waitForExists(waitingTime);
+        TestHelper.inlineAutocompleteEditText.clearTextField();
+        TestHelper.inlineAutocompleteEditText.setText("mozilla");
+        TestHelper.hint.waitForExists(waitingTime);
+        TestHelper.pressEnterKey();
+        TestHelper.webView.waitForExists(waitingTime);
```
pocmo: For in-app tests I would prefer if we switch to Espresso which is way more powerful instead of going back to uiautomator.

app/src/androidTest/java/org/mozilla/focus/activity/screenshots/ScreenGrabTest.java
```diff
@@ -177,7 +175,7 @@ public void screenGrabTest() throws InterruptedException, UiObjectNotFoundExcept
         takeScreenshotOfErrorPages(context, device);
     }
 
-    private void takeScreenshotsOfFirstrun(Context context, UiDevice device) throws UiObjectNotFoundException {
+    private void takeScreenshotsOfFirstrun(Context context, UiDevice device) throws UiObjectNotFoundException, InterruptedException {
```
pocmo: I opened a new PR for separating this test class into multiple classes. Let's remove those changes here and apply them to the new classes if needed.

app/src/androidTest/java/org/mozilla/focus/activity/AddtoHSTest.java
npark-mozilla: I think .waitUntilGone gets awaken when the event (progressbar disappearing) gets triggered, so it's technically not a sleep().  At the time, I thought assert was failing when there is no progressBar to disappear.  (it was exiting right away)  I'll double check though because I can't find doc that confirms/denies what I saw.

app/src/androidTest/java/org/mozilla/focus/activity/SwitchLocaleTest.java
npark-mozilla: oh I suppressed it because fix for the SwitchLocaleTest is in other PR, and failing this test will affect all subsequent tests (since it changes Locale).  One of the reason why this PR wasn't ready for the review :)

app/src/androidTest/java/org/mozilla/focus/activity/TestHelper.java
npark-mozilla: yes, I thought this can help our notification issue, then in some sim, where there were multiple notifications, this 'clear all' button was not visible.  I left it but since I'm not using it, perhaps I should remove it.

app/src/androidTest/java/org/mozilla/focus/activity/TestHelper.java
npark-mozilla: will do

app/src/androidTest/java/org/mozilla/focus/activity/screenshots/ScreenGrabTest.java
npark-mozilla: agree, will wait until https://github.com/mozilla-mobile/focus-android/pull/1505 and https://github.com/mozilla-mobile/focus-android/pull/1480 are merged.

pocmo:
@npark-mozilla Are there still things in here that need to land or can this be closed?

npark-mozilla:
No I think with the separation of tests, it resolved most of the issues seen here.

Issue #1470, opened 2017-10-12
Create decision task for screenshots


pocmo:
meta: #862
- [x] script the decision task to schedule dummy tasks.
- [x] modify the decision task to chunk screenshots.
- [ ] modify the decision task to support ChainOfTrust
- [ ] modify the decision task to implement the real graph.
- [ ] reach out to the TC team to get dedicated AMI. (They might allow you to use the same AMIs)

pocmo:
Closing this - Can't continue with the ChainOfTrust part without time the release engineering team. Will focus on releases for now.

Issue #1415, opened 2017-09-28
OnWebView() call in ScreenGrabTest times out intermittently


nojunpark:
On average about once per 60 run, this (https://github.com/mozilla-mobile/focus-android/blob/master/app/src/androidTest/java/org/mozilla/focus/activity/screenshots/ScreenGrabTest.java#L492) call will time out with below error after repeated calls in a loop.  This is different from element not being found, it seems to me it's waiting for the lock to be freed.  Looking at the video, the elements are present as well.

This does not seem to happen with .waitForExists call. is there a proper way to fix it, or should we go back to .waitForExists for all APIs? @pocmo, thoughts?

java.lang.RuntimeException: java.util.concurrent.TimeoutException: Waited 10 seconds for android.support.test.espresso.core.internal.deps.guava.util.concurrent.AbstractTransformFutureXTransformFuture@982ad95[status=PENDING, info=[inputFuture=[android.support.test.espresso.core.internal.deps.guava.util.concurrent.SettableFuture@3c57eaa[status=PENDING]], function=[android.support.test.espresso.web.action.AtomActionX3@92aea9b]]]
	at android.support.test.espresso.web.sugar.WebXWebInteractionXExceptionPropagator.<init>(Web.java:326)
	at android.support.test.espresso.web.sugar.WebXWebInteraction.doEval(Web.java:298)
	at android.support.test.espresso.web.sugar.WebXWebInteraction.withElement(Web.java:208)
	at org.mozilla.focus.activity.screenshots.ScreenGrabTest.takeScreenshotOfErrorPages(ScreenGrabTest.java:493)
	at org.mozilla.focus.activity.screenshots.ScreenGrabTest.screenGrabTest(ScreenGrabTest.java:168)
	at java.lang.reflect.Method.invoke(Native Method)
	at org.junit.runners.model.FrameworkMethodX1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at android.support.test.rule.ActivityTestRuleXActivityStatement.evaluate(ActivityTestRule.java:433)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
[...]

PR #1505, opened 2017-10-19
Split ScreenGrabTest into multiple tests. (#1470)

pocmo:
This has two advantages:
* If one of the test fails then the others can still run and we can get
  screenshots for them. Currently if ScreenGrabTest fails then all
  following screenshots are missing.
* Independent tests are more stable and easier to tweak.


pocmo:
Interestingly I see uiautomator sometimes randomly failing on the emulator locally (API 21). The UI is clearly showing up on the emulator but for some reason uiautomator can't see it. :(

app/src/androidTest/java/org/mozilla/focus/activity/screenshots/ErrorPagesScreenshots.java
```diff
@@ -0,0 +1,101 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.activity.screenshots;
+
+import android.os.Build;
+import android.support.test.espresso.Espresso;
```
sputnikci: \[Checkstyle\] ERROR: Unused import \- android\.support\.test\.espresso\.Espresso\.

app/src/androidTest/java/org/mozilla/focus/activity/screenshots/ErrorPagesScreenshots.java
```diff
@@ -0,0 +1,101 @@
...
+import static android.support.test.espresso.assertion.ViewAssertions.matches;
+import static android.support.test.espresso.matcher.ViewMatchers.hasFocus;
+import static android.support.test.espresso.matcher.ViewMatchers.isDisplayed;
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+import static android.support.test.espresso.web.sugar.Web.onWebView;
+import static android.support.test.espresso.web.webdriver.DriverAtoms.findElement;
+import static android.support.test.espresso.web.webdriver.DriverAtoms.webClick;
+import static android.support.test.espresso.web.webdriver.DriverAtoms.webScrollIntoView;
+import static junit.framework.Assert.assertTrue;
+import static org.mozilla.focus.activity.TestHelper.browserURLbar;
```
sputnikci: \[Checkstyle\] ERROR: Unused import \- org\.mozilla\.focus\.activity\.TestHelper\.browserURLbar\.

app/src/androidTest/java/org/mozilla/focus/activity/screenshots/ErrorPagesScreenshots.java
```diff
@@ -0,0 +1,99 @@
...
+            assertTrue(TestHelper.progressBar.waitUntilGone(waitingTime));
+
+            // Android O has an issue with using Locator.ID
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                UiObject tryAgainBtn = device.findObject(new UiSelector()
+                        .descriptionContains(getString(R.string.errorpage_refresh))
+                        .clickable(true));
+                assertTrue(tryAgainBtn.waitForExists(waitingTime));
+            } else {
+                onWebView()
```
npark-mozilla: So when I run a suite locally on the sim, there are no issues, but when I run this test alone, I always run into an error here: 

Caused by: java.util.concurrent.TimeoutException: Waited 10 seconds for android.support.test.espresso.core.internal.deps.guava.util.concurrent.AbstractTransformFuture$TransformFuture@2c24f97[status=PENDING, info=[inputFuture=[android.support.test.espresso.core.internal.deps.guava.util.concurrent.SettableFuture@6c61584[status=PENDING]], function=[android.support.test.espresso.web.action.AtomAction$3@4fd916d]]]

This seems like some internal espresso process is blocked by something.  This is also why I replaced this part with the uiautomator PR. (https://github.com/mozilla-mobile/focus-android/issues/1415)

app/src/androidTest/java/org/mozilla/focus/activity/screenshots/BrowserScreenScreenshots.java
```diff
@@ -0,0 +1,207 @@
...
+        onView(withId(R.id.url_edit))
+                .check(matches(isDisplayed()))
+                .check(matches(hasFocus()))
+                .perform(click(), replaceText(webServer.url("/").toString()), pressImeActionButton());
+
+        device.findObject(new UiSelector()
+                .resourceId("org.mozilla.focus.debug:id/webview")
+                .enabled(true))
+                .waitForExists(waitingTime);
+
```
npark-mozilla: let's also add check of progressBar to disappear. Sometimes webview appears, but it's still loading.  (esp. on slower sims)

public static UiObject progressBar = mDevice.findObject(new UiSelector()
            .resourceId("org.mozilla.focus.debug:id/progress")
            .enabled(true));

app/src/androidTest/java/org/mozilla/focus/activity/screenshots/ErrorPagesScreenshots.java
pocmo: On what configuration are you running the tests?

app/src/androidTest/java/org/mozilla/focus/activity/screenshots/NotificationScreenshots.java
```diff
@@ -0,0 +1,108 @@
...
+                .enabled(true));
+
+        device.openNotification();
+
+        try {
+            if (!openAction.waitForExists(waitingTime)) {
+                // The notification is not expanded. Let's expand it now.
+                device.findObject(new UiSelector()
+                        .text(getString(R.string.app_name)))
+                        .swipeDown(5);
```
npark-mozilla: I noticed low step number of .swipeDown() can cause unpredictable results, try experimenting with double digit numbers (for me 50 seems pretty safe)

pocmo: Oh, I forgot to update this PR. Locally I was using 50. On some emulators it seems to trigger a long-press which is super awkward.

pocmo:
> Also, just want to double check, are all onView() enclosed checks come with waiting for a few seconds? (they probably are, but if not, we should make them wait)

This shouldn't be needed in Espresso world: Espresso will wait for all UI thread operations to be completed before moving on (idle state). And if this is not enough then there are ways to add additional state checks that need to be done before the device is considered idle (e.g. I wrote SessionLoadedIdlingResource which - when activated for the test - will wait for the current website to be loaded completely before moving on).

However in some tests we are mixing uiautomator and espresso which might lead to problems. That's one of the reason why I would like to go with espresso-only tests where applicable (only UI in-app).

Issue #1516, opened 2017-10-19
Adaptive shortcut icons do not adapt


pocmo:
Google Pixel XL, Google Launcher

I switched the icon shape in the launcher from circle to squircle. The home screen shortcuts didn't adapt to that. They are still circles, but on a white squircle. I added a new shortcut after the change and it also isn't a squircle.

pocmo:
![screenshot_20171019-204452](https://user-images.githubusercontent.com/89638/31788442-f9055086-b50e-11e7-9df7-726626d1583b.png)

mcomella:
Originally implemented in #1214.

pocmo:
9d34d919b265a245bf95dcca110b0925ca559cbb

npark-mozilla:
@pocmo When I run the latest build, I see the circle shortcut (but no white squircle) on Pixel 2 device, Android 8.0, instead of squircle. Is this specific to certain android api version?

ekager:
Hi @npark-mozilla ! You have to [enable UI Tuner on Pixel ](https://www.howtogeek.com/270568/how-to-enable-android%E2%80%99s-%E2%80%9Csystem-ui-tuner%E2%80%9D-for-access-to-experimental-features/) and then [change icon shape](https://www.androidauthority.com/change-icon-shape-pixel-launcher-773473/).

pocmo:
If you have the Google launcher installed you can also long press the home screen -> "home settings" -> Change icon shape

npark-mozilla:
Ahh, thanks!

PR #1519, opened 2017-10-19
Adaptive pinned icons squircles (#1516)

ekager:


pocmo:
Forwarded the review to @mcomella who wrote the initial patch for adaptive shortcuts.

mcomella:
> so we'll have to figure out how to create the bitmap differently.

We do this in [`IconGenerator.generateAdaptiveLauncherIcon`](https://github.com/mozilla-mobile/focus-android/blob/01ec0e6cf8c260af62f20911b164c1feb0276517/app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.kt#L65). Some approaches you could try:
- I could be getting the icon dimensions incorrect (though I took them from the AdaptiveIconDrawable source so probably not)
- I think converting the AdaptiveIconDrawable into the bitmap applies the transformation – we could just create the icon inside a regular Bitmap and pass that into `IconCompat`.

To be explicit, the transformation includes both rounding the corners and adjusting the padding, which is why I think the text size increases (it crops out the padding twice with your code).

ekager:
Okay I changed this to create a Bitmap with background and foreground and removed all Drawable conversions! Sorry I didn't delve into your comments the first time around! :)

app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.kt
```diff
@@ -65,28 +59,37 @@ class IconGenerator {
...
-         * is applied so it'd be incorrect for us to use createWithAdaptiveBitmap, which would apply the formatting
-         * again.
-         *
-         * Unfortunately, we also can't just pass our legacy icon into createWithAdaptiveBitmap because the icon
-         * background has the corners removed and some device might display icons that have less of the corner missing
-         * than we do. In that case, we'd be unable to add the corners back and the icon wouldn't conform.
          */
         @TargetApi(26)
-        private fun generateAdaptiveLauncherIcon(context: Context, url: String?): AdaptiveIconDrawable {
+        private fun generateAdaptiveLauncherIcon(context: Context, url: String?): Bitmap {
```
mcomella: Most of this code is duplicated from `generateIcon`: can we merge the implementations? Duplication is bad because if we update one implementation, we often forget to update the others!

app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.kt
```diff
@@ -40,60 +33,36 @@ class IconGenerator {
...
+            return drawCharacterOnBitmap(context, url, bitmap)
         }
 
-        /**
-         * Generate an icon for this website based on the URL.
-         */
-        private fun generateIcon(context: Context, url: String?, iconSize: Int): Bitmap {
-            val bitmap = Bitmap.createBitmap(iconSize, iconSize, Bitmap.Config.ARGB_8888)
+        private fun drawCharacterOnBitmap(context: Context, url: String?, bitmap: Bitmap): Bitmap {
             val canvas = Canvas(bitmap)
```
mcomella: I don't think anything needs to change here but I just want to note it: it's a little weird that we create a second Canvas (in the Adaptive icon case). But [the docs say](https://developer.android.com/reference/android/graphics/Canvas.html#Canvas(android.graphics.Bitmap)), "Construct a canvas with the specified bitmap to draw into." so it sounds like all the state is in the Bitmap, not the canvas, and this would be fine. It's a one time method so I don't think we care about the performance implications either.

app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.kt
```diff
@@ -40,60 +33,36 @@ class IconGenerator {
...
-
             val adaptiveIconDimen = res.getDimensionPixelSize(R.dimen.adaptive_icon_drawable_dimen)
-            val foregroundLetter = BitmapDrawable(res, generateIcon(context, url, adaptiveIconDimen))
 
-            return AdaptiveIconDrawable(backgroundColor, foregroundLetter)
+            val bitmap = Bitmap.createBitmap(adaptiveIconDimen, adaptiveIconDimen, Bitmap.Config.ARGB_8888)
+            val canvas = Canvas(bitmap)
+
+            // Draw background to comply with Adaptive Icon guidelines
+            canvas.drawColor(res.getColor(R.color.add_to_homescreen_icon_background))
```
mcomella: nit: `getColor` is deprecated: you can use `ContextCompat.getColor`.

PR #1520, opened 2017-10-20
UI polishing: Add to home screen dialog etc (#1260) 

layely:
@mcomella, please have a look here.

screenshot-2017-10-20 30620468-61691760-9d5a-11e7-9ba5-400c67231760 png png image 1450 x 1490 pixels

## Note: 

Some corrections were made here -->

- **Add button** color must be 0A84FF (100%) instead of 009688
- **Scene.ca** text color is FFFFFF 70% (Not in the mock)

### DIFF
--- a/app/src/main/res/layout/add_to_homescreen.xml
+++ b/app/src/main/res/layout/add_to_homescreen.xml
@@ -12,7 +12,10 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:orientation="horizontal"
-        android:padding="16dp">
+        android:paddingStart="24dp"
+        android:paddingEnd="24dp"
+        android:paddingTop="16dp"
+        android:paddingBottom="10dp">
 
         <ImageView
             android:id="@+id/homescreen_icon"
@@ -27,8 +30,7 @@
             android:layout_gravity="center_vertical"
             android:backgroundTint="@color/add_to_homescreen_button_text"
             android:inputType="text"
-            android:paddingEnd="8dp"
-            android:paddingStart="8dp"
+            android:paddingStart="10dp"
             android:textAppearance="?android:attr/textAppearanceMedium"
             android:textColor="@color/add_to_homescreen_text"
             android:textColorHighlight="@color/colorAccent"
@@ -57,7 +59,9 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:textAllCaps="true"
-            android:textColor="@color/download_dialog_text" />
+            android:textColor="@color/download_dialog_text"
+            android:textSize="14sp"
+            android:fontFamily="@string/font_roboto_medium" />
 
         <Button
             android:id="@+id/addtohomescreen_dialog_add"
@@ -66,7 +70,9 @@
             android:layout_height="wrap_content"
             android:layout_gravity="end"
             android:textAllCaps="true"
-            android:textColor="@color/download_dialog_text" />
+            android:textColor="@color/download_dialog_text"
+            android:textSize="14sp"
+            android:fontFamily="@string/font_roboto_medium" />
     </LinearLayout>
 
 
@@ -95,8 +101,9 @@
             android:background="#ff262626"
             android:paddingStart="10dp"
             android:text="@string/dialog_addtohomescreen_tracking_protection"
-            android:textColor="#ff8b8b8b"
-            android:textSize="12sp" />
+            android:textColor="#66ffffff"
+            android:textSize="12sp"
+            android:fontFamily="@string/font_roboto_regular" />
     </LinearLayout>
 
 </LinearLayout>
--- a/app/src/main/res/layout/add_to_homescreen.xml
+++ b/app/src/main/res/layout/add_to_homescreen.xml
@@ -12,8 +12,8 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:orientation="horizontal"
-        android:paddingStart="24dp"
-        android:paddingEnd="24dp"
+        android:paddingStart="@dimen/dialogHorizontalPadding"
+        android:paddingEnd="@dimen/dialogHorizontalPadding"
         android:paddingTop="16dp"
         android:paddingBottom="10dp">
 
@@ -28,7 +28,7 @@
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:layout_gravity="center_vertical"
-            android:backgroundTint="@color/add_to_homescreen_button_text"
+            android:backgroundTint="@color/dialogAccent"
             android:inputType="text"
             android:paddingStart="10dp"
             android:textAppearance="?android:attr/textAppearanceMedium"
@@ -59,7 +59,7 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:textAllCaps="true"
-            android:textColor="@color/download_dialog_text"
+            android:textColor="@color/dialogAccent"
             android:textSize="14sp"
             android:fontFamily="@string/font_roboto_medium" />
 
@@ -70,7 +70,7 @@
             android:layout_height="wrap_content"
             android:layout_gravity="end"
             android:textAllCaps="true"
-            android:textColor="@color/download_dialog_text"
+            android:textColor="@color/dialogAccent"
             android:textSize="14sp"
             android:fontFamily="@string/font_roboto_medium" />
     </LinearLayout>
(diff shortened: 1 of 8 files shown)


app/src/main/res/layout/add_to_homescreen.xml
```diff
@@ -57,7 +59,9 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:textAllCaps="true"
-            android:textColor="@color/download_dialog_text" />
+            android:textColor="@color/download_dialog_text"
```
mcomella: nit: you don't need to fix this because it's an existing issue and it's problematic all over the code but I thought I'd point it out because you seem interested in learning. :)

imo, when a color is given a name (i.e. used in resources), it should be named for the place it's used. If it's used in multiple places, it should be named after all of them, or given a generic name (e.g. "dialogText", or "greyText").

In this specific case, this isn't the Download dialog so I think this name needs to change.

The reason for this is that if you have a well-named color, you can change it and have a good idea of what it will be changing (e.g. if I change a color called `dialogText`, most dialogs should correct themselves). In this case, it's misleading: if I change download_dialog_text, the add to homescreen dialog text color will change!

That being said, I think the comprehensive way to fix this is a larger issue: having a color palette, set by UX, and referencing that in all places, creating new colors for special cases.

This guideline also applies to other resources: if you use a dimension in one place, it should be named for the place. In multiple places, it should be named for all of them or generically (see my comment about creating a shared dimension above :).

mcomella:
Another weird thing that I'm not sure if we can fix: 

screenshot_1508788554

The correction underline on the keyboard is not the same color as the text. Could you look into this too, @layely ?

layely:
> The correction underline on the keyboard is not the same color as the text. Could you look into this too, @layely ?
 
I think it would be complicated to fix from my first investigation. But I will try to do more later and tell you if I find a way as it concerns also other inputs.

app/src/main/res/layout/add_to_homescreen.xml
```diff
@@ -12,7 +12,10 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:orientation="horizontal"
-        android:padding="16dp">
+        android:paddingStart="24dp"
```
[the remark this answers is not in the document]
layely: Yeah I like the idea especially for next when the need to create other dialogs comes.

app/src/main/res/layout/add_to_homescreen.xml
layely: Thank you for pointing this issue, I noticed it and thought there was a good reason that it was made this way. 

Now when it comes to should it be named for this specific case or for all dialog, I think it should be global (all dialog). what do you think about the name "dialog_input_text" ?

>That being said, I think the comprehensive way to fix this is a larger issue: having a color palette, set by UX, and referencing that in all places, creating new colors for special cases.

Absolutely agree with you here. It would be way more simple to respect the specifications for UI colors then. Maybe you will need to make them aware of the need of this.

mcomella:
>> The correction underline on the keyboard...
> 
> I think it would be complicated to fix from my first investigation.

If something is non-essential (like this), you can always file a follow-up issue and you (or someone else!) can tackle it later: there's no obligation for you to finish it.

app/src/main/res/layout/add_to_homescreen.xml
mcomella: > I noticed it and thought there was a good reason that it was made this way.

Sometimes there is, sometimes there isn't - sometimes someone just has a different opinion (I guess I haven't talked to anyone else about this yet!). In any case, if you have questions, feel free to ask! You can do so in issue or find us on IRC (we're generally around M-F GMT working hours through PST working hours).

> what do you think about the name "dialog_input_text"

I think "input" isn't quite right: input is usually a term used when the user enters information. Maybe "dialog_text" or "dialog_body_text", depending on which one is more relevant?

> Maybe you will need to make them aware of the need of this.

I'll bring it up. fwiw, [we did this with Firefox for Android](http://searchfox.org/mozilla-central/source/mobile/android/app/src/photon/res/values/colors.xml#125) but at some point the convention started being ignored so I don't know how practical keeping the colors consistent actually is.

layely: I think that "dialog_text" or "dialog_body_text" can also be misleading as this ressource is only used for : 

- buttons text (eg: CANCEL) 
- editText background tint

What I am proposing is to add shared ressources like "dialog_button_text" and "dialog_edit_background_tint"

### Note: 
This ressourse is also used for download snackbar text (notify the user when the download is finished) in DownloadBroadcastReceiver line 88

`snackbar.setActionTextColor(ContextCompat.getColor(context, R.color.download_dialog_text));`

which has no effect as the text color of the snackbar is still **white**. (like erase snackbar btw)

screenshot_1509076290


As a ressource have already been added for snackbar ("snackbarTextColor"), I propose to use it here.

What do you think about it ?

mcomella: > I think that "dialog_text" or "dialog_body_text" can also be misleading as this ressource is only used for :

Oh, it wasn't used for what I thought it was used for – thanks for catching that!

> What I am proposing is to add shared ressources like "dialog_button_text" and "dialog_edit_background_tint"

> As a ressource has already been added for snackbar ("snackbarTextColor"), I propose to use it here.

My understanding of your proposal is that you want to add a separate color resource for each use case:
- dialog_button_text
- dialog_edit_background_tint
- snackbarTextColor

If so, I think:
- I'd try to merge `dialog_button_text` and `dialog_edit_background_tint` into one resource: the colors are linked in their use so we should link it in the code (i.e. its unlikely we'd want to make them two separate colors in the same dialog). Perhaps `dialog_accent`? Like you alluded to, `dialog_tint` could be used too. Android uses `colorAccent` defined in an app's theme for similar purposes such as text field colors, preference title colors, switch colors, etc.
- I like your proposal to have a separate resource for the Snackbar. To be explicit, `snackbarTextColor` is used to set the text color `download_dialog_text` will set the action text color (such as "OPEN" in a download dialog). I'd use a name like `snackbarActionText` for this second resource.
    - Note that including "color" in the name of a color resource is redundant because you need to access color resources with `R.color.` or with `Resources.getColor`. Since it's already in the code, it looks like someone already overlooked this

What do you think? Thanks for looking into this!

layely: >(i.e. its unlikely we'd want to make them two separate colors in the same dialog). Perhaps dialog_accent?

`dialog_accent` is just fine I think 👍.

>To be explicit, snackbarTextColor is used to set the text color download_dialog_text will set the action text color (such as "OPEN" in a download dialog). I'd use a name like snackbarActionText for this second resource. 

Yes, I got it now, I was clearly confused here. Thank You
Let's use the name `snackbarActionText`.

>Note that including "color" in the name of a color resource is redundant because you need to access color resources with R.color. or with Resources.getColor. Since it's already in the code, it looks like someone already overlooked this

Well I took note, but what do you suggest here now. Maybe It would be interesting to create an issue proposing to rename it (that would involve several other colors btw). On the other hand it's not a priority I think but it could be a good issue for new contributors right ??? 


About resources you suggested above (`dialogAccent` and  `snackbarActionText`), I will add to this PR another commit very soon.

layely:
@mcomella  another commit added.

app/src/main/res/layout/add_to_homescreen.xml
mcomella: > > Note that including "color" in the name of a color resource is redundant because you need to access color resources with R.color. or with Resources.getColor. Since it's already in the code, it looks like someone already overlooked this
> 
> Maybe It would be interesting to create an issue proposing to rename those redundant resources (that would involve several other colors btw). On the other hand it's not a priority I think but it could be a good issue for new contributors right ???

A related issue we have is that some colors are camelCase and some are underlined_between_words.

One concern I have with changing the naming is that we will make the `git blame` history much harder to follow. That being said, it's a really good idea to make this a contributor issue because we need more of those! :)

@pocmo What do you think? Shall we create a good-first-issue to 1) remove "Color" from our color names (I find it redundant, see my comment above) and another (probably after the first is complete) to rename them so they're consistently camelCase or underlined?

layely:
>Have you had a chance to make sure these match the mocks? It'd really help us out!

Before I commit I check off course if the changes match the mock. 

> For the colors, even if the UI doesn't change, we should probably use a single resource for each of them to ensure consistency, e.g. so blue links could be a resource called "blue_links" with the color: #0A84FF

good idea, let's add it next with the overflow menu

>Imaybe we could work through #1529.

I will dive there for sure.

Thank You for your reviews, awesome

mcomella:
>> Have you had a chance to make sure these match the mocks? It'd really help us out!
> 
> Before I commit I check off course if the changes match the mock. 

Oh, sorry – I didn't mean to imply that! It'd be really impressive if you weren't trying to match the mocks and you did anyway. :P

Basically, I was wondering if we could close #1260: that the work for the overflow menu and making sure the colors throughout the app were consistent was done. It looks like you're about to go work on that next though, so you answered my question!

> Thank You for your reviews, awesome

And thanks for your help! :)

mcomella:
Adding this to the milestone because we don't have an issue for this small amount of work.

PR #1521, opened 2017-10-20
Release builds automation

pocmo:
After landing this taskcluster will automatically build release builds of Focus/Klar when a [release is created on GitHub](https://github.com/mozilla-mobile/focus-android/releases). With that all the manual building and taking care of the Adjust token etc. will not be needed anymore.

For now those builds will be unsigned. That's one of the next steps. However now we only need to create a release on GitHub, open a Bugzilla bug for signing & uploading and drop the taskcluster link with the builds attached into the Bugzilla bug.

### DIFF
--- a/tools/l10n/filter-release-translations.py
+++ b/tools/l10n/filter-release-translations.py
@@ -1,0 +1,49 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""
+This script takes the list of release locales defined in locales.py
+and removes all non-release translations from the application's
+resource folder.
+This script is run before building a release version so that
+those builds only contain locales we actually want to ship.
+"""
+
+import os
+import re
+import shutil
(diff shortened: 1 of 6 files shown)


pocmo:
@mastizada Added you as a second reviewer for the Python code again! :)

tools/l10n/filter-release-translations.py
```diff
@@ -0,0 +1,57 @@
...
+import shutil
+
+from locales import RELEASE_LOCALES
+
+LANGUAGE_REGEX = re.compile('^[a-z]{2,3}$')
+LANGUAGE_REGION_REGEX = re.compile('^([a-z]{2})-r([A-Z]{2})$')
+
+# Get all resource directories that start with "values-"
+resources_dir = os.path.join(os.path.dirname(__file__), '..', '..', 'app', 'src', 'main', 'res')
+locale_dirs = [localeDir for localeDir in os.listdir(resources_dir)
```
mastizada: You can append `localeDir.replace('values-', '')` instead of `localeDir` to avoid using line `locale_dirs = map(lambda x: x[7:], locale_dirs)`

PR #1524, opened 2017-10-20
Increase apk size limit to 4.5MB

npark-mozilla:


tools/metrics/apk_size.py
```diff
@@ -7,7 +7,7 @@
 from os import path, listdir, stat
 from sys import exit
 
-SIZE_LIMIT = 4194304
+SIZE_LIMIT = 4500000
```
pocmo: That's not exactly 4.5 MB (1024 * 1024 * 4.5 = 4718592). However I'll merge it anyways to unblock our builds and then we can update it once we made a decision.

PR #1526, opened 2017-10-21
fix bug : URL bar and status bar flickering when scrolling horizontal…

layely:
… images (#1179)

@mcomella this is a fix for bug #1179

### DIFF
--- a/app/src/main/java/org/mozilla/focus/session/SessionCallbackProxy.java
+++ b/app/src/main/java/org/mozilla/focus/session/SessionCallbackProxy.java
@@ -37,7 +37,10 @@ public class SessionCallbackProxy implements IWebView.Callback {
 
     @Override
     public void onPageFinished(boolean isSecure) {
-        session.setLoading(false);
+        if (session.getLoading().getValue()) {
+            //We should only set "session.setLoading(false);" if we are loading
+            session.setLoading(false);
+        }
         session.setSecure(isSecure);
     }
 
--- a/app/src/test/java/org/mozilla/focus/session/SessionCallbackProxyTest.java
+++ b/app/src/test/java/org/mozilla/focus/session/SessionCallbackProxyTest.java
@@ -12,6 +12,9 @@ import org.robolectric.RobolectricTestRunner;
 
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
 
 @RunWith(RobolectricTestRunner.class)
@@ -33,12 +36,20 @@ public class SessionCallbackProxyTest {
 
     @Test
     public void testOnPageFinished() {
-        final Session session = mock(Session.class);
+        final Session session = spy(new Session(Source.NONE, TEST_URL));
         final SessionCallbackProxy proxy = new SessionCallbackProxy(session, mock(IWebView.Callback.class));
 
+        session.setLoading(true);
         proxy.onPageFinished(true);
 
         verify(session).setLoading(false);
+        verify(session).setSecure(true);
+
+        session.setLoading(false);
+        reset(session);
+        proxy.onPageFinished(true);
+
+        verify(session, never()).setLoading(false);
         verify(session).setSecure(true);
     }
 


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -224,10 +224,15 @@ public void onValueChanged(@NonNull Boolean loading) {
                     progressView.setProgress(5);
                     progressView.setVisibility(View.VISIBLE);
                 } else {
-                    backgroundTransitionGroup.startTransition(ANIMATION_DURATION);
 
-                    progressView.setProgress(progressView.getMax());
-                    progressView.setVisibility(View.GONE);
+                    if(progressView.getVisibility() == View.VISIBLE) {
```
sputnikci: \[Checkstyle\] ERROR: 'if' is not followed by whitespace\.

layely: @sputnikci it's fixed thank you

mcomella:
@pocmo Please take a look at this – I have my hands full atm.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -224,10 +224,15 @@ public void onValueChanged(@NonNull Boolean loading) {
                     progressView.setProgress(5);
                     progressView.setVisibility(View.VISIBLE);
                 } else {
-                    backgroundTransitionGroup.startTransition(ANIMATION_DURATION);
 
-                    progressView.setProgress(progressView.getMax());
-                    progressView.setVisibility(View.GONE);
+                    if (progressView.getVisibility() == View.VISIBLE) {
```
pocmo: I think we could go one level higher and apply this patch to SessionCallbackProxy:
https://github.com/mozilla-mobile/focus-android/blob/d94251c4d7c262f26bb8ba22e5a218cff2e7354f/app/src/main/java/org/mozilla/focus/session/SessionCallbackProxy.java#L40

As it looks like onPageFinished() gets called multiple times by the WebView here. We should only set `session.setLoading(false);` if we are loading (`session.getLoading().getValue()`). We should still always do the setSecure() call in the next line though (in case it changes).

layely: Well thank you, I will come back with a commit very soon.

layely:
@pocmo changes are made.

app/src/main/java/org/mozilla/focus/session/SessionCallbackProxy.java
```diff
@@ -37,7 +37,10 @@ public void onPageStarted(String url) {
 
     @Override
     public void onPageFinished(boolean isSecure) {
-        session.setLoading(false);
+        if (session.getLoading().getValue() == true) {
```
pocmo: small nit: There's no need to write "== true" here as the result of the method call is already a boolean:
> if (session.getLoading().getValue()) {

Apart from that this is great and seems to work fine. Thank you! :)

layely: Oops...

pocmo:
The change looks good now. But apparently this broke on of our unit tests (SessionCallbackProxyTest.html#testOnPageFinished):
https://public-artifacts.taskcluster.net/Se3CPQpHQb2oRmsKvRhHsg/0/public/tests/testFocusGeckoBetaUnitTest/classes/org.mozilla.focus.session.SessionCallbackProxyTest.html#testOnPageFinished

Can you try to modify the test to cover the new behavior? :)

pocmo:
@layely Still interested in pushing this over the finish line? :)

layely:
Well @pocmo I modified the test now, tell me if it's ok ?

app/src/test/java/org/mozilla/focus/session/SessionCallbackProxyTest.java
```diff
@@ -6,13 +6,17 @@
 
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mozilla.focus.architecture.NonNullLiveData;
```
sputnikci: \[Checkstyle\] ERROR: Unused import \- org\.mozilla\.focus\.architecture\.NonNullLiveData\.

sputnikci: \[Checkstyle\] ERROR: Unused import \- org\.mozilla\.focus\.architecture\.NonNullLiveData\.

app/src/test/java/org/mozilla/focus/session/SessionCallbackProxyTest.java
```diff
@@ -6,13 +6,17 @@
...
 import org.mozilla.focus.architecture.NonNullMutableLiveData;
 import org.mozilla.focus.web.IWebView;
 import org.robolectric.RobolectricTestRunner;
 
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
```
sputnikci: \[Checkstyle\] ERROR: Unused import \- org\.mockito\.Mockito\.never\.

app/src/test/java/org/mozilla/focus/session/SessionCallbackProxyTest.java
```diff
@@ -36,10 +38,15 @@ public void testOnPageFinished() {
         final Session session = mock(Session.class);
         final SessionCallbackProxy proxy = new SessionCallbackProxy(session, mock(IWebView.Callback.class));
 
+        when(session.getLoading()).thenReturn(new NonNullMutableLiveData<Boolean>(Boolean.TRUE));
         proxy.onPageFinished(true);
 
-        verify(session).setLoading(false);
-        verify(session).setSecure(true);
```
pocmo: You call something on the proxy here but you do not verify after that. Later you do verify that the APIs have been called a specific amount of times - but this doesn't guarantee that it happened at the right time.

Instead of mocking the Session object and setting all the responses manually it might make sense to use an actual Session object here:
> final Session session = spy(new Session(Source.NONE, TEST_URL));

Then verify two things:
* Set loading to true and call onPageFinished -> Verify that setLoading and setSecure are called correctly
* Set loading to false and call onPageFinished -> Verify that setLoading is not called, but setSecure.

layely: @pocmo test modified

app/src/test/java/org/mozilla/focus/session/SessionCallbackProxyTest.java
```diff
@@ -33,13 +36,21 @@ public void testOnPageStarted() {
...
         final SessionCallbackProxy proxy = new SessionCallbackProxy(session, mock(IWebView.Callback.class));
 
+        session.setLoading(true);
         proxy.onPageFinished(true);
 
         verify(session).setLoading(false);
         verify(session).setSecure(true);
+
+        session.setLoading(false);
+        reset(session);
```
pocmo: Ah, interesting - Haven't seen/used "reset()" before. :)

PR #1530, opened 2017-10-23
Update "Architecture components" dependencies

pocmo:


pocmo:
> I can't say I fully understand the annotation processing changes, but seems reasonable.

Me neither. But according to the Kotlin docs this is required so that the annotation processors work with Kotlin code. And they mentioned that kapt will take care of the Java code too, so it completely replaces the "annotationProcessor" line in the build files. So far we do not use Architecture Components from Kotlin code. So this is mostly to get gradle to shut up and to be prepared for the future.

Issue #898, opened 2017-07-11
TimeoutException: AssetManager$AssetInputStream.close


pocmo:
```
java.util.concurrent.TimeoutException: 
  at android.content.res.AssetManager$AssetInputStream.close(AssetManager.java:577)
  at android.content.res.AssetManager$AssetInputStream.finalize(AssetManager.java:610)
  at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:191)
  at java.lang.Daemons$FinalizerDaemon.run(Daemons.java:174)
  at java.lang.Thread.run(Thread.java:818)
```

Reports: 916
Impacted users: 363

pocmo:
Not sure what to do here. The exception seems to happen outside of our code - so that we can't just ignore it. We are reading assets for two reasons: Domain lists for autocomplete and search plugins.

pocmo:
There's nothing helpful when searching for this problem.

We saw this exception a couple of times in Fennec too:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1212204
* https://bugzilla.mozilla.org/show_bug.cgi?id=1298298

pocmo:
Removing 1.1 label -> Right now I don't see any way how we can fix this on our side.

pocmo:
This is currently our top crash (which isn't as dramatic - we are at 99.6% crash-free daily sessions).

pocmo:
AssetInputStream code:
https://github.com/aosp-mirror/platform_frameworks_base/blob/master/core/java/android/content/res/AssetManager.java#L597

We open assets from two locations:
* UrlAutoCompleteFilter: To load the list of domains.
* SearchEngineManager: To load search engine plugins

For those we could think about wrapping AssetInputStream, overriding finalize() and suppressing the TimeoutException. But I'm not sure what state the device is in when this exception is thrown and we might keep an app alive that will crash with a different error later (and then things get really confusing to debug).

In addition to the things mentioned above there's also WebView loading some internal pages like licenses.html or gpl.html from the assets folder and it might be the case that the problem lies here (and only in specific WebView implementations).

bbinto:
@pocmo mentioned we have yet to find a solution for it and we need a bit more investigation.

PR #1533, opened 2017-10-23
Add test that verifies that IOUtils.readAsset() closes the stream. (#898)

pocmo:
Don't have a fix for bug #898. But as I wrote this test to verify something I thought I could just land it.


pocmo:
Now I felt like quickly converting IOUtils to Kotlin while I'm in that file already. Will update the PR. :)

app/src/test/java/org/mozilla/focus/ext/AssetManagerTest.kt
```diff
@@ -0,0 +1,42 @@
...
+
+        val stream = Mockito.spy(ByteArrayInputStream("Hello".toByteArray()))
+
+        val assetManager = Mockito.mock(AssetManager::class.java)
+        Mockito.`when`(assetManager.open(ArgumentMatchers.anyString())).thenReturn(stream)
+
+        val context = Mockito.mock(Context::class.java)
+        Mockito.`when`(context.assets).thenReturn(assetManager)
+
+        // Now use our mock classes to call IOUtils.close()
```
mcomella: nit: IOUtils -> AssetManager (fwiw, I feel like most of the comments here are redundant with the code)

pocmo: Ops, too fast..

pocmo:
.. and I need to fix this test. Pretty sure this passed before. Weird. :)

pocmo:
Fixed the test and added another one that verifies that the JSON is transformed into a JSONObject correctly.

pocmo:
Now findbugs is complaining:
> Self assignment of it in org.mozilla.focus.ext.AssetManagerKt.readJSONObject(AssetManager, String)
https://public-artifacts.taskcluster.net/O5kTi4O4Spi_lOKUgEum1g/0/public/findbugs/findbugs.html

This is a bit unfortunate as the code findbugs is complaining about is the result of compiling the Kotlin code.

pocmo:
Seems to be a known problem in Kotlin 1.1:
https://discuss.kotlinlang.org/t/should-we-be-reporting-findbugs-issues/2105

According to the thread above this is fixed in Kotlin 1.2. Kotlin 1.2 Beta2 was released recently - so 1.2 might not be that far away.

pocmo:
For now I added the @SuppressFBWarnings annotation with an explanation of the problem.

PR #1537, opened 2017-10-23
Import localized strings for Granite.

liuche:
(Should this even be a PR, or just a push to master? I guess it's good to run checks and sanity check that nothing got broken.)


liuche:
All green but rebased, so just going to merge.

Issue #1472, opened 2017-10-12
Investigate whether removing WebKit files on erase is a good idea


mcomella:
After a browser session, WebKit leaves around some files. I filed Chromium bug ...insert bug here.... We remove the files on erase after #433 (PR #476) as a preventative measure but it can create problems, like #1306:

> deleting the "Local Storage" files appears to block the [ed. WebView] from deleting the localStorage data exposed to the web page (until the process is killed), which is why I suggest that we should, in the simplest case, leave the files when hitting erase but delete them at some later point [ed. when the process is dead]

This makes me question whether or not it is a good idea to be removing the files WebKit expects to be there because it may create these side-effects.

Notes:
- WebKit will only create its files the first time the WebView is used in this process: if you delete files, the WebView will not recreate them until the process is restarted. This implies it doesn't check for their existence and may perform unexpected behavior when they're missing
- If we don't delete the files, WebKit may leak history data to the filesystem. For example, in "Local Storage", the files are named after the domain they come from such as "support.google.com".

# Guideline
As a guidline from https://github.com/mozilla-mobile/focus-android/issues/1306#issuecomment-336227143:

> We say we do not store any history etc. - So I think our requirements are:
> - Make sure that we do not accumulate traces of browsing history on disk. Temporary caches are okay.
> - Do not leak any data to other apps.
>
> I think as long as we stay in those bounds we can be flexible.

# Approaches
I think some approaches are:
- Wait for Chromium to fix the issue
- Selectively fix issues as they come up (e.g. #1306: don't delete "Local Storage" files on erase)
- Find a comprehensive solution:
  - where we don't delete files on erase but clear them after the process ends
  - where we let WebKit let us delete the files. This is likely hacky. We might be able to do this by killing the Focus/Chrome (if it's separate) process, using ClassLoaders to unload the WebView classes, keeping a copy of its files in memory only and copying them to disk when we delete, etc.

In https://github.com/mozilla-mobile/focus-android/issues/1306#issuecomment-336226016 we began the discussion about some approaches we could take to not delete files on erase but clear them later:
> > - Deleting "Local Storage" contents when the process starts (in Application.onCreate; I confirmed this works)
> 
> That's what I was wondering too and I think we could try that.

> > - Deleting the "Local Storage" contents after the process ends (since there's no callback, we might have to schedule a repeating job in onStop)
> 
[...]

PR #1555, opened 2017-10-24
Closes #1306: ensure "Local Storage" is deleted by the time a new session starts

### LINKED ISSUE #1306: Google "search history" survives erasing browsing history
Please can you add the below to Firefox Focus.

1. Ability to remove search history. For example, my previous searches appears in the Google search bar when visiting www.google.com.

2. Menu -> Open with Firefox. Can this option only appear to those who have Firefox installed.

Thanks

mcomella:
#1306.

We do two things here:
1. Don't delete the `Local Storage` directory so deleting when the erase button works
1. Work-around certain sites, like Google Search, who will restore LocalStorage files asynchronously after we delete them by deleting the local storage when starting a new session

The code is a little hacky: to avoid scope creep, I opted for a smaller change and I think we should revise our erase strategy as part of #1472.


pocmo:
https://public-artifacts.taskcluster.net/S_gwYP5cR2C0eVCnS06xZQ/0/public/findbugs/findbugs.html

mcomella:
> https://public-artifacts.taskcluster.net/S_gwYP5cR2C0eVCnS06xZQ/0/public/findbugs/findbugs.html

The error is `BC_BAD_CAST_TO_ABSTRACT_COLLECTION`

>  Questionable cast from Collection to abstract class java.util.List  in org.mozilla.focus.utils.FileUtils$Companion.deleteContent(File, Set)

Looking at the decompiled bytecode from Kotlin:

```java
            Collection element$iv = (Collection)(new ArrayList());

            for(int it = 0; it < $receiver$iv$iv.length; ++it) {
               Object $i$a$1$all = $receiver$iv$iv[it];
               File $i$f$all = (File)$i$a$1$all;
               if(!doNotEraseWhitelist.contains($i$f$all.getName())) {
                  element$iv.add($i$a$1$all);
               }
            }

            List filesToDelete = (List)element$iv;
```

The corresponding kotlin code is:
```kotlin
val filesToDelete = directory.listFiles()?.filter { !doNotEraseWhitelist.contains(it.name) } ?: return false
return filesToDelete.all { it.deleteRecursively() }
```

So we're casting to `Collection` to filter but casting back to `List` to store the temporary variable.

It looks like others have seen this problem: here's [an issue filed for findbugs](https://youtrack.jetbrains.com/issue/KT-18311) and [the issue it was duped to](https://youtrack.jetbrains.com/issue/KT-16446). Neither appears to be verified by the core maintainers or prioritized (the original issue is filed in Feb 2017).

I think the best course of action here is to suppress the error.

PR #1557, opened 2017-10-25
Closes #1556. Update search engine string.

### LINKED ISSUE #1556: Export strings for 2.4
String should be "+ Add another search engine" - original string is missing "+".

liuche:
Missing "+" in original string.


liuche:
I'm seeing some problems here because this PR is fixing a string, so it's present in the individual localization directories, but has been removed from the main strings.xml. @pocmo what do you think, should I also remove the strings from the localization directories (or write a script for it), or just wait for the new string import next time (where I assume the localizers will have removed the old strings)?

pocmo:
> I'm seeing some problems here because this PR is fixing a string, so it's present in the individual localization directories, but has been removed from the main strings.xml. @pocmo what do you think, should I also remove the strings from the localization directories (or write a script for it), or just wait for the new string import next time (where I assume the localizers will have removed the old strings)?

When removing a string I run an export and immediately import again. This will remove the string from all translated files. I then create a separate commit for that (it might contain additionally newly translated strings).

pocmo:
I just saw that this failed again and then I remembered a problem. Sorry!

Our import script resets the state of the l10n repository here:
https://github.com/mozilla-mobile/focus-android/blob/master/tools/l10n/import-strings.sh#L18

This means the po files you just exported are get reverted again. You could either comment this out or run the steps for import manually:
```bash
python tools/l10n/android2po/a2po.py import
tools/l10n/fix_locale_folders.sh
```

Maybe we should make this indeed easier with a little script or parameter for the existing script.

PR #1559, opened 2017-10-25
Taskcluster: Print executed commands.

pocmo:
Using "-x" will print the commands that taskcluster executes. This will make it simpler to read the logs on taskcluster.


liuche:
oh no, I should not have clicked that button. Well now I know that it makes a merge commit :P

pocmo:
> oh no, I should not have clicked that button. Well now I know that it makes a merge commit :P

That's no problem - as soon as you click "Rebase and merge" the rebase will make the merge commit disappear! :)

Issue #1569, opened 2017-10-27
[meta] Rethink/refactor our erase strategy


mcomella:
Quick breakdown, but we should do more in #1822.
- [x] #1822: breakdown
- [ ] #2012 Document high-level goals for erasing
- [ ] #1472 Is erasing WebView files a good idea?
- [x] #2013 High-level understanding: summarize all the ways the browsing session can end (e.g. user erases, system OOMs, battery falls out)
- [x] #2014 Historical context on WebView behavior: summarize the existing ways we erase (why there?) and our history of hack fixes (e.g. #1875, #1306).
- [ ] #2015 Identify and implement the fix (can we make our two erase entry points to one entry point?)
- [ ] #2016 Adjust (and add to?) our tests to incorporate our changes
- [ ] #2017 Document the erase strategy caveats in the high-level goals document

---

Splitting from #1472 to avoid scope creep:

> Notes:
> - We have three kinds of erases: erases with a WebView instance, erases with static WebView methods, and erasing the files on disk. I'd suggest merging the last two.
> - We call each of these three at different times for what seem like a lot of redundant calls, e.g. when sessions becomes empty (delete files on disk), onPause and finishing (delete files on disk), when the erase button is actually hit (all three)
> - The methods are confusingly named, especially with the extra layers of abstraction like WebViewProvider
> 
> We should try to centralize the erase so it only happens in one place and perhaps when starting a completely fresh browsing session (since that's the workaround for #1306).
>
> Another note (for another bug?): @pocmo mentions we erase these files whenever a tab is removed (via back) and this seems extreme.

mcomella:
Sebastian wrote an analysis of our goals in https://github.com/mozilla-mobile/focus-android/issues/1306#issuecomment-336226016. Here's the short version:

> We say we do not store any history etc. - So I think our requirements are:
> - Make sure that we do not accumulate traces of browsing history on disk. Temporary caches are okay.
> - Do not leak any data to other apps.
>
> I think as long as we stay in those bounds we can be flexible.

I think we should document this guideline on the wiki so we can always look back to it when making decisions (such as the safety of this PR https://github.com/mozilla-mobile/focus-android/pull/1570#pullrequestreview-72872647 ).

liuche:
Decided to discuss this offline with engineers to:
* Explain user benefit
* Discuss breakdown: regression tests, better approaches

mcomella:
[...]

PR #1570, opened 2017-10-27
Closes #1492: Don't delete cache files on disk; update test.

### LINKED ISSUE #1492: WebView stops using disk cache on erase (was: Report of unexpectedly high Focus network data usage)
We received a user review where the user felt Focus was using an unreasonable amount of data. Let's investigate!

2 thoughts:
1. We erase the cache every time the erase button is pressed – I wonder how this would affect browser data usage. Should we make this configurable? Most users probably aren't concerned about the cache sticking around (and it probably sticks around for private browsing in other browsers).
2. WebKit won't recreate the folders we delete (note: may not affect temp files) until the process is restarted (cause of #1306 and further investigation required in #1472) so I wonder if this includes the cache. If WebKit can't save cache files, it may cache fewer files since it can only do so in memory and we may use more data to re-retrieve these files.

mcomella:
#1492

After we manually delete the cache files, WebView will no longer use the disk
cache, which means Focus uses much more network data. Therefore, we no longer
delete the cache.

While we could have erased all the contents of cache/ except the WebView
directory, I opted not to: different versions of WebView could have different
file hierarchies and so this bug would still happen if that's the case. Instead,
we no longer delete any of cache/. As far as I know, WebView is currently the
only user of cache/ so it should have the same effect.

I tested the updated test on an Android API 23 & 26 emulator.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/FileUtils.kt
+++ b/app/src/main/java/org/mozilla/focus/utils/FileUtils.kt
@@ -10,16 +10,20 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
 
 import java.io.File
 
+// WebView directory and contents.
 private const val WEBVIEW_DIRECTORY = "app_webview"
 private const val LOCAL_STORAGE_DIR = "Local Storage"
+
+// Cache directory contents.
+private const val WEBVIEW_CACHE_DIR = "org.chromium.android_webview"
 
 class FileUtils {
     companion object {
         @JvmStatic
-        fun truncateCacheDirectory(context: Context): Boolean {
-            val cacheDirectory = context.cacheDir
-            return cacheDirectory.exists() && deleteContent(cacheDirectory)
-        }
+        fun truncateCacheDirectory(context: Context) = deleteContent(context.cacheDir, doNotEraseWhitelist = setOf(
+                // If the folder or its contents are deleted, WebView will stop using the disk cache entirely.
+                WEBVIEW_CACHE_DIR
+        ))
 
         @JvmStatic
         fun deleteWebViewDirectory(context: Context): Boolean {
--- a/app/src/androidTest/java/org/mozilla/focus/activity/WebViewDataTest.java
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/WebViewDataTest.java
@@ -67,7 +67,7 @@ public class WebViewDataTest {
     private static final Set<String> WHITELIST_DATA_DIR_CONTENTS;
     static {
         final Set<String> whitelistDataDirContents = new HashSet<>(Arrays.asList(
-                "cache", // We assert that this folder is empty
+                "cache",
                 "code_cache",
                 "shared_prefs",
                 "app_dxmaker_cache",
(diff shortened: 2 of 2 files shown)


mcomella:
> Are we sure that we are not starting to accumulate browsing fragments in the cache directory now?

By "browsing fragments", do you mean files that the WebView saves?

In any case, no – I'm not sure if third-party code (or even our code) could be adding to `cache/`, especially for different versions of WebView. I was optimizing for making sure this bug is fixed on all devices:

> While we could have erased all the contents of cache/ except the WebView
directory, I opted not to: different versions of WebView could have different
file hierarchies and so this bug would still happen if that's the case. Instead,
we no longer delete any of cache/.

But I think you make a good point that this could make our users more vulnerable either:
1. WebView or other code reuses files we leave around in `cache/`, which exposes previous browsing history to:
    1. sites via web browsing
    1. Other users of focus on the same device who may see the browsing history reflected in page state
1. Users reading files left over on disk forever.

We previously agreed that we should avoid leaving files on disk (except for temporary caches) and this PR may fail to do that: I'll update it such that it will delete the maximum number of files and still fix the issue. If users still see issues, they can file a new bug and we'll fix it on a case-by-case basis.

mcomella:
via https://github.com/mozilla-mobile/focus-android/issues/1492#issuecomment-340613621, I should also verify service workers aren't storing files and look at WebViewDataTest for details.

I feel like we need some documentation to mention our erase goals – I added this to #1569 – and link to our current strategies so it'd be great if we could centralize the code to clean-up the browser.

pocmo:
> I feel like we need some documentation to mention our erase goals

This is a good idea. I wonder if such a documentation exists for Private Browsing on Firefox. We could use it as a baseline and document where we (have to) differ.

mcomella:
@pocmo Please re-review (and ignore top-most commit).

app/src/main/java/org/mozilla/focus/utils/FileUtils.kt
```diff
@@ -10,16 +10,20 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
...
     companion object {
         @JvmStatic
-        fun truncateCacheDirectory(context: Context): Boolean {
-            val cacheDirectory = context.cacheDir
-            return cacheDirectory.exists() && deleteContent(cacheDirectory)
-        }
+        fun truncateCacheDirectory(context: Context) = deleteContent(context.cacheDir, doNotEraseWhitelist = setOf(
+                // If the folder or its contents are deleted, WebView will stop using the disk cache entirely.
+                WEBVIEW_CACHE_DIR
+        ))
```
pocmo: If we do not remove the webview cache files then I do not see any need to truncate the cache directory at all. The only reason why we did that is to get rid of the webview files. And just to be sure we remove everything in there to avoid guessing the right folder.
So I guess if we want to keep the files then let's remove this completely.

mcomella: My initial PR did this but, after my interpretation of your comments, I decided we should be more aggressive about deleting another other files that make their way in there: https://github.com/mozilla-mobile/focus-android/pull/1570#issuecomment-340611638

@pocmo Are you sure you want to remove the cache deletion entirely?

mcomella:
> - We call clearCache() on the WebView object (in SystemWebView). Does this remove everything in the cache directory?

No.

> - If things are not completely deleted by WebView itself ... what remains?

Before deletion, inside `<data-dir>/cache/org.chromium.android_webview`, we see:
```sh
index
index-dir/
    the-real-index
002ee73b1c794adf_0 // <- many, many files named like this
```

After deletion, we see:
```sh
index
index-dir/
    the-real-index
```

Where `index` and `the-real-index` are binary files, the real index is 48 B and index is 20 B – i.e. nothing large can be stored inside. When the cache is not deleted, `the-real-index` grows (e.g. 1.4 KB in my case) while `index` remains the same size.

My understanding is that Chromium is keeping around some minimal index files after deletion, which point at nothing when the cache is deleted.

mcomella:
Since they don't contain user identifiable data (as best as I can tell based on their small size), I thought it'd be fine to leave around.

PR #1599, opened 2017-10-28
README.md: Fix link to "good first issue" label

alanorth:
Looking at the labels of closed issues, it appears that the project is actually using "good first issue" as the label for issues that would be good for new contributors. The README suggests that good issues for would-be new contributors are labeled "good first bug".


pocmo:
Thank you! We recently renamed the label to follow the naming scheme of GitHub. :)

PR #1619, opened 2017-11-01
Added Interlingua (ia) to the list of missing locales

### LINKED ISSUE #1601: (Interlingua) [failed] ia is not a valid locale
Just imported strings for #1572 and noticed that Android2Po complains about "ai".

Maybe some custom handling is needed, see #1148.

mastizada:
fix #1601

### DIFF
--- a/tools/l10n/android2po/env.py
+++ b/tools/l10n/android2po/env.py
@@ -56,6 +56,12 @@ to Babel locale mapping.
 """
 
 MISSING_LOCALES = {
+    'ia': {
+        'name': "Interlingua",
+        'local_name': "Interlingua",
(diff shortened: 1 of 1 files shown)


mastizada:
@pocmo Any updates on removing values (translations) from repository? I'm deleting new folders manually everytime when testing a2po

pocmo:
> @pocmo Any updates on removing values (translations) from repository (adding them to gitignore)? I'm deleting new folders manually everytime when testing a2po

For now we decided to keep them in the repository. Our release process for translations is changing currently - so we might be able to import more often in the future (without requiring translations to be at 100%).

Issue #1499, opened 2017-10-18
[UX] Invisible scrollbar


jenn-chaulk:
Design for issue #1447

pocmo:
@lime124 Here's a screenshot of the new scrollbar. The PR sets the scrollbar to be 4dp in size and to use color #aaaaaaaa (ARGB). The change is noticeable on white only really - where previously the scrollbar was invisible.

![nytimes](https://user-images.githubusercontent.com/89638/32556420-a9392348-c49f-11e7-9b42-b1c5cb55b180.png)

pocmo:
@lime124 Did you make a decision here? Should we land the contributor patch?

lime124:
Sorry for the delay! Can we try #b2b2b2 with 70% opacity? Thanks!

pocmo:
@Benestar Do you have time to update your PR #1621 with those values? :)

Benestar:
Sorry for my late response, quite busy irl right now. I will update the color code and create another screenshot.

lime124:
Awesome - looks great! Thanks so much!

Benestar:
Thanks for your quick feedback! :-)

pocmo:
@lime124 Merged the patch from @Benestar and closed the engineering issue #1447. Should we close this one here too or do you want to revisit this?

lime124:
oops good call :)

PR #1621, opened 2017-11-01
Make vertical scrollbar actually visible

### LINKED ISSUE #1447: Invisible scrollbar
UX Bug - #1499 

Firefox Focus just doesn't seem to display a vertical scrollbar for any web pages, no matter how long they are. This leads me to lose context of where I am in the page, how long it is, and how much more there is to read or see.

If I look closely, or if there's a coloured banner, I can just about see that there is in fact a white scrollbar on a white background. It auto-hides when not scrolling, but that's not the issue here. It's the fact that it's a white scrollbar on a white background that makes it invisible and therefor not useful.

To see what I mean you can visit any page, even [this list of Github issues](https://github.com/mozilla-mobile/focus-android/issues). I see no scrollbar, and have no context for where I am in the page.

It's much more of an issue with articles on Medium etc. With no proportionally-sized scrollbar, I have no idea how long the article is when I scroll down a bit. If I want to find out how long the article is, I have to repeatedly scroll down until I see the end, then scroll back up to where I was beforehand - something made more difficult by (again) not having a scrollbar to know my previous location.

Is this something peculiar to my setup, or is it generally true for all Focus users? I find it to be a real usability problem. Should it not show the scrollbar in a colour that contrasts sufficiently with the background?

Details:
Firefox Focus
Version: 2.1
Build: #12721912
Android Oreo
Pixel C tablet

Benestar:
Per default, the scrollbar is very light and almost
completely transparent. This patch changes the style
of the scrollbar to a less transparent and much darker
color so that it actually gets visible.

Fixes #1447

### DIFF
--- a/app/src/main/res/drawable/scrollbar_thumb.xml
+++ b/app/src/main/res/drawable/scrollbar_thumb.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<shape xmlns:android="http://schemas.android.com/apk/res/android">
+    <solid android:color="@color/scrollbarThumb" />
+    <size android:width="4dp" />
+</shape>
--- a/app/src/main/res/layout/fragment_browser.xml
+++ b/app/src/main/res/layout/fragment_browser.xml
@@ -37,7 +37,8 @@
                 <org.mozilla.focus.web.IWebView
                     android:id="@+id/webview"
                     android:layout_width="match_parent"
-                    android:layout_height="match_parent" />
+                    android:layout_height="match_parent"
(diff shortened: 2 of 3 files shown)


pocmo:
This is tracked by  #1447. We have no guidelines from the UX team here yet.

Benestar:
Hey @pocmo thanks for the fast response! What is the best channel to discuss this with the UX team? Is there someone in particular responsible for the Firefox Focus UX?

pocmo:
You could try pinging @lime124 in issue #1499. It's currently ranked as P5 - so it doesn't have a high priority. If you share some screenshots (maybe before/after) in the issue then it might be easier to make a quick decision like "Yep, that's better, let's ship that for now".

lime124:
Hi @Benestar ! I am the designer you seek :) If you can do as @pocmo advised and attach some screenshots I'll be happy to provide feedback. Thanks for helping us make Focus better!

Benestar:
Needs rebase

Benestar:
Rebased and color code adjusted, looking forward for another review :-)

Issue #1418, opened 2017-09-29
Swipe down to refresh web page (Pull-To-Refresh)


codebu5ter:
A Swipe-down-to-refresh gesture would be a great addition to Focus, just like in Google Chrome. 
It would make the act of refreshing a web page more user-intuitive.

Benestar:
Hi @lime124, I used the `photonMagenta70` color from `photon_colors.xml` which is the same color as the floating action button to erase browsing history. Its color code is `#ffb5007f`.

lime124:
Awesome! Thanks again @Benestar !

jolcsk:
I hate to disturb a development thread with my question, but can you make this feature optional (able to turn-off)? I absolutely hate the swipe-down-to-refresh feature (I trip it accidentally too often), that is one of the reasons I like using firefox (fennec) and focus too.

max2505:
Hey all,

I would also like the option to turn off this feature.
On a website page where a chat conversation is shown and by default you're at the bottom (end of the conversation), you don't have the options to go up (by scrollingen down) when this feature is enabled (I hope the explaination is somewhat clear).

Tester on Oneplus 3T.

SavageBiscuits:
I'd also like to chime in for an option to disable this feature. Already I've managed to trigger this feature a few times without wishing to. It gets a bit annoying. It's a useful feature to have, but it's not for everyone.

Benestar:
Hi @max2505 @SavageBiscuits I created #1776 to address the issues with pull to refresh. Can you maybe provide the links to the pages where this feature causes issues?

PR #1622, opened 2017-11-01
Fix scrolling in NestedWebView

Benestar:
Calling super.onTouchEvent in all cases fixes more
complicated gestures like zooming on Google Maps.

The other change in the ACTION_MOVE case is required
for the SwipeRefreshLayout to work (see follow up pr).


Benestar:
BTW it took me ages to find this fix and I spent lots of time trying to understand what the relevant bits of the NestedScrollView class from the support library are. In the end it just looks so simple... 😩

pocmo:
#1418

pocmo:
> BTW it took me ages to find this fix and I spent lots of time trying to understand what the relevant bits of the NestedScrollView class from the support library are. In the end it just looks so simple... 😩

Thank you so much for investing all this time. :)

Issue #1529, opened 2017-10-23
Custom Tabs: No option to open URL in Focus


pocmo:
Usually in Custom Tabs you can switch to the full version of the browser. Some time ago we removed Focus from the "Open with" option in the menu; because we thought that while inside Focus no one wants to switch to Focus (#894).

However this is a valid use case while you are in a Custom Tab. Third-party apps control what options are available in the UI and only by switching to the full browser you are able to perform all the available actions.

brampitoyo:
This patch looks good to me, and comprises a part of what I’m proposing with the custom tabs redesign (#1626).

Let’s make sure that the wording says “Open in Firefox Focus”.

pocmo:
Talked to Bram about this an other custom tabs things on Vidyo.

We decided that we want to have an additional menu item in custom tabs that opens Focus. This means we do not need to add Focus to the "Open with..." list.

> Let’s make sure that the wording says “Open in Firefox Focus”.

@brampitoyo For other apps we show a menu item with the string "Open with XXX". Do we want to re-use this one or explicitly say "in" here?

brampitoyo:
I use “Open in…” because it’s the string that other browsers’ Custom Tabs use. But if Focus already uses “Open with…”, and if this string aligns with Fennec, we should use “with”.

pocmo:
Follow-up issues: #2188, #2187

npark-mozilla:
#2335 opened

Issue #1334, opened 2017-09-18
[2.0 RC2] "search privately" then pressing the back button kills both tabs and exits the app


phunkular:
I would expect it to send me to the tab I just came from instead of killing the whole app

pocmo:
1. Going back from an externally opened tab (text selection, share, links, ..) always leaves Focus and ideally gets you back to the app (if you haven't performed a bunch of other navigations in between). This is the intended behavior currently.

2. This should only close the current tab and all others should stay intact. Removing all tabs would be a bug.

I tried to reproduce (2) but couldn't. Here's what I tried:
* Open Focus and open a couple of tabs
* Press home and open another app
* Select some text and press "Search privately"
* In Focus press back.
* -> The tab is closed and I'm back in the previous app. When going back to Focus the other tabs are still open.

@phunkular Can you provide me with the exact steps how to reproduce (2)? Also: What Android device and Android version are you using?

phunkular:
hi there, to reproduce (happens every time for me):

1. open focus (klar in my case)
2. Search for anything (one tab, not several)
3. highlight any text
4. select search privately
5. on newly opened tab press back
6. Focus closes with a toast saying private data cleared

miui 9 android 6.0.1

phunkular:
same issue on 2.0 final

MrGeneric:
@phunkular Not sure whether it's the intended behaviour or not but I was able to reproduce using your steps.
Nexus 5x Focus 2.0

phunkular:
I believe it's unexpected. Perhaps we're expected to search privately from external apps and not so much from within Firefox focus?.. In any case I have some more to add:

When several tabs are open and I search privately and then press back, I am sent to homescreen with a toast saying that my data has been cleared. However if I return to focus, my other tabs are still there. So to recap; with just one tab open everything disappears and I'm kicked out to homescreen, and; with several tabs I am kicked out to home screen though the other tabs persist.

liuche:
Depends on #1569

mcomella:
> Depends on #1569 

It could be fixed by this bug too

lobontiumira:
Hi all!
The behavior on the latest RC 92.1.1 is this:
- when the user chooses to search privately from Fenix browser, a new tab is opened in Focus - when tapping back, the tab closes, and the user remains in the Focus app,
[...]

PR #1634, opened 2017-11-02
Custom Tabs: Add option to open URL in Focus (#1529)

layely:
fix for #1529

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -812,6 +812,11 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
                 break;
             }
 
+            case R.id.open_in_firefox_focus: {
+                sessionManager.replaceSessionWithNoCustomTab(session);
+                break;
+            }
+
             case R.id.share: {
                 final String url = getUrl();
                 final Intent shareIntent = new Intent(Intent.ACTION_SEND);
--- a/app/src/main/java/org/mozilla/focus/session/SessionManager.java
+++ b/app/src/main/java/org/mozilla/focus/session/SessionManager.java
@@ -260,4 +260,17 @@ public class SessionManager {
 
         this.sessions.setValue(sessions);
     }
+
+    // Replaces a (custom-tab) session by its no custom-tab version
+    public void replaceSessionWithNoCustomTab(Session session) {
+        if (!session.isCustomTab()) {
+            return;
+        }
+
+        Session noCustomTabSession = new Session(Source.MENU, session.getUrl().getValue());
+        noCustomTabSession.setBlockingEnabled(session.isBlockingEnabled());
+
+        this.addSession(noCustomTabSession);
+        this.removeSession(session.getUUID());
+    }
 }
--- a/app/src/main/java/org/mozilla/focus/menu/browser/BrowserMenuAdapter.java
+++ b/app/src/main/java/org/mozilla/focus/menu/browser/BrowserMenuAdapter.java
@@ -64,6 +64,10 @@ public class BrowserMenuAdapter extends RecyclerView.Adapter<BrowserMenuViewHold
 
         this.items = new ArrayList<>();
 
+        if (customTabConfig != null) {
+            items.add(new MenuItem(R.id.open_in_firefox_focus, resources.getString(R.string.menu_open_in_firefox_focus)));
+        }
+
         if (customTabConfig == null || customTabConfig.showShareMenuItem) {
             items.add(new MenuItem(R.id.share, resources.getString(R.string.menu_share)));
         }
--- a/app/src/main/res/values/ids.xml
+++ b/app/src/main/res/values/ids.xml
@@ -13,4 +13,5 @@
     <item name="add_to_homescreen" type="id" />
     <item name="help" type="id" />
     <item name="whats_new" type="id" />
+    <item name="open_in_firefox_focus" type="id" />
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -35,6 +35,10 @@
       a screen with multiple browsers (or other apps that can open the current URL) to choose from. -->
     <string name="menu_open_with_a_browser">Open with…</string>
 
+    <!-- Menu item: Switch from custom tab to normal tab. Thus this menu item is only available
+        on custom tabs -->
+    <string name="menu_open_in_firefox_focus">Open in Firefox Focus</string>
+
     <!-- Menu item: Add the current page as a shortcut to the device's home screen (launcher). -->
     <string name="menu_add_to_home_screen">Add to Home screen</string>
 


app/src/main/java/org/mozilla/focus/utils/Browsers.java
```diff
@@ -111,15 +112,15 @@ private ActivityInfo findFirefoxBrandedBrowser() {
...
+            if (!context.getPackageName().equals(info.activityInfo.packageName) || isCustomTab) {
                 browsers.put(info.activityInfo.packageName, info.activityInfo);
             }
         }
 
         return browsers;
     }
 
-    private void findKnownBrowsers(PackageManager packageManager, Map<String, ActivityInfo> browsers, @NonNull Uri uri) {
+    private void findKnownBrowsers(Context context, PackageManager packageManager, Map<String, ActivityInfo> browsers, @NonNull Uri uri, boolean isCustomTab) {
```
sputnikci: \[PMD\] INFO: The method 'findKnownBrowsers' has a Cyclomatic Complexity of 9\.

app/src/main/java/org/mozilla/focus/telemetry/TelemetrySettingsProvider.java
```diff
@@ -50,7 +50,7 @@ public boolean containsKey(String key) {
             // The default browser is not actually a setting. We determine if we are the
             // default and then inject this into telemetry.
             final Context context = TelemetryHolder.get().getConfiguration().getContext();
-            final Browsers browsers = new Browsers(context, Browsers.TRADITIONAL_BROWSER_URL);
+            final Browsers browsers = new Browsers(context, Browsers.TRADITIONAL_BROWSER_URL, false /*doesn't matter here*/);
```
layely: I arbitrary set the `isCustomTab` to `false`, but whether it's `false` or `true` `Browsers.isDefaultBrowser` return the same value.

app/src/main/java/org/mozilla/focus/widget/DefaultBrowserPreference.java
```diff
@@ -55,7 +55,7 @@ protected void onBindView(View view) {
 
     public void update() {
         if (switchView != null) {
-            final Browsers browsers = new Browsers(getContext(), Browsers.TRADITIONAL_BROWSER_URL);
+            final Browsers browsers = new Browsers(getContext(), Browsers.TRADITIONAL_BROWSER_URL, false /*doesn't matter here*/);
```
layely: >I arbitrary set the isCustomTab to false, but whether it's false or true Browsers.isDefaultBrowser return the same value.

layely:
Would you like to review this @mcomella ?

mcomella:
Sure, I'll get to this on Monday (unless @liuche or @pocmo want to review sooner!).

mcomella:
Sorry, I overlooked this today – I added it to my list of things to work on tomorrow so I won't miss it!

app/src/main/java/org/mozilla/focus/utils/Browsers.java
```diff
@@ -237,12 +240,31 @@ public boolean isDefaultBrowser(Context context) {
 
     }
 
-    public ActivityInfo[] getInstalledBrowsers() {
-        final Collection<ActivityInfo> collection = browsers.values();
+    public ActivityInfo[] getInstalledBrowsers(boolean isCustomTab) {
```
mcomella: I think you chose the optimal implementation strategy, given the existing code, but I just want to make a note: `Browsers` probably shouldn't be an object that needs to be constructed and these should all be static methods. The current implementation creates some strangeness:
-  Some values are cached and some aren't
- Holding a long-lived reference to the object can make the cached values out of date.
- We have to [explicitly remove Focus from the initial cached browsers value](https://github.com/mozilla-mobile/focus-android/commit/c2664c9b12ad9e18907ada8c7af84620943b57fb) and then re-add it here in this method.

I assume the reason we do it this way is to allow mocking but I'd rather use PowerMockito to mock static methods than mush the code up like this. :)

mcomella:
By the way, thanks for being patient on for my review! :)

app/src/main/java/org/mozilla/focus/utils/Browsers.java
pocmo: Note that the data in this object is based on the URL. The set of default apps/browsers can be different depending on what URL you want to open. That's the major reason why some of this data is not cached statically.

mcomella: To be explicit, I was suggesting we should never cache this data (because we do in a Browsers instance) and use static functions instead.

layely:
> 1. If we select "Open With..." Focus, Focus will duplicate the tab: when we open Focus, we maintain the tab from the custom tab and duplicate it in a new tab (since we open links sent in the "Open with..." intent).
> 2.  After opening a custom tab and selecting "Open with..." Focus, clicking back will claim, "Browsing history has been erased" but the session still exists (only the current tab is closed). This is #1334
> 3. After ^, going to the recent app switcher and selecting Focus will reopen Focus in custom tab mode

Thank you for pointing these issues @mcomella , I think we should fix it but note here that 2 & 3 are the current behaviour of the app (it's note introduced by my commit).

>what does Chrome do?

- for 1, chrome will close custom tab before opening the URL in it.
- for 2, it's like chrome custom tab is part of its root app (I mean the app that launched the custom tab eg: reddit) and is unatached from full chrome.
eg: when you open chrome custom tab from reddit, then use app switcher and go to chrome you will not find there the custom tab, but going back to reddit still opens the chrome custom tab.
Compared to focus, the custom tab is atached not attached to it's root (ed: reddit) but to full focus.

What I suggest here is fixing 1 here (as it's directly related to my commit) and open issues for 2 and  3 (I could always work on them after this anyway).

Cc @pocmo : I think this might interrest you also.

app/src/main/java/org/mozilla/focus/utils/Browsers.java
pocmo: Generating this data is (or can be) super slow though. :)

mcomella:
Hey @layely – I'm going on vacation until 11/27 and don't have enough time to respond to your comment today (my last day before vacation). Sorry about that!

@pocmo Do you mind looking after this bug, and layely's future patches, while I'm gone? :) The most recent comment that I haven't had a chance to respond to [is here.](https://github.com/mozilla-mobile/focus-android/pull/1634#issuecomment-343038678)

layely:
Enjoy your vacations @mcomella 👍

mcomella:
Hey @layely – I'm back from vacation. I'm sorry that no one has gotten back to you yet. I'll be sure to look into this today or tomorrow.

layely:
Thanks @mcomella for thinking of me your first day back, hope you enjoyed your vaccation btw. In the mean while I restarted with school so I may not respond or submit work quickly, but I am still here.

mcomella:
> In the mean while I restarted with school so I may not respond or submit work quickly, but I am still here.

No worries – studies are important and we'll do our best to give you time and accommodate for them! Thanks for letting us know!

mcomella:
@layely Thanks for your analysis and looking into what Chrome does. Your description of what Chrome does for 2) makes me wonder whether or not we're providing the right UX for custom tabs, particularly within the app switcher. I'd like to have UX take a look at this and decide what's appropriate to ship – could you upload a build for @lime124 to look at?

@lime124 Once the build is uploaded, can you look at how the "Open with..." Focus behavior works? In particular, a few problems layely and I found can be seen in https://github.com/mozilla-mobile/focus-android/pull/1634#pullrequestreview-74970950 and https://github.com/mozilla-mobile/focus-android/pull/1634#issuecomment-343038678.

In the meantime, I have two thoughts on how we can continue in this bug:
1. We could put this feature behind a feature flag (like [`FLAG_MANUAL_SEARCH_ENGINE`](https://github.com/mozilla-mobile/focus-android/blob/5009ea3944c3c5103cf65ff1cb701cfb35d17d8c/app/src/main/java/org/mozilla/focus/utils/AppConstants.java#L39)) and land it preffed off (add a comment with the issue # for additional work!). We can file issues for 1) and 3), and work on them after (depending on Tif's response).
1. We could just go ahead and fix 1) because it's very likely we'll want to fix that before shipping. If Tif says it's good as is, we could file a follow-up for 3) and ship it. If there's more to do, we could do a feature flag and file other bugs, or fix them here.

Since you mentioned you may be pressed for time, I'd recommend 1. fwiw, I also find working on a large set of patches makes it harder to make additional changes due to:
- Bitrot
- Each change doesn't feel like a small incremental change, but a change to a larger, more complex change
- You don't get the same satisfaction of actually landing something. :)

So I'd probably also prefer the feature flag if I was working on this.

---

> Note: during the tests I once got focus crashed (didn't have the chance to reproduce it) with the error

Unfortunately, this looks like a system error unrelated to the actual crash: https://stackoverflow.com/a/8177859/2219998

layely:
> could you upload a build for @lime124 to look at?

Bellow a build for you @lime124, would you like to check it ?

https://drive.google.com/open?id=1lBOjWJRQ5qUEfBRFlmSAMy2UELgiMA09

>So I'd probably also prefer the feature flag if I was working on this.

Well, Let's do it this way. Thank you for your suggestion

lime124:
Thanks @layely. I'm going to ask my colleague, @brampitoyo to give it a look over. He's been working a lot with custom tabs and itsn't as buried in our 4.0 release. Thanks bram for helping out!

brampitoyo:
Hi @layely. It turned out that I had reviewed your issue as soon as I heard from @lime124 about it yesterday, but posted on the wrong page (#1529). Apologies for that.

The only two things we have to make sure:

1. The wording of the menu item should be “Open in Firefox Focus”
2. The menu item should sit directly underneath the “Trackers blocked” information

That’s it. Two fairly simple things. Thank you!

layely:
Thanks @brampitoyo for your review, I will take it into consideration. However a issue with the app switcher were pointed out by @mcomella. 
Please see [comment#1634](https://github.com/mozilla-mobile/focus-android/pull/1634#pullrequestreview-74970950) and [comment#1634](https://github.com/mozilla-mobile/focus-android/pull/1634#issuecomment-343038678) and [comment#1634](#issuecomment-347379295) for more information about problems @mcomella and I found.

pocmo:
@layely Thank you for your work so far. :)

After talking to @brampitoyo about the UX we decided that we want to show an addition "Open with/in Firefox Focus" menu item (see #1529). This means we do not need to modify Browsers.java and the "Open with..." sheet anymore. Would you like to continue working on this and modify your patch? :)
The only thing I'm not 100% sure about is the exact string (see comment in #1529).

mcomella:
@brampitoyo The change in UX Sebastian mentioned will still run into the same issues layely mentioned in https://github.com/mozilla-mobile/focus-android/pull/1634#issuecomment-348203547 – please still address them! Thanks!

layely:
>Would you like to continue working on this and modify your patch? :)

@pocmo I will submit a patch as soon as possible.

@mcomella maybe creating an issue pointing to problems that we found in others comments would be more convenient as they seem to be ignored here.

pocmo:
> @brampitoyo The change in UX Sebastian mentioned will still run into the same issues layely mentioned in #1634 (comment) – please still address them! Thanks!

If we have a separate menu item and not need to use the generic "Open with.." then we can customize the behavior easily and work around those issues. :)

Some random ideas:
* We could clear the custom tab config and re-add a BrowsingFragment for the current session. With that we do not duplicate the tab and are back in the normal browser view.
* Or: Remove the session and send a VIEW intent to ourselves.

brampitoyo:
Hi @layely, I finally had a look at the .apk you’ve provided.

Let’s address the problems @mcomella wrote.

> 1. If we select "Open With..." Focus, Focus will duplicate the tab: when we open Focus, we maintain the tab from the custom tab and duplicate it in a new tab (since we open links sent in the "Open with..." intent).
> 2. After opening a custom tab and selecting "Open with..." Focus, clicking back will claim, "Browsing history has been erased" but the session still exists (only the current tab is closed). This is #1334

Link duplication should happen, but if Custom Tabs and Focus App Tabs are still considered one entity in the app launcher, duplication will feel odd. I think this is what caused these two problems, right?

To fix it, we should separate Custom Tabs from Focus App Tabs (as @layely had documented, this is how Chrome Custom Tabs works). By separating, Focus Custom Tabs will be considered a part of the root third-party app that opened the link, not a part Focus App. In the app launcher, the two will appear as two different entitites.

When properly separated, then link duplication between Custom Tabs and Focus App Tabs isn’t an issue, and should actually happen, and problem 3 will be fixed:

> 3. After ^, going to the recent app switcher and selecting Focus will reopen Focus in custom tab mode

What do you think?

layely:
Thank you @brampitoyo for trying out the apk and you clearely understand the issue now.

>To fix it, we should separate Custom Tabs from Focus App Tabs (as @layely had documented, this is how Chrome Custom Tabs works). By separating, Focus Custom Tabs will be considered a part of the root third-party app that opened the link, not a part Focus App. In the app launcher, the two will appear as two different entitites.

+1 That's exactly the way it should be I think.

Let's flag @mcomella to see what he think.

mcomella:
This seems desirable to me – go for it @layely! To be clear, it sounds like the action items are:
- (https://github.com/mozilla-mobile/focus-android/pull/1634#issuecomment-348497853) Update this patch to show an addition "Open with/in Firefox Focus" menu item (see #1529). This means we do not need to modify Browsers.java and the "Open with..." sheet anymore.
- (https://github.com/mozilla-mobile/focus-android/pull/1634#issuecomment-350922061) Open custom tabs as a separate entity – this could be a follow-up

By the way, sorry for the delay in my response – we had a pretty crazy last week!

layely:
Thank you @mcomella. I think I will have some free time next week to submit a patch for the 1) at least or during the weekend.

layely:
>(#1634 (comment)) Update this patch to show an addition "Open with/in Firefox Focus" menu item (see #1529). This means we do not need to modify Browsers.java and the "Open with..." sheet anymore.

Hi @mcomella I have sent an update for this.

>Some random ideas:
>- We could clear the custom tab config and re-add a BrowsingFragment for the current session. With that we do not duplicate the tab and are back in the normal browser view.
>-  Or: Remove the session and send a VIEW intent to ourselves.

I have taken the first approach suggested by @pocmo in this update.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -812,6 +812,17 @@ public void onClick(View view) {
                 break;
             }
 
+            case R.id.open_in_firefox_focus: {
+                session.clearCustomTabConfig();
```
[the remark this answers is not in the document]
layely: I agree with you here @mcomella. In fact, I haven't thought about the side effects it could generate. I will send the changes soon.

layely:
Hey @mcomella, I did some changes. If it's ok then I would ping @liuche to help with the telemetry.

app/src/main/java/org/mozilla/focus/session/SessionManager.java
```diff
@@ -260,4 +260,17 @@ public void removeCurrentSession() {
...
+    public void replaceSessionWithNoCustomTab(Session session) {
+        if (!session.isCustomTab()) {
+            return;
+        }
+
+        Session noCustomTabSession = new Session(Source.MENU, session.getUrl().getValue());
+        noCustomTabSession.setBlockingEnabled(session.isBlockingEnabled());
+
+        this.addSession(noCustomTabSession);
+        this.removeSession(session.getUUID());
```
layely: Here I am forced to add the new session before removing the custom-tab session otherwise we have this exception:

image

Sdaswani:
@layely thanks for your contribution. I want one @pocmo to do the final review on this but he is on paternity leave, so please stay tuned (he should be back in about 5 weeks).

layely:
Thank you @Sdaswani. Let me flag @mcomella here again as he has already started to review this.

mcomella:
Hey @layely – sorry, I've gotten really busy (which is why pocmo would be taking over) but that being said, I previously r+'d this so I'll try to look by the end of the week!

app/src/main/java/org/mozilla/focus/session/SessionManager.java
```diff
@@ -260,4 +260,17 @@ public void removeCurrentSession() {
...
+    // Replaces a (custom-tab) session by its no custom-tab version
+    public void replaceSessionWithNoCustomTab(Session session) {
+        if (!session.isCustomTab()) {
+            return;
+        }
+
+        Session noCustomTabSession = new Session(Source.MENU, session.getUrl().getValue());
+        noCustomTabSession.setBlockingEnabled(session.isBlockingEnabled());
+
+        this.addSession(noCustomTabSession);
```
mcomella: As per your comment about crashing, add a comment mentioning this crash and why the ordering is necessary.

app/src/main/java/org/mozilla/focus/session/SessionManager.java
mcomella: I'm not sure why this would be throwing `IllegalStateException` – I'd expect `NullPointerException`. I also can't find where such an IllegalStateException would be thrown from.

I'm guessing that maybe we don't handle the case where all sessions are removed in a custom tab session properly.

It's best if we can figure out why this is happening but if you've already looked for a little bit and can't find anything, I think reversing the logic order seems sound and I'm okay landing it without knowing exactly why this is happening.

mcomella:
@brampitoyo This is what it looks like after Layely's patch:

device-2018-02-14-142844

Where should "Open in Firefox Focus" be in the list?

layely:
Thank you @mcomella and sorry for the late reply, I will certainly get to this the next week because I've gotten really busy too with school.

pocmo:
Thank you for your patches @layely! I landed them with some changes on top in #2180. I was able to re-use the existing Session by removing the CustomTabConfig from it and then re-creating a BrowserFragment for it.

PR #1646, opened 2017-11-03
Add taskcluster scripts to import strings and create pull requests.

pocmo:
Those scripts can run on taskcluster and will import strings, create a
commit, push them to the bot's repository and create a new pull request
to be merged by a human.

Example merge request:
https://github.com/mozilla-mobile/focus-android/pull/1643

This is not running automatically yet. If this is proving to be useful
I want to run it regularly via taskcluster's hooks service [1].

Later this commit could be merged automatically. But for now I think
it's a good idea to let a human review it first.

[1] https://tools.taskcluster.net/hooks/

### DIFF
--- a/tools/taskcluster/import_strings_and_create_pull_request.sh
+++ b/tools/taskcluster/import_strings_and_create_pull_request.sh
@@ -1,0 +1,32 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This script imports the latest strings, creates a commit, pushes
+# it to the bot's repository and creates a pull request.
+
+# If a command fails then do not proceed and fail this script too.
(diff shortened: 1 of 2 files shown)


tools/taskcluster/import_strings_and_create_pull_request.sh
```diff
@@ -0,0 +1,32 @@
...
+# If a command fails then do not proceed and fail this script too.
+set -e
+
+# Go to project root
+cd "$(dirname "$0")"
+cd ../..
+
+# Update local checkout
+git fetch origin
+git reset --mixed origin/master
```
liuche: Is this --mixed because there is other development going on in the local user master?

pocmo: I used "--hard" first and somehow couldn't figure out why everything behaved really weirdly. Then I noticed that this will also reset the state of the scripts I'm currently writing and testing. To prevent this I switched to "--mixed" which will not reset the local changes :D

PR #1648, opened 2017-11-03
Update Search Engine Icon Size closes #1618

### LINKED ISSUE #1618: Update padding, size, and color of the Search Engine Selection screen
Make it match the mocks :)

visual design

...and the coding style.

NOTE: this is just for the padding, sizing, color polish (the parts with colorful labels). The 3-dot menu and the "Add search engine" are handled by other issues.

A few comments:
* If a dimension seems like it will be reused in different places, add it to dimens.xml and reference it - that way we don't have one-off numbers everywhere in the code
* Colors should almost always be added to colors.xml - please name it generically. In this case, it can be something like settings_selection_highlight.
* When you've finished, please add a screenshot with the changes so UX can visually verify.

Please ask if you have any other questions!

ekager:
Needed to force the icon to be 24dp


liuche:
Lint errors should be fixed in intermediate PR.

Issue #1613, opened 2017-10-31
Screen does not go to sleep


pocmo:
From whuang via Slack:

> it actually brings a regression that the browser avoids the screen from going to sleep even not in media playback.
> 
> STR:
> 1. config your settings: let the screen sleep after 15 sec of inactivity
> 2. use Focus to play a youtube video and make it fullscreen
> 3. before the video ends, quickly switch to other app (such as Facebook) to open a link (website not a media), and open it with Focus
> Then, Focus avoids the screen to go to sleep even after 15 secs. The “track can” icon is gone as well, even if you navigates to other pages.

Could be causing battery problems: #1216

npark-mozilla:
This fix is verified, but there is another potential issue.  When I open a media, and make it fullscreen, the screen does not sleep even though it's stopped/paused.  (This does not happen when the media is not in fullscreen mode) Is this expected? (similar battery drain issue might occur if one just pauses in fullscreen mode and leave it) @pocmo what do you think?

Just double checked that in Chrome paused fullscreen mode enters sleep mode normally.

PR #1652, opened 2017-11-06
BrowserFragment: Leave immersive mode if still active when fragment gets destroyed. (#1613)

pocmo:


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -535,6 +535,22 @@ private void switchToImmersiveMode() {
                         | View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY);
     }
 
+    @Override
+    public void onDestroy() {
```
mcomella: nit: For correctness, I think we should set a boolean, `isImmersive`, when entering/exiting immersive mode and deactivate/activate as necessary in `onStart`/`onStop`.

I'm concerned that `onDestroy` is not guaranteed to be called. Sensically, the app isn't being closed so the system should clean up the resources when transitioning between fragments but since it's not contractually obligated to call `onDestroy`, I'm concerned it won't, especially on obscure devices with modified Android.

That being said, if you disagree and stick with onDestroy:
- I think you should add a comment (maybe above `exitImmersive...`) about why we don't care that `onDestroy` isn't guaranteed to be called
- Consider using `onDetach` - wouldn't it be more correct? According [to the lifecycle](https://cdn-images-1.medium.com/max/1600/0*YWCr1M1pwWfvCzvZ.png), there are times when `onDetach` will be called but `onDestroy` won't be. It's also documented [in the `setRetainInstance` javadoc](https://developer.android.com/reference/android/app/Fragment.html#setRetainInstance(boolean)).

mcomella: > onStart/onStop would mean we leave immersive mode if the app is not visible anymore. However right now it's perfectly fine to leave the app and then get back to your full screen video.

True. I guess the flag we really want to clear onStart/onStop is the SCREEN_ON one, which breaks the abstraction we have around immersiveMode but there's probably a way to clean it up (e.g. an object we delegate onStart/onStop to).

> This is not a fragment that retains the instance (You have to manually call setRetainInstance yourself for that).

I was just future proofing. :)

pocmo: Android will take care of that. No need to manually remove the flag when switching apps etc.: https://developer.android.com/training/scheduling/wakelock.html

Issue #1523, opened 2017-10-20
Increase 4MB apk size limit


npark-mozilla:
Currently build is failing on master because it's over 4MB limit

pocmo:
Background: For our first release we decided to not exceed a limit of 4 MB. Back then we added this check that will fail the build once we exceed it. With a small app we wanted to explore whether this improves our install rate (especially in countries where mobile traffic is expensive or limited).

We deployed a quick fix to raise this limit to 4.5 MB - But we need to make a decision whether we still care about this limit (or pick a new one). With multitasking and other features we managed to squeeze a lot of things into 4 MB and it might be impossible to keep this forever. Some previous ideas and investigations are documented in #624.

bbinto:
I think there is nothing wrong with keeping a cap on the size of the app. We should not deviate more than +1MB from the original APK size limit.

So can we up it to 5MB?

pocmo:
> So can we up it to 5MB?

Alright!

mcomella:
Closed in 3c06788eef9ddfa8cc628c1d7c628073ec7fd50b.

PR #1653, opened 2017-11-06
Set APK size limit to 5 MB. (#1523)

pocmo:


mcomella:
Oh, can we remove the extra merge commit?

pocmo:
> Oh, can we remove the extra merge commit?

Yep, this one will be gone when rebasing automatically.

Issue #1662, opened 2017-11-07
python: specify max APK size in MiB


mcomella:
Right now, we use [apk_size.py](https://github.com/mozilla-mobile/focus-android/blob/master/tools/metrics/apk_size.py) to ensure we don't exceed a particular APK size. We specify the maximum size of the APK in bytes but that's error prone. We should:
- Specify the `SIZE_LIMIT` in MiB
- Convert it to bytes to compare to [the file size returned by `st_size`](https://github.com/mozilla-mobile/focus-android/blob/5490ca45f036145980cf6617c875c0223b43b65d/tools/metrics/apk_size.py#L24)

Preferably, if Python has an API that is aware of bytes/MiB/etc., we should use that (unfortunately, my quick searches show that there isn't one built-in and afaik, we don't support packages atm).

pocmo:
> We specify the maximum size of the APK in bytes but that's error prone

I can't follow why this is error prone?

mcomella:
> I can't follow why this is error prone?

We want to specify the value in MiB/MB but the program requires the value as bytes so we need to do a conversion each time the value is updated. There's room for error in the conversion, especially with potential confusion over MiB vs. MB (e.g. how "5mb to bytes" in DDG says "5000000" and we probably want "5MiB to bytes").

JordanShaak:
I'm still shaky on the process of submitting my stuff for review when trying to contribute to a project but I made a pull request since I think that's what you're supposed to do? I don't really get this process very well since I'm still spending more time wrestling with Git than doing anything else.

mcomella:
> I don't really get this process very well since I'm still spending more time wrestling with Git than doing anything else.

I've commented in the PR: #1672.

mcomella:
Closed by 90b0803.

PR #1663, opened 2017-11-07
Issue#1662 ?

JordanShaak:
I might be misunderstanding something but it looks like this would achieve the functionality of having SIZE_LIMIT be defined in MiB instead of bytes as per #1662. By upload since I'm having a difficult time trying to mange using git.


JordanShaak:
I didn't want that other part to be part of this pull, how do I remove it?

Issue #1617, opened 2017-10-31
Add Manual Search Engine page


liuche:
Meta: #961

Add preference screen for adding a new search engine manually.

liuche:
Same as #1660 , this is Part 1 of the #961 meta - @npark-mozilla should we flag for QA upon partial landing of features, or should it be part of the meta when everything lands? Flagging QA+ for now to be safe.

npark-mozilla:
@liuche I think flagging it for each bug is a good idea too. will check.

PR #1669, opened 2017-11-07
Closes #1668 - Use Fragment newInstance and add null checks.

### LINKED ISSUE #1668: Null check in SettingsFragment
Broken SwitchLocaleTest because getArguments is null.

liuche:
Fallout from #1617.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
+++ b/app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
@@ -58,11 +58,9 @@ public class SettingsFragment extends PreferenceFragment implements SharedPrefer
         super.onCreate(savedInstanceState);
 
         final Bundle args = getArguments();
-        int prefResId = R.xml.settings;
-
-        if (args != null) {
-            prefResId = args.getInt(PREFERENCES_RESID_INTENT_EXTRA, R.xml.settings);
-        }
+        final int prefResId = args != null ?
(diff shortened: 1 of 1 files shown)


liuche:
Green BB UITests: https://dashboard.buddybuild.com/apps/584f67f6f3d6eb01000842d6/build/5a0228c991f506000193118b#uitests

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -185,7 +191,7 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
 
             // The easiest way to ensure we update the language is by replacing the entire fragment:
             getFragmentManager().beginTransaction()
-                    .replace(R.id.container, new SettingsFragment())
+                    .replace(R.id.container, SettingsFragment.newInstance(null, R.xml.settings, R.string.menu_settings))
```
mcomella: nit:  We use duplicate these resource arguments twice (putting them in danger of getting out of sync), here and in `SettingsActivity.onCreate`, when all we're really trying to say is, "Show top level settings".

It may be outside the scope of this quick fix, but I feel like `newInstance` should instead take a single argument with the screen we want, e.g. `enum SettingsScreen { TOP_LEVEL, ... }`.

liuche: I like that idea! But yeah, I don't want to do it in this fix so I filed #1671 for it.

PR #1673, opened 2017-11-08
Save custom search engine

liuche:
This adds the functionality behind the "Save" on the (manual) add search engine page.

The search engines are saved in the OpenSearch xml format, so we can reuse code and to be forward-compatible if we fetch OpenSearch search engines from webpages that supply them for search engines.

(Sorry about the commits covering multiple topics - I accidentally combined some commits during a rebase!)

### DIFF
--- a/app/src/main/java/org/mozilla/focus/search/SearchEngine.java
+++ b/app/src/main/java/org/mozilla/focus/search/SearchEngine.java
@@ -7,14 +7,32 @@ package org.mozilla.focus.search;
 
 import android.graphics.Bitmap;
 import android.net.Uri;
+import android.util.Log;
 
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+
+import java.io.StringWriter;
+import java.io.Writer;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.OutputKeys;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerConfigurationException;
+import javax.xml.transform.TransformerException;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 public class SearchEngine {
+    private static final String LOGTAG = "SearchEngine";
+
     // Parameters copied from nsSearchService.js
     private static final String MOZ_PARAM_LOCALE = "\\{moz:locale\\}";
     private static final String MOZ_PARAM_DIST_ID = "\\{moz:distributionID\\}";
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/mozilla/focus/search/SearchEngine.java
```diff
@@ -101,4 +130,79 @@ private String paramSubstitution(String template, String query) {
 
         return template;
     }
+
+    public static boolean addSearchEngine(SharedPreferences sharedPreferences, Context context, String engineName, String searchQuery) {
```
pocmo: SearchEngine is a simple model class / POJO. I wouldn't overload its meaning by adding additional static methods to it. Maybe this could be added to `SearchEngineManager` (or its own class; or the `Settings` helper class)?

app/src/main/java/org/mozilla/focus/search/SearchEngine.java
```diff
@@ -101,4 +130,79 @@ private String paramSubstitution(String template, String query) {
...
+
+        // Force SearchEngineManager to refetch, to get the newest search engine.
+        SearchEngineManager.getInstance().init(context);
+        return true;
+    }
+
+    public static String buildSearchEngineXML(String engineName, String searchQuery, Bitmap iconBitmap) {
+        Document document = null;
+        try {
+            document = DocumentBuilderFactory.newInstance().newDocumentBuilder().newDocument();
```
pocmo: I have been using `XmlSerializer` for writing XML to disk in the past. The advantage is that it can write to a stream immediately without the need to build up the document in memory first. But then again those files are rather small anyways.

app/src/main/java/org/mozilla/focus/search/SearchEngineParser.java
```diff
@@ -39,7 +38,7 @@ public static SearchEngine load(AssetManager assetManager, String identifier, St
         }
     }
 
-    @VisibleForTesting static SearchEngine load(String identifier, InputStream stream) throws IOException, XmlPullParserException {
+    public static SearchEngine load(String identifier, InputStream stream) throws IOException, XmlPullParserException {
```
pocmo: Does this need to be public? SearchEngineManager is in the same package and should have access already.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -44,10 +54,35 @@ public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
...
+                final String searchQuery = ((EditText) rootView.findViewById(R.id.edit_search_string)).getText().toString();
+
+                final SharedPreferences sharedPreferences = getActivity()
+                        .getSharedPreferences(SearchEngine.PREF_FILE_SEARCH_ENGINES, Context.MODE_PRIVATE);
+                if (!validateSearchFields(engineName, searchQuery, sharedPreferences)) {
+                    Snackbar.make(rootView, R.string.search_add_error, Snackbar.LENGTH_SHORT).show();
+                } else {
+                    SearchEngine.addSearchEngine(sharedPreferences, getActivity(), engineName, searchQuery);
+                    Snackbar.make(rootView, R.string.search_add_confirmation, Snackbar.LENGTH_SHORT).show();
+                    getActivity().onBackPressed();
```
pocmo: Should this be finish() or fragmentManager.popBackStack() ? Or does onBackPressed() do something additionally?

app/src/main/java/org/mozilla/focus/search/SearchEngine.java
liuche: Cool, it looks like that's nice because I wouldn't need to do the conversion with this transformer factory (which seemed weirdly complicated, but I assumed it was just "because Java"). I won't do it here but I'll file it in a follow-up with the unit test.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
liuche: onBackPressed does the combination of popBackStack and finish - for this case, I guess it doesn't really matter because it's not the last view, but maybe it makes more sense for the Fragment to use the FragmentManager call.

app/src/main/java/org/mozilla/focus/search/SearchEngineParser.java
liuche: Ah, I guess not! There was an AS error highlight, but I guess technically it's fine. Will change it back.

PR #1679, opened 2017-11-08
Closes #1671. Use enum for settings screen types.

### LINKED ISSUE #1671: Use enums for Settings screen types
From comment in #1669, the Settings fragments could use enums to declare the different settings screens rather than pairs of resource ids (which could be error prone).

liuche:
Use an enum instead of crazy pairings of resource ids.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
+++ b/app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
@@ -4,6 +4,7 @@
 
 package org.mozilla.focus.settings;
 
+import android.app.Fragment;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -29,41 +30,64 @@
...
+        switch (settingsType) {
+            case MAIN:
+            case SEARCH_ENGINES:
+                f = new SettingsFragment();
+                break;
+            case ADD_SEARCH:
+                f = new ManualAddSearchEngineSettingsFragment();
+                break;
+            default:
+                f = new SettingsFragment();
```
mcomella: You should throw here, right?

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -29,41 +30,64 @@
...
     public void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
         final Bundle args = getArguments();
-        final int prefResId = args != null ?
-                args.getInt(PREFERENCES_RESID_INTENT_EXTRA, R.xml.settings) :
-                R.xml.settings;
+        settingsScreen = args != null ?
+                SettingsScreen.valueOf(args.getString(SETTINGS_SCREEN_NAME, SettingsScreen.MAIN.name())) :
+                SettingsScreen.MAIN;
```
mcomella: nit: When can args be null if we always set args? Add a comment.

I assume it's just because they're fragments and they're bullshit. x_x

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -139,15 +147,10 @@ public void onResume() {
...
-        final int titleResId = args != null ?
-                args.getInt(TITLE_RESID_INTENT_EXTRA, R.string.menu_settings) :
-                R.string.menu_settings;
-        updater.updateTitle(titleResId);
+        updater.updateTitle(settingsScreen.titleResId);
         updater.updateIcon(R.drawable.ic_back);
 
-        // TODO: #1671 would add Enum types so figuring out the current Settings screen is cleaner
-        if (args != null && args.getInt(PREFERENCES_RESID_INTENT_EXTRA, 0) == R.xml.search_engine_settings_featureflag_manual) {
+        if (settingsScreen == SettingsScreen.SEARCH_ENGINES) {
```
mcomella: Do we also need to check the feature flag here? Before we were only checking one of the values behind SEARCH_ENGINES.

PR #1682, opened 2017-11-09
Add support for removing search engines

liuche:


app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -95,6 +107,40 @@ public void onAttach(Context context) {
     }
 
     @Override
+    public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
+        switch(settingsScreen) {
```
sputnikci: \[Checkstyle\] ERROR: 'switch' is not followed by whitespace\.

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -95,6 +107,40 @@ public void onAttach(Context context) {
...
+                break;
+            case REMOVE_ENGINES:
+                inflater.inflate(R.menu.menu_remove_search_engines, menu);
+                break;
+        }
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        switch(item.getItemId()) {
```
sputnikci: \[Checkstyle\] ERROR: 'switch' is not followed by whitespace\.

app/src/main/java/org/mozilla/focus/search/SearchEngineListPreference.java
```diff
@@ -40,24 +39,24 @@ public SearchEngineChooserPreference(Context context, AttributeSet attrs, int de
...
             // There is no search engine group yet.
             return;
         }
-        final SearchEngineManager sem = SearchEngineManager.getInstance();
         final Context context = searchEngineGroup.getContext();
         final Resources resources = context.getResources();
+        final SearchEngineManager sem = SearchEngineManager.getInstance();
+
+        // Force SearchEngineManager to refetch, to get the newest search engine.
+        sem.init(context);
```
pocmo: Those init calls are problematic. SearchEngineManager registers a broadcast receiver and now this happens multiple times. Previously this happened only for the Application context (as init was called from FocusApplication). Now we register it multiple times for the Activity, which at some point will get finished. But the registration still exists and keeps a reference to the Activity causing:

> android.app.IntentReceiverLeaked: Activity org.mozilla.focus.activity.SettingsActivity has leaked IntentReceiver org.mozilla.focus.search.SearchEngineManager@5f00ea4 that was originally registered here. Are you missing a call to unregisterReceiver()?

There's no need to register/unregister new receivers, so I guess it would be enough to call loadSearchEngines() instead of init(). Also we might want to use the Application Context as the load operation on a background thread might again live longer than the activity.

pocmo:
> I added a search engine locally, but whenever I select "Remove" from the menu I can only select engines to remove from the remaining list of default engines. My custom one just disappears in "remove mode"...?

Oh, I'm stupid. This is because my custom engine is my default engine..... :)

app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
```diff
@@ -339,4 +340,8 @@ public static void removeSearchEngines(Set<String> engineIdsToRemove, SharedPref
     public static void restoreDefaultSearchEngines(SharedPreferences sharedPreferences) {
         sharedPreferences.edit().remove(PREF_KEY_HIDDEN_DEFAULT_ENGINES).apply();
     }
+
+    private void invalidateSearchEngines() {
```
pocmo: I think this method needs to be `synchronized`. Otherwise another thread reading it might not see it that is has changed, e.g.:
https://www.javacodegeeks.com/2014/08/java-concurrency-tutorial-visibility-between-threads.html

PR #1690, opened 2017-11-10
Search engine polish

liuche:
Fixes:
- Show disabled default search engine instead of hiding it
- Show default search engine icon matching title instead of url
- Screenshots UITest


app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
```diff
@@ -31,8 +31,9 @@ protected int getItemResId() {
 
     @Override
     protected void updateDefaultItem(CompoundButton defaultButton) {
-        // Hide default engine so it can't be removed
-        defaultButton.setVisibility(View.GONE);
+        defaultButton.setClickable(false);
+        // Use state_activated works because state_enabled doesn't work in state list.
```
mcomella: What does this  comment mean? Is `.setActivated` really necessary? I'd think you'd just need to set it not clickable.

app/src/main/res/drawable/foreground_disabled.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mcomella: nit: `foreground_disabled.xml` is a pretty generic name. Can we name it like `preference_multiselect...` da-da-da if it's not used anywhere else?

app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
liuche: The reason for 'activated' is so that we can have use a statelistdrawable - there isn't a state_clickable, and for some reason state_enabled was just not working (perhaps it needed a default drawable state).

https://developer.android.com/reference/android/graphics/drawable/StateListDrawable.html

app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.kt
```diff
@@ -23,37 +23,55 @@ class IconGenerator {
 
     companion object {
         private val TEXT_SIZE_DP = 36f
-        private val DEFAULT_ICON_CHAR = "?"
+        private val DEFAULT_ICON_CHAR = '?'
 
         /**
-         * Use the given raw website icon and generate a launcher icon from it. If the given icon is null
-         * or too small then an icon will be generated based on the website's URL. The icon will be drawn
+         * Generate an icon based on the website's URL. The icon will be drawn
```
mcomella: nit: We should say, "See [generateLaunchIcon...] for more details" rather than duplicating the comment.

app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.kt
```diff
@@ -23,37 +23,55 @@ class IconGenerator {
...
+        @JvmStatic
+        fun generateCharacterIcon(context: Context, character: Char) = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            generateAdaptiveLauncherIcon(context, character)
         } else {
-            generateLauncherIconPreOreo(context, url)
+            generateLauncherIconPreOreo(context, character)
+        }
+
+        @JvmStatic
+        fun generateLauncherIconPreOreo(context: Context, url: String?): Bitmap {
```
mcomella: I'd rather expose `getRepresentativeCharacter` to the code calling this and make it use the Char method than to duplicate the code

app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
mcomella: Can you make your comment sound more like this explanation? :)

app/src/androidTest/java/org/mozilla/focus/activity/screenshots/SettingsScreenshots.java
```diff
@@ -79,14 +82,26 @@ public void takeScreenShotsOfSettings() throws Exception {
...
+            TestHelper.pressBackKey();
+
+            /* Manual Search Engine page */
             final String addEngineLabel = getString(R.string.preference_search_add2);
             onData(withTitleText(addEngineLabel))
                     .perform(click());
             TestHelper.settingsHeading.waitUntilGone(waitingTime);
             Screengrab.screenshot("SearchEngine_Add_Search_Engine");
             TestHelper.pressBackKey();
+
```
mcomella: nit: unnecessary newline

liuche:
Hm, seeing a bunch of UITest failures that seem unrelated. Since this is already done and just needs to land, I'll land when I get back from PTO.

liuche:
Rebased and ran a bunch of these UITests locally and they pass, so going to repush and see if they work on BB.

liuche:
Tests still failing on BB, despite passing locally. Master also looks like it has a lot of red (UiObjectNotFoundException, failed assertions). @npark-mozilla let me know if you can think of anything that I might be missing. I'm hesitant to land this because it hasn't gone green on BB yet (even though they don't seem related), but I'm wondering if I should just land it because things are landing on a red master anyway.

liuche:
I started to break up the commits to run separately but realized that the flavor that was built here and having UITests run on was GeckoDebug - going to switch it to WebviewDebug to re-run.

liuche:
One timing failure, but otherwise green!

Thanks No-Jun for finding that for the physical devices, build-variant was empty - should be set to FocusWebviewDebug, and kicked off with "Rebuild with no cache".

PR #1693, opened 2017-11-10
Disable swipe to refresh on error pages closes #1685

### LINKED ISSUE #1685: Pull down to refresh animation freezes on error page
screen shot 2017-11-09 at 11 03 53 am

Found when investigating #1610

STR: 
1. Go to invalid address
2. Pull down to refresh 

Expected Behavior : Error page is refreshed 
Behavior : Spinning loading animation never goes away

ekager:
This would be a quick fix to disable the swipe to refresh on error pages. I think a problem lies in that the second onPageFinished is not called for error pages, and so the swipe reload animation never ends?


pocmo:
> I think a problem lies in that the second onPageFinished is not called for error pages, and so the swipe reload animation never ends?

Does `onPageStarted()` get called? If not then we might not call `onPageFinished()` anymore in `SessionCallbackProxy`.

PR #1721, opened 2017-11-15
Update default search engines for US/CA/HK/TW/UA

mkaply:
This patch adds a new engine with our code and uses that engine for the US.

It switches away from using the no codes engine in Canada, the zh-TW build and in the Ukraine.

### DIFF
--- a/app/src/main/assets/search/default/google-2018.xml
+++ b/app/src/main/assets/search/default/google-2018.xml
@@ -1,0 +1,17 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<SearchPlugin xmlns="http://www.mozilla.org/2006/browser/search/">
+<ShortName>Google</ShortName>
+<InputEncoding>UTF-8</InputEncoding>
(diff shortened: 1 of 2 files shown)


pocmo:
@mkaply Thanks! Is this already in mozilla-central / Fennec?

list.json is not used directly in Focus. Instead it will be processed by scrape_plugins.py which will do some modifications (e.g. remove bing) and pull the search plugins from mozilla-central. It will then generate search_configuration.json which is used by Focus at runtime.

As a quick fix we can land this and modify search_configuration.json  manually (I can put up a follow-up patch). As soon as mozilla-central is up-to-date I would like to run scrape_plugins.py again and this will hopefully lead to the same result (or maybe pick up even more changes that we haven't ported to Focus yet).

pocmo:
Looking at the code of scrape_plugins.py ... it's super outdated. It points to mozilla-aurora and the repository structure has changed since then (Apparently the plugins used to live in the L10N repo?).

Landing this and modifiying search_configuration.json manually seems to be the way to go for this release.

pocmo:
@mkaply I wonder why we completely removed "UA", "HK" and "TW" from list.json?

pocmo:
Updating search_configuration.json in #1725

mkaply:
> @mkaply I wonder why we completely removed "UA", "HK" and "TW" from list.json?

You only need to add region specific chnages if they differ from the default configuration. For these three regions (UA is Ukraine), we are now using google (not google-nocodes) so we no longer need a custom configuration.

PR #1731, opened 2017-11-16
Record when user opens Focus via tap on Focus app icon closes #1635

### LINKED ISSUE #1635: Add probe to record when user opens Focus via tap on Focus app icon
Besides knowing when a user opened Focus via shortcut or intent (via different apps), I'd like know when a user actually clicks on the Focus app icon.

https://github.com/mozilla-mobile/focus-android/wiki/Telemetry#browsing

ekager:
@bbinto This would add a telemetry probe to record opening Focus ("actual app launches") from an icon if it is *not* currently open; the first option @pocmo gave you in #1635 . Wasn't sure from the discussion if we want two probes or to record all clicks on the App Icon, so I can change this to include "app resumed from the launcher icon" case as well if you like!

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -62,6 +62,10 @@ public class MainActivity extends LocaleAwareAppCompatActivity {
 
         final SafeIntent intent = new SafeIntent(getIntent());
 
+        if (intent.isLauncherIntent()) {
+            TelemetryWrapper.openFromIconEvent();
+        }
+
         sessionManager.handleIntent(this, intent, savedInstanceState);
 
         sessionManager.getSessions().observe(this,  new NonNullObserver<List<Session>>() {
@@ -145,6 +149,10 @@ public class MainActivity extends LocaleAwareAppCompatActivity {
 
         if (ACTION_ERASE.equals(action)) {
             processEraseAction(intent);
+        }
+
+        if (intent.isLauncherIntent()) {
+            TelemetryWrapper.resumeFromIconEvent();
         }
     }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -62,6 +63,13 @@ protected void onCreate(Bundle savedInstanceState) {
 
         final SafeIntent intent = new SafeIntent(getIntent());
 
+        final Set<String> intentCategories = intent.getCategories();
+        if (intentCategories != null) {
+            if (intentCategories.contains(Intent.CATEGORY_LAUNCHER) && intent.getAction().equals(Intent.ACTION_MAIN)) {
```
sputnikci: \[PMD\] INFO: These nested if statements could be combined

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -62,6 +63,11 @@ protected void onCreate(Bundle savedInstanceState) {
 
         final SafeIntent intent = new SafeIntent(getIntent());
 
+        final Set<String> intentCategories = intent.getCategories();
+        if (intentCategories != null && intentCategories.contains(Intent.CATEGORY_LAUNCHER) && intent.getAction().equals(Intent.ACTION_MAIN)) {
```
pocmo: We could move all this inside SafeIntent behind a method. Something like isLauncherIntent() maybe?

bbinto:
Not sure if my input is still needed, I'm fine with one (1) but if you can add the second probe, I won't complain :) Thanks for checking in

pocmo:
Perfect. Thanks!

PR #1745, opened 2017-11-19
Swipe to close tab

codebu5ter:
Implemented a way to close individual tabs, wherein the user has to swipe left on the tab to close it.
It is consistent with material design guidelines as Google also uses this functionality in its apps, like the Gmail app.

Closes #1362

PR #1758, opened 2017-11-21
WIP Refactor SessionManager

Benestar:
This patch converts SessionManager to Kotlin, makes use of the nice Lambda supporting functions on Iterables from Kotlin and adds support to remove Session objects by position (needed by #1745).


Benestar:
Sorry that this Pull Request grow so large, I tried to make reviewing easier by splitting the changes into several commits that can be reviewed independently. However, I know that this is still a big patch so already thanks for looking at this. :-)

pocmo:
Is this still a WIP or should this be reviewed, @Benestar? :)

app/src/main/java/org/mozilla/focus/session/SessionManager.kt
```diff
@@ -0,0 +1,227 @@
...
+import org.mozilla.focus.customtabs.CustomTabConfig
+import org.mozilla.focus.shortcut.HomeScreen
+import org.mozilla.focus.utils.SafeIntent
+import org.mozilla.focus.utils.UrlUtils
+import java.util.*
+
+/**
+ * Sessions are managed by this global SessionManager instance.
+ */
+class SessionManager private constructor() {
```
pocmo: This should be an `object` instead of a `class`. With that it's automatically a singleton and the INSTANCE reference can go.

app/src/main/java/org/mozilla/focus/session/SessionManager.kt
```diff
@@ -0,0 +1,227 @@
...
+    /**
+     * Get the current session. This method will throw an exception if there's no active session.
+     */
+    val currentSession: Session
+        get() {
+            if (currentSessionUUID == null) {
+                throw IllegalAccessError("There's no active session")
+            }
+
+            return getSessionByUUID(currentSessionUUID!!)
```
pocmo: We should avoid using `!!` as much as possible.


Maybe helpful here:
```
currentSessionUUID?.let {
  ...
}
```

app/src/main/java/org/mozilla/focus/session/SessionManager.kt
```diff
@@ -0,0 +1,227 @@
...
+    fun handleNewIntent(context: Context, intent: SafeIntent) {
+        createSessionFromIntent(context, intent)
+    }
+
+    private fun createSessionFromIntent(context: Context, intent: SafeIntent) {
+        val action = intent.action
+
+        if (Intent.ACTION_VIEW == action) {
+            val dataString = intent.dataString
+            if (TextUtils.isEmpty(dataString)) {
```
pocmo: I think Kotlin has a extension function for that. Something like isNullOrEmpty()

pocmo:
No activity. I'll close this PR for now. We can always re-open it.

PR #1827, opened 2017-11-28
Telemetry for autocomplete feature

pocmo:
📈 📉 📊 🤓


app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -0,0 +1,593 @@
...
+    @JvmStatic
+    fun urlBarEvent(isUrl: Boolean, autocompleteResult: AutocompleteResult) {
+        if (isUrl) {
+            TelemetryWrapper.browseEvent(autocompleteResult)
+        } else {
+            TelemetryWrapper.searchEnterEvent()
+        }
+    }
+
+    private fun browseEvent(autocompleteResult: AutocompleteResult) {
```
liuche: Cool, was looking to see what data was collected here, just size, custom/default, and isEmpty is fine, just Type 2 data.

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteSettingsFragment.kt
```diff
@@ -41,4 +51,12 @@ class AutocompleteSettingsFragment : PreferenceFragment() {
 
         return super.onPreferenceTreeClick(preferenceScreen, preference)
     }
+
+    override fun onSharedPreferenceChanged(sharedPreferences: SharedPreferences?, key: String?) {
+        if (key == null || sharedPreferences == null) {
+            return
+        }
+
+        TelemetryWrapper.settingsEvent(key, sharedPreferences.all[key].toString())
```
liuche: Since this is returning a setting value, this will never be something like a custom domain that's saved, right? Just looking at these TelemetryWrapper changes in this patch for data review it's not obvious to me if the custom domains are saved as values in settings (but probably not).

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -0,0 +1,593 @@
...
+        val CUSTOM_TAB_CLOSE_BUTTON = "custom_tab_close_but"
+        val CUSTOM_TAB_ACTION_BUTTON = "custom_tab_action_bu"
+        val FIRSTRUN = "firstrun"
+        val DOWNLOAD_DIALOG = "download_dialog"
+        val ADD_TO_HOMESCREEN_DIALOG = "add_to_homescreen_dialog"
+        val HOMESCREEN_SHORTCUT = "homescreen_shortcut"
+        val TABS_TRAY = "tabs_tray"
+        val RECENT_APPS = "recent_apps"
+        val APP_ICON = "app_icon"
+        val AUTOCOMPLETE_DOMAIN = "autocomplete_domain"
```
liuche: Should add this to the telemetry wiki documentation.

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteSettingsFragment.kt
pocmo: Correct. In this screen we only show two switches to enable/disable autocomplete for default and custom domains.
The actual domains are added in another screen and are not stored in the app's default shared preferences (although in a different shared preferences file). Domains are never reported in any of the telemetry probes.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
pocmo: Yep. I was waiting for the merge before updating the wiki. :)

Issue #1567, opened 2017-10-26
update search engine manual entry error strings


lime124:
We're not in love with the error strings shown during manual entry - they aren't helpful. This ticket to update those. 

Depending on what @BrianNJones comes up with, this may need a little bit of eng effort in addition to just updating the strings

pocmo:
@lime124 @BrianNJones This is the only P1 with "needs string" label. Is there anything we can export today so that the L10N team can start translating it?

lime124:
i think @BrianNJones is still working on one string. i'll let him update. thanks!

BrianNJones:
@pocmo, if you can confirm that tapping "Save" validates that text exists in the name and string fields, we'd need error messages for their empty states:
If no text entered in "Search engine name" field, string would be:
Enter search engine name
If no text entered in "Search string to use" field, string would be:
Enter search string

liuche:
Looks like these strings are landed so going to close this - they'll get picked up in the next export.

liuche:
Oh wait, the string and feature issues are the same, reopening.

liuche:
Just a string change, should be picked up in UX review as the last step of #961.

PR #1844, opened 2017-11-30
Remove unused string + string import 

liuche:
Removing an unused string from #1567 and re-importing to sync the strings files.


liuche:
Also filed two bugs for missing placeholders in the strings:
https://bugzilla.mozilla.org/show_bug.cgi?id=1421851
https://bugzilla.mozilla.org/show_bug.cgi?id=1421852

liuche:
Oops, forgot to comment out the lines of the import script that pulls a fresh copy of l10n master, thereby overwriting the exported removal of the string.

PR #1849, opened 2017-11-30
Closes #1656 "Open with..." does not show Rocket when user has default browser

### LINKED ISSUE #1656: "Open with..." does not show Rocket when user has default browser
via https://github.com/mozilla-tw/Rocket/issues/1149#issuecomment-341329353

> Here's the brief:
> 1. If no default browser is set, we can easily identify all browsers that can handle this open intent.
> 2. However, when a default browser is set, this does not work anymore, Focus try to send intent to all those known browsers which is hardcoded to try find if these exist on user's environment. Firefox Rocket is not listed (yet) in this hardcoded known browsers.

If I understand this correctly, we should add Rocket to the known browsers list.

nosark:
… #1656 

I noticed while fixing #1656 that the two functions below do not include Firefox Rocket:
findFirefoxBrandedBrowser()
isFirefoxDefaultBrowser()

Should these also be updated with Rocket?

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/Browsers.java
+++ b/app/src/main/java/org/mozilla/focus/utils/Browsers.java
@@ -29,6 +29,7 @@ public class Browsers {
         FIREFOX_BETA("org.mozilla.firefox_beta"),
         FIREFOX_AURORA("org.mozilla.fennec_aurora"),
         FIREFOX_NIGHTLY("org.mozilla.fennec"),
+        FIREFOX_ROCKET("org.mozilla.rocket"),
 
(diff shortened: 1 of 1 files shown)


pocmo:
> I noticed while fixing #1656 that the two functions below do not include Firefox Rocket:
findFirefoxBrandedBrowser()
isFirefoxDefaultBrowser()
> 
> Should these also be updated with Rocket?

No. The name is probably not perfect. Those are about "Firefox for Android" only.

nosark:
okie dokie thanks @pocmo  !

PR #1857, opened 2017-11-30
Telemetry for SSL Errors Received

### LINKED ISSUE #1775: Telemetry probe for SSL errors
See #1338. We want to see how often our users run into SSL errors. If possible we would like to even see categories of SSL errors.

ekager:
We didn't have error telemetry values so I had to add some but not sure if it's the right format for Category etc.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
+++ b/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
@@ -6,6 +6,7 @@
 package org.mozilla.focus.telemetry
 
 import android.content.Context
+import android.net.http.SslError
 import android.os.StrictMode
 import android.preference.PreferenceManager
 import android.support.annotation.CheckResult
@@ -44,6 +45,7 @@ object TelemetryWrapper {
 
     private object Category {
         val ACTION = "action"
+        val ERROR = "error"
     }
 
     private object Method {
@@ -71,6 +73,8 @@ object TelemetryWrapper {
         val REMOVE = "remove"
         val REORDER = "reorder"
         val RESTORE = "restore"
+        val PAGE = "page"
+        val RESOURCE = "resource"
     }
 
     private object Object {
@@ -137,6 +141,7 @@ object TelemetryWrapper {
         val AUTOCOMPLETE = "autocomplete"
         val SOURCE = "source"
         val SUCCESS = "success"
+        val ERROR_CODE = "error_code"
     }
 
     @JvmStatic
@@ -581,6 +586,23 @@ object TelemetryWrapper {
     @JvmStatic
     fun menuReloadEvent() {
         TelemetryEvent.create(Category.ACTION, Method.CLICK, Object.MENU, Value.RELOAD).queue();
+    }
+
+    @JvmStatic
+    fun sslErrorEvent(fromPage: Boolean, error: SslError) {
+        // SSL Errors from https://developer.android.com/reference/android/net/http/SslError.html
+        val primaryErrorMessage = when (error.primaryError) {
+            SslError.SSL_DATE_INVALID -> "SSL_DATE_INVALID"
+            SslError.SSL_EXPIRED -> "SSL_EXPIRED"
+            SslError.SSL_IDMISMATCH -> "SSL_IDMISMATCH"
+            SslError.SSL_NOTYETVALID -> "SSL_NOTYETVALID"
+            SslError.SSL_UNTRUSTED -> "SSL_UNTRUSTED"
+            SslError.SSL_INVALID -> "SSL_INVALID"
+            else -> "Undefined SSL Error"
+        }
+        TelemetryEvent.create(Category.ERROR, if (fromPage) Method.PAGE else Method.RESOURCE, Object.BROWSER)
+                .extra(Extra.ERROR_CODE, primaryErrorMessage)
+                .queue()
     }
 
     fun saveAutocompleteDomainEvent() {
--- a/app/src/webview/java/org/mozilla/focus/webview/FocusWebViewClient.java
+++ b/app/src/webview/java/org/mozilla/focus/webview/FocusWebViewClient.java
@@ -18,6 +18,7 @@ import android.webkit.WebView;
 import android.webkit.WebViewClient;
 
 import org.mozilla.focus.browser.LocalizedContent;
+import org.mozilla.focus.telemetry.TelemetryWrapper;
 import org.mozilla.focus.utils.IntentUtils;
 import org.mozilla.focus.utils.UrlUtils;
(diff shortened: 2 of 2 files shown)


app/src/webview/java/org/mozilla/focus/webview/FocusWebViewClient.java
```diff
@@ -263,6 +264,27 @@ public void onReceivedSslError(WebView view, SslErrorHandler handler, SslError e
         }
     }
 
+    private void recordSSLErrors(SslError error) {
```
liuche: What does the sslerror.toString return? I'd assume it gives a string representation of the error so we don't need to translate these to strings ourselves.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -70,6 +71,7 @@ object TelemetryWrapper {
         val SHARE_INTENT = "share_intent"
         val REMOVE = "remove"
         val REORDER = "reorder"
+        val RECEIVE = "receive"
```
liuche: Receive isn't very specific, and it looks like maybe we want to differentiate between an error from the current page or a page resource. I'd actually add two methods, PAGE and RESOURCE, and add a comment about what they are (see the onReceivedSslError method and comment for the two cases).

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -120,6 +122,12 @@ object TelemetryWrapper {
         val WHATS_NEW = "whats_new"
         val RESUME = "resume"
         val RELOAD = "refresh"
+        val SSLDataInvalidError = "SSLDateInvalidError"
```
liuche: I don't think we should add these all manually - either the toString method of SslError gives us something useful (which I'm pretty sure is the case), or we just return the code and handle it on the telemetry end.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -576,6 +584,36 @@ object TelemetryWrapper {
         TelemetryEvent.create(Category.ACTION, Method.CLICK, Object.MENU, Value.RELOAD).queue();
     }
 
+    @JvmStatic
```
liuche: Don't need these then. In general, if you find yourself copying a lot of information that is encapsulated in an API, there should be a better way.

app/src/webview/java/org/mozilla/focus/webview/FocusWebViewClient.java
```diff
@@ -252,7 +253,7 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
     @Override
     public void onReceivedSslError(WebView view, SslErrorHandler handler, SslError error) {
         handler.cancel();
-
+        recordSSLErrors(error);
```
liuche: Since there are two states here, we probably want to differentiate between the two.

app/src/webview/java/org/mozilla/focus/webview/FocusWebViewClient.java
sputnikci: \[PMD\] INFO: The method 'recordSSLErrors' has a Cyclomatic Complexity of 7\.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
liuche: Actually, on second thought, if the toString method doesn't return anything useful, then we should have this int => string handled in TelemetryWrapper. These don't need to be variables because we don't expect to use them anywhere else, so if we NEED to because the API isn't returning a useful string result for the errors, add a function like sslErrorCodeToString that switches through the code and returns a string.

liuche:
I'll get to this next week, sorry!

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -576,6 +581,25 @@ object TelemetryWrapper {
...
+    fun SSLError(fromPage: Boolean, error: SslError) {
+        val primaryErrorMessage:String
+        when (error.primaryError) {
```
liuche: Nice use of kotlin state. Optionally include a link to the API here that lists these ssl errors.

pocmo: nit: You can use `when` directly in an assignment:

```
val x = when(y) {
  a -> ".."
  b -> ".."
}
```

ekager: Very cool!

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -582,6 +587,25 @@ object TelemetryWrapper {
         TelemetryEvent.create(Category.ACTION, Method.CLICK, Object.MENU, Value.RELOAD).queue();
     }
 
+    @JvmStatic
+    fun SSLError(fromPage: Boolean, error: SslError) {
```
pocmo: detekt (on taskcluster) is throwing a warning:
> FunctionNaming - [SSLError] at focus-android/app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt:590:5

Probably because it doesn't start with a lowercase character. Maybe sslErrorEvent(...)?

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -582,6 +587,25 @@ object TelemetryWrapper {
...
+        val primaryErrorMessage = when (error.primaryError) {
+            SslError.SSL_DATE_INVALID -> "SSL_DATE_INVALID"
+            SslError.SSL_EXPIRED -> "SSL_EXPIRED"
+            SslError.SSL_IDMISMATCH -> "SSL_IDMISMATCH"
+            SslError.SSL_NOTYETVALID -> "SSL_NOTYETVALID"
+            SslError.SSL_UNTRUSTED -> "SSL_UNTRUSTED"
+            SslError.SSL_INVALID -> "SSL_INVALID"
+            else -> {
+                "Undefined SSL Error"
+            }
```
pocmo: nit: No block needed:

```Kotlin
else -> "Undefined SSL Error"
```

liuche:
TC failure due to "unhandled exception" in http validation. I'll fix it - it's breaking every CI build.

Issue #1755, opened 2017-11-20
[ux] search engine - implementation phase


lime124:
Track work done during implementation. Track with @liuche polish bug since there isn't a meta issue.

lime124:
@liuche Feedback on build:

manual entry screen
- [ ] for the text fields - I want to use standard Android components. They should follow these states: https://material.io/guidelines/components/text-fields.html#text-fields-states
- [ ] can we get rid of the line under the example?
- [ ] is it possible to give 16dp padding on both sides of the text fields?

remove screen
- [ ] should be an X instead of a back

main search screen
- [ ] what size is the font in the new search engine icon? we need to make it a lot bigger :)
- [ ]  in the overflow menu - can we disable restore until one of the ones that ships is deleted

liuche:
@lime124 made some changes, a few questions for you:

1. Standard Android Components:
There is a standard Android component for the label/text behavior of Material Design!...except it's a known issue that there are no "Placeholder" strings: https://issuetracker.google.com/issues/37065330

This means we don't get useful hint strings about copy-pasting the text in, and it looks like this:
![screenshot_1512115960](https://user-images.githubusercontent.com/1082253/33473746-7aaa6174-d62c-11e7-8b64-1b904c3d0b15.png)

I will take some time tomorrow to see if I can hack the hint strings in if there is no text, but which would you prefer?
a) No hint text but standard components
b) The way it is (non-standard text fields)

2. I have two icons with different text sizes, the "n" is 100f and the "u" is 64f. Do you like one of them, or do you have another size I should try?
![screenshot_1512115642](https://user-images.githubusercontent.com/1082253/33473867-fbb38cbe-d62c-11e7-827c-9dbbb12de6d8.png)

3. Having some trouble getting rid of the line under "example", if I don't figure out a way to do it, can we move it to the next sprint? or WONTFIX (but I have a few more things to try)

liuche:
Okay, fixed all the issues but the line under, and after talking to @lime124 decided we can punt that for this release.

liuche:
Closed by #1865.

PR #1865, opened 2017-12-01
Address UX feedback

liuche:
First patch is experimental, unsure if UX wants the tradeoffs, and I've asked in the bug.

The rest are listed in #1755.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
+++ b/app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
@@ -337,6 +337,10 @@ public class SearchEngineManager extends BroadcastReceiver {
         enginesEditor.apply();
     }
 
+    public static boolean hasAllDefaultSearchEngines(SharedPreferences sharedPreferences) {
+        return !sharedPreferences.contains(PREF_KEY_HIDDEN_DEFAULT_ENGINES);
+    }
+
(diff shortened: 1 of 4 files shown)


pocmo:
Adding @mcomella: You recently converted the icon code to Kotlin and refactored it to support adaptive icons. Can you take a look too?

app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.kt
```diff
@@ -72,16 +72,16 @@ class IconGenerator {
...
 
-        private fun drawCharacterOnBitmap(context: Context, character: Char, bitmap: Bitmap): Bitmap {
+        private fun drawCharacterOnBitmap(context: Context, character: Char, bitmap: Bitmap, textSize: Float?): Bitmap {
             val canvas = Canvas(bitmap)
 
             val paint = Paint()
 
             val textSize = TypedValue.applyDimension(
-                    TypedValue.COMPLEX_UNIT_DIP, TEXT_SIZE_DP, context.resources.displayMetrics)
+                    TypedValue.COMPLEX_UNIT_DIP, textSize ?: TEXT_SIZE_DP, context.resources.displayMetrics)
```
pocmo: You could set the default value for the textSize in the top most function right in the parameter list:
> fun generate...(..., textSize : Float = TEXT_SIZE_DP) {

With that you can remove the `?` from all other functions so that they do not have to deal with null values.

PR #1873, opened 2017-12-01
Custom Tabs: Always show the "Share" menu item by default. (#1358)

pocmo:


app/src/main/java/org/mozilla/focus/customtabs/CustomTabConfig.java
```diff
@@ -162,7 +162,11 @@ public static CustomTabConfig parseCustomTabIntent(final @NonNull Context contex
 
         // Share is part of the default menu, so it's simplest just to toggle it off as necessary instead
         // of creating a fake menu item here, hence we keep this as  aboolean for now:
-        final boolean showShareMenuItem = intent.getBooleanExtra(CustomTabsIntent.EXTRA_DEFAULT_SHARE_MENU_ITEM, false);
+        final boolean showShareMenuItem = intent.getBooleanExtra(CustomTabsIntent.EXTRA_DEFAULT_SHARE_MENU_ITEM,
+                // Other than Chrome we will show the share menu by default. Note that currently the
```
mcomella: nit: "Other than" -> "Unlike" - I'll file a good first issue.

app/src/main/java/org/mozilla/focus/customtabs/CustomTabConfig.java
```diff
@@ -162,7 +162,11 @@ public static CustomTabConfig parseCustomTabIntent(final @NonNull Context contex
...
+                // Custom Tabs support library has no option to explicitly hide this menu item. So
```
mcomella: Can apps develop without the custom tabs support library? e.g. send the intend manually and override with `false`? In that case, I think we may we want to clarify the comment (good first bug?).

pocmo: Yes, they could. At then end of the day a custom tab is mostly an Intent with extras.

pocmo: But I doubt that any app is actually doing this. They would need to do the Intent routing ("find browser with custom tabs implementation") etc. themselves.

Issue #1836, opened 2017-11-29
When no Custom URL is selected, tapping trash can exits the Custom URLs dialog


npark-mozilla:
STR:
Add custom URL
Tap 3-dot menu and select 'Remove'
Tap Trash can icon without selecting custom URL checkbox

Expected:
- Since no url is selected, either the trash can should be greyed out or have no action
Actual:
- exits 'Remove custom URLs' dialog

pocmo:
I followed the behavior of the "remove search engines" screen here.

@lime124 Is this a bug? What's the expected behavior?

lime124:
Nice catch @npark-mozilla ! I didn't call out what should happen in this case and that's my fault. 

In this case we should disable the trash can until at least one item is selected. @pocmo can we make sure this behaviour is replicated for autocomplete too? Thanks!

pocmo:
Okay, I will have a look.

CC @liuche In case you have time to do this for search engines. :)

pocmo:
@npark-mozilla I have a patch for autocomplete in review. Can you file a separate issue for search engines?

liuche:
@npark-mozilla One thing I found was that the trashcan doesn't look disabled when there are no items selected, but it **is** disabled - and maybe this varies by device?

mcomella:
Closed with https://github.com/mozilla-mobile/focus-android/commit/e34247507e9915267c4d63bd73d93a9c46ead51d.

mcomella:
At least, I think this is closed by ^.

@pocmo Can you use the "Closes #xxx" tag so we know which PRs close which issues (as opposed to some PRs which might partially address an issue)? Just in case you're ever not around to close issues.

pocmo:
> @pocmo Can you use the "Closes #xxx" tag so we know which PRs close which issues (as opposed to some PRs which might partially address an issue)? Just in case you're ever not around to close issues.

I don't like auto-closing issues most of the time because landing the code is not always sufficient. I always check if there are additional tasks in the issue. E.g. this one doesn't have all labels set correctly. It's missing the sprint label. :)

npark-mozilla:
@liuche Now I notice that with API 23, the trashcan is not greyed out, but it's disabled. In API 26 prior to the fix, it was actually exiting when pressed. Which I just verified that it's now greyed out and disabled.

mcomella:
[...]

PR #1874, opened 2017-12-01
Autocomplete: Disable remove icon if no domain is selected. (#1836)

pocmo:


app/src/main/res/drawable/ic_delete.xml
```diff
@@ -5,5 +5,5 @@
         android:viewportHeight="24.0">
     <path
         android:pathData="M6,19c0,1.1 0.9,2 2,2h8c1.1,0 2,-0.9 2,-2V7H6v12zM19,4h-3.5l-1,-1h-5l-1,1H5v2h14V4z"
-        android:fillColor="#ffffff"/>
+        android:fillColor="@drawable/state_colors"/>
```
liuche: Are color state lists supported in vector drawables? On a Nexus 5 / 6.0 the trashcan icon isn't getting set to the !isEnabled drawable listed in the state color list, even though the states are being updated correctly. I wonder if state color lists need to wrap individual drawables/svgs, rather than the other way around.

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
```diff
@@ -130,7 +130,11 @@ open class AutocompleteListFragment : Fragment() {
     }
 
     override fun onPrepareOptionsMenu(menu: Menu?) {
-        menu?.findItem(R.id.remove)?.isVisible = isSelectionMode() || domainList.adapter.itemCount > 1
+        val removeItem = menu?.findItem(R.id.remove)
+
+        removeItem?.isVisible = isSelectionMode() || domainList.adapter.itemCount > 1
```
liuche: I see, this is actually the same R.id.remove being used for (the different) menus of both fragments (List and Remove), and the "prepareMenu" logic is being shared between them, so it's a little more more complex but does work :)

PR #1877, opened 2017-12-01
Ktlint

pocmo:


build.gradle
```diff
@@ -45,3 +45,19 @@ allprojects {
...
+
+dependencies {
+    ktlint "com.github.shyiko:ktlint:0.14.0"
+}
+
+task ktlint(type: JavaExec, group: "verification") {
+    description = "Check Kotlin code style."
+    classpath = configurations.ktlint
+    main = "com.github.shyiko.ktlint.Main"
+    args "**/*.kt"
```
mcomella: The docs also include:
```
    // to generate report in checkstyle format prepend following args:
    // "--reporter=plain", "--reporter=checkstyle,output=${buildDir}/ktlint.xml"
    // see https://github.com/shyiko/ktlint#usage for more
```

I wonder if this is desirable so if we ever parse checkstyle results, we can just re-use that code.

mcomella:
I rebased on master and ran `./gradlew ktlint` without errors so I'm going to merge.

Issue #1837, opened 2017-11-29
Loading bar occassionally disappears during load after deleting data


mcomella:
The loading bar will disappear until I try to load the next page.

I can't reproduce it consistently but I've been seeing it for 3-ish weeks and liuche can also see this (I asked via IRC).

@pocmo @ekager Any ideas what could have caused this?

pocmo:
2.5 on Google Play is affected too.

There's some new code in BrowserFragment that does not update the loading bar and background etc. if we haven't been loading before. Maybe that's causing this?

nosark:
@mcomella I would like to take this if it is available :)

mcomella:
@nosark Sure, go for it! Thanks for your help! Please note that this is likely to be a high-priority issue so if it's not fixed quickly, one of the core maintainers may need to take it over from you. Our next release to get this issue in is in January though so you've got a bit of time. :)

Unfortunately, I'm not sure what could be the matter here but you could start investigating by:
- (from pocmo's suggestion) Looking at the git log for `BrowserFragment` - does anything look suspicious? Perhaps back out a changeset and see if it fixes the problem
- Bisecting: check out different changesets on master to find out which reproduce and which don't to narrow down which changeset could have caused the regression (this is less viable because the issue isn't consistently reproducible yet)
- Just reading the code! :)

Let us know if you need any assistance and don't forget you can find us at #focus on IRC generally during PST working hours.

nosark:
ok well i'll get to it and pose any questions in IRC(irc handle: codeasaurusHex).

mcomella:
Discussion regarding some tentative STR: https://mozilla.logbot.info/focus/20171201#c13944153

mcomella:
When I reproduce, I'm also noticing that the URL bar doesn't seem to be changing to the multi-color gradient – this is probably related.

pocmo:
> Our next release to get this issue in is in January (v4.1, see the calendar)

Depending on the fix and timing we should try to uplift to 4.0 though :)

npark-mozilla:
I am not sure whether it's a same bug or not, but on latest FocusWebViewRelease build, the progress bar is only shown covering about 5% of the horizontal length of the screen, then disappears.  I'm on Pixel 2 API 26.

nosark:
[...]

PR #1887, opened 2017-12-02
Closes Loading bar occassionally disappears during load after deleting data #1837

nosark:
Setting the default session.loading to true alleviates a race case that hits the else statement inside of session.getLoading().observe(this, new NonNullObserver<Boolean>() {} correcting the disappearing progressView and failing background transition that was occurring concurrently with the progressView issue. If this causes other issues I can find another look into other possible fixes :)


mcomella:
Thanks for the PR and the research! I'll have time to take a look today or tomorrow.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -235,8 +235,7 @@ public void onChanged(@Nullable String url) {
             @Override
             public void onValueChanged(@NonNull Boolean loading) {
                 if (loading) {
-                    backgroundTransitionGroup.resetTransition();
-
+                     backgroundTransitionGroup.resetTransition();
```
mcomella: Please don't make changes to unrelated code – it makes it harder to track the changes in code over time with `git blame`!

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -344,7 +343,6 @@ private void initialiseCustomTabUi(final @NonNull View view) {
         eraseContainer.removeView(sessions);
 
         final int textColor;
-
```
mcomella: Same.

pocmo:
I'm also a bit concerned. While this might fix the issue here I'm not sure what the side effects are. In the past such patches have caused other bugs so we should take some time to investigate and test thoroughly.

Issue #1876, opened 2017-12-01
When no search engine is selected for deletion, tapping trash icon exits 'Remove Search Engines' dialog


npark-mozilla:
STR:
Go to search engine list in Settings menu
Tap 3-dot menu and select 'Remove'
Tap Trash can icon without selecting search engine checkbox

Expected:
Since no search engine is selected, either the trash can should be greyed out or have no action
Actual:
exits 'Remove search engines' dialog

Copied from #1836

pocmo:
UX from #1836:
> In this case we should disable the trash can until at least one item is selected.

liuche:
Also, just to be explicit because we've mentioned "short on time" here and there, but this is a `release-blocker` and we want to uplift this patch to ship with the 4.0 release. I think the pace that we've been reviewing PR #1898 is good, and I expect this last round of changes will fix everything and allow us to land this Dec 6 on schedule. But on the off-chance that this gets more complicated pushes beyond Dec 6, either @mcomella or I will make any outstanding changes in a commit on top so we can get this into 4.0.

Again, I think @layely is doing great on this patch and I'm pretty confident he will have a finished patch that's ready land in time for Dec 6, but just to document this in the issue.

liuche:
@npark-mozilla should be in the next master - if you can verify this I'll uplift it! Also, #1915 may be a dupe, but filed before this was finished and merged.

npark-mozilla:
👍

PR #1898, opened 2017-12-04
Disable the trash can until at least one engine  is selected for deletion (#1876)

layely:
fix for #1876
Also the trash is greyed when disabled


app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
```diff
@@ -48,4 +49,36 @@ protected void updateDefaultItem(CompoundButton defaultButton) {
         }
         return engineIdSet;
     }
+
+    // Whenever an engine is checked or unchecked, we adapt the MenuItem state (enable or disable) given in param
+    // Though some conditions are added just for optimisation
+    public void bindEnginesCheckboxesTo(final MenuItem menuDeleteItem, final int alpha_default, final int alpha_greyed) {
```
sputnikci: \[PMD\] INFO: The method 'bindEnginesCheckboxesTo' has a Cyclomatic Complexity of 7\.

app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
```diff
@@ -48,4 +49,36 @@ protected void updateDefaultItem(CompoundButton defaultButton) {
...
+        for (int i = 0; i < searchEngineGroup.getChildCount(); i++) {
+            final CompoundButton engineButton = (CompoundButton) searchEngineGroup.getChildAt(i);
+            engineButton.setOnCheckedChangeListener(new CompoundButton.OnCheckedChangeListener() {
+                @Override
+                public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
```
sputnikci: \[PMD\] INFO: The method 'onCheckedChanged' has a Cyclomatic Complexity of 5\.

app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
```diff
@@ -48,4 +49,36 @@ protected void updateDefaultItem(CompoundButton defaultButton) {
...
+                    if (buttonView.isChecked() || atLeastOneEngineChecked()) {
```
liuche: Nit: this first buttonView.isChecked can just be the isChecked boolean in the arguments.

app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
liuche: Nit: I think the "To" naming convention is more of a Swift/iOS pattern, so I'd remove the "to", but up to you.

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -37,6 +37,8 @@
 
 public class SettingsFragment extends PreferenceFragment implements SharedPreferences.OnSharedPreferenceChangeListener {
     public static final String SETTINGS_SCREEN_NAME = "settingsScreenName";
+    public static final int ALPHA_DEFAULT = 255;
```
liuche: I would name these as "ALPHA_ENABLED" and "ALPHA_DISABLED".

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -104,13 +106,27 @@ public void onAttach(Context context) {
...
                 break;
             case REMOVE_ENGINES:
                 inflater.inflate(R.menu.menu_remove_search_engines, menu);
+
+                // We disable the trash icon when no engine is checked, enable it otherwise
+                getView().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        MenuItem menuDeleteItem = menu.findItem(R.id.menu_delete_items);
+                        menuDeleteItem.getIcon().mutate().setAlpha(SettingsFragment.ALPHA_GREY);
```
mcomella: You shouldn't need to modify the alpha directly: this menu item uses `icon="@drawable/ic_delete"`, which has `fillColor=@drawable/state_colors` which is:

```xml
<selector xmlns:android="http://schemas.android.com/apk/res/android">
    <item android:color="#4Cffffff" android:state_enabled="false" />
    <item android:color="#ffffffff" />
</selector>
```

That is when the state is disabled, we get the disabled color. When the state matches no other clauses (i.e. is enabled), we get the enabled color.

This line, and the related lines, can change to `setEnabled(...)`

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -104,13 +106,27 @@ public void onAttach(Context context) {
...
+                        final MultiselectSearchEngineListPreference pref = (MultiselectSearchEngineListPreference) getPreferenceScreen()
+                                .findPreference(getResources()
+                                        .getString(R.string.pref_key_multiselect_search_engine_list));
+
+                        pref.bindEnginesCheckboxesTo(menuDeleteItem, SettingsFragment.ALPHA_DEFAULT, SettingsFragment.ALPHA_GREY);
```
mcomella: This works and we're pressed for time, so I'd take it but I'll defer to liuche.

That being said, this breaks best practices on view hierarchies: views shouldn't know about their parent views implementation. In this case, we're telling the child view (`MultiselectSearchEngineListPreference`) about `menuDeleteItem` from its parent view. Instead, best practices would dictate:
1. We add a high-level listener to the child view (`onCheckboxChecked`) and the the child view notifies the parent when an event occurs
1. (less ideal but less boilerplate) The parent reaches into the child implementation and adds its own listeners (this is essentially what I'm trying to stop with this comment, but in the other direction)

Why does this matter? The more dependencies there are between different components of the code, the more difficult it becomes to change it. For example:
- If we go with 1), when we change a view, we largely don't have to consider how any views interact with us - we're a black box!
- If 2), you have to consider how the parent interacts with this view's state (one dependency, in one direction down the tree)
- If you let the child know about the parent and you're changing the parent, you have to consider the above and how its children might be touching the parent's internals (two dependencies, but in both directions on the view tree!)

app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
layely: Oh I never tried Swift, I wanted just to make it clear that we are binding the engines to the trash menuItem in param.

What about the name `bindEnginesCheckboxesToMenuItem` ?

app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
layely: OMG didn't even noticed that param :) . Thank you

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
layely: Agree 👍

mcomella: @layely Please read https://github.com/mozilla-mobile/focus-android/pull/1898#discussion_r154808808 - I don't think setting the alpha directly is necessary and it's a bit more error prone than `setEnabled`.

app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
liuche: That sounds good to me! Though to be specific, probably Engine singular, since each checkbox has one engine.

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
liuche: I actually thought that `setEnabled` by itself doesn't work - @layely have you tried the simpler version @mcomella  suggests, and does it work?

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
liuche: These are good tips! In general, I agree with this - but sometimes, the APIs don't work as one would expect. In those situations, it's also helpful to include a short comment explaining something like "colorStateLists don't work with vector drawables" (if that is in fact the case!) so other people reading the code will know why the more complicated, indirect version was used.

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
layely: 👍  It seems not working though

layely:
Hi @pocmo , I was sending the changes I made while you were commenting. Let me look into it quickly

layely:
I second version of the commit sent to take into account @pocmo  and @mcomella suggestions.
@liuche what do you think about this one ?

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -104,13 +106,30 @@ public void onAttach(Context context) {
     }
 
     @Override
-    public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
+    public void onCreateOptionsMenu(final Menu menu, MenuInflater inflater) {
```
sputnikci: \[PMD\] INFO: The method 'onCreateOptionsMenu' has a Cyclomatic Complexity of 5\.

mcomella:
FYI: Buddybuild probably won't build this latest push but I enabled BBs for layely's branches in the future. :)

app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
```diff
@@ -48,4 +52,36 @@ protected void updateDefaultItem(CompoundButton defaultButton) {
         }
         return engineIdSet;
     }
+
+    // Whenever an engine is checked or unchecked, we notify the menu
+    public void bindEngineCheckboxesToMenu() {
```
liuche: Maybe rename this to bindOnEngineCheckboxesChanged now that we're not binding to the menu any more, to remove confusion.

app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
```diff
@@ -48,4 +52,36 @@ protected void updateDefaultItem(CompoundButton defaultButton) {
...
+        for (int i = 0; i < searchEngineGroup.getChildCount(); i++) {
+            final CompoundButton engineButton = (CompoundButton) searchEngineGroup.getChildAt(i);
+            engineButton.setOnCheckedChangeListener(new CompoundButton.OnCheckedChangeListener() {
+                @Override
+                public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
+                    ((Activity) getContext()).invalidateOptionsMenu();
```
liuche: Nice removing the binding to the menu item - however, I'm now seeing a flicker on every checkbox click because we're invalidating every time the items are clicked.

We should only invalidate in two cases:
1) if the button is newly checked, and it is the ONLY item checked
2) if the button is newly unchecked, and there are NO items checked

so something like:
```
if (isChecked && numItemsChecked() == 1) || (!isChecked && numItemsChecked() == 0) {
    invalidateOptionsMenu()
}
```
(sorry for writing the code already! but I didn't want to make you do something impossible x_x and then the code just happened)

app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.java
```diff
@@ -48,4 +52,36 @@ protected void updateDefaultItem(CompoundButton defaultButton) {
...
+            }
+        }
+        return false;
+    }
+
+    private void setSearchEnginesBindingToMenu(boolean binding) {
+        this.searchEnginesBindingToMenu = binding;
+    }
+
+    public boolean isSearchEnginesBindingToMenu() {
```
liuche: Is this so that we don't repeatedly bind an onCheckedChangeListener? Since this doesn't actually depend on the MenuItem anymore (it just sends an "invalidate" message) we can just move this into the construction of the Preference.

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -104,13 +106,30 @@ public void onAttach(Context context) {
...
+                        final MultiselectSearchEngineListPreference pref = (MultiselectSearchEngineListPreference) getPreferenceScreen()
+                                .findPreference(getResources()
+                                        .getString(R.string.pref_key_multiselect_search_engine_list));
+
+                        if (!pref.isSearchEnginesBindingToMenu()) {
+                            pref.bindEngineCheckboxesToMenu();
+                        }
+
+                        MenuItem menuDeleteItem = menu.findItem(R.id.menu_delete_items);
+                        SettingsFragment.this.setMenuItemEnabled(menuDeleteItem, pref.atLeastOneEngineChecked());
```
liuche: You can move this "setMenuItemEnabled" line into onPrepareMenuOptions - that way, you don't need to post it as a Runnable on the view, because in onPrepareMenuOptions, the menu items are already created and you can simply call methods on the menuItem itself.

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -104,13 +106,30 @@ public void onAttach(Context context) {
...
+                // We disable the trash icon when no engine is checked, enable it otherwise
+                getView().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        final MultiselectSearchEngineListPreference pref = (MultiselectSearchEngineListPreference) getPreferenceScreen()
+                                .findPreference(getResources()
+                                        .getString(R.string.pref_key_multiselect_search_engine_list));
+
+                        if (!pref.isSearchEnginesBindingToMenu()) {
+                            pref.bindEngineCheckboxesToMenu();
```
liuche: Since this no longer has any dependency on the Menu, and we only want to call it once, we can move this to the MultiselectPreference itself :)

app/src/main/res/menu/menu_remove_search_engines.xml
```diff
@@ -8,5 +8,6 @@
     <item android:id="@+id/menu_delete_items"
         android:title="@string/preference_search_remove"
         android:icon="@drawable/ic_delete"
-        app:showAsAction="always"/>
+        app:showAsAction="always"
+        android:enabled="false" />
```
liuche: Actually, if you just remove this "false" then the flicker goes away. What's happening is that the `invalidate` is recreating the menu each time, and setting it false, and then the code is re-enabling it, so we're seeing a flicker.

liuche:
@layely just a heads up that we'd like to get this fix into 4.0 by Dec 6: https://github.com/mozilla-mobile/focus-android/issues/1876#issuecomment-349459812

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
layely: Here the problem is not the menuItem but the preference view. If we call the method `atLeastOneEngineChecked` outside the runnable, the activity will crash cause engines checkboxes would be not inflated yet.

app/src/main/res/menu/menu_remove_search_engines.xml
layely: +1

layely:
Thanks @liuche I made some changes to match your suggestions. But I think the Runnable can't be avoided [please read my feedback about it ](#discussion_r155238410)

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -118,13 +120,25 @@ public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
     }
 
     @Override
-    public void onPrepareOptionsMenu(Menu menu) {
+    public void onPrepareOptionsMenu(final Menu menu) {
```
sputnikci: \[PMD\] INFO: The method 'onPrepareOptionsMenu' has a Cyclomatic Complexity of 5\.

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -120,13 +120,25 @@ public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
...
```
sputnikci: \[PMD\] INFO: The method 'onPrepareOptionsMenu' has a Cyclomatic Complexity of 5\.

liuche:
Needed on `break` statement in the switch statement caught by checkstyle, added and squashed in #1938.

Merged to master! Thanks for your help, @layely!

Issue #1888, opened 2017-12-02
'Custom URLs' menu item inside URL Autocomplete menu needs a unique identifier for UI Test


npark-mozilla:
In some locales, like th, zh-cn, zh-tw, the translated text for 'Custom URL' and 'Custom URLs' are identical.  This causes screenshot test to fail since it locates more than one element.  For now the workaround is to use the parent index, but ideally it should come with a unique identifier.

pocmo:
Can you select additionally by type? One of them should be a Switch, I guess?

npark-mozilla:
Actually from the uiautomatorviewer app, they seem identical, since it's not a toggle switch- it's a menu item.

<img width="712" alt="screen shot 2017-12-02 at 1 09 41 pm" src="https://user-images.githubusercontent.com/6930774/33518383-16410a26-d762-11e7-8584-9f4cbc76eaea.png">

pocmo:
Hmm. I see. Those settings views are created by Android. So I would like to avoid that we need to write complex code to modify the default behavior.
I still wonder if we could select the element more reliably? For now maybe just selecting "the last one"?

npark-mozilla:
In UiAutomator, you can you UiSelector.instance(), but with Espresso, I think the only option is to use the index value of the parent.  Luckily in this case, the index values were different.

pocmo:
Espresso has a PreferenceMatchers helper class. Especially withKey() looks interesting:
https://developer.android.com/reference/android/support/test/espresso/matcher/PreferenceMatchers.html

I'll try and let you know if it works.. :)

pocmo:
@npark-mozilla  This seems to work:

```java
        final String key = InstrumentationRegistry
                .getTargetContext()
                .getString(R.string.pref_key_screen_custom_domains);

        onData(withKey(key))
                .perform(click());
```

npark-mozilla:
#1902 created to fix this issue

PR #1902, opened 2017-12-04
Use withKeys to locate an element for ScreenshotTest

npark-mozilla:


app/src/androidTest/java/org/mozilla/focus/activity/screenshots/SettingsScreenshots.java
```diff
@@ -133,8 +134,12 @@ public void takeScreenShotsOfSettings() throws Exception {
         onView(withText(getString(R.string.preference_autocomplete_custom_summary)))
                 .perform(click());
         /* Add custom URL */
-        onView(withText(getString(R.string.preference_autocomplete_subitem_customlist)))
+        final String key = InstrumentationRegistry
+                .getTargetContext()
+                .getString(R.string.pref_key_screen_custom_domains);
```
mcomella: nit: Why isn't `this.getString` usable anymore?

app/src/androidTest/java/org/mozilla/focus/activity/screenshots/SettingsScreenshots.java
```diff
@@ -133,8 +134,12 @@ public void takeScreenShotsOfSettings() throws Exception {
...
+        onData(withKey(key))
```
mcomella: Why do we need to change this away from `onView(withText(`? The PR doesn't provide any context.

`onData` says it's useful for AdapterViews but we're in a preference (which may actually use an AdapterView under the hood though).

npark-mozilla: @mcomella Oh, sorry about missing the background info on this. so for some locales, 'Custom URL' and 'Custom URLs' have identical text. so for those locales, the screenshot test fails if we only use the text to locate the identifier. #1888 is the bug that I raised for this.

PR #1903, opened 2017-12-04
Closes #1880: Provide error msg when user adds duplicate search engine

### LINKED ISSUE #1880: Incorrect error msg when a search engine with same name is being added
STR:
add a search engine with name 'Test'
add another search engine with sam name

Expected:
rejects with the error message saying "there is already a search engine with same name"
Actual:
There is only one error msg shown, which says "Check that search string matches example format" which is misleading in this case.

mcomella:
#1880.

This assumes Tif will just want to add a string, rather than change the existing behavior.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
+++ b/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
@@ -54,17 +54,17 @@ public class ManualAddSearchEngineSettingsFragment extends SettingsFragment {
                 final SharedPreferences sharedPreferences = getSearchEngineSharedPreferences();
                 boolean isSuccess = false;
                 if (TextUtils.isEmpty(engineName)) {
-                    Snackbar.make(rootView, R.string.search_add_error_empty_name, Snackbar.LENGTH_SHORT).show();
+                    showSnackbar(R.string.search_add_error_empty_name);
                 } else if (TextUtils.isEmpty(searchQuery)) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -71,12 +74,13 @@ public boolean onOptionsItemSelected(MenuItem item) {
         }
     }
 
-    private static boolean validateSearchFields(String engineName, String searchString, SharedPreferences sharedPreferences) {
-        if (sharedPreferences.getStringSet(SearchEngineManager.PREF_KEY_CUSTOM_SEARCH_ENGINES,
-                Collections.<String>emptySet()).contains(engineName)) {
-            return false;
-        }
+    private void showSnackbar(@StringRes final int errStrRes) {
```
liuche: Nice pulling out all the Snackbar creation into a single function.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -71,12 +74,13 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+        Snackbar.make(getView(), errStrRes, Snackbar.LENGTH_SHORT).show();
+    }
```
pocmo: Unrelated to this change. But note that there's also ViewUtils.showBrandedSnackbar() that is used in other places.

liuche:
We've made so many changes to the error message handling that this PR should probably be completely redone, @mcomella.

PR #1912, opened 2017-12-05
Closes #1906. Update trashcan to reflect disabled state.

### LINKED ISSUE #1906: In API 23, the trashcan is not greyed out when nothing is selected.
From #1836.  

This can be reproduced on API 23 simulator.  Go to Settings -> URL Autocomplete -> Custom URLs, and select 'Delete' from More Options button.

liuche:
This uses the constants from @layely's patch in #1898 .

### DIFF
--- a/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
@@ -133,8 +133,10 @@ open class AutocompleteListFragment : Fragment() {
         val removeItem = menu?.findItem(R.id.remove)
 
         removeItem?.isVisible = isSelectionMode() || domainList.adapter.itemCount > 1
-        removeItem?.isEnabled = !isSelectionMode()
+        val isEnabled = !isSelectionMode()
                 || (domainList.adapter as DomainListAdapter).selection().isNotEmpty()
+        removeItem?.isEnabled = isEnabled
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
```diff
@@ -133,8 +133,10 @@ open class AutocompleteListFragment : Fragment() {
         val removeItem = menu?.findItem(R.id.remove)
 
         removeItem?.isVisible = isSelectionMode() || domainList.adapter.itemCount > 1
-        removeItem?.isEnabled = !isSelectionMode()
+        val isEnabled = !isSelectionMode()
                 || (domainList.adapter as DomainListAdapter).selection().isNotEmpty()
+        removeItem?.isEnabled = isEnabled
+        removeItem?.icon?.alpha = if (isEnabled) SettingsFragment.ALPHA_ENABLED else SettingsFragment.ALPHA_DISABLED
```
mcomella: nit: If it works for you, works for me so r+ but can we use [`android:tintList`](https://developer.android.com/reference/android/view/MenuItem.html#attr_android:iconTint) instead?

liuche: After your suggestion I gave that a try (copied the state list into @color, used app:itemIconTint) and while that works for r25...it doesn't for r23.

I also tried setting the tint programmatically but at that point, since we're invalidating every time, it's not really worth it.

pocmo:
@liuche Thank you for taking this over.

* Can't we remove @drawable/state_colors from ic_delete now (and state_colors.xml)?

* The build failed on taskcluster with a style violation found by detekt. Was this fixed?

PR #1914, opened 2017-12-06
Closes #1875: Remove cookies on new session.

### LINKED ISSUE #1875: Closing app in "recent apps" screen does not remove cookies (and maybe more)
Android 8.1 Beta, Pixel XL.

STR:
* Got to twitter.com
* Login
* Go to "recent apps" screen
* Close Focus by swiping it away
* Open Focus again
* Go to twitter.com

Expected: You are not logged in anymore.

Actual: You are still logged in.

We handle this situation in SessionNotificationService.onTaskRemoved():
https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/session/SessionNotificationService.java#L81

mcomella:
#1875

This is the simplest fix that aims to avoid unwanted side effects: this means
there may be other hidden issues we're not fixing here. We should try to
comprehensively fix this bug in issue #1569.

A better fix with potential side effects would be forcing WebView.cleanup to
run on a new browser session, not just deleting cookies.

@liuche, @pocmo: if we review this soon, we could consider uplifting and fixing it for v4.0. CC @Sdaswani @npark-mozilla

### DIFF
--- a/app/src/webview/java/org/mozilla/focus/web/WebViewProvider.java
+++ b/app/src/webview/java/org/mozilla/focus/web/WebViewProvider.java
@@ -14,6 +14,7 @@ import android.support.annotation.VisibleForTesting;
 import android.text.TextUtils;
 import android.util.AttributeSet;
 import android.view.View;
+import android.webkit.CookieManager;
 import android.webkit.WebSettings;
 import android.webkit.WebStorage;
 import android.webkit.WebView;
@@ -48,6 +49,9 @@ public class WebViewProvider {
      * This function must be called before WebView.loadUrl to avoid erasing current session data.
      */
     public static void performNewBrowserSessionCleanup() {
+        // If the app is closed in certain ways, WebView.cleanup will not get called and we don't clear cookies.
+        CookieManager.getInstance().removeAllCookies(null);
+
         // We run this on the main thread to guarantee it occurs before loadUrl so we don't erase current session data.
         final StrictMode.ThreadPolicy oldPolicy = StrictMode.allowThreadDiskWrites();
 


mcomella:
> LGTM - But why not cleanup from SessionNotificationService.onTaskRemoved()?

I didn't consider it but I think it would work as well.

That being said, now that I'm thinking about it, this bug exhibits really weird behavior: cookies are persisting even though the files are permanently deleted and the process, with WebView, is dead – so where are the cookies being stored between runs? **Note to self: figure this out before landing.**

Maybe it's due to the cache, that we're no longer deleting the files of on session end (after #1492)?

This is making me realize that these small hack fixes to correct the WebView behavior are creating scary side effects and we need to prioritize the comprehensive fix: #1569. But at least they're giving us more information about what's important... :)

pocmo:
> cookies are persisting even though the files are permanently deleted and the process, with WebView, is dead – so where are the cookies being stored between runs? Note to self: figure this out before landing.

Can you try killing Chrome on the device? I recently noticed an odd behavior where some data remained in memory as long as the Chrome process was still alive. I wonder if this is the same here.

mcomella:
> this bug exhibits really weird behavior: cookies are persisting even though the files are permanently deleted and the process, with WebView, is dead – so where are the cookies being stored between runs?

It looks like swipe-to-close doesn't actually clear the process from memory and **if you forcibly remove Focus from memory, it won't remember your cookies**. For example, Focus won't store your cookies with this STR:
- Open Focus
- Go to a page that stores cookies, create cookies so they'll be noticeable when you come back (I logged into spotify)
- Swipe close on the recent app switcher
- Settings -> developer -> View running services -> View cached services

Unexpectedly, Firefox Focus is present here. If you force-stop Focus, it won't remember your cookies. Note that I also used `adb shell` to run `top` and `org.mozilla.focus.debug` appears in the list (0% CPU) after a swipe-to-close until we force stop the service.

I wonder if the service being cached is our NotificationService, while swipe-to-close would ordinarily evict an Activity from memory.

mcomella:
In any case, I don't think there's anything different to do here: the system is caching WebView data in memory and (I assume) we're not guaranteed to get any callbacks that'd allow us to clear it. Our best choice is to clear it on startup, which we do here. Ideally, we'd comprehensively clear the WebView data on starting a new session, rather than just these cookies, but we can resolve that later.

mcomella:
> > LGTM - But why not cleanup from SessionNotificationService.onTaskRemoved()?
> 
> I didn't consider it but I think it would work as well.

Sorry that I didn't reconsider using this approach (I forgot about it in my testing). That being said, I'll consider it for #1569.

PR #1921, opened 2017-12-06
Detekt: No thresholds

pocmo:
By default detekt has some thresholds defined. For example it won't fail the build if there are less than 5 warnings. I think we should see and deal with every warning (fix or suppress) and therefore removed all threshold in this PR.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
@@ -136,7 +136,10 @@ open class AutocompleteListFragment : Fragment() {
         val isEnabled = !isSelectionMode()
                 || (domainList.adapter as DomainListAdapter).selection().isNotEmpty()
         removeItem?.isEnabled = isEnabled
-        removeItem?.icon?.mutate()?.alpha = if (isEnabled) SettingsFragment.ALPHA_ENABLED else SettingsFragment.ALPHA_DISABLED
+        removeItem?.icon?.mutate()?.alpha = if (isEnabled)
+            SettingsFragment.ALPHA_ENABLED
+        else
+            SettingsFragment.ALPHA_DISABLED
     }
 
     override fun onOptionsItemSelected(item: MenuItem?): Boolean = when (item?.itemId) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.kt
```diff
@@ -38,11 +38,11 @@ class IconGenerator {
          * on top of a generic launcher icon shape that we provide.
          */
         @JvmStatic
```
mcomella: nit: I think we could remove this annotation too if this fn is private

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
```diff
@@ -136,7 +136,10 @@ open class AutocompleteListFragment : Fragment() {
         val isEnabled = !isSelectionMode()
                 || (domainList.adapter as DomainListAdapter).selection().isNotEmpty()
         removeItem?.isEnabled = isEnabled
-        removeItem?.icon?.mutate()?.alpha = if (isEnabled) SettingsFragment.ALPHA_ENABLED else SettingsFragment.ALPHA_DISABLED
+        removeItem?.icon?.mutate()?.alpha = if (isEnabled)
```
mcomella: nit: I dislike these brief, 1 line if statements becoming 4 lines. :\ Could we try:
```kotlin
removeItem?.icon?.mutate()?.alpha =
    if (isEnabled) SettingsFragment.ALPHA_ENABLED else SettingsFragment.ALPHA_DISABLED
```

or

```kotlin
removeItem?.icon?.mutate()?.alpha = if (isEnabled) SettingsFragment.ALPHA_ENABLED
    else SettingsFragment.ALPHA_DISABLED
```

or similar?

mcomella: Intellij automatically formats the second one so that `removeItem` and `else` are in the same vertical line, which is awful so I don't think it's conventional.

I went with the former.

mcomella:
Merged with https://github.com/mozilla-mobile/focus-android/commit/11dceb95e1efe9711e60733984f83dfcf116bf46.

mcomella:
Note that I added my review nits on top without review.

PR #1929, opened 2017-12-06
Closes #1886: Use constant for detekt version in build.gradle

### LINKED ISSUE #1886: Use constant for detekt version in build.gradle
Right now, we specify it individually: e.g. the last time we updated the version is https://github.com/mozilla-mobile/focus-android/pull/1878/files

Since the values are duplicated, we should use a constant, like `kotlin_version`.

siddware:
Added comments explaining the rationale for persisting with the duplicate values for Detekt versions

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,15 +23,15 @@ buildscript {
 }
 
 plugins {
-    //The version number is duplicated in the detekt block. We are unable to use constants due to
-    //limitations of the gradle plugin. We've decided to stick with duplication for now since the
-    //other methods involve using undocumented APIs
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -23,10 +23,15 @@ buildscript {
 }
 
 plugins {
+    //The version number is duplicated in the detekt block. We are unable to use constants due to
```
mcomella: nit: There should be a space between the comment markers and the comments, e.g. `// The version number...`

siddware: Done :)

PR #1932, opened 2017-12-06
 Keyboard stays up after adding new search engine or autocomplete domain

### LINKED ISSUE #1923: Keyboard stays up after adding new search engine or autocomplete domain
Not critical but a nice thing that a contributor could fix.

STR:
* Go to Settings -> Search -> Add another search engine
* Enter some random search engine URL and name.
* Press "Save"

Expected: Hide keyboard and return to the list of search engines.

Actual: Return to the list of search engines but keyboard is still up.

The same issue can be reproduced when adding a new autocomplete domain.

Jacquester:
Closes #1923 
This should hide the keyboard when closing the fragments: ManualAddSearchEngineSettingsFragment and AutocompleteAddFragment

### DIFF
--- a/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
@@ -8,6 +8,7 @@ import android.app.Fragment
 import android.content.Context
 import android.os.Bundle
 import android.view.*
+import android.view.inputmethod.InputMethodManager
 import org.mozilla.focus.R
 import org.mozilla.focus.settings.SettingsFragment
 
@@ -34,6 +35,16 @@ class AutocompleteAddFragment : Fragment() {
         val updater = activity as SettingsFragment.ActionBarUpdater
         updater.updateTitle(R.string.preference_autocomplete_title_add)
         updater.updateIcon(R.drawable.ic_close)
+    }
+
+    override fun onPause() {
+        super.onPause()
+        val view = activity.currentFocus
+        if (view != null) {
+            val imm = activity
+                    .getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
+            imm.hideSoftInputFromWindow(view.windowToken, 0)
+        }
     }
 
     override fun onCreateView(inflater: LayoutInflater?, container: ViewGroup?, savedInstanceState: Bundle?): View =
--- a/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
@@ -8,7 +8,6 @@ import android.app.Fragment
 import android.content.Context
 import android.os.Bundle
 import android.view.*
-import android.view.inputmethod.InputMethodManager
 import org.mozilla.focus.R
 import org.mozilla.focus.settings.SettingsFragment
 
@@ -39,12 +38,7 @@ class AutocompleteAddFragment : Fragment() {
 
     override fun onPause() {
         super.onPause()
-        val view = activity.currentFocus
-        if (view != null) {
-            val imm = activity
-                    .getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
-            imm.hideSoftInputFromWindow(view.windowToken, 0)
-        }
+        ViewUtils.hideKeyboard(activity.currentFocus)
     }
 
     override fun onCreateView(inflater: LayoutInflater?, container: ViewGroup?, savedInstanceState: Bundle?): View =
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
```diff
@@ -36,6 +37,16 @@ class AutocompleteAddFragment : Fragment() {
...
     }
 
+    override fun onPause() {
+        super.onPause()
+        val view = activity.currentFocus
+        if (view != null) {
+            val imm = activity
+                    .getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
+            imm.hideSoftInputFromWindow(view.windowToken, 0)
+        }
```
pocmo: Please use our helper here: ViewUtils.hideKeyboard()

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
```diff
@@ -36,6 +37,16 @@ class AutocompleteAddFragment : Fragment() {
         updater.updateIcon(R.drawable.ic_close)
     }
 
+    override fun onPause() {
```
pocmo: I wonder why you chose onPause() and not for example saveDomainAndClose(). Doesn't using onPause() mean we will also dismiss the keyboard if the user leaves the app and returns back to it?

Jacquester: Hi @pocmo 
I used onPause() to ensure the keyboard is hidden even when the user decides not to add a domain or search engine. 
Example:

* go to Settings > URL Autocomplete > Custom URLs > Add custom URL
* Enter a random URL (not required)
* Press "Cancel" or X

If the code to hide the keyboard is located in saveDomainAndClose() the keyboard will remain open until closed by the user.

I tested the changes I made by closing and re-opening the app and the code in onPause() does not seem to have any effect although I understand that it is an unnecessary call when the app closes

pocmo: I see. This makes sense. Thank you for testing this.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -71,6 +72,12 @@ public boolean onOptionsItemSelected(MenuItem item) {
         }
     }
 
+    @Override
+    public void onPause() {
+        super.onPause();
+        ViewUtils.hideKeyboard(getActivity().getCurrentFocus());
```
pocmo: Is there any chance getCurrentFocus() could return null here?

Jacquester: From the testing I've done getCurrentFocus() never returned null. The tests I tried was:

* Rotating the screen whilest ManualAddSearchEngineSettingsFragment was being displayed
* Closing the app and reopening after some time
* Closing the app, rotating the screen then returning to the app
* Clicking "cancel" when the fragment was being displayed
* Clicking "save" when the fragment was being displayed

To be safe I'll wrap it within an if-statement ensuring the view returned is not null

pocmo: Great. That's definitely more than I tested.. :)

Jacquester: Now that I think of it let's change the code in ViewUtils.hideKeyboard() to check that the View passed to it is not null, thus allowing us to call hideKeyboard() without having to check if the view passed is null every time

pocmo: Yep, sounds good!

app/src/main/java/org/mozilla/focus/utils/ViewUtils.java
```diff
@@ -104,12 +104,14 @@ public static void showKeyboard(View view) {
...
     public static void hideKeyboard(View view) {
-        InputMethodManager imm = (InputMethodManager) view.getContext().getSystemService(Context.INPUT_METHOD_SERVICE);
-        if (imm == null) {
-            return;
-        }
+        if (view != null) {
+            InputMethodManager imm = (InputMethodManager) view.getContext().getSystemService(Context.INPUT_METHOD_SERVICE);
+            if (imm == null) {
+                return;
+            }
```
pocmo: In our code we mostly prefer early returns (like with imm == null).

Jacquester: I'll just ammend this to the previous commit

Jacquester:
Commit has been renamed and the early return has been added :)

PR #1937, opened 2017-12-06
Use new URLs for downloading Search Plugins

### LINKED ISSUE #1724: Rewrite scrape_plugins.py
Initially we imported this script from the Focus for iOS repository. The code is still pointing to the aurora repository and a file structure that doesn't exist anymore.

mozilla-central paths:

list.json:
https://hg.mozilla.org/mozilla-central/raw-file/default/mobile/locales/search/list.json

Search plugins:
https://hg.mozilla.org/mozilla-central/raw-file/default/mobile/locales/searchplugins

mastizada:
Fix #1724 - Rewrite scrape_plugins.py


mastizada:
There is `regionOverrides` in list.json file, should we also change google to google-nocode for those regions?

tools/search/scrape_plugins.py
```diff
@@ -44,38 +38,43 @@ def main():
         for region in regions:
             if region == "default":
                 code = locale
+            elif region == "experimental-hidden":
```
mastizada: seen that in en-us

tools/search/scrape_plugins.py
```diff
@@ -44,38 +38,43 @@ def main():
...
+                continue
             else:
                 language = locale.split("-")[0]
                 code = ("%s-%s" % (language, region))
 
             print("adding %s..." % code)
 
             visibleEngines = regions[region]["visibleDefaultEngines"]
-            downloadEngines(code, L10nScraper(locale), visibleEngines)
+            downloadEngines(code, Scraper(locale), visibleEngines)
```
mastizada: we can also remove locale variable from Scraper, it is not used

tools/search/scrape_plugins.py
```diff
@@ -88,15 +87,15 @@ def downloadEngines(locale, scraper, engines):
             break
 
     for engine in engines:
-        file = engine + ".xml"
-        path = os.path.join(directory, file)
-        downloadedFile = scraper.getFile(file)
-        if downloadedFile == None:
-            print("  skipping: %s..." % file)
+        engine_file = engine + ".xml"
```
mastizada: `file` is built-in in python

pocmo:
@mastizada Thank you for writing this!

For your questions I'd like to defer to @mkaply:
* Should we just ignore `experimental-hidden` in Focus?
* How should we deal with `regionOverrides`?

@mastizada Can you run the script and create a readable diff between the result and the search_configuration.json file in the repo? This would help me and @mkaply to verify that the output is what we expect.

mkaply:
Yes, you can ignore experimental-hidden.

The way regionOverride should be used is that for a given region, you should substitute that engine. So in RU, for instance, you should use google-nocodes instead of google.

Note that there is a patch still coming for central that will affect this somewhat - https://bugzilla.mozilla.org/show_bug.cgi?id=1417678

Also looking at my Firefox IOS version of the patch to use list.json might make things a little clearer with regards to region overrides.

https://github.com/mozilla-mobile/firefox-ios/pull/3529

mcomella:
Hey @mastizada – thanks for the pull request! We really appreciate it! Sorry this review has been taking so long – we're having a colocated team meeting and our schedule has been a bit crazy. I'll get to this review either later today or tomorrow morning.

mastizada:
@mkaply There is problem with region and locale in list.json:
For example, there is ru locale, but there is no RU region (only default). For US region, in en-US there is US region and default, override for US region is google-2018 for google, US region has google-nocode in visibleDefaultEngines and google in default. How this regionOverride should be handled?

Changes to list.json will be handled as long as structure of the file will be same (checked diff, the only changes to file is the name of the plugin). Script loads the list.json file from url (https://hg.mozilla.org/mozilla-central/raw-file/default/mobile/locales/search/list.json) each time.

Will upload search_configuration.json file to gist after implementing regionOverride rules.

mcomella:
^ Awaiting mkaply's response and an additional commit before reviewing (I spoke to mastizada about this on IRC).

Correct me if I'm mistaken @mastizada!

mkaply:
Please see the latest version of list.json that just landed on tree for Fennec:

https://searchfox.org/mozilla-central/source/mobile/locales/search/list.json

There doesn't need to be an RU region because regionOverrides handles that. ru always uses the default list except when a regionOverride is specified, and then in that region, it uses the value in the override. So in ru-RU, google-nocodes is used, in ru-DE, google is used. The same is true for the US. in en-DE, google is used, in en-US, google-2018 is used.

mastizada:
@pocmo @mkaply I have analyzed both iOS (`SearchEnginesJSON.swift`) and Android (`SearchEngineManager.java`) codes, also currently used output (`search_configuration.json` for Android, `SearchEngines.plist` for iOS). As I understood, whenever region is, for example, `RU` app should use override value, previously those values were in `list.json` file as language_code-region. In current version, possible outcome can be list of locales and regions if provided as it currently is, and list of region overrides, and both Android and iOS code (java and swift) should also check for those overrides as all locales should use these override values.

Example:

```
{"default": [...]},
{"en": [...]},
{"en-CA": [...]},
{"overrides": {
    "US": {
        "google": "google-2018"
    }
}
```

Is it correct way to solve?

Note: Currently regions values of `regionOverrides` are dictionaries, should we assume they will never be array lists?

mcomella:
@mastizada We've been on vacation for the holidays – sorry for the delay! We're back now so @mkaply should get back to you soon!

mkaply:
I'm confused why you're not using list.json directly instead of creating a new file?

> @pocmo @mkaply I have analyzed both iOS (SearchEnginesJSON.swift) and Android (SearchEngineManager.java) codes, also currently used output (search_configuration.json for Android, SearchEngines.plist for iOS). As I understood, whenever region is, for example, RU app should use override value, previously those values were in list.json file as language_code-region. In current version, possible outcome can be list of locales and regions if provided as it currently is, and list of region overrides, and both Android and iOS code (java and swift) should also check for those overrides as all locales should use these override values.

That is correct. regionOverrides replace a value in the list (versus having to have a new list for every language/locale).

So overrides says "Any time you're in this region, use A instead of B"

mastizada:
@mkaply idea in #1724 is to use the new list.json file from mozilla-mobile instead of local list.json file. We can also just change url for downloading search engines and manually maintain local list.json file.

mkaply:
So do you have the answers you need for this bug?

Issue #1936, opened 2017-12-06
Add additional validation (just making a network request?) to add search engine


mcomella:

Sdaswani:
Yes @mcomella let's look for a 200. Do what is _reasonable_ about disabling the 'Save' button / UI while the network request is in flight and time limit it (e.g., max 5 seconds waiting on the request).

liuche:
Going to switch to #1939 because it's a slight dependency.

pocmo:
Are you really sure that's a reason to block the release? This looks like we are rushing something that has a bunch of side effects, e.g.:

* Redirects!
* Search engines do fancy things based on the client accesing them (user agent etc.) - They don't want you to scrape search results.

Sdaswani:
@pocmo both of those are real risks, but I'd rather they are false negatives (i.e., don't report back 200) blocking the experience rather than leading to the white screen of nothingness (which makes Focus look totally broken).

pocmo:
Agreed. Although we are not fixing the random redirect to a white about:blank here.

Sdaswani:
@pocmo wouldn't we get a redirect at first and then say it isn't valid? So not sure how we would redirect to a random white screen. But we realize there is a risk that we get a 200 to a page the user doesn't expect (i.e., non search page).

liuche:
Talked to @antlam about the UX, and the outcome is:
1) User fills out all the fields, but uses something like imdb.com/%s (which is incorrect)
2) Click "Save"
3) Show a spinner (for feedback)
4) If not successful, display the error message "Server not found" under the search string entry (this string is already shipped with Focus). (If successful, just add search engine and show confirmation)

Sdaswani:
I like the spinner and the error message. The error message captures the 200 check and if the device doesn't have a connection.

Sdaswani:
Cleared by security

liuche:
@npark-mozilla I'm especially interested in how this tests on oreo - I thought I had an infinite timeout on the oreo emulator, but didn't have a physical device.

npark-mozilla:
It seems to work well on actual device too - I can't make it fail.  It does complain if there is a suffix after %s like %s&where=aaa but this is probably expected.

mcomella:
>  It does complain if there is a suffix after %s like %s&where=aaa but this is probably expected.

#1960

Issue #1660, opened 2017-11-07
Save newly added search engines


liuche:
Meta: #961 

#1617 added the screen to create new search engines, this issue is for actually saving them.

This bug should validate the search string, and then save the search engines in opensearch format, and then show them in the search engines page.

liuche:
This is part 2, and only handles saving search engines.

I think we'll need to add some more strings because our error messages are not satisfactory/complete ( #1567 ) and there are still more parts, see #961 for meta and current progress.

@npark-mozilla alternatively I can flag you when all of #961 is done, what would you prefer?

npark-mozilla:
@liuche i think current approach is good, if it becomes too cumbersome, let me know.

PR #1942, opened 2017-12-07
Use Material Design errors in "Add search engine"

liuche:
Thought this was needed for #1936 , because of the way "Save" should behave according to @antlam.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -81,7 +81,7 @@
         <activity android:name=".activity.SettingsActivity"
             android:theme="@style/SettingsTheme"
             android:configChanges="locale"
-            android:windowSoftInputMode="adjustResize" />
+            android:windowSoftInputMode="adjustPan"/>
 
         <activity android:name=".activity.InfoActivity"
                   android:theme="@style/InfoTheme"/>
--- a/app/src/main/java/org/mozilla/focus/search/ManualAddSearchEnginePreference.java
+++ b/app/src/main/java/org/mozilla/focus/search/ManualAddSearchEnginePreference.java
@@ -1,0 +1,92 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.search;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+import android.preference.Preference;
+import android.support.design.widget.TextInputLayout;
+import android.text.Editable;
+import android.text.TextUtils;
+import android.text.TextWatcher;
+import android.util.AttributeSet;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.EditText;
+
+import org.mozilla.focus.R;
+import org.mozilla.focus.utils.UrlUtils;
+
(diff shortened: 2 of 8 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -81,7 +81,7 @@
         <activity android:name=".activity.SettingsActivity"
             android:theme="@style/SettingsTheme"
             android:configChanges="locale"
-            android:windowSoftInputMode="adjustResize" />
+            android:windowSoftInputMode="adjustPan"/>
```
pocmo: Seeing that this was just added for #1660 - Was this purely for the Snackbars?

app/src/main/java/org/mozilla/focus/search/ManualAddSearchEnginePreference.java
```diff
@@ -0,0 +1,88 @@
+package org.mozilla.focus.search;
```
pocmo: nit: License header.

app/src/main/java/org/mozilla/focus/search/ManualAddSearchEnginePreference.java
```diff
@@ -0,0 +1,88 @@
...
+import android.widget.EditText;
+
+import org.mozilla.focus.R;
+import org.mozilla.focus.utils.UrlUtils;
+
+public class ManualAddSearchEnginePreference extends Preference {
+    EditText engineNameEditText;
+    EditText searchQueryEditText;
+    TextInputLayout engineNameErrorLayout;
+    TextInputLayout searchQueryErrorLayout;
```
pocmo: Can they be ```private```?

app/src/main/java/org/mozilla/focus/search/ManualAddSearchEnginePreference.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    private TextWatcher buildTextWatcherForErrorLayout(final TextInputLayout errorLayout) {
+        return new TextWatcher() {
+            @Override
+            public void beforeTextChanged(CharSequence charSequence, int i, int i1, int i2) {}
+
+            @Override
+            public void onTextChanged(CharSequence charSequence, int i, int i1, int i2) {
+                errorLayout.setError(null);
```
pocmo: Oh, is this needed? I've been calling setError() on an EditText somewhere and it seemed to clear automatically when the text was changed. Is this because the error is set on the TextInputLayout?

app/src/main/AndroidManifest.xml
liuche: I found this was a fix for the weird keyboard behavior - I assumed from the name and descriptions that adjustResize was the right flag but adjustPan fixed all the problems...

app/src/main/java/org/mozilla/focus/search/ManualAddSearchEnginePreference.java
liuche: The error state looks a little different when called with setError on the EditText - it's a "!" and a hovertext, and doesn't match the material design docs. I do think the APIs for EditText.setError are nicer, because the error autohides, so maybe there's an update to those APIs that removes the need to add a TextWatcher.

liuche:
Commented in #1923, but that patch creates conflicts with this patch that needs to be uplifted into release today (and might be fixed anyway by this patch) - will reland them after, if still an issue!

PR #1958, opened 2017-12-08
Closes #1936: Validate add search engine URL with request. 

### LINKED ISSUE #1936: Add additional validation (just making a network request?) to add search engine

mcomella:


app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -65,18 +82,101 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
-                    Snackbar.make(rootView, R.string.search_add_confirmation, Snackbar.LENGTH_SHORT).show();
-                    getFragmentManager().popBackStack();
+                if (isPartialSuccess) {
+                    // TODO: add spinner, disable button until async task completes.
+                    // Hide the keyboard because:
+                    // - It's awkward to show the keyboard while waiting for a response
+                    // - We want it hidden when we return to the previous screen (on success)
+                    // - An expanded keyboard hides the success snackbar
+                    ViewUtils.hideKeyboard(rootView);
+                    new ValidateSearchEngineAsyncTask(this, engineName, searchQuery).execute();
```
liuche: Can you cancel this task in the fragment if the fragment gets destroyed? This should be a short-running task, and requires user action to trigger, but it might still be weird if the user 1) rotates the device, and then 2) goes back to the search engines. Settings doesn't handle rotation configChanges, so probably want to have an onCancelled that *doesn't* add the search engine.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -65,18 +82,101 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+            TelemetryWrapper.saveCustomSearchEngineEvent(isValidSearchQuery);
+            return isValidSearchQuery;
+        }
+
+        @Override
+        protected void onPostExecute(final Boolean isValidSearchQuery) {
+            super.onPostExecute(isValidSearchQuery);
+
+            final ManualAddSearchEngineSettingsFragment that = thatWeakReference.get();
+            if (that == null) {
```
liuche: Oh I see, you handle the cancelled state here. Are Fragments always gc-ed quickly after they're destroyed? Might be better to also explicitly setCancelled in the Fragment's onDestroy.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -22,8 +26,21 @@
...
+import java.lang.ref.WeakReference;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
 
 public class ManualAddSearchEngineSettingsFragment extends SettingsFragment {
+    private static String LOGTAG = "ManualAddSearchEngine";
+
+    // Set so the user doesn't have to wait *too* long. It's used twice: once for connecting and once for reading.
+    private static int SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS = 4000;
```
liuche: Is 4 seconds too long?

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
liuche: Also, optionally check for network connectivity first.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -65,18 +82,101 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+        // TODO: we should share the code to substitute and normalize the search string (see SearchEngine.buildSearchUrl).
+        final String encodedTestQuery = Uri.encode("test");
+        final String normalizedSearchURLStr = UrlUtils.normalize(query);
+        final String searchURLStr = normalizedSearchURLStr.replaceAll("%s", encodedTestQuery);
+
+        final URL searchURL;
+        try {
+            searchURL = new URL(searchURLStr);
+        } catch (final MalformedURLException e) {
+            // Don't log exception to avoid leaking URL.
```
liuche: Good.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
mcomella: I didn't want to prevent users with poor internet connections from using this feature and tried to balance it with how long users will realistically wait.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
mcomella: Added a commit.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
mcomella: fwiw, onDestroy doesn't work – e.g. if the home button is pressed, the fragment is not destroyed. Note that FragmentManager.popBackStack (which we call after the async task finishes) throws if it's called when the app is in the background.

mcomella:
Note: antlam wants:
1. An overlay with the spinner
1. No spinner

Unfortunately, I ran out of time so we have neither option, but tbh, I think it's good enough.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -65,18 +112,141 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+
+        private void showServerError(final ManualAddSearchEngineSettingsFragment that) {
+            final ManualAddSearchEnginePreference pref = (ManualAddSearchEnginePreference) that.findPreference(
+                    that.getString(R.string.pref_key_manual_add_search_engine));
+            pref.setSearchQueryErrorText(that.getString(R.string.error_hostLookup_title));
+        }
+    }
+
+    @WorkerThread // makes network request.
+    private static boolean isValidSearchQueryURL(final String query) {
```
sputnikci: \[PMD\] INFO: The method 'isValidSearchQueryURL' has a Cyclomatic Complexity of 5\.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -65,18 +112,141 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+
+        } catch (final IOException e) {
+            // Don't log exception to avoid leaking URL.
+            Log.d(LOGTAG, "Failure to get response code from server: returning invalid search query");
+            return false;
+        } finally {
+            if (connection != null) {
+                try {
+                    connection.getInputStream().close(); // HttpURLConnection.getResponseCode opens the InputStream.
+                } catch (final IOException e) { } // Whatever.
```
sputnikci: \[PMD\] WARNING: Must handle exceptions

liuche:
One problem is a failure with IMDB, so I'm going to repush with a patch that 1) removes the ensureHTTPS and then 2) accepts redirects (anything <400), matching what Focus iOS does.

liuche:
Repushed as #1965 with the same patches.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -102,14 +102,24 @@ public boolean onOptionsItemSelected(MenuItem item) {
         }
     }
 
+    private void setUiIsValidatingAsync(final boolean isValidating, final MenuItem saveMenuItem) {
+        final ManualAddSearchEnginePreference pref =
+                (ManualAddSearchEnginePreference) findPreference(getString(R.string.pref_key_manual_add_search_engine));
+        pref.setProgressViewShown(isValidating);
+
+        saveMenuItem.setEnabled(!isValidating);
```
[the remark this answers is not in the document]
mcomella: I didn't see it in my testing; not sure why (maybe higher APIs use an iconTint that was added later?)

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
mcomella: logcat StrictMode yelled at me that the InputStream wasn't closed and so I assume disconnect doesn't do it.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -136,9 +164,13 @@ protected Boolean doInBackground(final Void... voids) {
         protected void onPostExecute(final Boolean isValidSearchQuery) {
             super.onPostExecute(isValidSearchQuery);
 
+            if (isCancelled()) {
```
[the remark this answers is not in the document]
mcomella: > Documentation says that a cancelled AsyncTask should never go to onPostExecute

Good to know.

> (though who knows if it can be called while in isCancelled! I'm going to ignore that case.)

I assume it can be, however:
- From the docs you pointed out, it sounds like the intention of cancelling is ending the long running background task, rather than stopping the (hopefully) short UI thread operation
- In this case, I call cancel from the UI thread so it can't happen inside onPostExecute.

> So basically calling AsyncTask.cancel only sets a flag, and you have to check the flag yourself. 

Not always – you can tell AsyncTask.cancel to interrupt the thread (which stops execution and throws an InterruptedException) and which I did here

> And at this point, since this call is only 4s long anyway...

I don't think there's a way to stop a blocking call anyway, beyond the interrupt method mentioned above.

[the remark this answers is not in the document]
pocmo: > Not always – you can tell AsyncTask.cancel to interrupt the thread (which stops execution and throws an InterruptedException) and which I did here

Although note that the thread is not stopped magically (with an InterruptedException). Code inside the Thread might need to check for the interrupted state to stop running _early_. I'm not sure if HttpUrlConnection actually does that (like for example sleep()).

https://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html

PR #1961, opened 2017-12-08
Update spacing between TextInputLayout.

liuche:
Following conversation with @antlam.

### DIFF
--- a/app/src/main/res/layout/preference_manual_add_search_engine.xml
+++ b/app/src/main/res/layout/preference_manual_add_search_engine.xml
@@ -13,7 +13,8 @@
     <android.support.design.widget.TextInputLayout
         android:id="@+id/edit_engine_name_layout"
         android:layout_width="match_parent"
-        android:layout_height="wrap_content">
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/preference_manual_add_search_engine.xml
```diff
@@ -13,7 +13,8 @@
     <android.support.design.widget.TextInputLayout
         android:id="@+id/edit_engine_name_layout"
         android:layout_width="match_parent"
-        android:layout_height="wrap_content">
+        android:layout_height="wrap_content"
+        android:paddingBottom="@dimen/textinputlayout_spacing">
```
mcomella: nit: why not just hard-code this in the XML document?

I find it's easier to hardcode styles that you don't share:
- You don't have to enter a new file to understand what the layout is doing
- The dimens file stays much more readable

liuche: Didn't hard-code at @antlam's request - he said we'll need this later so I made it a dimen.

Issue #1712, opened 2017-11-13
Downloads Do Not Save With Correct File Extenstion


gdevitis:
For example, if you open this link : 
http://www.giappichelli.it/index.php?znfModule=public&znfAction=loadIndice&idArticolo=11173681
you obtain an "index.pdf" file in Firefox Focus and an "indice_9210787.pdf" file in other mobile/desktop browsers. Is this solvable without changing the security and privacy standards provided by browser?

gdevitis:
why "help wanted" label? maybe "bug" label is more correct.

mcomella:
> why "help wanted" label? maybe "bug" label is more correct.

"help wanted" means we don't have enough engineering resources to tackle this bug in a timely fashion and so we're requesting help from our contributor community. :)

mcomella:
To fix this bug, take a look at the URL given at the top and verify you can reproduce this issue. If it's reproducible, I think Focus is either:
1. Downloading the wrong file
1. Changing the file name when downloading

Figure out what's going wrong and fix the issue! :)

pocmo:
Some additional hints:

DownloadUtils is determining the file name based on the content disposition header or if absent based on the URL:
https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/utils/DownloadUtils.java

There are unit tests for DownloadUtils. Please extend the unit test to cover this case (assuming the problem is in DownloadUtils):
https://github.com/mozilla-mobile/focus-android/blob/master/app/src/test/java/org/mozilla/focus/utils/DownloadUtilsTest.java

siddware:
Thanks for the hints @pocmo :) The problem stems from the fact that the aforementioned URL has inline in its content disposition instead of attachment. As a result, we get a null [here](https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/utils/DownloadUtils.java#L57) and the method resorts to guessing the name instead of using the filename in content disposition. This pattern seems to be [based on WebKit](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/webkit/URLUtil.java) (Line: 387). [A browser](https://github.com/anthonycr/Lightning-Browser) using the same engine suffers from the same issue. But Chrome seems to work fine and so does Firefox(Gecko?). Should we change the pattern to handle inline as well?

mcomella:
@siddware Just a few questions to get more context:

[...]

Issue #887, opened 2017-07-10
Use LeakCanary in Dev (And Beta?) builds


pocmo:
In #877 we saw multiple instances of MainActivity. LeakCanary can help us to see whether it's our code that keeps references to old instances.

We should enable it in Dev builds. In beta builds it would be helpful too - but it might be confusing to our testers.

mcomella:
Some additional comments in #1901.

Issue #732, opened 2017-06-08
Check if Adjust token exists at build time (release builds)


pocmo:
Currently we have a runtime check to crash Focus release builds if no Adjust token is set.. So that we do not accidentally ship without a token. It might be more helpful to "crash" the build of release builds if no token is set. Otherwise we create APKs that just crash - and there's the risk that we might accidentally ship such an APK. So let's no create them in the first place.

mcomella:
Note: this will require some gradle knowledge – check for the existence of our adjust token on the release variant and fill the build if the file does not exist or is not empty!

bansalayush:
hey @pocmo  @mcomella  , I would like to work on this issue . Can you guys drop a few hints on what to do ?

mcomella:
@bansalayush Unfortunately, I don't know gradle well enough to know the correct answer. I think it involves waiting for the task graph to complete, checking if "release" is present in the set of running tasks and, if so, ensuring the config has set the adjust token, failing the build if not.

bansalayush:
@mcomella Understood the first two parts . What's an adjust token ? Can you shed some light on this ?

mcomella:
Adjust is an SDK we use and it requires an API token. We get the token at
compile time from an optional file on desk: here's the build.gradle file
where we slurp it up:
https://github.com/mozilla-mobile/focus-android/blob/308093ecbd82f755f56b0a44f4dd27930adaa208/app/build.gradle#L317

mcomella:
The change is happening at the build script's configure time: *all* tasks
are affected for exceptions thrown during configure. You want to do this
check at Execution time, only if the appropriate release tasks are being
executed, to avoid breaking the build for every task if this token is
missing.

See this doc (disclaimer: haven't read it yet) on the build lifecycle for
notes on configure and execution time:
https://docs.gradle.org/current/userguide/build_lifecycle.html

bansalayush:
@mcomella Thanks again , above link helped . Narrowed the solution down to 
```
gradle.taskGraph.beforeTask {
    Task task ->
        def isRelease = task.toString().contains("assembleFocus") && task.toString().contains("Release")
        println(isRelease)
        if(isRelease){
            try {
                def token = new File("${rootDir}/.adjust_token").text.trim()
                buildConfigField 'String', 'ADJUST_TOKEN', '"' + token + '"'
                println "(Added from .adjust_token file)"
            } catch (java.io.FileNotFoundException e) {
[...]

PR #1966, opened 2017-12-08
Closes #1900: Kotlin: treat warnings as errors

### LINKED ISSUE #1900: Kotlin: treat warnings as errors
Kotlin 1.2 introduces a compiler flag to treat warnings as errors: https://kotlinlang.org/docs/reference/whatsnew12.html#warnings-as-errors

I've found this to be a good practice for toy programs in other languages so we should consider it here!

siddware:
Added flag to treat Kotlin errors as warnings. Changed the use of kotlin-kapt from apply to using the plugins block to avoid a builds failing due to sources output directory not being specified: https://github.com/mozilla-mobile/focus-android/issues/1900#issuecomment-349306400 Closes #1900

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -138,9 +138,7 @@ kotlin {
 }
 
 tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
-    kotlinOptions {
         kotlinOptions.allWarningsAsErrors = true
-    }
 }
 
 jacocoAndroidUnitTestReport {
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -6,8 +6,6 @@ apply plugin: 'jacoco-android'
 apply plugin: 'pmd'
 apply plugin: 'checkstyle'
 apply plugin: 'kotlin-android-extensions'
-apply plugin: 'kotlin-kapt'
-
 apply from: "$project.rootDir/tools/gradle/versionCode.gradle"
 
 android {
@@ -137,6 +135,12 @@ android {
 
 kotlin {
     experimental { coroutines 'enable' }
+}
+
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+    kotlinOptions {
+        kotlinOptions.allWarningsAsErrors = true
+    }
 }
 
 jacocoAndroidUnitTestReport {
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,7 @@ buildscript {
 
 plugins {
     id "io.gitlab.arturbosch.detekt" version "1.0.0.RC5-6"
+    id "org.jetbrains.kotlin.kapt" version "1.2.0"
 }
 
 detekt {


siddware:
I was facing the same error @pocmo was facing and with the flag set to on, it was resulting in builds failing. I tried several things from creating source sets for the build variant (focusGeckoBeta) to turning on the [generateStubs for kapt](https://github.com/griffio/dagger2-kotlin) to adding some code[ based on an idea sample](https://github.com/JetBrains/kotlin-examples/blob/master/gradle/kotlin-querydsl/build.gradle). They didn't have any effect, some were outdated and some just didn't work. The only way this works is by using an [alternate way ](https://kotlinlang.org/docs/reference/kapt.html) to apply the kapt plugin, and a shot-in-the-dark theory is that kapt seems to have a [few versions in the wild](https://medium.com/@workingkills/pushing-the-limits-of-kotlin-annotation-processing-8611027b6711) and explicitly specifying the version fetches the right one?

app/build.gradle
```diff
@@ -139,6 +137,12 @@ kotlin {
     experimental { coroutines 'enable' }
 }
 
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
```
mcomella: nit: I actually don't know that much gradle – why does this diverge from [the docs](https://kotlinlang.org/docs/reference/whatsnew12.html#warnings-as-errors), which state:
```groovy
compileKotlin {
    kotlinOptions.allWarningsAsErrors = true
}
```

Furthermore, I'd expect tasks.CompileKotlin, rather than `KotlinCompile` – why is it KotlinCompile?

app/build.gradle
```diff
@@ -139,6 +137,12 @@ kotlin {
...
+    kotlinOptions {
+        kotlinOptions.allWarningsAsErrors = true
```
mcomella: nit: I'd also expect the duplicated `kotlinOptions` to be unnecessary

build.gradle
```diff
@@ -24,6 +24,7 @@ buildscript {
 
 plugins {
     id "io.gitlab.arturbosch.detekt" version "1.0.0.RC5-6"
+    id "org.jetbrains.kotlin.kapt" version "1.2.0"
```
mcomella: If we apply kapt this way, do we also have to specify the kapt dependency, like they mention in [the docs](https://kotlinlang.org/docs/reference/kapt.html#using-in-gradle)?
```groovy
dependencies {
    kapt 'groupId:artifactId:version'
}
```

mcomella:
Once again, thanks for your help! :) If you want to continue fixing up bugs while I'm off over the weekend, perhaps #887 if you want to do more build stuff or #1982 for some Android UI work?

app/build.gradle
siddware: The documentation seems to be aimed at pure Java + Kotlin projects, and since Android's gradle implementation produces tasks with a different naming schema, it threw a "Could not find compileKotlin()..." error and I based my code on the solution proposed [here](https://stackoverflow.com/a/44143253). I tried tasks.CompileKotlin now and it results in an error("Could not get unknown property 'org' for project..."). I couldn't find a reason why, the closest I got was this [official doc](https://kotlinlang.org/docs/reference/using-gradle.html) while it doesn't have an explanation it uses the same code(under Compiler Options).

build.gradle
siddware: To the best of my understanding, it seems `kapt` just replaces the usage of `annotationProcessor` in gradle. I looked for the usages before raising the PR but I couldn't find any. Please let me know if this needs to be corrected. 
Sources: [1](https://github.com/JetBrains/kotlin-examples/blob/master/gradle/android-dagger/app/build.gradle) [2](https://github.com/JetBrains/kotlin-examples/blob/master/gradle/android-databinding/app/build.gradle)

app/build.gradle
siddware: Thanks, fixed it :)

siddware:
Not that I am a Gradle ninja but I tried my best to explain the rationale behind my choices :) Please let me know if you need any other information. 
Thanks for your continued suggestions about apt issues. You and @pocmo have been really welcoming!

mcomella:
Hey @siddware – thanks for the comments and updated commit! We really appreciate it! Sorry this review has been taking so long – we're having a colocated team meeting and our schedule has been a bit crazy. I'll get to this review either later today or tomorrow morning.

app/build.gradle
mcomella: This makes sense – we don't have the compileKotlin task due to the build variants and instead we have e.g. compileFocusWebviewDebugKotlin so we assign it to all tasks.

Thanks for the sources! It's great that I'm learning about gradle by reviewing this too! :)

build.gradle
mcomella: In the two sources, they use the `apply plugin ...kapt` model so they're different than our `plugins` block.

That being said, we apply detekt in the `plugins` block without having a separate dependencies entry so this wfm.

mcomella:
Thanks again for your thorough research, @siddware !

If you want to continue to hack with us, maybe #732 for build related issues and #1712 for UI issues.

siddware:
@mcomella Thanks :) Sorry about the lack of communication been a busy week, will take a look at those two issues.

Issue #1972, opened 2017-12-08
ManualAddSearchEngineSettingsFragment: Add 'final' keyword to constant


pocmo:
A good first issue for new contributors.

SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS in ManualAddSearchEngineSettingsFragment should be final:
https://github.com/mozilla-mobile/focus-android/blob/e79c081104f741bafa75d30ff3b0729a0daf9b1c/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java#L42

Jacquester:
If no one wants this in about a day I'll take it :)

Issue #1945, opened 2017-12-07
Reduce complexity of SettingsFragment by not overloading responsibility


pocmo:
SettingsFragment used to be responsible for showing the main settings screen only.

Now it's also responsible for showing a bunch of sub-screens. Those screens share almost no functionality so having a lot of switch cases in newInstance, onCreateOptionsMenu and so on make this class super complex and I fear we end up with monster like in Fennec.

For the autocomplete screens I added separate fragments and this seemed to work nicely. The only thing I borrowed was a little bit of code to send a telemetry event in case a setting was changed.

I think it would be cleaner to move the code into separate fragments for separate screens. If there's more code they need to share then it might make sense to introduce a base class for them?

Jacquester:
Hi @pocmo 

https://github.com/Jacquester/focus-android/commits/Jacquester
Is this basically what you mean ? I'm essentially moving the sub-views to their own fragments (Later removing unnecessary classes and code from SettingsFragment)

mcomella:
Hey @Jacquester – it looks like you're in the right direction. To summarize, what this issue is about, it's primarily intended to simplify SettingsFragment by assigning only one layout to one fragment, whereas `SettingsFragment` now inflates multiple (4?) separate layouts.

Note that PR #1992 is currently open for this issue, but it's not complete. If you keep working on this issue, you may conflict with that user and your changes may not get in. That being said, if that PR is inactive for long enough, we'll assume it won't get worked and close it, giving you space to work on this.

By the way, I'd recommend working in git branches – they allow you to separate different, unrelated bits of work so they can all land independently on the main line of code (master).

Jacquester:
> Note that PR #1992 is currently open for this issue

Yeah I'll leave this to @michaelsam94 to figure out lol
I only played with it to see if I understand the issue correctly and to figure out if this could possibly solve #1974

Jacquester:
I don't really want to make a new issue since this is mostly based off of this issue:

Is there a possibility of merging ManualAddSearchEngineSettingsFragment and ManualAddSearchEnginePreference into one class ?

Reasons why I'm asking:
* They both represent the same view and I can't really see a reason for ManualAddSearchEnginePreference existing.
[...]

PR #1992, opened 2017-12-09
ManualAddSearchEngineSettingsFragment: Add 'final' keyword to constan…

michaelsam94:
…t #1972


mcomella:
Hey @michaelsam94 – thanks for the comments and updated commit! We really appreciate it! Sorry this review has been taking so long – we're having a colocated team meeting and our schedule has been a bit crazy. I'll get to this review either later today or tomorrow morning.

michaelsam94:
@mcomella never mind I really happy to contribute in this project thanks.

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -72,9 +72,9 @@ public static SettingsFragment newInstance(Bundle intentArgs, SettingsScreen set
             case REMOVE_ENGINES:
                 f = new SettingsFragment();
                 break;
-            case ADD_SEARCH:
+            /*case ADD_SEARCH:
```
mcomella: When removing code, you can just delete it – no need to comment it out! If we need to revive the code, we can get it from version control.

That being said, I'm not convinced we should delete this – see below.

app/src/main/java/org/mozilla/focus/settings/ValidateSearchEngineAsyncTask.java
```diff
@@ -0,0 +1,130 @@
...
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+
+
+
+public class ValidateSearchEngineAsyncTask extends AsyncTask<Void, Void, Boolean> {
```
mcomella: I don't think it's worth removing this from `Manual...Fragment` – we had to change the visibility of two mutable variables, which is already a really fragile pattern, so I don't want to expose them to other classes. The first alternative I think of, to use a listener to change the values, is a lot of boilerplate. I would prefer to just leave this as a private inner class of `Manual...Fragment`.

It's true that Manual... becomes a really long file and that this validation logic is a separate responsibility from managing the settings UI (so it does make sense to separate) but I think it'd be better to try to fix it by:
- Using more expressive constructs than AsyncTasks, e.g. kotlin coroutines.
- Taking *all* of the UI logic away from the validate async task
- Separating the validation logic from the async-handling logic as much as possible

Note that this async task was written last minute so it is a bit of a mess. :\

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -7,53 +7,41 @@
...
-
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-
-public class ManualAddSearchEngineSettingsFragment extends SettingsFragment {
+public class ManualAddSearchEngineSettingsFragment extends PreferenceFragment {
     private static String LOGTAG = "ManualAddSearchEngine";
 
-    // Set so the user doesn't have to wait *too* long. It's used twice: once for connecting and once for reading.
-    private static int SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS = 4000;
+
```
mcomella: nit: there should only be one line of whitespace between any part of the code.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -135,107 +123,17 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
     }
 
-    private void setUiIsValidatingAsync(final boolean isValidating, final MenuItem saveMenuItem) {
+    void setUiIsValidatingAsync(final boolean isValidating, final MenuItem saveMenuItem) {
         final ManualAddSearchEnginePreference pref =
                 (ManualAddSearchEnginePreference) findPreference(getString(R.string.pref_key_manual_add_search_engine));
         pref.setProgressViewShown(isValidating);
 
         saveMenuItem.setEnabled(!isValidating);
     }
```
mcomella: nit: there's a lot of newline whitespace at the end of the file

app/src/main/java/org/mozilla/focus/settings/SettingsFragment.java
```diff
@@ -194,7 +194,10 @@ public boolean onPreferenceTreeClick(PreferenceScreen preferenceScreen, Preferen
             showSettingsFragment(SettingsScreen.SEARCH_ENGINES);
             TelemetryWrapper.openSearchSettingsEvent();
         } else if (preference.getKey().equals(resources.getString(R.string.pref_key_manual_add_search_engine))) {
-            showSettingsFragment(SettingsScreen.ADD_SEARCH);
+            getFragmentManager().beginTransaction()
```
mcomella: This statement is equivalent to the one you removed but less high-level – did you have a particular reason for changing the code? I find the high-level `showSettingsFragment(SettingsScreen.ADD_SEARCH)` to be clearer and it lets our SettingsScreen enum tell us all the different search screens we can show (if it's kept up to date).

mcomella:
Just an FYI that #1972 was closed but #1945 is still open for completion.

mcomella:
This PR seems inactive: closing. @michaelsam94 please let us know if you're still working on this and we can reopen.

Issue #1996, opened 2017-12-13
Focus crashes when switching tabs


npark-mozilla:
Device: Pixel 2
STR:

Open 4+ tabs.  Do it fast so the new tab is created before the parent page is fully loaded.  (e.g. www.thestar.com)
switch between tabs.  Switch to another tab before the site is fully loaded.
Expected:
Switches fine
Actual:
Crashes.

adb logcat shows following:
12-13 11:28:12.623 18337 18489 W System.err: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference
12-13 11:28:12.623 18337 18489 W System.err: 	at org.mozilla.focus.utils.UrlUtils.urlsMatchExceptForTrailingSlash(UrlUtils.java:108)
12-13 11:28:12.623 18337 18489 W System.err: 	at org.mozilla.focus.webview.FocusWebViewClient.shouldInterceptRequest(FocusWebViewClient.java:124)
12-13 11:28:12.623 18337 18489 W System.err: 	at com.android.webview.chromium.Ap.shouldInterceptRequest(WebViewContentsClientAdapter.java:54)
12-13 11:28:12.623 18337 18489 W System.err: 	at org.chromium.android_webview.AwContents$BackgroundThreadClientImpl.shouldInterceptRequest(AwContents.java:9)
12-13 11:28:12.623 18337 18489 W System.err: 	at org.chromium.android_webview.AwContentsBackgroundThreadClient.shouldInterceptRequestFromNative(AwContentsBackgroundThreadClient.java:11)
12-13 11:28:12.624 18337 18489 F chromium: [FATAL:jni_android.cc(243)] Please include Java exception stack in crash report
12-13 11:28:12.636 18337 18337 D GeckoLocales: No selected locale. No correction needed.
12-13 11:28:12.638 18337 18337 W Notification: Use of stream types is deprecated for operations other than volume control
12-13 11:28:12.638 18337 18337 W Notification: See the documentation of setSound() for what to use instead with android.media.AudioAttributes to qualify your playback use case

npark-mozilla:
@liuche FYI

BTW, this shows up on 2.5

npark-mozilla:
Hmm, it looks like the crash is still there, here's the logcat:

12-13 14:48:33.160   757  2908 D audio_hw_primary: out_write: retry previous failed cal level set
[...]

PR #1998, opened 2017-12-13
Bandaid for #1996 - Add null check for urlutils.

liuche:
This is a quick fix that we want to uplift to 4.0, for RC2.

We should do a proper fix, so not closing this issue yet.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
+++ b/app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
@@ -123,7 +123,16 @@ public class UrlUtils {
         return "data:text/html;charset=utf-8;base64,".equals(url);
     }
 
-    public static boolean urlsMatchExceptForTrailingSlash(final @NonNull String url1, final @NonNull String url2) {
+    /**
+     * Checks that urls are non-null and are the same aside from a trailing slash.
+     *
+     * @return true if urls are the same except for trailing slash, or if either url is null.
+     */
+    public static boolean urlsMatchExceptForTrailingSlash(final String url1, final String url2) {
+        // This is a hack to catch a NPE in issue #26.
+        if (url1 == null || url2 == null) {
+            return false;
+        }
         int lengthDifference = url1.length() - url2.length();
 
         if (lengthDifference == 0) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -123,7 +123,10 @@ public static boolean isInternalErrorURL(final String url) {
         return "data:text/html;charset=utf-8;base64,".equals(url);
     }
 
-    public static boolean urlsMatchExceptForTrailingSlash(final @NonNull String url1, final @NonNull String url2) {
+    public static boolean urlsMatchExceptForTrailingSlash(final String url1, final String url2) {
+        if (url1 == null || url2 == null) {
```
mcomella: It may be unexpected that this returns false if they're both null – can you add a javadoc comment?

mcomella: Can you add a comment explaining why we changed this and mention it's hacky? I think we should revert this change when we fix it (hopefully with kotlin's null type safety).

app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -123,7 +123,10 @@ public static boolean isInternalErrorURL(final String url) {
         return "data:text/html;charset=utf-8;base64,".equals(url);
     }
 
-    public static boolean urlsMatchExceptForTrailingSlash(final @NonNull String url1, final @NonNull String url2) {
+    public static boolean urlsMatchExceptForTrailingSlash(final String url1, final String url2) {
```
mcomella: **Analysis**:
Assuming `request.getUrl().toString()` is not null (and assuming the impl of abstract Uri is OpaqueUri, it's not: http://androidxref.com/8.0.0_r4/xref/frameworks/base/core/java/android/net/Uri.java#961), `currentPageURL` must be null. It's assigned in two places which I analyzed:
- TrackingProtectionWebViewClient.notifyCurrentURL: all of the callers null check this value before setting it
- TPWVC.onPageStarted: this value is *not* null checked and the value is passed from the webView APIs (maybe it crashes on only some API levels?)

So I think the null is getting set from the latter.

---

It looks like the code won't crash (it seems like currentURL basically gets ignored if it's null), but I think we should fix the assumption that currentURL should not be null: can you file a bug to look into this further and fix it? (and CC Sebastian)

PR #1999, opened 2017-12-14
Closes #1985 encodedTestQuery changed to F0cusRu1e5, Closes #1972 SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS changed to final

### LINKED ISSUE #1972: ManualAddSearchEngineSettingsFragment: Add 'final' keyword to constant

### LINKED ISSUE #1985: Url live validation check for imdb.com/%s passes, but is not a valid search engine
It looks like our http validation substitutes "test" for %s to check for validity, and adds the engine if it's not an error (400+ response code).

However, in the case of imdb.com/%s (and possibly other sites), imdb.com/test *is* a valid site, but it's not a valid search query (e.g. imdb.com/movie will 404).

We should switch our test query to something else more randomized.

Jacquester:
#1985

### DIFF
--- a/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
+++ b/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
@@ -42,7 +42,7 @@ public class ManualAddSearchEngineSettingsFragment extends SettingsFragment {
     private static String LOGTAG = "ManualAddSearchEngine";
 
     // Set so the user doesn't have to wait *too* long. It's used twice: once for connecting and once for reading.
-    private static int SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS = 4000;
+    private static final int SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS = 4000;
 
     /**
      * A reference to an active async task, if applicable, used to manage the task for lifecycle changes.
--- a/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
+++ b/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
@@ -202,7 +202,7 @@ public class ManualAddSearchEngineSettingsFragment extends SettingsFragment {
     @WorkerThread // makes network request.
     @VisibleForTesting static boolean isValidSearchQueryURL(final String query) {
         // TODO: we should share the code to substitute and normalize the search string (see SearchEngine.buildSearchUrl).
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -202,7 +202,7 @@ private void showServerError(final ManualAddSearchEngineSettingsFragment that) {
     @WorkerThread // makes network request.
     @VisibleForTesting static boolean isValidSearchQueryURL(final String query) {
         // TODO: we should share the code to substitute and normalize the search string (see SearchEngine.buildSearchUrl).
-        final String encodedTestQuery = Uri.encode("test");
+        final String encodedTestQuery = Uri.encode("F0cu5Ru1e5");
```
mcomella: Sorry we weren't clearer about this.

I think we have an opportunity to make this string more helpful:
- Servers who get repeated requests to q="F0cu5Ru1e5" might curious why but "f0cu5Ru1e5" doesn't indicate what this string is
- If this code gets accidentally used elsewhere and we see "F0cu5Ru1e5", we wouldn't really know where that value is coming from

Ideally, I think the string is something like "testSearchEngineValidation" but I think the query should at least indicate that this is a test string/query.

mcomella:
By the way, sorry for the delay – we had a colocated work meetup last week and it was hectic so we've fallen behind on issues – it shouldn't happen again!

mcomella:
Also, it's best to separate multiple issue-fixing commits into separate pull requests so they can get landed/reviewed independently (e.g. your first commit is okay and could land, but it's stuck on second commit which needs changes).

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
Jacquester: I'll get that sorted just now thanx :)

Would you recomend we close this PR so I can create two separate ones or shall we continue with this one ?

mcomella: It'd be more overhead for both of us to review and merge two PRs so let's just keep this one. Thanks! :)

PR #2020, opened 2017-12-21
Closes #1979 , Automatically bring up the keyboard when the user pres…

### LINKED ISSUE #1979: Automatically bring up the keyboard when the user presses "+ Add another search engine"
When the user presses the button to Add search engine, the form has 2 input fields and the user has to select one to get the keyboard to show up.

A simple convenience thing we could do is already bring it up once they press the button so they can start typing in the new view right away.

Thoughts @lime124 ?

rucquoy:
…ses + Add another search engine
focus_1979

I also added code to make the keyboard hide when the app is paused and show when it's resumed.
regards, 
Alex :)

### DIFF
--- a/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
+++ b/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
@@ -89,13 +89,13 @@ public class ManualAddSearchEngineSettingsFragment extends SettingsFragment {
             activeAsyncTask = null;
             menuItemForActiveAsyncTask = null;
         }
-        imm.hideSoftInputFromWindow(getView().getWindowToken(),0);
+        imm.hideSoftInputFromWindow(getView().getWindowToken(), 0);
     }
 
     @Override
     public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
         inflater.inflate(R.menu.menu_search_engine_manual_add, menu);
-        imm.toggleSoftInput(InputMethodManager.SHOW_FORCED,InputMethodManager.HIDE_IMPLICIT_ONLY);
+        imm.toggleSoftInput(InputMethodManager.SHOW_FORCED, InputMethodManager.HIDE_IMPLICIT_ONLY);
     }
 
     @Override
--- a/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
+++ b/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
@@ -20,6 +20,8 @@ import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.view.View;
+import android.view.WindowManager;
+import android.view.inputmethod.InputMethodManager;
 import android.widget.EditText;
 import org.mozilla.focus.R;
 import org.mozilla.focus.activity.InfoActivity;
@@ -43,6 +45,7 @@ public class ManualAddSearchEngineSettingsFragment extends SettingsFragment {
 
     // Set so the user doesn't have to wait *too* long. It's used twice: once for connecting and once for reading.
     private static final int SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS = 4000;
+    private InputMethodManager imm;
 
     /**
      * A reference to an active async task, if applicable, used to manage the task for lifecycle changes.
@@ -54,13 +57,14 @@ public class ManualAddSearchEngineSettingsFragment extends SettingsFragment {
     @Override
     public void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        this.imm = (InputMethodManager) getActivity().getSystemService(Context.INPUT_METHOD_SERVICE);
         setHasOptionsMenu(true);
     }
 
     @Override
     public void onResume() {
         super.onResume();
-
+        getActivity().getWindow().setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_STATE_ALWAYS_VISIBLE);
         // We've checked that this cast is legal in super.onAttach.
         ((ActionBarUpdater) getActivity()).updateIcon(R.drawable.ic_close);
    }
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -85,11 +89,13 @@ public void onPause() {
             activeAsyncTask = null;
             menuItemForActiveAsyncTask = null;
         }
+        imm.hideSoftInputFromWindow(getView().getWindowToken(),0);
```
sputnikci: \[Checkstyle\] ERROR: ',' is not followed by whitespace\.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -85,11 +89,13 @@ public void onPause() {
...
     }
 
     @Override
     public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
         inflater.inflate(R.menu.menu_search_engine_manual_add, menu);
+        imm.toggleSoftInput(InputMethodManager.SHOW_FORCED,InputMethodManager.HIDE_IMPLICIT_ONLY);
```
sputnikci: \[Checkstyle\] ERROR: ',' is not followed by whitespace\.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -54,13 +57,14 @@
...
         super.onCreate(savedInstanceState);
+        this.imm = (InputMethodManager) getActivity().getSystemService(Context.INPUT_METHOD_SERVICE);
         setHasOptionsMenu(true);
     }
 
     @Override
     public void onResume() {
         super.onResume();
-
+        getActivity().getWindow().setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_STATE_ALWAYS_VISIBLE);
```
mcomella: Would there be any adverse affects if we specified this in the XML instead? e.g. in MainActivity, we have:
```xml
android:windowSoftInputMode="adjustResize|stateAlwaysHidden"
```

Could we include stateAlwaysVisible in SettingsActivity?

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
mcomella: nit: Please use `ViewUtils.hideKeyboard`

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
mcomella: Why is toggle necessary? Also, why are the SHOW_FORCED & HIDE_IMPLICIT flags necessary? WHen can a user force the soft input open?

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
rucquoy: To be honnest, that's the only workaround i found for onResume.. It seems like anything you do with InputMethodManager in onResume is simply ignored.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
rucquoy: Because showSoftInput doesn't work here. I think the reason is the Activity is not "focused" so to use showSoftInput, i first need to call setFocusable and setFocusableInTouchMode. But if i toggle, it works just like that.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
rucquoy: I wanted to use this at first, but using ViewUtils' classes and functions in ManualAddSearchEngineSettingsFragment.java does nothing at all. Do you have any idea why ?

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
rucquoy: Sorry i didn't even answered your question. It's just my preference to use java instead of xml for everything else than "core" UI elements. I can try to use the xml statement you quoted if u want me to :) . But if we put this in SettingsActivity, isn't the keyboard going to show up everywhere in the SettingsActivity, even if no keyboard input is necessary ?

mcomella:
Hey @rucquoy – sorry it's taking so long to get back to you. We're on vacation for the holidays so at worst, we'll get to this when we come back around 1/5.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
mcomella: > It's just my preference to use java instead of xml for everything else than "core" UI elements. I can try to use the xml statement you quoted if u want me to :) .

Please do! We'd prefer XML because there's really only one place to change this attribute whereas it's harder to track down a single line of code.

> But if we put this in SettingsActivity, isn't the keyboard going to show up everywhere in the SettingsActivity, even if no keyboard input is necessary ?

My understanding of this attribute is that it will take affect only when keyboard input is necessary (i.e. it won't show a keyboard when the activity is shown if there's no input on the screen). But please correct me if I'm wrong!

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
mcomella: Unfortunately, I don't. Have you tried using the debugger to step through the call to see what is (or isn't) being called?

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
mcomella: Maybe this is related to why ViewUtils isn't working...

What else is focused? Would it be appropriate to allow focus on the activity? I'm surprised anything else besides the activity can be focused. Note: you can use the [UI hierarchy tool](https://developer.android.com/studio/profile/hierarchy-viewer.html) to figure out what is focused.

mcomella:
@rucquoy Responded to the comments – thanks for being patient!

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
rucquoy: Hey @mcomella ! I tried with an XML statement as u quoted above, but i'm afraid it's not working. Because if we add this statement in the manifest for SettingsActivity, the keyboard shows up at the start of the activity, even tho there's no input field on screen). The problem is that ManualAddSearchEngine is a fragment and not an activity itself. Therefore, i think we don't have the choice, we have to do it programatically. Any suggestions or ideas ? Please let me know.

pocmo:
Progress has stalled here and the code is going to be replaced by #2108 soon.

Issue #1981, opened 2017-12-08
Write script that compares imported locales with locales.py


pocmo:
To prevent something like #1975 in the future:

Let's write a script that checks if all the locales in the repo (values-*/strings.xml) are listed in [locales.py](https://github.com/mozilla-mobile/focus-android/blob/master/tools/l10n/locales.py) ... either in the release list or in the screenshot list.

Ignore non-language folders like `values-sw600dp`.

Let's run this at least on taskcluster.

pocmo:
@mastizada Maybe something for you. :)

pocmo:
To clarify:

* values-*/strings.xml exists but no entry in locales.py -> Error
* Entry in locales.py exists but no values-*/strings.xml -> Warning (or ignore this case if too complex)

mastizada:
Yes, assign it to me, a bit busy these days, will try to solve till next Sunday.
Are non-language folders (`values-sw600dp`) fixed? - If yes, will write them in a list to ignore.

PR #2025, opened 2017-12-25
Check imported locales with RELEASE_LOCALES

mastizada:
Fix issue #1981 - Write script that compares imported locales with locales.py
Current local output:
```
Checking for imported locales...
Error: * [values-lt] missing in RELEASE_LOCALES
Error: * [values-am] missing in RELEASE_LOCALES
Warning: * [some] missing in imported locales
```


.taskcluster.yml
```diff
@@ -118,6 +118,7 @@ tasks:
           && git checkout {{ event.version }}
           && python tools/taskcluster/get-adjust-token.py
           && python tools/l10n/filter-release-translations.py
+          && python tools/l10n/check_locales.py
```
mastizada: Will it stop if output is 1? (`exit(1)`)
Is it correct place to add? :)

tools/l10n/check_locales.py
```diff
@@ -0,0 +1,40 @@
...
+# -*- coding: utf-8 -*-
+#  This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""Check imported locales with release locales."""
+
+from os import path, listdir
+from sys import exit
+from locales import RELEASE_LOCALES
```
mastizada: Should we check for `SCREENSHOT_LOCALES` or this one?

mcomella:
Hey @mastizada – sorry it's taking so long to get back to you. We're on vacation for the holidays so at worst, we'll get to this when we come back around 1/5.

mcomella:
@mastizada Sorry, I don't have a grasp on how the l10n code works at the moment. What problem is this PR trying to solve? Also, where does the output from the first comment come from?

mcomella:
(fwiw, it might be more appropriate for @liuche to review this)

tools/l10n/check_locales.py
```diff
@@ -0,0 +1,40 @@
...
+
+PATH = path.join(path.dirname(path.abspath(__file__)), '../../app/src/main/res/')
+IGNORED_DIRECTORIES = ['sw600dp']
+
+def check_locales():
+    print("Checking for imported locales...")
+    got_error = False
+    imported_locales = [
+        directory[7:].replace('-r', '-') for directory in listdir(PATH) if directory.startswith('values-')
```
mcomella: Why `directory[7:]`? What are we removing (iiuc, we're getting the list from index 7 to the end)?

Indexing is a fragile way to remove these entries – is there a better way to get these locales?

tools/l10n/check_locales.py
```diff
@@ -0,0 +1,40 @@
...
+    ]
+    for ignored in IGNORED_DIRECTORIES:
+        try:
+            imported_locales.remove(ignored)
```
mcomella: nit: fwiw, if imported_locales and IGNORED_DIRECTORIES were sets, I think you could simplify this with `imported_locales -= IGNORED_DIRECTORIES`.

.taskcluster.yml
mcomella: I'm not sure – can you test this? Your PRs should build on task cluster so you should be able to add 1's, push a PR, and see if it stops execution correctly.

That being said, stopping execution on 1 makes sense to me.

mcomella:
My understanding of this PR:
- `imported_locales` is the set of locales that we have data for (extracted from the values folders)
- `RELEASE_LOCALES` is the set of locales that we've whitelisted to bundle and ship with the app
- On release builds, this change will fail the build if `imported_locales` is not the same as `RELEASE_LOCALES` (what is the intended problem we're trying to solve here?)

Questions:
- Are my assumptions of what `imported_locales` and `RELEASE_LOCALES` are correct?
- Can `imported_locales` diverge from `RELEASE_LOCALES`? e.g. imported_locales has locales with partial translations that isn't ready to be shipped yet
- Is there a better time to fix/verify this? e.g. on string import

tools/l10n/check_locales.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#  This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""Check imported locales with release locales."""
```
mcomella: As someone with little context on our l10n setup, this description isn't clear what this script is trying to do.

tools/l10n/check_locales.py
mcomella: I'm not sure. @liuche Do you know?

tools/l10n/check_locales.py
mastizada: It gets list of directories starting with `values-` name, and then removes that `values-` part from their name. Locale directories are like `values-<locale>`. Usually `replace` is used, but we already filtered list before doing any operation, so it will always work.

tools/l10n/check_locales.py
mastizada: `IGNORED_DIRECTORIES` is a list of ignored directories starting with `values-` as locales. Currently we are only aware of 1 directory, but list can be extended in future. These are default directories for Android, and I'm not sure if they will be same always and not generated while building application. `try` block was used to avoid unexpected issues with these directories. We just need to have a list of currently imported locales in project. If anyone with Android development experience can give a good info for possible folders with `values-` in their name, that will help a lot.

mastizada:
@mcomella This PR tries to fix issue #1981 - please, check its details for better information.

`imported_locales` - list of locales imported from `po` files to Android source directory as `values-<locale code>` as xml files.
`RELEASE_LOCALES` - the list of locales that are expected to exists in imported locales list.

The issues has better description of problem, that problem already happened before and we are trying to automatically test and be sure that it will not happen again.

Example outcomes are from my local tests.

Sdaswani:
@mastizada thanks for your contribution. I want one @pocmo to take a look at this but he is on paternity leave, so please stay tuned (he should be back in about 5 weeks).

pocmo:
@mastizada Hey! I'm back and will have a look at this soon. Is this from your side done? :)

mastizada:
Hi @pocmo , its nice to have you back :) Just two questions, first about taskcluster and second about using `RELEASE_LOCALES` or `SCREENSHOT_LOCALES`. Also I think it will be better to update branch with latest commits.

pocmo:
Thank you for your work, @mastizada! :)

* Yes. If the taskcluster command returns a non-zero exit code then the task will be marked as failed.
* Let's use SCREENSHOT_LOCALES (which includes RELEASE_LOCALES). What we want to make sure is that all imported locales are listed in locales.py somewhere
* Let's only execute this for pull requests for now (.taskcluster.yml)

mastizada:
It works. In Taskcluster:
```
+ python tools/l10n/check_locales.py
Checking for imported locales...
Error: * [values-hr] missing in SCREENSHOT_LOCALES
Error: * [values-am] missing in SCREENSHOT_LOCALES
```

pocmo:
@mastizada Thank you! :)

I'll take care of the missing locales and after that I'll merge your PR.

pocmo:
There's a merge conflict. Will fix in a separate PR and land patches.

Issue #1962, opened 2017-12-08
Custom search engines aren't selected after creation even though they can't be used until selected


mcomella:
after a user creates a search engine, we don’t select it for them. It’s seems wrong to do so but feels weird that I can’t use this search engine unless it’s the default so what’s the point of adding it?

antlam and liuche agreed on slack.

pocmo:
Yeah, for now auto-selection might be a good idea. In the long run there was some UX discussion about adding something to quickly search with other search engines too (like with the quick search bar in fennec).

liuche:
@lime124 says that this is supposed to happen.

vickz84259:
Would like to work on this if no one is

mcomella:
@ekager @boek ^

ekager:
Hey @vickz84259 , you are welcome to work on this! Let's get @brampitoyo (UX) to confirm the new behavior before you start :)

brampitoyo:
@ekager @vickz84259 Yes. I can confirm that we should set the newly-created custom search engine as a default, right after you finish adding it.

sblatz:
@bbinto Do you want Telemetry to be updated in any way for this? In other words: the default search engine will now be set every time the user adds a new search engine, thus pinging two Telemetry events. Would you prefer that it _not_ ping the "change default search engine" Telemetry event?

PR #2026, opened 2017-12-26
Closes #1982 Increase search engine icon size and center character

### LINKED ISSUE #1982: Custom search engine icon should be consistent with default icons
screenshot_20171208-115358

The icon is a different size than the rest, and the single character isn't centered :)

tvalle:
This issue was caused because `ic_homescreen_shape.png` had a padding in the image, so it made the icon in the "add search engine" appear smaller.

One problem is that `ic_homescreen_shape.png` was also being used in the "Add to home screen dialog". One could either add another image to use specifically in the search engine or remove the padding in the image entirely and then add the padding in the xml, which is the route I decided to go.
If the image that I updated is not OK please send another one so I can change it.

Regarding centering the character, I went with the second answer on this [StackOverflow question](https://stackoverflow.com/questions/11120392/android-center-text-on-canvas). The character was being centered using the marked solution, which sometimes is not OK as explained in the second answer.

Search engine screen:
screenshot-focus

Add to home screen dialog:
screenshot-focus2


mcomella:
Hey @tvalle – sorry it's taking so long to get back to you. We're on vacation for the holidays so at worst, we'll get to this when we come back around 1/5.

mcomella:
> remove the padding in the image entirely and then add the padding in the xml, which is the route I decided to go.

Sounds good, that's what I would do.

> If the image that I updated is not OK please send another one so I can change it.

@lime124 Can you give us an asset for [ic_homescreen_shape.png](https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/res/drawable-nodpi/ic_homescreen_shape.png) without padding? Looking at the asset tvalle supplied, it's the same dip as the original asset (according to github's UI) so I'm concerned it may have been scaled and have artifacting.

app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.kt
```diff
@@ -76,21 +77,22 @@ class IconGenerator {
...
-            paint.textAlign = Paint.Align.CENTER
+            paint.textAlign = Paint.Align.LEFT
             paint.textSize = textSize
             paint.isAntiAlias = true
 
-            canvas.drawText(character.toString(),
-                    canvas.width / 2.0f,
-                    canvas.height / 2.0f - (paint.descent() + paint.ascent()) / 2.0f,
-                    paint)
+            paint.getTextBounds(character.toString(), 0, 1, rect)
```
mcomella: nit: micro-optimization, but maybe we should cache `character.toString()` in a local variable?

app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.kt
```diff
@@ -76,21 +77,22 @@ class IconGenerator {
...
+
+            val x = canvas.width / 2f - rect.width() / 2f - rect.left
```
mcomella: nit: add a comment to the SO post you found this from.

mcomella:
tvalle, if you'd like to get hacking on something else with us while we wait for lime, maybe #1962?

tvalle:
I'm addressing the comments on the code but I have a question: it's better to address those on a follow-up commit, or rewrite the history of this branch with -amend and a push -f?

mcomella:
> t's better to address those on a follow-up commit, or rewrite the history of this branch with -amend and a push -f?

It depends on the situation:
- Generally, when writing code, it's hard to follow the history of code where commits write new code and then subsequent commits rewrite that code so `--amend` and `-f` is preferred
- But as a reviewer, seeing separate commits can be helpful if there are only a few small changes. (if there are many changes, the amend works better)
- But as a developer creating a PR, amending takes more time when you have multiple commits and need to figure out the appropriate one to make the changes in
- There's also the possibility of creating commits as a separate commit in review and then squashing when pushing, but this is even more time.

To be honest, I generally optimize for my time and create new commits – it's more common for me to respond to reviews than for someone to carefully look through my commit history. You can use your judgement and take your own pick though!

antlam:
hey @tvalle @mcomella 

I don't think we should adjust the UI in the Add to Homescreen dialog with these changes. The design over there is to spec and we dont want to adjust something about that design just to fit this use-case over here.

Instead, I can make the asset (the purple rounded-square) for this list. Try this!

Question for @tvalle , in the "Installed search engines" list, what size is the single letter 'w' ? and is it upper or lowercase?

[Search engine list.zip](https://github.com/mozilla-mobile/focus-android/files/1623989/Search.engine.list.zip)

tvalle:
Sorry for the delay,

>> I don't think we should adjust the UI in the Add to Homescreen dialog with these changes.

> afaik, we generate the icons the same way and it'd have the same problems but tvalle can say more.

Yes, exactly, @mcomella is right. We generate the icons the same way, if we use these new icons that @antlam provided without adjusting the ui in the Add to Home Screen dialog, the dialog would not be ok since it was expecting a padding inside the .png, which now we don't have anymore.

I think the right approach is to always have images without paddings and if needed adjust paddings in the .xml layouts.

tvalle:
@antlam In the "Installed search engines" the single letter is lowercase. In the code it says the size is 36dp.

Regarding the icons, wouldn't it be better for the icon to be a vector instead of 5 .pngs?

antlam:
> Yes, exactly, @mcomella is right. We generate the icons the same way, if we use these new icons that @antlam provided without adjusting the ui in the Add to Home Screen dialog, the dialog would not be ok since it was expecting a padding inside the .png, which now we don't have anymore. I think the right approach is to always have images without paddings and if needed adjust paddings in the .xml layouts.

I'm sorry but I don't follow... To be clear, I just want to make sure that we dont' do anything _here_ that would affect the UI _over there_ (Add to Homescreen dialog). As long as that's still the case then we should be OK.

> @antlam In the "Installed search engines" the single letter is lowercase. In the code it says the size is 36dp.

Let's update it to upper case. But keep it the same size.

> Regarding the icons, wouldn't it be better for the icon to be a vector instead of 5 .pngs?

Vector would be great. But I was under the impression we had to use .PNGs for now.

antlam:
@tvalle whenever you're ready, can I see a full screenshot of this (emulator is fine) on device? Thanks!

tvalle:
Here are the screenshots using the provided icon and with upper case:

screenshot
screenshot2

> I'm sorry but I don't follow... To be clear, I just want to make sure that we dont' do anything here that would affect the UI over there (Add to Homescreen dialog). As long as that's still the case then we should be OK.

The issue is that the code that generates icons is the same for both screens, on "Add to home" it was expecting the padding that existed in the icon so there was no padding in the .xml layout, on "Search Engines" it wasn't expecting the padding in the image, so it created this bug that we are fixing.

In order to just fix the layout in "Search Engines" and not affect "Add to home" we would have to make the icon bigger to compensate for the padding that existed in the .png, but this is messy(I don't even know if this is possible). The correct solution was to remove the padding in the image icon, but by doing that we end up with a broken layout in "Add to home", that's why we have to add the padding in the "Add to home" xml layout.

antlam:
@tvalle both screenshots look great! thanks!

brampitoyo:
Just to clarify that @tvalle’s commit on this issue will fix #1982. When this is closed, the other issue should be closed, too.

Issue #1433, opened 2017-10-04
Switch from findbugs to spotbugs?


pocmo:
Following https://github.com/mozilla-mobile/telemetry-android/issues/41 - Should we switch from findbugs to spotbugs?

Spotbugs:
https://github.com/findbugsproject/findbugs#readme

liuche:
This involves a small change to our build config.

grubdragon:
Could I take this up?

mcomella:
@grubdragon Sure! The steps for this bug:
- Summarize why findbugs development stopped: https://github.com/findbugsproject/findbugs#readme
- Switch to spotbugs

Thanks for your help! :)

vladioan3:
Why to switch to Spotbugs?
- It can analyze programs compiled for any version of Java, from 1.0 to 1.9.
- Checks for more than 400 bug patterns
- Is extensible. New detectors can be added through plugins.
- Findbugs seems to be dead: https://github.com/findbugsproject/findbugs/issues/161
- Supports FindBugsFilter

grubdragon:
Thanks @vladioan3, I'm short of time these days.

Issue #1085, opened 2017-08-04
Alpha animation of wordmark after erasing


pocmo:
Follow-up from #832: After erasing a session add an alpha animation (0-1) to the wordmark translate animation.

imjacobclark:
@pocmo @jenn-chaulk @bbinto is this something external Mozilla folk can work on (such as myself?)

pocmo:
@imjacobclark In theory yes. When erasing the browsing history we go back to the home screen. The keyboard comes up and we animate the word mark moving up as well. Only in this situation we want to add an alpha animation (0f - 1f) to it.
There's currently a refactoring of certain parts of the app underway. So we probably can't merge this right now. But you could definitely investigate how to add this and then provide a patch on top of the refactored implementation once it lands. :)
Thank you for your help so far!

grubdragon:
Hey can I take this up?

bbinto:
Yes, @grubdragon -- Thanks, let @lime124 or @Sdaswani know if you need any assistance.

mcomella:
@grubdragon Sorry we missed you on IRC!

The wordmark is the "Firefox Focus" text on the homescreen (the file for the asset is actually called `wordmark.png`). Note that the text under it will probably need an alpha animation too.

mcomella:
@grubdragon On that note, I see you've already started working on another `good first issue` – it'd be great if you pull something from our `help wanted` issues for your next bug (if I remember, I'll probably include one in your PR once it's completed, if I haven't already).

grubdragon:
Okay, I'll try something else :)

noloman:
Hi! is this issue still valid, open and available for doing some work on it?

grubdragon:
Hey, yes it is! You can start right away!

noloman:
yes but not for this issue, or you mean that there's already a PR for this issue in particular?

grubdragon:
This issue is for beginners, why don't you try some help-wanted ones?

noloman:
yes sure, that'd work too. I consider myself a beginner in this repository, and that's why I chose it :)

rohitramkumar308:
@noloman are you working on this one? Can I take it if you are not. I am a beginner too. Thanks

shikhart98:
Can I just try it out? I'm new to open source.

pocmo:
Hey @shikhart98! There's a pull request for this issue in flight already (#2268).

[...]

PR #2027, opened 2017-12-28
Closes #1976, HttpUrlConnection now allowed to follow redirects

### LINKED ISSUE #1976: Search engine validation: Follow redirects
We should allow HttpUrlConnection to follow redirects:

https://developer.android.com/reference/java/net/HttpURLConnection.html#setInstanceFollowRedirects(boolean)

Currently we accept a redirect as successful. That should be okay in most cases. I wouldn't be surprised if there are some http->https redirects that will end up in a 404 though.

grubdragon:
@pocmo pointed out that instead of accepting a redirect as successful, we could follow redirects.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
+++ b/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
@@ -219,6 +219,7 @@ public class ManualAddSearchEngineSettingsFragment extends SettingsFragment {
         HttpURLConnection connection = null;
         try {
             connection = (HttpURLConnection) searchURL.openConnection();
+            connection.setInstanceFollowRedirects(true);
             connection.setConnectTimeout(SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS);
             connection.setReadTimeout(SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS);
 
--- a/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
+++ b/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
@@ -223,8 +223,8 @@ public class ManualAddSearchEngineSettingsFragment extends SettingsFragment {
             connection.setConnectTimeout(SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS);
             connection.setReadTimeout(SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS);
 
-            // A non-error HTTP response is good enough as a sanity check, some search engines redirect to https.
-            return connection.getResponseCode() < 400;
+            // Now that redirects are followed, 300 is a better and stronger sanity check, checks for a non error and non redirect response
+            return connection.getResponseCode() < 300;
 
         } catch (final IOException e) {
             // Don't log exception to avoid leaking URL.
--- a/app/src/test/java/org/mozilla/focus/settings/SearchEngineValidationTest.kt
+++ b/app/src/test/java/org/mozilla/focus/settings/SearchEngineValidationTest.kt
@@ -24,9 +24,9 @@ class SearchEngineValidationTest {
     }
 
     @Test
-    fun `URL using HTTP redirect is valid`() = withMockWebServer(responseWithStatus(301)) {
-        // Currently we accept redirects as valid. We might want to follow the redirect (Issue #1976)
-        assertTrue(isValidSearchQueryURL(it.rootUrl()))
+    fun `URL using HTTP redirect is invalid`() = withMockWebServer(responseWithStatus(301)) {
+        // We now follow redirects(Issue #1976). This test now asserts false.
+        assertFalse(isValidSearchQueryURL(it.rootUrl()))
     }
 
     @Test


mcomella:
Hey @grubdragon – we're on vacation for the holidays so at worst, we'll get to this when we come back around 1/5.

btw, from a glance, the code looks good but I'd like to test it, which is why I can't just land it now.

grubdragon:
Yeah cool! I'll try to fix some other bugs till then! :smile:

mcomella:
> Yeah cool! I'll try to fix some other bugs till then!

@grubdragon Have you found enough bugs or do you want me to find something for you?

I'll get around to reviewing this soon.

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -219,6 +219,7 @@ private void showServerError(final ManualAddSearchEngineSettingsFragment that) {
         HttpURLConnection connection = null;
         try {
             connection = (HttpURLConnection) searchURL.openConnection();
+            connection.setInstanceFollowRedirects(true);
             connection.setConnectTimeout(SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS);
             connection.setReadTimeout(SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS);
```
mcomella: A few lines below this, we check against responseCode < 400 – we should check < 300 now because we're following redirects. If we get a 300 even after following redirects, something must be crazy.

grubdragon: Ahh right, covering all bases

grubdragon:
@mcomella Hey could you suggest me some bugs?

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -223,8 +223,8 @@ private void showServerError(final ManualAddSearchEngineSettingsFragment that) {
...
 
-            // A non-error HTTP response is good enough as a sanity check, some search engines redirect to https.
-            return connection.getResponseCode() < 400;
+            // Now that redirects are followed, 300 is a better and stronger sanity check, checks for a non error and non redirect response
```
mcomella: nit: general coding tip: this comment references the old state of the code "now that redirects...", "...is a better..." but someone reading this code has no idea what the previous code looked like. Thus, comments should either:
- Only reference the current code
- Explain the previous state of the code and describe why the current code doesn't do that anymore.

In this case, a dev doesn't need to understand the old state to understand the code so I'd only reference the current code (e.g. "We follow redirects, check for success codes only"). Note that your comment could be a good extended commit message though! That's usually the place to write, "I changed this code for these reasons".

That being said, I don't think this is worth creating another commit for - I'm just going to land it.

mcomella:
> @mcomella Hey could you suggest me some bugs?

@grubdragon I saw you commenting on #1085 and #1433 – those sound like they'd be good to tackle! Thanks for hacking on code with us! :)

app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.java
```diff
@@ -219,11 +219,12 @@ private void showServerError(final ManualAddSearchEngineSettingsFragment that) {
...
         try {
             connection = (HttpURLConnection) searchURL.openConnection();
+            connection.setInstanceFollowRedirects(true);
             connection.setConnectTimeout(SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS);
             connection.setReadTimeout(SEARCH_QUERY_VALIDATION_TIMEOUT_MILLIS);
 
-            // A non-error HTTP response is good enough as a sanity check, some search engines redirect to https.
-            return connection.getResponseCode() < 400;
+            // Now that redirects are followed, 300 is a better and stronger sanity check, checks for a non error and non redirect response
+            return connection.getResponseCode() < 300;
```
mcomella: Actually, this change I request breaks the tests: https://travis-ci.org/mozilla-mobile/focus-android/builds/325730045?utm_source=github_status&utm_medium=notification since the test explicitly checks request codes in 3xx – do you mind updating the tests too? Thanks!

grubdragon: Oh yeah sure!

grubdragon: I'm not sure if I've done that right though

app/src/test/java/org/mozilla/focus/settings/SearchEngineValidationTest.kt
```diff
@@ -24,9 +24,9 @@ class SearchEngineValidationTest {
     }
 
     @Test
-    fun `URL using HTTP redirect is valid`() = withMockWebServer(responseWithStatus(301)) {
-        // Currently we accept redirects as valid. We might want to follow the redirect (Issue #1976)
-        assertTrue(isValidSearchQueryURL(it.rootUrl()))
+    fun `URL using HTTP redirect is invalid`() = withMockWebServer(responseWithStatus(301)) {
+        // We now follow redirects(Issue #1976). This test now asserts false.
```
mcomella: nit: similar to my comment above, this comment references the past state of the code, "We **now** follow redirects... This test **now** asserts...". I think a better description might be: "We follow redirects so we should never receive a 3xx. If we do, we want to fail." Or something.

But I don't think it's worth the turn around time to fix now.

grubdragon: Will keep in mind the next time 😄 Thanks for helping me through!

PR #2031, opened 2018-01-02
Closes #887: Use LeakCanary in Dev (And Beta?) builds

### LINKED ISSUE #887: Use LeakCanary in Dev (And Beta?) builds

siddware:
Integrated Leakcanary for debug builds. Closes #887

### DIFF
--- a/app/src/main/java/org/mozilla/focus/FocusApplication.java
+++ b/app/src/main/java/org/mozilla/focus/FocusApplication.java
@@ -9,6 +9,8 @@ import android.arch.lifecycle.LiveData;
 import android.os.StrictMode;
 import android.preference.PreferenceManager;
 
+import com.squareup.leakcanary.LeakCanary;
+
 import org.mozilla.focus.locale.LocaleAwareApplication;
 import org.mozilla.focus.search.SearchEngineManager;
 import org.mozilla.focus.session.NotificationSessionObserver;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/FocusApplication.java
```diff
@@ -6,9 +6,12 @@
 package org.mozilla.focus;
 
 import android.arch.lifecycle.LiveData;
+import android.content.Context;
```
sputnikci: \[Checkstyle\] ERROR: Unused import \- android\.content\.Context\.

siddware: Thanks, resolved it :)

app/src/main/java/org/mozilla/focus/FocusApplication.java
```diff
@@ -25,9 +27,12 @@
 
 public class FocusApplication extends LocaleAwareApplication {
     private VisibilityLifeCycleCallback visibilityLifeCycleCallback;
-
```
mcomella: nit: please keep the whitespace

build.gradle
```diff
@@ -56,6 +57,16 @@ task clean(type: Delete) {
 configurations {
     ktlint
 }
+// To ensure unit tests and instrumentation tests remain consistent after the introduction of leakcanary
```
mcomella: nit: whitespace above this line

mcomella: nit: I agree with the sentiment – let's not change the tests just yet. Can you:
- file an issue (you don't have to work on it) to investigate whether we want to add this to unit or UI tests
- Mention "See #xxx" in this comment where xxx is the issue I mention above.

siddware: See #2041

mcomella:
I can confirm this appears when using a debug build on a device (but not on the emulator, for some reason).

Issue #1840, opened 2017-11-29
WebViewDataTest fails on clean install / delete all data


mcomella:
It appears the test asserts that the LocalStorage directory exists, which it will if we've used LocalStorage before but if we haven't, the LocalStorage directory will not exist and the test will fail.

Devs may run into this in testing but this passes in CI because we run a LocalStorage-based test before running this one.

To fix this bug, you should add a LocalStorage component to the test page used in WebViewDataTest and run it on a clean install / data deleted install to verify it works.

mcomella:
You can learn how to create LocalStorage in a test in `LocalSessionStorageTest`

mcomella:
FYI: JohnKal mentioned they want to work on this in https://github.com/mozilla-mobile/focus-android/pull/1854#issuecomment-348493416.

PR #2036, opened 2018-01-05
Fixed failing test DeleteWebViewDataTest by adding a test item to …

shahbazahmed1269:
…local storage in test.html.
This fixes Issue #1840 .

### DIFF
--- a/app/src/androidTest/assets/test.html
+++ b/app/src/androidTest/assets/test.html
@@ -26,7 +26,8 @@
 <p id="status"></p>
 
 <script type="text/javascript">
-        // Add a key value pair to local storage since WebViewDataTest asserts that the local storage directory should exist.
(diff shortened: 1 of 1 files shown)


app/src/androidTest/assets/test.html
```diff
@@ -26,6 +26,7 @@ <h2>Service worker</h2>
 <p id="status"></p>
 
 <script type="text/javascript">
+        localStorage.setItem('focusLocalStorageTest', 'testValue');
```
mcomella: nit: can you add a comment explaining why we do this?

mcomella: Also, it'd be great to add a newline below this to separate the intentions between the two sections of code.

shahbazahmed1269:
@mcomella Made the changes you requested. Let me know if the comment explains the issue appropriately. Thanks.

PR #2043, opened 2018-01-11
Closes #2042: Don't collect default search engine name.

### LINKED ISSUE #2042: For custom search engines, only collect "custom" when sending default search engine telemetry
In #961 I didn't remove the default search engine name that we're collecting from telemetry, and we can't collect that.

If the user is using a default search engine, we should send that name (Google, Yandex etc) and if a custom search engine is used, send "custom".

liuche:
I removed this from the clicks on the preferences, but didn't remove it from the preferences dump.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
+++ b/app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
@@ -241,6 +241,7 @@ public class SearchEngineManager extends BroadcastReceiver {
         try {
             for (String engine : engines) {
                 final InputStream engineInputStream = new ByteArrayInputStream(prefs.getString(engine, "").getBytes(StandardCharsets.UTF_8));
+                // Search engine identifier is the search engine name.
                 customEngines.add(SearchEngineParser.load(engine, engineInputStream));
             }
         } catch (IOException e) {
--- a/app/src/main/java/org/mozilla/focus/telemetry/TelemetrySettingsProvider.java
+++ b/app/src/main/java/org/mozilla/focus/telemetry/TelemetrySettingsProvider.java
@@ -61,6 +61,9 @@ import org.mozilla.telemetry.measurement.SettingsMeasurement;
                 // However we still want to report the current search engine of the user.
                 // Therefore we inject this value at runtime.
                 value = SearchEngineManager.getInstance().getDefaultSearchEngine(context).getName();
+            } else if (SearchEngineManager.getInstance().isCustomSearchEngine((String) value, context)) {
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
```diff
@@ -241,6 +241,7 @@ private synchronized void loadFromDisk(Context context) {
         try {
             for (String engine : engines) {
                 final InputStream engineInputStream = new ByteArrayInputStream(prefs.getString(engine, "").getBytes(StandardCharsets.UTF_8));
+                // Search engine identifier is the search engine name.
```
mcomella: nit: what is the search engine identifier? `engine`? Maybe it should be renamed to `searchEngineName` or `searchEngineID` to make the comment clearer?

Maybe it'd be good to elaborate on how the code works. Right now I see:
- we get a value from the shared prefs about the engine
- This is passed into the engine object and becomes its identifier
- a name is extracted from the engine XML and that's the name field

wrt your comment, is the value in shared prefs the name such that `engine.identfier == engine.name`? If so, maybe the comment is too far extracted from the code where the names are set equal to each other that it'd go out of date... I'm not sure. Maybe there's a better way/place to specify this assumption?

app/src/main/java/org/mozilla/focus/telemetry/TelemetrySettingsProvider.java
```diff
@@ -61,6 +61,9 @@ public boolean containsKey(String key) {
                 // However we still want to report the current search engine of the user.
                 // Therefore we inject this value at runtime.
                 value = SearchEngineManager.getInstance().getDefaultSearchEngine(context).getName();
+            } else if (SearchEngineManager.getInstance().isCustomSearchEngine((String) value, context)) {
+                // We only want to collect "custom" or "bundled", not the search engine itself.
```
mcomella: Why is "custom" or "bundled" relevant if we just specify "custom"? I'd suggest removing "bundled", if it's relevant.

Also, I'd include our motivation, i.e. explain that we don't want to collect personal info, like a custom search engine, so we replace it with a static value.

app/src/main/java/org/mozilla/focus/search/SearchEngineManager.java
liuche: This is a throwback from integrating with the existing `SearchEngine` format for non-custom search engines. For that there's a separate engineName and engineIdentifier, but for custom engines, we just use engineName for both (because the user enters a name and a url). I'll add a comment into the SearchEngine class.

Issue #1826, opened 2017-11-28
Consider adding screenshot-tests for regression testing


mcomella:
[`Facebook's screenshot-tests-for-android`](https://github.com/facebook/screenshot-tests-for-android) is a library for running a test that will take a screenshot of various screens in the app and return no error if all of the screenshots are identical – this prevents accidental UI changes, which is helpful for not regressing polish changes after we make them. If the changes were intentional, you can ignore the errors.

We should consider adding this to our CI suite.

mcomella:
triage: No-Jun agrees this would be nice to have eventually.

grubdragon:
Hey so I'm trying to add this to the gradle file right now

Sdaswani:
Thanks @grubdragon I'm hoping @npark-mozilla can review for you.

Sdaswani:
@boek No-Jun is away, can you evaluate the old PR and see if we even get value from doing this?

npark-mozilla:
This is a good research project to perform, (if we incorporate this to our l10n screenshot tests, then we can identify UI issues right away - which'll be awesome) but currently we don't have a bandwidth to properly work on it at the moment.  @AaronMT , if you have some time, could you assess how much effort we would need to use this?

PR #2060, opened 2018-01-19
Added basic gradle setup for screenshot-tests-for-android

grubdragon:
#1826 Tried to add the gradle setup, hopefully didn't mess up. Would love a review. @mcomella @Sdaswani

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -6,6 +6,7 @@ apply plugin: 'jacoco-android'
 apply plugin: 'pmd'
 apply plugin: 'checkstyle'
 apply plugin: 'kotlin-android-extensions'
+apply plugin: 'com.facebook.testing.screenshot'
 apply from: "$project.rootDir/tools/gradle/versionCode.gradle"
 
 android {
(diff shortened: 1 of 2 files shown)


Sdaswani:
@npark-mozilla is this something you can review while @pocmo is on pat leave?

npark-mozilla:
@Sdaswani : @pocmo would know better regarding the gradle stuff, but this seems pretty harmless, I can take a look.

app/build.gradle
```diff
@@ -183,6 +184,10 @@ dependencies {
     debugCompile "com.squareup.leakcanary:leakcanary-android:$leakcanary_version"
     compile "com.squareup.leakcanary:leakcanary-android-no-op:$leakcanary_version"
 
+    // screenshot-tests-for-android : Layout Hierarchy Plugin dependencies
```
npark-mozilla: If this is used for UI Tests, then perhaps we could use androidTestCompile instead of compile command?

grubdragon: Seems like http://facebook.github.io/screenshot-tests-for-android/ this page compiles normally. But I have very little idea about gradle, so I'm willing to change it, if need be.

npark-mozilla: it is a convention for libs that are used for UI Tests - it's more for the style of the gradle script.  If it works, it'd be great if we change it.

Sdaswani:
@grubdragon i think @npark-mozilla had a suggestion here? https://github.com/mozilla-mobile/focus-android/pull/2060#discussion_r162790267

pocmo:
Thank you for this PR @grubdragon. Do you have experience in running this tool? Specifically in automation. This PR looks like a good first step - but without actually running the tool (and reporting the results) I'm not sure if we gain much by merging this as-is.

PR #2063, opened 2018-01-20
Make some static fields final

gaul:
Adding final prohibits modification and allows initialization of
primitive and String fields at compile time instead of runtime in
clinit:

https://developer.android.com/training/articles/perf-tips.html#UseFinal

Found via error-prone.


Sdaswani:
Feel free to merge after CI is green.

gaul:
@Sdaswani buddybuild never completed; how can I restart CI?

Sdaswani:
@npark-mozilla how can we get BB to run CI on this PR? thanks for the assist.

npark-mozilla:
Ah, it's because the PR is in gaul, instead of created directry as a branch in this repo.  There is a setting in Buddybuild that enables this as an exception, but I can't seem to find it. @liuche , I remember you fixed this for Emily right? Could you give us a help?

mcomella:
This looks fine with BB: it just runs build distribution and UI tests and we can back it out of master if the UI tests fail.

mcomella:
Thanks gaul!

PR #2064, opened 2018-01-22
Closes #1974 Changing orientation while entering search engine details wipes entered data

### LINKED ISSUE #1974: Changing orientation while entering search engine details wipes entered data
This is on Pixel 2, Android O
STR:
Go to Settings -> Search Engine List -> Add Search engines
Enter details of new search engine
without saving or exiting, with the entered data shown, change orientation

Expected:
- Entered data survives
Actual:
Both fields are wiped clean

This bug is probably uncovered by #1924 fix.  Not sure this is a release blocker, but it would make users quite annoyed if they ever change the screen orientation for whatever reason during the add search engine step. @bbinto , @Sdaswani thoughts?

siddware:
Implemented saving and restoring of view state during configuration changes. Closes #1974


app/src/main/java/org/mozilla/focus/search/ManualAddSearchEnginePreference.java
```diff
@@ -102,4 +135,48 @@ public void setSearchQueryErrorText(final String err) {
...
+            return "ManualAddSearchEnginePreference.SavedState{"
+                    + Integer.toHexString(System.identityHashCode(this))
+                    + " search_engine_name=" + searchEngineName
+                    + " search_query=" + searchQuery
+                    + "}";
+        }
+
+        public static final Parcelable.Creator<SavedState> CREATOR =
+                new Parcelable.Creator<SavedState>(){
+                    @Override
```
sputnikci: \[Checkstyle\] ERROR: '\{' is not preceded with whitespace\.

app/src/main/java/org/mozilla/focus/search/ManualAddSearchEnginePreference.java
```diff
@@ -102,4 +135,48 @@ public void setSearchQueryErrorText(final String err) {
...
+        public String toString() {
+            return "ManualAddSearchEnginePreference.SavedState{"
+                    + Integer.toHexString(System.identityHashCode(this))
+                    + " search_engine_name=" + searchEngineName
+                    + " search_query=" + searchQuery
+                    + "}";
+        }
+
+        public static final Parcelable.Creator<SavedState> CREATOR =
+                new Parcelable.Creator<SavedState>(){
```
sputnikci: \[Checkstyle\] ERROR: '\{' is not preceded with whitespace\.

app/src/main/java/org/mozilla/focus/search/ManualAddSearchEnginePreference.java
```diff
@@ -49,9 +54,37 @@ protected View onCreateView(ViewGroup parent) {
...
+    protected Parcelable onSaveInstanceState() {
+        Parcelable superState = super.onSaveInstanceState();
+        SavedState savedState = new SavedState(superState);
+        savedState.searchEngineName = engineNameEditText.getText().toString();
+        savedState.searchQuery = searchQueryEditText.getText().toString();
+        return savedState;
+    }
+
+    private void updateState() {
+        if (savedSearchEngineName != null) {
```
noloman: I think it would be better to null-check on the `engineNameEditText` and the `searchQueryEditText` themselves, as if you do not and the `EditText` is null, you'll crash the app; while if you simply set a `null` `String` on an existing `EditText` you'll see an empty text in the worst case scenario

siddware: Thanks, fixed it :)

siddware:
@pocmo You are right, thanks :)

PR #2072, opened 2018-01-26
Google test cloud migration from BB - WIP

npark-mozilla:


npark-mozilla:
@pocmo I added the scopes to all tasks in taskcluster.yml but it still complains of insufficient scopes - where did I miss?

npark-mozilla:
Pending work here is as follows:
1. Decide which PR job can access the taskcluster secrets (I understand that it'll be only the Master builds task)
2. Make sure get-google-firebase-token.py can properly retrieve the secret
3. Implement pass/fail status report in the taskcluster

@pocmo , probably need to run by you at least for tasks 1 and 2.

pocmo:
Merged this as part of my PR. Thank you @npark-mozilla :)

Issue #2128, opened 2018-02-21
Google maps get location should be inactive 


sv-ohorvath:
Build: Klar+GeckoView #3898

STR:
1. Go to maps.google.com
2. Tap the get location button.

Expected results:
As on Focus, the button is inactive. Can't ask for location. 

Actual results:
You can ask to get location, but with no result, it just keeps searching.
Recording: https://drive.google.com/open?id=1AaLQX8qsPUiYaVUZqJh4th2CtJeMT824

ekager:
Some more information - with Android WebView, we can disable these permissions so that the WebView cannot even request them (https://github.com/mozilla-mobile/focus-android/blob/master/app/src/webview/java/org/mozilla/focus/web/WebViewProvider.java#L119), but with GeckoView, we can't disable these currently and if we do not implement the callbacks, then the user will just see the browser as not working. GeckoSessionSettings would need to add a way for us to disable these permissions.

pocmo:
I filed an issue to determine whether GeckoView can do the same as WebView:
https://bugzilla.mozilla.org/show_bug.cgi?id=1440710

To reduce risk I would avoid adding location etc. as a new feature unless @bbinto thinks it's a feature we want to ship in 5.0.

ekager:
I think without setting any PermissionDelegate, this issue is now solved. It's not "disabled" but it doesn't continue to search or prompt the user, which seems to be what Focus does as well. Closing.

PR #2136, opened 2018-02-21
Adding Permission Delegate to Klar+GV

ekager:


app/src/gecko/java/org/mozilla/focus/web/GeckoViewPermissionsDelegate.kt
```diff
@@ -0,0 +1,115 @@
...
+
+import org.mozilla.focus.R
+import org.mozilla.focus.permissions.IPermissionsDelegate
+import org.mozilla.gecko.GeckoSession
+
+import java.util.Locale
+
+class GeckoViewPermissionsDelegate(private val context: Context,
+                                   private val geckoSession: GeckoSession) :
+        GeckoSession.PermissionDelegate, IPermissionsDelegate {
```
pocmo: nit: Following the coding conventions the formatting should be:

```Kotlin
class GeckoViewPermissionsDelegate(
    private val context: Context,
    private val geckoSession: GeckoSession
) : GeckoSession.PermissionDelegate,
    IPermissionsDelegate {
```


https://kotlinlang.org/docs/reference/coding-conventions.html#formatting

app/src/gecko/java/org/mozilla/focus/web/GeckoViewPermissionsDelegate.kt
```diff
@@ -0,0 +1,115 @@
...
+    private var callback: GeckoSession.PermissionDelegate.Callback? = null
+
+    override fun onRequestPermissionsResult(permissions: Array<String>, grantResults: IntArray) {
+        if (callback == null) {
+            return
+        }
+
+        val cb = callback
```
pocmo: You can rewrite this as:

```Kotlin
val cb = callback ?: return
```

With that `cb` should never be null and you can avoid using `!!`

app/src/gecko/java/org/mozilla/focus/web/GeckoViewPermissionsDelegate.kt
```diff
@@ -0,0 +1,115 @@
...
+                Log.w(LOGTAG, "Unknown permission: " + type)
+                callback.reject()
+                return
+            }
+        }
+
+        val title = context.getString(resId, Uri.parse(uri).authority)
+        val prompt = geckoSession.promptDelegate as GeckoViewPrompt
+        prompt.promptForPermission(session, title, callback)
+    }
```
pocmo: What permissions do we need this for? I think we shouldn't introduce location/camera/audio as new features at this time - unless we absolutely need to.

ekager: With Android WebView, we can disable these permissions so that the WebView cannot even request them (https://github.com/mozilla-mobile/focus-android/blob/master/app/src/webview/java/org/mozilla/focus/web/WebViewProvider.java#L119), but with GeckoView, we can't disable these currently and if we do not implement the permissions, then the user will just see the browser as not working - see #2128.

pocmo:
Afaik this isn't needed anymore?

Issue #1405, opened 2017-09-27
Please include F-Droid as one of the options for getting Mozilla Klar.


eighthave:
Firefox Klar has been included in F-Droid since 2017-07-04, built directly from source using the official Mozilla build system. As with all apps included in f-droid.org, we also run our automatic verification process on it, you can see Klar diffs here:

    https://verification.f-droid.org/org.mozilla.klar_11.apk.diffoscope.html
    https://verification.f-droid.org/org.mozilla.klar_10.apk.diffoscope.html

Once the last reproducibility issues are fixed, like #1401, then the APK that f-droid.org ships will be exactly the same, except for the APK signature.

It is also possible to include the Mozilla APK signature in f-droid.org provided that Klar builds reproducibly. We test this by building from source on our own infrastructure, then copying the developer's APK signature into our APK. If it verifies, we ship it.

bbinto:
Do we know how many users we will be supporting by doing this?

eighthave:
A couple of years ago, we turned off the weblogs.  We had about 50
millions app downloads from the main f-droid.org app repository at that
time.  Its hard to say more from the f-droid side of things since we
actively avoid tracking users.  Now that we've implemented a bunch of
tracking protections, we plan on adding back some truly anonymous usage
tracking:
https://guardianproject.info/2017/06/08/tracking-usage-without-tracking-people/

bbinto:
To confirm @eighthave - you'd like us to promote it here? https://www.mozilla.org/en-US/firefox/focus/

pocmo:
It's about adding a badge (next to the Google Play one) to the README that is shown here:
https://github.com/mozilla-mobile/focus-android

bbinto:
@eleemoz -- Elvin, is that ok?

eighthave:
Yes, that would be amazing!  I initially asked about getting it on the
Github README, but yes, ideally the F-Droid badge would be next to the
Google Play badge wherever those are used.

easyKL:
F-Droid badge still not there 😞

eighthave:
we would love to have a build flavor that builds Focus without proprietary libs, 
so we can ship it on F-Droid.

cadeyrn:
[...]

PR #2157, opened 2018-02-23
Add F-Droid badge in README.md

yvan-sraka:
https://www.f-droid.org/packages/org.mozilla.klar/


pocmo:
Thank you for this PR. 👍 

This is still something we are trying to figure out in #1405. Even though those builds are using the code as-is, they are not official builds from Mozilla. So we are not sure yet whether we want to link to them from the README.

PR #2208, opened 2018-03-02
Fix typo

GammaBurst101:
There was a typo where "thus" was spelled as "thos".
I corrected it and added a comma after "thus" because of grammar.


GammaBurst101:
Welcome :+1:

Issue #2207, opened 2018-03-02
Fix TypicalUseTest


pocmo:
This test has been failing in automation frequently.

```
android.support.test.uiautomator.UiObjectNotFoundException: UiSelector[DESCRIPTION=Navigate up]
	at android.support.test.uiautomator.UiObject.click(UiObject.java:416)
	at org.mozilla.focus.activity.TypicalUseScenarioTest.TypicalUseTest(TypicalUseScenarioTest.java:138)
	at java.lang.reflect.Method.invoke(Native Method)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at android.support.test.internal.runner.junit4.statement.RunAfters.evaluate(RunAfters.java:61)
	at android.support.test.rule.ActivityTestRule$ActivityStatement.evaluate(ActivityTestRule.java:433)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
```

pocmo:
@npark-mozilla I'll disable this test temporarily. We didn't get a single successful test run on all devices today. Can you investigate what's going wrong here?

pocmo:
[...]

PR #2216, opened 2018-03-02
Fix UITest Failures in Firebase

npark-mozilla:
Use back key instead of matching by description
Match new title text on Open In dialog


npark-mozilla:
Also fixes #2207

npark-mozilla:
From the latest test result, the screenshoterror is caused by Espresso not being ready, and not exactly sure why WebViewDataTest is failing. it seems that the file is not created?

pocmo:
LGTM 👍

pocmo:
Oh man - next run failed because of another intermittent.. :)

https://tools.taskcluster.net/groups/Q8ETRxVLQUaerIjozs1-0g/tasks/QBd5IETPQ8apu_nEth0I-A/runs/0/logs/public%2Flogs%2Flive.log

Issue #2229, opened 2018-03-05
App crashes in AverageLoadTimeObserver.java during test


npark-mozilla:
When running SwitchContextTest in Firebase, the app crashes during the test and fails.  The stacktrace shows that it's failing here:

FATAL EXCEPTION: main
Process: org.mozilla.focus.debug, PID: 6452
java.lang.IllegalAccessError: There's no active session
	at org.mozilla.focus.session.SessionManager.getCurrentSession(SessionManager.java:117)
	at org.mozilla.focus.observer.AverageLoadTimeObserver.onValueChanged(AverageLoadTimeObserver.java:28)
	at org.mozilla.focus.observer.AverageLoadTimeObserver.onValueChanged(AverageLoadTimeObserver.java:10)
	at org.mozilla.focus.architecture.NonNullObserver.onChanged(NonNullObserver.kt:14)
	at android.arch.lifecycle.LiveData.considerNotify(LiveData.java:109)
	at android.arch.lifecycle.LiveData.dispatchingValue(LiveData.java:126)
	at android.arch.lifecycle.LiveData.setValue(LiveData.java:282)
	at org.mozilla.focus.architecture.NonNullLiveData.setValue(NonNullLiveData.kt:40)
	at org.mozilla.focus.architecture.NonNullMutableLiveData.setValue(NonNullMutableLiveData.kt:19)
	at org.mozilla.focus.session.Session.setLoading(Session.java:88)
	at org.mozilla.focus.session.SessionCallbackProxy.onPageFinished(SessionCallbackProxy.java:40)
	at org.mozilla.focus.webview.FocusWebViewClient.onPageFinished(FocusWebViewClient.java:197)
	at com.android.webview.chromium.WebViewContentsClientAdapter.onPageFinished(WebViewContentsClientAdapter.java:531)
	at org.chromium.android_webview.AwContentsClientCallbackHelper$MyHandler.handleMessage(AwContentsClientCallbackHelper.java:188)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loop(Looper.java:154)
	at android.app.ActivityThread.main(ActivityThread.java:6119)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)

[...]

PR #2232, opened 2018-03-05
Pass session object to AverageLoadTimeObserver. (#2229)

pocmo:
When running UI tests it sometimes has happened that there was no "current session"
anymore at the time a page finished loading. In addition to that there can be
multiple concurrent sessions - so keeping track of the session this observer
observes should make this more robust.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/observer/AverageLoadTimeObserver.java
+++ b/app/src/main/java/org/mozilla/focus/observer/AverageLoadTimeObserver.java
@@ -1,9 +1,11 @@
 package org.mozilla.focus.observer;
 
 import android.os.SystemClock;
+import android.support.annotation.NonNull;
 import android.util.Log;
 
 import org.mozilla.focus.architecture.NonNullObserver;
+import org.mozilla.focus.session.Session;
 import org.mozilla.focus.session.SessionManager;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/observer/AverageLoadTimeObserver.java
```diff
@@ -24,8 +31,7 @@ protected void onValueChanged(Boolean loading) {
             }
         } else {
             if (loadStarted) {
-                Log.i(LOG_TAG, "Loaded page at " +
-                        SessionManager.getInstance().getCurrentSession().getUrl().getValue());
+                Log.i(LOG_TAG, "Loaded page at " + session.getUrl().getValue());
```
ekager: We don't need to null check session here?

pocmo: We registered this observer on a property of the session object in BrowserFragment. So for the current version of the code the reference can't be null when AverageLoadTimeObserver is created (and we never clear it inside the observer).

But yeah, I think I'm going to add @NonNull to the constructor to make this more clear and avoid mistakes in the future.

Issue #2047, opened 2018-01-12
BUG: App incompatiblity with Camera cutout of new Essential PH-1


vivekkrish:
- Firefox Focus 4.0.1 (Build #13472157)

- [Essential PH-1](https://www.essential.com/)
	Android 8.0.0
	Security Patch Dec 5, 2017
	Build number OPM1.170911.213

- User agent: Mozilla/5.0 (Linux; Android 8.0.0) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Focus/4.0.1 Chrome/63.0.3239.111 Mobile Safari/537.36

	Other HTTP headers:
	```	
	Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
	Accept-Encoding: gzip, deflate
	Accept-Language: en-US
	Host: duckduckgo.com
	User-Agent: Mozilla/5.0 (Linux; Android 8.0.0) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Focus/4.0.1 Chrome/63.0.3239.111 Mobile Safari/537.36
	UPGRADE-INSECURE-REQUESTS: 1
	X-REQUESTED-WITH:
	```

- How to reproduce:
	1. Buy Essential PH-1
	2. Install Firefox Focus
	3. Profit?? 😝  
	
The app itself runs fine, but there are minor rendering issues of the Android status bar when the app is active (please see screenshots below). The height/padding of the bar is not being automatically adjusted to account for the unusually tall status bar on this phone (due to the in-display camera cutout).

Screenshot 1 (Homeview) | Screenshot 2 (Loaded page)
------------- | --------------
![Screenshot 1](https://i.imgur.com/gyCZM0Z.png) | ![Screenshot 2](https://i.imgur.com/1Ncj5T9.png)

Please do let me know if the above screenshots don't convey the issue properly (since you cannot see the camera cutout). If needed, I can take (and share) a photo of my actual phone running the app.

The dev team at Essential has asked owners of this phone to notify app developers directly when encountering this type of issue and make you aware of this thread on the [r/Essential](https://www.reddit.com/r/essential) subreddit (https://www.reddit.com/r/essential/comments/7amvxr/app_compatibility_with_camera_cutout_in_essential/).

In the OP of this thread, please refer to the section "For Developers", which explains the necessary steps to be taken to enable apps to dynamically set the status bar height.

[...]

PR #2264, opened 2018-03-08
Support for camera cutouts

fercarcedo:
This PR fixes content overlapping with the status bar on devices with a camera cutout.

The fix consists of replacing the hardcoded 25dp values with the actual status bar height via the WindowInsets API.

Below you can find screenshots of before and after the change was made (taken from an Android P emulator thanks to P's notch simulation feature).

Before:
focus_cameracutout_before_1
focus_cameracutout_before_2
focus_cameracutout_before_3
focus_cameracutout_before_4

After:
focus_cameracuout_after_1
focus_cameracutout_after_2
focus_cameracutout_after_3
focus_cameracutout_after_4

#2047

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -34,7 +34,6 @@ import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.Window;
-import android.view.WindowInsets;
 import android.view.WindowManager;
 import android.webkit.CookieManager;
 import android.widget.FrameLayout;
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -67,6 +67,7 @@ import org.mozilla.focus.utils.ColorUtils;
 import org.mozilla.focus.utils.DownloadUtils;
 import org.mozilla.focus.utils.DrawableUtils;
 import org.mozilla.focus.utils.Features;
+import org.mozilla.focus.utils.StatusBarUtils;
 import org.mozilla.focus.utils.UrlUtils;
 import org.mozilla.focus.web.Download;
 import org.mozilla.focus.web.IWebView;
@@ -206,11 +207,10 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
 
         urlBar = view.findViewById(R.id.urlbar);
         statusBar = view.findViewById(R.id.status_bar_background);
-        statusBar.setOnApplyWindowInsetsListener(new View.OnApplyWindowInsetsListener() {
+        StatusBarUtils.getStatusBarHeight(statusBar, new StatusBarUtils.StatusBarHeightListener() {
             @Override
-            public WindowInsets onApplyWindowInsets(View v, WindowInsets insets) {
-                v.getLayoutParams().height = insets.getSystemWindowInsetTop();
-                return insets;
+            public void onStatusBarHeightFetched(int statusBarHeight) {
+                statusBar.getLayoutParams().height = statusBarHeight;
             }
         });
 
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -34,6 +34,7 @@ import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.Window;
+import android.view.WindowInsets;
 import android.view.WindowManager;
 import android.webkit.CookieManager;
 import android.widget.FrameLayout;
@@ -205,6 +206,13 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
 
         urlBar = view.findViewById(R.id.urlbar);
         statusBar = view.findViewById(R.id.status_bar_background);
+        statusBar.setOnApplyWindowInsetsListener(new View.OnApplyWindowInsetsListener() {
+            @Override
+            public WindowInsets onApplyWindowInsets(View v, WindowInsets insets) {
+                v.getLayoutParams().height = insets.getSystemWindowInsetTop();
+                return insets;
+            }
+        });
 
         urlView = (TextView) view.findViewById(R.id.display_url);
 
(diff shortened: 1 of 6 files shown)


pocmo:
@fercarcedo detekt complained about two smells:
https://taskcluster-artifacts.net/Ip-z8ZPqT_mJi9aYQrWnWQ/0/public/detekt/detekt-report.html

pocmo:
I noticed two things while testing the patch:

* The search hint seems to depend on the height too and is off now:

search-hint

* When clicking on the URL to edit it, the input field is at the top first and then moves to the correct position with a delay. Is this something we can change? Maybe we need to delay the animation here until we know the height?

pocmo:
> Maybe we need to delay the animation here until we know the height?

Or pass the height to the other fragment (like we do with other coordinates for the animation)

pocmo:
ktlint found two issues:
> /opt/focus-android/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt:186:88: Unnecessary semicolon
> /opt/focus-android/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt:187:19: Unnecessary semicolon

pocmo:
Other than that this looks pretty sweet! Thank you so far! :)

fercarcedo:
I have solved both problems. For the second one (input field first at the top and then moves to the correct position) I have created a new class ``StatusBarUtils`` which caches the status bar height so that it isn't necessary to wait for the listener every time we need that value

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -34,6 +34,7 @@
 import android.view.View;
 import android.view.ViewGroup;
 import android.view.Window;
+import android.view.WindowInsets;
```
sputnikci: \[Checkstyle\] ERROR: Unused import \- android\.view\.WindowInsets\.

app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.java
```diff
@@ -18,10 +18,12 @@
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.WindowInsets;
```
sputnikci: \[Checkstyle\] ERROR: Unused import \- android\.view\.WindowInsets\.

app/src/main/java/org/mozilla/focus/utils/StatusBarUtils.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class StatusBarUtils {
+    private static int STATUS_BAR_SIZE = -1;
+
+    public interface StatusBarHeightListener {
+        void onStatusBarHeightFetched(int statusBarHeight);
+    }
+
+    public static void getStatusBarHeight(View view, final StatusBarHeightListener listener) {
+        if (STATUS_BAR_SIZE > 0) {
```
sputnikci: \[Checkstyle\] ERROR: 'if' construct must use '\{\}'s\.

sputnikci: \[PMD\] INFO: Avoid using if statements without curly braces

fercarcedo: This seems like an error, because, in fact, there are curly brances on the if statement

app/src/main/java/org/mozilla/focus/utils/StatusBarUtils.java
```diff
@@ -0,0 +1,31 @@
+package org.mozilla.focus.utils;
+
+import android.view.View;
+import android.view.WindowInsets;
+
+/**
+ * Created by Fer on 08/03/2018.
+ */
+
```
pocmo: Please remove this auto-generated comment :)

app/src/main/java/org/mozilla/focus/utils/StatusBarUtils.java
```diff
@@ -0,0 +1,31 @@
+package org.mozilla.focus.utils;
```
pocmo: Please add our license header at the top of the file:

```
```

app/src/main/java/org/mozilla/focus/utils/StatusBarUtils.java
```diff
@@ -0,0 +1,31 @@
...
+    public static void getStatusBarHeight(View view, final StatusBarHeightListener listener) {
+        if (STATUS_BAR_SIZE > 0) {
+            listener.onStatusBarHeightFetched(STATUS_BAR_SIZE);
+        } else {
+            view.setOnApplyWindowInsetsListener(new View.OnApplyWindowInsetsListener() {
+                @Override
+                public WindowInsets onApplyWindowInsets(View v, WindowInsets insets) {
+                    STATUS_BAR_SIZE = insets.getSystemWindowInsetTop();
+                    listener.onStatusBarHeightFetched(STATUS_BAR_SIZE);
+                    return insets;
```
pocmo: Is it required to remove the listener here again?

app/src/main/java/org/mozilla/focus/utils/StatusBarUtils.java
fercarcedo: Done

app/src/main/java/org/mozilla/focus/utils/StatusBarUtils.java
fercarcedo: Done

app/src/main/java/org/mozilla/focus/utils/StatusBarUtils.java
fercarcedo: I'm not sure if I'm understanding the question correctly, what this code does is if we don't have the status bar height cached it registers a listener and when it gets the status bar height saves it for later usage and sends it back to the caller.

pocmo: What I mean is: Should we call view.setOnApplyWindowInsetsListener(null) here once we got a value or do we need to stay subscribed?

pocmo: I was thinking of things like OnPreDrawListener where you explicitly clear the listener after receiving what you want. To avoid keeping strong references in memory.

fercarcedo: Yes, after all it's no longer needed, so why keep references to the listener (and thus to the Activity). Added on latest commit

pocmo:
Thank you @fercarcedo! I landed this as a squashed commit: ba973149aa72fb227fadb8c2878d605f5408379c

PR #2268, opened 2018-03-09
Alpha animation of wordmark after erasing

vladioan3:
I have investigated multiple scenarios where this animation would occour and how it intersects with the translation animation and fragment creation.
The implementation in this PR is more consistent because the animation is started when the fragment is created.


pocmo:
#1085

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -218,6 +221,16 @@ class UrlInputFragment :
...
+    private fun animateWordMark() {
+        keyboardLinearLayout.alpha = 0f
+        keyboardLinearLayout.animate()
+                .setInterpolator(AccelerateDecelerateInterpolator())
+                .alpha(0f)
+                .alphaBy(1f)
+                .setDuration(WORD_MARK_ANIMATION_DURATION.toLong())
+                .start()
+    }
+
```
pocmo: Could this be added to the animation code we already have in place for the translation animation of the word mark?

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -176,6 +178,7 @@ class UrlInputFragment :
             clearView.visibility = View.VISIBLE
             searchViewContainer.visibility = View.GONE
         }
+        animateWordMark()
```
pocmo: This is a great idea. One drawback here is that the animation will also play when the app launches the first time. The bug description is not really clear about that - sorry! What UX was asking for is to only do this alpha animation if the user comes back to this screen after erasing the browsing session. In other situations (like the app start) there should no alpha animation be added.

vladioan3: Thank you for this informations. I will do this changes.

Issue #1216, opened 2017-08-28
Reported Battery Issues


bbinto:
Via GP
-  #https://play.google.com/store/apps/details?id=org.mozilla.focus&reviewId=Z3A6QU9xcFRPSFRXSVVuQ0g0VnNaRjNPcjQ3SXJuc1QtSmtOa2c1TTAwYlp4VS04R0IwWVJSbFkzM3RmRWJ5SVhaaTdMT2g4R1dQNm5vZ09wWDVTcVEzRENR
- https://play.google.com/store/apps/details?id=org.mozilla.focus&reviewId=Z3A6QU9xcFRPR3NqZDB4YzdKMEtQZ1NiZ3pSa3h3Y1g3QlZlYnp1WVYxSVhNRDg1SWtpM1VnXzlhd09lUWpiOUFTZ3pLNWFGMjVOdTd5TmRxYlloS1JvZXpB
- https://play.google.com/store/apps/details?id=org.mozilla.focus&reviewId=Z3A6QU9xcFRPRmJSVVFJNHMtNUt0RGJoemk1NERfVmVsSWNacFJPV3Ewd2M1Tk4zc1E0MU5wMHdPNXQtNVdYYmxoR2dFaXRqUkQ1Q0l5MkIzeWN5U1h1T1NJ
- https://play.google.com/store/apps/details?id=org.mozilla.focus&reviewId=Z3A6QU9xcFRPRzJ2WDRIbm5BZmV3UjFqWEJ6N0w2bDJ4MVhHdGRMRlFJd1B5bGxTeUZLa2UxSFR2WVlhZHNrZWhuUkR1VHRMeGJwZW5DVEcweGhzNDYyTGxF

bbinto:
@pocmo to run some preliminary tests

pocmo:
Probably related: #1333

bbinto:
More reports via GPS
"Draining battery on 7.1.1 stock, Xperia xz as Stamina recognized it."

mcomella:
This could also be related to #1492 – we would stop using the disk cache, so we would use the network more often and thus be using more battery than requesting the data from disk.

bbinto:
Setting this P2, and revisit again once the P1 items have been fixed to see if issues have been resolved (#1492 and #1613)

bbinto:
[...]

PR #2274, opened 2018-03-10
Pause and resume WebView investigating #1216

ekager:
Do you think this would help #1216 at all? I looked into it a bit.. Looks like a fairly common pain, there was one other solution I found here http://www.mcseven.me/2011/12/how-to-kill-an-android-webview/ that loads a static asset in onPause and saves the current URL.

### DIFF
--- a/app/src/webview/java/org/mozilla/focus/webview/SystemWebView.java
+++ b/app/src/webview/java/org/mozilla/focus/webview/SystemWebView.java
@@ -110,6 +110,18 @@ public class SystemWebView extends NestedWebView implements IWebView, SharedPref
     }
 
     @Override
+    public void onPause() {
+        super.onPause();
+        pauseTimers();
+    }
+
+    @Override
+    public void onResume() {
+        super.onResume();
+        resumeTimers();
+    }
+
+    @Override
     public void restoreWebViewState(Session session) {
         final Bundle stateData = session.getWebViewState();
 


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -186,6 +186,12 @@ public String getInitialUrl() {
     @Override
     public void onPause() {
         super.onPause();
+
+        final IWebView webView = getWebView();
+        if (webView != null) {
+            webView.onPause();
+        }
+
```
pocmo: This shouldn't be needed. WebFragment (parent) already takes care of forwarding the lifecycle methods: https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/fragment/WebFragment.java#L109

app/src/webview/java/org/mozilla/focus/webview/SystemWebView.java
```diff
@@ -109,6 +109,18 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
...
+    @Override
+    public void onPause() {
+        super.onPause();
+        pauseTimers();
+    }
+
+    @Override
+    public void onResume() {
+        super.onResume();
+        resumeTimers();
```
pocmo: This is super interesting. I haven't noticed resumeTimers()/pauseTimers() before. There's not much written about this - but it looks like it will pause JavaScript timers and this might be a good idea in Focus whenever the WebView is not in the foreground. Let's try that!

Issue #2275, opened 2018-03-10
Firefox Focus drawing under android status bar


KazaHesto:
Device: Essential PH-1
Android version: 7.1.1
Firefox Focus version: 4.1 (Build 11)

**Steps to reproduce:**
1. Open any other app
2. Long-press recent apps button to enter multi-window
3. Open Firefox Focus in the lower part of the screen
4. Drag window separator up to have Firefox Focus take the whole screen

**Actual results:**
Firefox Focus draws into the status bar

**Expected results:**
A margin is kept so content is below the status bar

**Screenshot**
![screenshot_20180310-193548](https://user-images.githubusercontent.com/19279387/37240094-5e82867c-249a-11e8-9c6b-1a62c23c4b24.png)

Maybe this is related to  #2047?

fercarcedo:
It's a problem with the cache implementation of my last pull request. The problem is now fixed. I'm going to send a pull request with the solution

pocmo:
Thank you for taking care of this, @fercarcedo!

vivekkrish:
Hi Everyone,

Not sure if this might be related, but running Focus (4.1 Build #11) in multi-window mode (Random app ABOVE, Focus BELOW), we see a static bar equivalent to the height of default status bar.

Please see screenshot below:
![image](https://user-images.githubusercontent.com/711637/37357558-248686ea-26bf-11e8-8c31-aeaf3f4e7284.png)

Would this have been addressed by either of the 2 commits above?

Thank you!

fercarcedo:
Yes, this is addressed by the commit I submitted

PR #2278, opened 2018-03-10
Correct status bar height on multiwindow

fercarcedo:
This PR solves the problem of the status bar height not adjusting when going from multiwindow to fullscreen or vice versa.

The solution consists of querying the status bar height on onResume since it's the callback we get when entering/leaving multiwindow, and we cannot remove the WindowInsets listener due to the fact that when resizing the window (such as, for example, when resizing to go fullscreen) we don't get any lifecycle callback (such as onStart or onResume), instead the listener gets triggered.

#2275


pocmo:
detekt is complaining about a wildcard import:
https://taskcluster-artifacts.net/IdHSjVd3Rq-Kex_yRxpKlw/0/public/detekt/detekt-report.html

fercarcedo:
The wildcard import is now removed

pocmo:
LGTM. Found another issue and filed #2316.

pocmo:
Due to a merge conflict I'm landing this separately in #2317

PR #2293, opened 2018-03-13
Closes #907 Add option to disable JavaScript

### LINKED ISSUE #907: [Meta] Add option for disabling JavaScript
Pros:
* It makes loading a website faster
* It is a good protection against fingerprinting

Cons:
* It breaks some websites

ekager:
I also need to file an issue to add to GV WebViewProvider when we can

### DIFF
--- a/app/src/webview/java/org/mozilla/focus/web/WebViewProvider.java
+++ b/app/src/webview/java/org/mozilla/focus/web/WebViewProvider.java
@@ -127,6 +127,7 @@ public class WebViewProvider {
     public static void applyAppSettings(Context context, WebSettings settings) {
         // We could consider calling setLoadsImagesAutomatically() here too (This will block images not loaded over the network too)
         settings.setBlockNetworkImage(Settings.getInstance(context).shouldBlockImages());
+        settings.setJavaScriptEnabled(!Settings.getInstance(context).shouldBlockJavaScript());
     }
 
     /**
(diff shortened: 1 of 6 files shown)


pocmo:
> I also need to file an issue to add to GV WebViewProvider when we can

I filed one some time ago:
https://bugzilla.mozilla.org/show_bug.cgi?id=1440708

pocmo:
Looks like Brian provided final strings. Can you update the PR for merging? :)

app/src/webview/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -127,6 +127,7 @@ private static void configureDefaultSettings(Context context, WebSettings settin
     public static void applyAppSettings(Context context, WebSettings settings) {
         // We could consider calling setLoadsImagesAutomatically() here too (This will block images not loaded over the network too)
         settings.setBlockNetworkImage(Settings.getInstance(context).shouldBlockImages());
+        settings.setJavaScriptEnabled(!Settings.getInstance(context).shouldBlockJavaScript());
```
pocmo: This would only work for new WebView instances, right? So if I browse, go to settings, flips this toggle and go back to my browsing session then this WebView doesn't know about that, right? I think we had code somewhere that checks if a blocking setting has changed and updates the existing WebView - do we need to do that for JavaScript too?

ekager: I think this is being handled already. `applyAppSettings()` in `WebViewProvider` is called by `onSharedPreferenceChanged()` in `SystemWebView`.

pocmo: Okay, perfect. :)

PR #2304, opened 2018-03-15
The uneraseable "y"

vladioan3:
I have investigated some scenarios and I was able to reproduce this issue on Samsung Galaxy Tab 2, Samsung Galaxy S7 and Android Emulator using stock keyboard and Gboard.
This issue can be reproduced by enabling predictive text or autosuggest for your keyboard.
In this PR I have added TYPE_TEXT_VARIATION_PASSWORD flag to the view. This will disable both predictive text and autosuggest.


app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -146,6 +147,7 @@ class UrlInputFragment :
 
         urlView.setOnFilterListener(this)
         urlView.imeOptions = urlView.imeOptions or ViewUtils.IME_FLAG_NO_PERSONALIZED_LEARNING
+        urlView.inputType = InputType.TYPE_TEXT_VARIATION_PASSWORD
```
pocmo: Interesting that this is fixed by this. Are there any other options? Setting the input type to password might have other unwanted side effects.

vladioan3: You can find below all the solutions that I found and tried 

1) I have tried to use in some cases SpannableStringBuilder instead of Editable in the code. I have found some information related to SpannableStringBuilder where the tip was to use this instead Editable to be able to solve similar issues. No success with this.

2) Using TYPE_TEXT_VARIATION_PASSWORD flag I was able to fix this issue because seems to prevent word prediction. As a down side on non-English keyboards this may disable some of the features such as toggling the language.

3) Using TYPE_TEXT_FLAG_NO_SUGGESTIONS may solve this issue on some devices where the IME implementation supports that. Taking this into account, we can suggest to the IME this flag but this issue wont be fixed for all devices.

PR #2305, opened 2018-03-15
Switch from findbugs to spotbugs #1433 

vladioan3:
This PR contains the migration from Findbugs to Spotbugs.


pocmo:
Please add the issue number in you commit message so that it will be linked to the issue.

One part of #1433 is exploration/verification. Have you used spotbugs before? Does it find more or less than findbugs (I see you added one error to the exclude list)? Is it faster? How is the adoption in other Android projects? You provided the *how* (and that's great! 👍). The only thing that's missing for actually switching is the *why* to decide *if*. Please comment on #1433 with everything you know! That would really help us here. :)

app/build.gradle
```diff
@@ -125,6 +127,9 @@ android {
     }
 
     sourceSets {
+        main {
+            java.srcDirs = ['src/main/java']
+        }
```
pocmo: Why is this needed? Doesn't this source set exist by default? Is spotbugs not able to find the default source sets? Please add a comment here explaining why we do this. :)

app/build.gradle
```diff
@@ -270,6 +275,7 @@ repositories {
 }
 
 dependencies {
+
```
pocmo: nit: unnecessary whitespace.

app/build.gradle
```diff
@@ -278,11 +284,8 @@ dependencies {
 
     compile "android.arch.lifecycle:extensions:$architecture_components_version"
 
-    compile ("com.google.code.findbugs:annotations:3.0.1") {
-        // We really only need the SuppressFBWarnings annotation, everything else can be ignored.
-        // Without this we get weird failures due to dependencies.
-        transitive = false
-    }
+    compileOnly 'net.jcip:jcip-annotations:1.0'
```
pocmo: This replaces the suppress annotations etc.? Is there no official package from spotbugs?

build.gradle
```diff
@@ -5,10 +5,14 @@ buildscript {
     ext.support_libraries_version = '27.0.2'
     ext.espresso_version = '3.0.1'
     ext.kotlin_version = '1.2.30'
+    ext.spotbugs_version = '3.1.2'
 
     repositories {
         google()
         jcenter()
+        maven {
+            url "https://plugins.gradle.org/m2/"
```
pocmo: Oh, this library is not on jcenter?

app/build.gradle
vladioan3: We should depend on both of spotbugs-annotations and net.jcip:jcip-annotations:1.0 instead of findbugs:annotations. This can be found in Spotbugs documentation.

build.gradle
vladioan3: Is on jcenter but I forgot to cleanup the file. Sorry for that.

app/build.gradle
vladioan3: Spotbugs is able to find the default source sets.

vladioan3:
"I see you added one error to the exclude list"
Seems like Spotbugs can find new errors that doesn't exists in Findbugs.

pocmo:
Thank you! I landed this with a conflict resolved in #2354.

vladioan3:
Cool. Thank you! 👍

Issue #2342, opened 2018-03-19
After opening a page in custom tab with "Open in Firefox Focus" it still appears to be in custom tab mode


npark-mozilla:
On:
FocusWebViewUniversalDebug
STR:
Open app, click a link to open in custom tab, then select 'Open in Firefox Focus' 
After the page is opened in Focus app, press Home to exit
Now select the app from step 1
Expected:
Since it is opened in Focus app, it should either show custom tab view, or the original app view
Actual:
It shows the Firefox Focus instance of the page

pocmo:
Oh, yes, that's right. In this case we'd need to switch tasks to separate ourselves from the third-party app task.

Potentially a release blocker.

Issue #2340, opened 2018-03-19
Opening in Focus from custom tab does not display system notification item


npark-mozilla:
On:
FocusWebviewUniversalDebug

STR:
Open custom tab from an app, and select 'Open in Firefox Focus'
Expected:
System notification icon shows up
Actual
no system notification icon shows up

Issue #2336, opened 2018-03-19
Pressing back button after opening a site in Focus from custom tab causes Focus to crash


npark-mozilla:
On:
FocusWebviewUniversalDebug

STR:
Open an app, click a link to open in custom tab
From custom tab, select "Open in Focus"
When Focus is opened, press back button

Expected:
Focus goes back to either the originating app, or doesn't do anything
Actual:
Focus crashes

Logcat:
55.873 14528 14528 E AndroidRuntime: java.lang.RuntimeException: Unable to pause activity {org.mozilla.focus.debug/org.mozilla.focus.activity.CustomTabActivity}: java.lang.NullPointerException: Attempt to read from field 'java.lang.String org.mozilla.focus.customtabs.CustomTabConfig.id' on a null object reference
03-19 16:41:55.873 14528 14528 E AndroidRuntime: 	at android.app.ActivityThread.performPauseActivityIfNeeded(ActivityThread.java:3912)
03-19 16:41:55.873 14528 14528 E AndroidRuntime: 	at android.app.ActivityThread.performPauseActivity(ActivityThread.java:3878)
03-19 16:41:55.873 14528 14528 E AndroidRuntime: 	at android.app.ActivityThread.performPauseActivity(ActivityThread.java:3852)
03-19 16:41:55.873 14528 14528 E AndroidRuntime: 	at android.app.ActivityThread.handlePauseActivity(ActivityThread.java:3826)
03-19 16:41:55.873 14528 14528 E AndroidRuntime: 	at android.app.ActivityThread.-wrap15(Unknown Source:0)
03-19 16:41:55.873 14528 14528 E AndroidRuntime: 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1609)
03-19 16:41:55.873 14528 14528 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:106)
03-19 16:41:55.873 14528 14528 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:164)
03-19 16:41:55.873 14528 14528 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:6494)
03-19 16:41:55.873 14528 14528 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
[...]

Issue #2335, opened 2018-03-19
Opening a site using 'Open in Focus' in custom view causes tabs to be inaccessible


npark-mozilla:
STR:
Open focus, open two tabs
Go to homescreen, open another app, and open a link to be opened in Focus custom tab.
From the menu, select 'Open in Focus'

Expected:
Site opens in Focus app, and previously opened tabs are accessible
Actual:
Site opens in Focus app, but selecting sites in the tab list does nothing.

pocmo:
I can reproduce this and this looks like a release blocker for 4.1.

PR #2353, opened 2018-03-20
Switch to MainActivity when opening a custom tab in Focus. (#2342, #2340, #2336, #2335)

pocmo:
#2342, #2340, #2336, #2335

Previously we just turned the session into a non-custom-tab and re-created the
fragment. Now that custom tabs and the browser are separated we need to
explicitly switch to the browser task.


pocmo:
This should fix the issues No-Jun found. This should be uplifted to 4.1 too.

PR #2367, opened 2018-03-21
WIP Updating GeckoView and using loadData methods for Error Pages

ekager:


app/build.gradle
```diff
@@ -294,15 +294,15 @@ repositories {
...
     }
     maven {
         // x86 GeckoView builds
-        url "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.nightly.${geckoview_nightly_version}.latest.mobile.android-x86-opt/artifacts/public/android/maven"
+        url "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.nightly.latest.mobile.android-x86-opt/artifacts/public/android/maven"
     }
     maven {
         // aarch64 builds
-        url "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.nightly.${geckoview_nightly_version}.latest.mobile.android-aarch64-opt/artifacts/public/android/maven"
+        url "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.nightly.latest.mobile.android-aarch64-opt/artifacts/public/android/maven"
```
pocmo: We'll still have to use a fixed URL here. "latest" is still only a repository with only the latest version and not the specific one you are referencing below. So for now we will have to track two version numbers (for the repository and the artifact). :(

app/src/gecko/java/org/mozilla/focus/web/GeckoErrorPage.kt
```diff
@@ -0,0 +1,117 @@
...
+import android.content.Context
+import android.support.v4.util.ArrayMap
+import android.support.v4.util.Pair
+import android.webkit.WebViewClient
+import org.mozilla.focus.R
+import org.mozilla.focus.utils.HtmlLoader
+import org.mozilla.geckoview.GeckoSession
+import java.util.HashMap
+
+object GeckoErrorPage {
```
pocmo: I started to create a separate library for the error pages (so that UX can design them once for all our products). Let's see if we can mix our work together.

app/build.gradle
```diff
@@ -294,15 +294,15 @@ repositories {
         //   https://tools.taskcluster.net/index/gecko.v2.mozilla-central.nightly
 
         // ARM GeckoView builds
-        url "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.nightly.${geckoview_nightly_version}.latest.mobile.android-api-16-opt/artifacts/public/android/maven"
+        url "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.nightly.latest.mobile.android-api-16-opt/artifacts/public/android/maven"
```
pocmo: Those changes (geckoview update + error pages) might be worth two separate commits tbh. :)

PR #2371, opened 2018-03-22
Use new API for version code override

pocmo:
Since updating the Android build plugin our version code override was failing silently (sigh). However there's a new API that we can use.


mcomella:
It turns out we did run into this issue: https://github.com/mozilla-mobile/firefox-tv/issues/831

Maybe it'd be worth sharing this build code. :)

PR #2373, opened 2018-03-22
Update search config

mkaply:


Sdaswani:
@ekager And @boek the sooner this is in master the better

boek:
rebasing master now. Will fix the Taskcluster failures as well 👍

PR #2382, opened 2018-03-25
Enable error-prone plugin

gaul:
This can find a variety of bugs at compile-time.


boek:
This looks interesting 😄, but I'll have to defer to @pocmo to add any other dependencies

PR #2392, opened 2018-03-27
New String "Custom URLs" vs "Custom URL" Closes #2343

### LINKED ISSUE #2343: Inconsistencies in string use: "Custom URLs" vs "Custom URL"
It seems like there are inconsistencies in the app between "Custom URLs" and "Custom URL".
Flod mentioned that the header used here, with single case, sounds particularly confusing to him:
https://npark-mozilla.github.io/en-US/images/phoneScreenshots/Autocomplete_Menu_Item_1521232078475.png

So we wanted to make sure these inconsistencies were expected or not?
You can find the list of occurrences for both here: https://transvision.mozfr.org/?recherche=custom+url&repo=focus_android&sourcelocale=en-US&locale=fr&search_type=strings

ekager:


boek:
LGTM

app/src/main/res/values/strings.xml
```diff
@@ -128,6 +128,9 @@
     <!-- Preference category: Custom (user-defined) autocomplete URLs. -->
     <string name="preference_category_autocomplete_custom">Custom URL</string>
```
pocmo: Can we remove this one now?

ekager: I tried to remove all declarations but it formatted it... Do we need to edit all files or does removing it from the main strings.xml remove it from our translations?

boek: IIRC removing it from the main `strings.xml` will remove it from our translations on the next export/import

ekager: Had lint errors, so going to open a new issue to address unused strings

ekager: Or I can take it out and we can override the lint errors?

pocmo: I usually remove the string from strings.xml and then run export/import again - That will remove it from the translations.

Issue #2394, opened 2018-03-27
Refactor URLBar states


boek:
While fixing #1837 it feels like this area could be simplified a bit.

Assigning this to myself to expand on this thought later

bbinto:
@pocmo could this be handled by the android components

pocmo:
Yes, I expect the toolbar component to be usable in Focus too. However it will take some time until we can do the switch.

boek:
Sounds good, I'm going to close this and when the component is done we can talk about pulling it into Focus 🙌

PR #2395, opened 2018-03-27
#1837 - disappearing progress bar

boek:


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -257,7 +257,6 @@ public void onValueChanged(@NonNull Boolean loading) {
                         // We start a transition only if a page was just loading before
                         // allowing to avoid issue #1179
                         backgroundTransitionGroup.startTransition(ANIMATION_DURATION);
-                        progressView.setProgress(progressView.getMax());
```
boek: This was the simplest fix for the issue while also not regressing #1179. Ultimately I would love to revisit with #2394.

Will this potentially break something else? I ran tested it a couple times and nothing jumped out at me.

pocmo: Mhhhh. I think we added that because otherwise the "finished loading" animation was missing. Can you still see it? If it's still there and "git blame" for this line doesn't reveal any other reason then let's land this. :)

In any other case I think it would be better to regress #1179 (~ polish) than to keep #1837 (broken). :)

app/src/main/java/org/mozilla/focus/widget/AnimatedProgressBar.kt
```diff
@@ -0,0 +1,195 @@
+/* -*- Mode: Java; c-basic-offset: 4; tab-width: 4; indent-tabs-mode: nil; -*-
```
pocmo: Please remove this mode line. It's a relic from Fennec and we try to get rid of it whenever we can :)

Issue #2055, opened 2018-01-16
What's new notification shown for dot releases, despite no what's new page for the release


mcomella:
e.g. I saw this locally for the 4.0.1 -> 4.0.2 release from the play store.

ekager:
Thinking of implementing this by saving the What's New URL and check that to see if we should show highlight. Will we always get a new What's New URL for highlighted version, @pocmo ?

pocmo:
Good question. Updating the "What's new" URL for every release is an annoying workaround because SUMO can't redirect to the appropriate page based on the version number in the URL.

In an ideal world (and hopefully in a not to distant future) we wouldn't need to update the URL. That's why I would be hesitant to use the URL as an indicator.

We never really defined the desired outcome here. I assume what we want is to keep looking at the version number and only compare the major version (which should nicely work with the existing implementation).

For example:
4.2.0 -> 4.2.1 - No highlight
4.2.0 -> 4.3 - No highlight
4.3 -> 5.0 - Highlight!
(Hey, that's a great thing for writing unit tests 😄)

@bbinto Is the assumption that we only want to highlight the "What's new" item for major versions correct?
@brampitoyo Do you agree?
@boek What does the iOS version do?

ekager:
Yeah that was what I was originally going to try, but I thought that we have been creating new links for 4.0 -> 4.1 -> 4.2?

boek:
@pocmo We show it for every new version of the application right now. But I think we intended to change it to something a little more robust in the future.

brampitoyo:
I would recommend only showing what’s new if there’s something notable and user-visible. Your assumption about having one for each major version – but not minor ones – is correct.

pocmo:
> Yeah that was what I was originally going to try, but I thought that we have been creating new links for 4.0 -> 4.1 -> 4.2?

Yeah, we haven't been really consistent. Mostly because the current implementation forces us to show something though. :)
For 4.1 we are now re-using the page URL for "version 4" which should work nicely with just highlighting for major versions.

boek:
Yeah, I was thinking about this yesterday; if the need ever arises for minor version release notes we can always do something about it :)

PR #2400, opened 2018-03-28
For #2055 - Only show the whats new dot on major versions

boek:
@ekager @pocmo I would love to add another test to test the negativity of a minor version update. But couldn't figure out how to stub `context.appVersionName`. Any ideas?


boek:
Gonna write some new test cases real quick

PR #2406, opened 2018-03-29
1647: Order browser list alphabetically

MatonAnthony:
The order of the browser is based on the definition in a XML file.
This commit makes it ordered by alphabetical order.


boek:
Thanks again @MatonAnthony!

PR #2410, opened 2018-03-30
Ordered search engines alphabetically

### LINKED ISSUE #1647: fix search engine ordering
#961 [meta] Ability to add search engines and or remove existing search engines

The current ordering seems arbitrary and will be difficult for people to find what they are looking for if the method of ordering isn't apparent. 
- Yahoo
- Google
- Amazon.com
- DuckDuckGo
- Twitter
- Wikipedia
See also #1618 for a screenshot.

It seems the current order is based on the order of entry in the xml file that contains the locale search engine list. 

I would advocate to use alphabetical ordering instead. And newly added search engines would fit in rather than be appended to the bottom of the list. This will make it much easier for people to quickly find the option they are looking for. 

Alphabetical ordering is also advocated for in the material guidelines: https://material.io/guidelines/components/lists.html#lists-behavior

@bbinto do you see any issue with this? Also, can we drop the .com in Amazon? It seems odd since the other don't use that.

shikhart98:
Fixes #1647 
Signed-off-by: Shikhar Tyagi <shikhart98@gmail.com>

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -318,6 +318,7 @@ dependencies {
     implementation "com.android.support:cardview-v7:$support_libraries_version"
     implementation "com.android.support:recyclerview-v7:$support_libraries_version"
 
+
     implementation "android.arch.lifecycle:extensions:$architecture_components_version"
 
     compileOnly 'net.jcip:jcip-annotations:1.0'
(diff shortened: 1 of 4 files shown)


shikhart98:
Closing, because this issue has been already solved.

PR #2427, opened 2018-04-05
Fixed addition of duplicate custom autocomplete URL 

### LINKED ISSUE #2363: 4.1 RC - A User Should Not Be Able to Add Duplicate Custom URLs
Build from:
https://bugzilla.mozilla.org/show_bug.cgi?id=1447461  - app-focus-webview-universal-release-signed-and-aligned.apk
Pixel 2 - Android 8.1

Steps:
1) Go to Settings -> URL Autocomplete -> Custom URL 
2) Enable Custom URL and tap on it
3) add several custom URLs ( try same one different times)
4) Tap 3 dots menu and choose remove
5) Choose one of the duplicates

Expected:
- at step 3 you should not be able to add duples 

Actual:
- at step 3 you are able to add dupes

Note:
- I have added the deletion steps because i got to this enhancement by mistake - I have added Mozilla.org and mozilla.org and thought they are displayed the same name but read different but when deleting one mozilla.org both are deleted. 
- if we keep allowing dupes we should not delete all instance at once

shikhart98:
…mobile/focus-android/issues/2363

Fixes #2363 

Screenshots for the changes:
whatsapp image 2018-04-05 at 1 10 42 pm

Signed-off-by: Shikhar Tyagi <shikhart98@gmail.com>

### DIFF
--- a/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
@@ -60,16 +60,20 @@ class AutocompleteAddFragment : Fragment() {
                     .toLowerCase()
                     .removePrefixesIgnoreCase("http://", "https://", "www.")
 
-            launch(UI) {
-                when {
-                    domain.isEmpty()
-                    -> domainView.error = getString(R.string.preference_autocomplete_add_error)
-                    CustomDomains
-                            .load(activity.applicationContext)
-                            .contains(domain)
-                    -> domainView.error = getString(R.string.preference_autocomplete_duplicate_url_error)
-                    else
-                    -> saveDomainAndClose(activity.applicationContext, domain)
+            launch(CommonPool) {
+                val domains = CustomDomains.load(activity)
+                val error = when {
+                    domain.isEmpty() -> getString(R.string.preference_autocomplete_add_error)
+                    domains.contains(domain) -> getString(R.string.preference_autocomplete_duplicate_url_error)
+                    else -> null
+                }
+
+                launch(UI) {
+                    if (error != null) {
+                        domainView.error = error
+                    } else {
+                        saveDomainAndClose(activity.applicationContext, domain)
+                    }
                 }
             }
 
--- a/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
@@ -15,6 +15,7 @@ import android.view.View
 import android.view.ViewGroup
 import kotlinx.android.synthetic.main.fragment_autocomplete_add_domain.*
 import kotlinx.coroutines.experimental.CommonPool
+import kotlinx.coroutines.experimental.android.UI
 import kotlinx.coroutines.experimental.launch
 import mozilla.components.domains.CustomDomains
 import org.mozilla.focus.R
@@ -29,7 +30,6 @@ import org.mozilla.focus.utils.ViewUtils
 class AutocompleteAddFragment : Fragment() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-
         setHasOptionsMenu(true)
     }
 
@@ -60,10 +60,17 @@ class AutocompleteAddFragment : Fragment() {
                     .toLowerCase()
                     .removePrefixesIgnoreCase("http://", "https://", "www.")
 
-            if (domain.isEmpty()) {
-                domainView.error = getString(R.string.preference_autocomplete_add_error)
-            } else {
-                saveDomainAndClose(activity.applicationContext, domain)
+            launch(UI) {
+                when {
+                    domain.isEmpty()
+                    -> domainView.error = getString(R.string.preference_autocomplete_add_error)
+                    CustomDomains
+                            .load(activity.applicationContext)
+                            .contains(domain)
+                    -> domainView.error = getString(R.string.preference_autocomplete_duplicate_url_error)
+                    else
+                    -> saveDomainAndClose(activity.applicationContext, domain)
+                }
             }
 
             return true
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -593,4 +593,5 @@ be temporary, and you can try again later.</li>
 
     <!-- Content Description of Security Icon in toolbar -->
     <string name="site_security_icon_content_description">Site Security</string>
+    <string name="preference_autocomplete_duplicate_url_error">URL already exists</string>
 </resources>


shikhart98:
I don't understand! what should I do about this?

shikhart98:
Can someone review and tell whats wrong with this? My code worked fine for me! tried it on 2 different devices.

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
```diff
@@ -7,14 +7,17 @@ package org.mozilla.focus.autocomplete
 import android.app.Fragment
 import android.content.Context
 import android.os.Bundle
+import android.util.Log
```
boek: Looks like you have an unused import here

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
```diff
@@ -59,10 +61,14 @@ class AutocompleteAddFragment : Fragment() {
                     .toLowerCase()
                     .removePrefixesIgnoreCase("http://", "https://", "www.")
 
-            if (domain.isEmpty()) {
-                domainView.error = getString(R.string.preference_autocomplete_add_error)
-            } else {
-                saveDomainAndClose(activity.applicationContext, domain)
+            launch(UI) {
+                if (domain.isEmpty()) {
```
boek: This if/else expression looks like it could be cleaned up with a [when](https://kotlinlang.org/docs/reference/control-flow.html#when-expression) :)

app/src/main/res/values/strings.xml
```diff
@@ -577,4 +577,6 @@ be temporary, and you can try again later.</li>
 
     <!-- Content Description of Security Icon in toolbar -->
     <string name="site_security_icon_content_description">Site Security</string>
+    <string name="duplicate_error">URL already exists</string>
```
boek: Did you mean to have two new strings here?

shikhart98:
@boek  fixed all the issues mentioned by you :)

shikhart98:
Ruleset: style
        MaxLineLength - [AutocompleteAddFragment.kt] at focus-android/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt:67:125

@boek  what should I do about this?

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
```diff
@@ -59,10 +59,12 @@ class AutocompleteAddFragment : Fragment() {
...
                     .removePrefixesIgnoreCase("http://", "https://", "www.")
 
-            if (domain.isEmpty()) {
-                domainView.error = getString(R.string.preference_autocomplete_add_error)
-            } else {
-                saveDomainAndClose(activity.applicationContext, domain)
+            launch(UI) {
+                when {
+                    domain.isEmpty() -> domainView.error = getString(R.string.preference_autocomplete_add_error)
+                    CustomAutocomplete.loadCustomAutoCompleteDomains(activity.applicationContext).contains(domain) -> domainView.error = getString(R.string.preference_autocomplete_duplicate_url_error)
```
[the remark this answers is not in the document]
shikhart98: I fixed this too :)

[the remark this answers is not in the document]
shikhart98: see my latest commit!

boek:
Awesome! Gonna try it out and get it merged.

Thanks again for the PR!

shikhart98:
@boek Should I commit the changes requested by @csadilek now?

boek:
@shikhart98 Yep! Sorry for the delay, this process has been kind of a moving target.

shikhart98:
So what I need to do now is : 
1) import `mozilla.components.domains`
2) change function name from `CustomAutocomplete.loadCustomAutoCompleteDomains(context) to CustomDomains.load(context)`

right?

boek:
yep!

shikhart98:
Thanks :) I'll do it right now 👍

shikhart98:
Okay thanks 👍

shikhart98:
@boek  Yeah! that helped, thanks! 
I am thinking to updated my current branch for this issue with upstream and then re-write my code (not a problem for me as there are few line only )  and then commit!

shikhart98:
@boek I have made necessary changes! now you can review :)

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
```diff
@@ -60,10 +60,17 @@ class AutocompleteAddFragment : Fragment() {
...
                     .removePrefixesIgnoreCase("http://", "https://", "www.")
 
-            if (domain.isEmpty()) {
-                domainView.error = getString(R.string.preference_autocomplete_add_error)
-            } else {
-                saveDomainAndClose(activity.applicationContext, domain)
+            launch(UI) {
+                when {
+                    domain.isEmpty()
+                    -> domainView.error = getString(R.string.preference_autocomplete_add_error)
```
mcomella: I don't think this is correct style. You want to keep the arrows on the same line like:
```kotlin
domain.isEmpty() -> {
    // Multiline things here
}
```

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
```diff
@@ -60,10 +60,17 @@ class AutocompleteAddFragment : Fragment() {
...
+                    CustomDomains
```
mcomella: I don't think this is correct style. You want to keep the arrows on the same line like:
```kotlin
domain.isEmpty() -> {
    // Multiline things here
}
```

And I'd avoid having multi-line expressions as the switch expression in the when statement (e.g. store the result as a variable above). I think they should be pretty literal: "when this is an error", "when this is a success", etc.

boek: @mcomella good catch, would this be difficult to add to ktlint or detektcheck?

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
```diff
@@ -60,10 +60,17 @@ class AutocompleteAddFragment : Fragment() {
                     .toLowerCase()
                     .removePrefixesIgnoreCase("http://", "https://", "www.")
 
-            if (domain.isEmpty()) {
-                domainView.error = getString(R.string.preference_autocomplete_add_error)
-            } else {
-                saveDomainAndClose(activity.applicationContext, domain)
+            launch(UI) {
```
mcomella: Discussed on Slack:

>  I actually don’t think CustomDomains.load needs to be a suspending function: it’s not calling anything that suspends, afaict so I don’t think coroutines are necessary here
>
> That being said, using a suspending function on the UI thread creates a lot of complexity: you allow the UI thread to return to the system before your code completes. For example, one code snippet I had once… 

```kotlin
    // In a coroutine context...
     
    val screenshot = async { getScreenshot() }
    val title = async { getTitle() }
     
    screenshotView.drawable = screenshot.await() // <-- suspends here
    titleView.text = title.await() // <-- suspends here
     
    // Control returns to system at each suspension so ^ screenshot appears,
    // then title appears, and now we hide them: there's an unintended pop-in effect.
    screenshotView.alpha = 0
    titleView.alpha = 0
     
    screenshotView.animate()
    titleView.animate() 
```

> mcomella [3 minutes ago]
wrt my example, in this case, it looks like it’ll be **okay** but that may not be true if the code changes and in your code the user will have an opportunity to interact with the app (unexpectedly) before the fragment manager code gets to run (in the final case).

> I think a generally safer pattern is: 
```kotlin
    launch(CommonPool) {
        // Do suspending work here
        launch(UI) {
            // Do non-suspending UI work here to avoid accidentally giving control to system
        }
    } 
```

> Suspending on the UI thread also forces scheduling on the UI, hurting its performance but that could be negligible

boek: @mcomella Tracking auditing usage of suspend in `android-components` here: https://github.com/mozilla-mobile/android-components/issues/79. If the API changes then we can remove the async code here 👍 

I'll go ahead and address the other comments as well

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteAddFragment.kt
mcomella: I don't know how easy ktlint or detektCheck are to extend. detektCheck may have a check like this built in: I'm not sure how strict our configuration is.

PR #2430, opened 2018-04-05
Support Cookie 🍪 blocking Closes #2330

### LINKED ISSUE #2330: Support blocking of 3rd party cookies
Why
- Provide additional tracking protection for users
- Protects users during an active session

What
- User can turn on 3rd party cookie blocking

Acceptance Criteria (how do I know when I’m done?)
- 3rd party cookies are blocked and off by default
- User can turn this feature on/off via settings
- Toggle off in menu will turn off 3rd party cookie blocking as well
- Probes to see if users have setting turned off/on

@bbinto to check with Softvision to test blockage for top 10 Alexa websites

-----
More info:

We do block trackers and some ads, but not 3rd party cookies by default.

After talking to @snorp and @pocmo, it seems to not be a huge effort to include this, and promote it once GDPR goes into effect. 

Will it cause more breakage?

Hoping for 5.0

ekager:


app/src/main/res/values/strings.xml
```diff
@@ -189,6 +189,20 @@
...
+        <item>@string/preference_privacy_should_block_cookies_yes_option</item>
+        <item>@string/preference_privacy_should_block_cookies_third_party_only_option</item>
+        <item>@string/preference_privacy_should_block_cookies_no_option</item>
+    </string-array>
+
+    <!-- Three options for Cookie blocking. Yes (All cookies Blocked), Block 3rd-party cookies
+    only (Only 3rd-party Cookies Blocked), No (No cookies blocked) -->
+    <string name="preference_privacy_should_block_cookies_no_option">No</string>
+    <string name="preference_privacy_should_block_cookies_third_party_only_option">Block
+        3rd-party cookies only</string>
```
pocmo: nit: This line break inside the string is a bit unusual.

PR #2441, opened 2018-04-09
CA name can wrap in Security Popup Closes #2436

### LINKED ISSUE #2436: Klar 4.1: CA name is cut off
Please see the attached screenshot. It should be "Verfied by: Let's Encrypt" and not only "Verfied by: Let's".

screenshot_20180407-214416

ekager:
Also waiting for confirmation from UX that wrapping is okay in this case


ekager:
Got the okay for wrapping this line from UX!

ekager:
Added to 4.2 as well

Issue #2326, opened 2018-03-19
Add crash reporting via Sentry


pocmo:
Why
- GeckoView cannot leverage GPS crash reporting
- Baseline for comparing GPS crash reporting for WebView vs. Sentry crash reporting for GeckoView
- Provide stable product for Gecko-view powered Klar
- Better crash handling (e.g. search for crashes etc.)
- With GeckoView we will need our own custom crash reporting and cannot rely on Google Play crash reporting anymore.

What/ Requirements:
* [x] Enable crash reporting with sentry for release builds
* [x] Use sentry instance hosted by Mozilla
* [ ] Make sure that this works with GeckoView crash reporting ([Bug 1433968](https://bugzilla.mozilla.org/show_bug.cgi?id=1433968)) and native crashes show up in Sentry
* [x] Make Sentry DSN configurable at build time (like Adjust token)
 * [x] Configure taskcluster to pick up Sentry DSN from secrets service and configure build
* [x] Figure out with product/ux/legal if/what user-facing toggles are needed
* [x] Document in wiki (see https://github.com/mozilla-mobile/firefox-ios/wiki/Crash-Reporting-with-Sentry)
* [ ] Look into reporting to Sentry AND Google Play crash reporting

Acceptance Criteria (how do I know when I’m done?)
- Crash reporting in Sentry is accurate
- APT to view and analyze crash reporting
- Receiving crash reporting by default (in Klar, currently crash reporting is connected to "share usage data" which is off by default in Klar)

Sdaswani:
@bbinto Sentry is approved in our mobile products - can you confirm we don't need a separate toggle for this?

Sdaswani:
If we do need a toggle @bbinto then we may need UX work from @brampitoyo

mcomella:
Firefox TV's requirements are fewer but our Sentry PR is up: https://github.com/mozilla-mobile/firefox-tv/issues/429

ekager:
@Sdaswani looks like we want to receive crash reporting by default, but since Klar is opt-in for all usage-data, we may need to create a new opt-out toggle for Focus+Klar.

Sdaswani:
Can we just shut it off for Klar @ekager ?

ekager:
[...]

PR #2443, opened 2018-04-09
Add crash reporting via Sentry 🎈

ekager:


app/src/main/java/org/mozilla/focus/telemetry/SentryWrapper.kt
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+/**
+ * A Sentry user that does not contain identifiable information like UUID.
+ *
+ * Use this instead of [io.sentry.context.Context.clearUser] because that
+ * will just clear any user we set and use a Sentry-generated UUID instead.
+ */
+private class SentryUnknownUser : User(null, null, null, null)
```
sputnikci: \[detekt\] INFO: \[style/NewLineAtEndOfFile\] Checks whether files end with a line separator\.

app/src/main/java/org/mozilla/focus/activity/MainActivity.java
```diff
@@ -32,6 +33,7 @@
 
 import java.util.List;
 
+import io.sentry.Sentry;
```
sputnikci: \[Checkstyle\] ERROR: Unused import \- io\.sentry\.Sentry\.

app/src/main/java/org/mozilla/focus/telemetry/SentryWrapper.kt
```diff
@@ -0,0 +1,52 @@
...
+ */
+object SentryWrapper {
+
+    fun init(context: Context) {
+        onIsEnabledChanged(context, TelemetryWrapper.isTelemetryEnabled(context))
+    }
+
+    internal fun onIsEnabledChanged(context: Context, isEnabled: Boolean) {
+        // The BuildConfig value is populated from a file at compile time.
+        // If the file did not exist, the value will be null.
```
pocmo: But above you run "&& echo "--" > .sentry_token" on taskcluster. Why? If this is copied from Adjust: On release builds we crash the app if there's no adjust token set. To run unit tests on release builds (without the actual token) we write a dummy token on taskcluster. This doesn't seem to be needed.

app/build.gradle
```diff
@@ -469,6 +474,30 @@ android.applicationVariants.all { variant ->
...
+// -------------------------------------------------------------------------------------------------
+// Sentry: Read token from locale file if it exists (Only release builds)
+// -------------------------------------------------------------------------------------------------
+
+android.applicationVariants.all { variant ->
+    def variantName = variant.getName()
+
+    print("Sentry token: ")
+
+    if (variantName.contains("Release")) {
```
pocmo: We only want the Adjust token in release builds. However here I think we can be more open and allow the usage of sentry whenever there's a .sentry_token file present. It would be pretty nice to get crashes for our preview builds too.

tools/taskcluster/schedule-master-build.py
```diff
@@ -24,7 +24,7 @@ def generate_build_task():
 	return taskcluster.slugId(), generate_task(
 		name = "(Focus for Android) Build",
 		description = "Build Focus/Klar for Android from source code.",
-		command = ('echo "--" > .adjust_token'
+		command = ('echo "--" > .adjust_token && echo "--" > .sentry_token'
```
pocmo: See above. I do not think this is needed?

tools/taskcluster/get-sentry-token.py
```diff
@@ -0,0 +1,22 @@
...
+
+# Get JSON data from taskcluster secrets service
+secrets = taskcluster.Secrets({'baseUrl': 'http://taskcluster/secrets/v1'})
+data = secrets.get('project/focus/tokens')
+
+token_file_path = os.path.join(os.path.dirname(__file__), '../../.sentry_token')
+with open(token_file_path, 'w') as token_file:
+    token_file.write(data['secret']['sentryToken'])
+
+print("Imported sentry token from secrets service")
```
pocmo: This code has been copied a thousand times. Eventually someone should just write a script that takes three arguments:
* Name of secret (project/focus/tokens)
* Name of key (sentryToken)
* File to write to (.sentry_token)

But this would be a new issue to file... :)

pocmo:
"Closes #2326" is not completely correct. The issue mentiones GeckoView crash handling too (We can file a new issue for that!) and looking into sending crashes to Sentry *and* Google Play.

app/src/main/java/org/mozilla/focus/telemetry/SentryWrapper.kt
pocmo: I assume this is from Fire-TV? Isn't the "users affected" metric completely useless without the UUID?

pocmo: We discussed this for FireTV (see https://github.com/mozilla-mobile/firefox-tv/issues/817 and https://github.com/mcomella/firefox-tv/commit/ebd335dc425a847366b342be47ffb4f1518ca0c6#commitcomment-28574513). We will put this back in.

tools/taskcluster/schedule-master-build.py
```diff
@@ -33,15 +33,17 @@ def generate_unit_test_task(buildTaskId):
...
 
 
 def generate_code_quality_task(buildTaskId):
 	return taskcluster.slugId(), generate_task(
 		name = "(Focus for Android) Code quality",
 		description = "Run code quality tools on Focus/Klar for Android code base.",
-		command = 'echo "--" > .adjust_token && ./gradlew --no-daemon clean detektCheck ktlint lint pmd checkstyle spotbugs',
+		command = 'echo "--" > .adjust_token && ./gradlew '
+				  '--no-daemon clean detektCheck ktlint lint pmd checkstyle spotbugs',
 		dependencies = [ buildTaskId ])
```
pocmo: nit: The changes in this file are only line breaks now and are not really needed. :)

Issue #1933, opened 2017-12-06
Quick "Add" URL to autocomplete list 


antlam:
Why
- Users are currently unable to easily add URLs to the autocomplete list

What / Requirements
- [ ] Provide a convenient and quick way to add the current visited URL to autocomplete list
- [ ] Allow user to save path with URL
- [ ] (optional) Fuzzy autocomplete?

Acceptance Criteria (how do I know when I’m done?)
- User can add any URL successfully to the autocomplete list
- Current list management should stay the same (delete, view etc, via settings)
- Probes should be added and working

-----

We discussed this briefly in issue #1609, leaving it to @lime124 for more thoughts 

Should we try to add (and of course, measure) another context menu item for "Add search engine" and "Add to URL autocomplete"?

> maybe we could use the long-press context menu for this?
1. User navigates to somesportswebsite.com/traderumours
2. They decide they want to save this frequented site of theirs
3. Long-press on the URL bar 
4. User presses a button to quickly add the site they're on to this "URL autocomplete" list

pocmo:
The autocomplete case seems to be straight-forward. "Search engine" needs more thought: How do we know where to put the placeholder for the search query?

lime124:
@brampitoyo please take a look! thanks!

brampitoyo:
The UI for adding custom URL via the address bar is actually quite simple and straightforward:

![url - add via address bar](https://user-images.githubusercontent.com/51007/37436355-7265e7f0-284c-11e8-89bd-d4b6cbd38406.png)

As @pocmo wrote, if all we do is add the character content of the address bar into our database of URL to autocomplete, it’s straight-forward. Let’s definitely do this.

brampitoyo:
However, if we want to add a search engine, we’ll need to know to put the `%s` placeholder in the right place.

I did a little investigation yesterday, as well as talked with @ekager, who has been an awesome help.

Hypothesis: search URLs declare their queries in a standardised way `domain.url/[whatever]?[search/q]=%s`

Strawman solution: if this is the case, then we can just look for the `q` bit in the URL to determine whether the site is a search engine or not, and where to put our `%s` query placeholder.

Results:

* Google, Yahoo, Bing, DuckDuckGo, Twitter, Ask.com, AOL, IMDb and Shopify stores uses `q=%s`
* Wikipedia uses `search=%s`
[...]

PR #2450, opened 2018-04-11
For #1933 - Adds string for the quick add dialog

boek:


app/src/main/res/values/strings.xml
```diff
@@ -140,6 +140,9 @@
     <!-- Title of the screen for adding custom autocomplete URLs. -->
     <string name="preference_autocomplete_title_add">Add custom URL</string>
 
+    <!-- Title of the screen for adding custom autocomplete URLs. -->
```
ekager: Can we change the comment to give the new context?

PR #2462, opened 2018-04-16
Properly fetch the nimbledroid key

npark-mozilla:


tools/taskcluster/upload_apk_nimbledroid.sh
```diff
@@ -0,0 +1,16 @@
...
+This script executes upload_apk_nimbledroid.py and returns nonzero when the upload fails
+"""
+
+returnValue=$(python ./tools/taskcluster/upload_apk_nimbledroid.py)
+
+if [ ! -z "$returnValue" ]
+then
+    echo $returnValue
+    exit 1
+fi
```
pocmo: This script doesn't make any sense. You are checking whether the python script returned a non-zero exit code and if so return 1. A non-zero exit code is already enough to signal an error to taskcluster. No need to translate everything to 1 instead of the actual error.

npark-mozilla: @pocmo but it seems that despite the print lines, TC thinks the job has passed, so probably python call doesn't 'return' anything to the shell? https://tools.taskcluster.net/groups/LUhigomASxyi6HH9jqgKow/tasks/Rd58QQ1jQGClOZoJBFwqqg/runs/0/logs/public%2Flogs%2Flive.log

npark-mozilla: What the script does is to fetch the print statement from python, and convert it to shell exit value, since python doesn't seem to be able to provide return value to shell. (to my knowledge)

tools/taskcluster/upload_apk_nimbledroid.py
```diff
@@ -21,23 +21,15 @@ def uploadApk(apk,key):
 	response = requests.post(url, auth=(key, ''), headers=headers, files=apk, data=payload)
 
 	if response.status_code != 201:
-		print('Status:', response.status_code, 'Headers:', response.headers, 'Error Response:',response.json())
+		print('Error Response:',response.json())
 		exit()
```
pocmo: exit() is short for exit(0). You may want to use something like exit(1) here and then your shell script is not needed anymore.

PR #2479, opened 2018-04-18
Fix IllegalAccessError in UrlInputFragment Closes #2417

### LINKED ISSUE #2417: Crash IllegalAccessError
Don't have STR. It appears in GP and also have been able to trigger it with Monkey. 

```
2018-03-30 16:15:01.022 14940-14940/org.mozilla.focus.debug E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.mozilla.focus.debug, PID: 14940
    java.lang.IllegalAccessError: There's no active session with UUID f19478de-74f4-4a4a-a72f-543eaca1a021
        at org.mozilla.focus.session.SessionManager.getSessionByUUID(SessionManager.java:207)
        at org.mozilla.focus.fragment.UrlInputFragment.onCreate(UrlInputFragment.kt:130)
        at android.support.v4.app.Fragment.performCreate(Fragment.java:2246)
        at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1377)
        at android.support.v4.app.FragmentTransition.addToFirstInLastOut(FragmentTransition.java:1187)
        at android.support.v4.app.FragmentTransition.calculateFragments(FragmentTransition.java:1070)
        at android.support.v4.app.FragmentTransition.startTransitions(FragmentTransition.java:115)
        at android.support.v4.app.FragmentManagerImpl.executeOpsTogether(FragmentManager.java:2374)
        at android.support.v4.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManager.java:2332)
        at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:2239)
        at android.support.v4.app.FragmentManagerImpl$1.run(FragmentManager.java:700)
        at android.os.Handler.handleCallback(Handler.java:751)
        at android.os.Handler.dispatchMessage(Handler.java:95)
        at android.os.Looper.loop(Looper.java:154)
        at android.app.ActivityThread.main(ActivityThread.java:6165)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:888)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:778)
```

ekager:


ekager:
@pocmo just want to make sure this is an okay approach using `onDestroy`. I tried removing the fragment with fragment manager, but it still crashes.

boek:
@ekager now that you say that, can you call framework lifecycle methods yourself? Some other frameworks I've worked with didn't like that.

ekager:
And I realized that it also discolors the URL bar on the home screen.. Back to the drawing board.

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -126,7 +126,13 @@ class UrlInputFragment :
...
         // Get session from session manager if there's a session UUID in the fragment's arguments
         arguments?.getString(ARGUMENT_SESSION_UUID)?.let {
-            session = SessionManager.getInstance().getSessionByUUID(it)
+            try {
+                session = SessionManager.getInstance().getSessionByUUID(it)
+            } catch (i: IllegalAccessError) {
+                // When Erase FAB and Url Bar are touched in quick succession, it is possible for
+                // the session we want to not exist anymore, so we can destroy the fragment
+                onDestroy()
+            }
```
pocmo: Oh nice, you can reproduce that!? I haven't been able to do that and that issue has been around for quite some time.

I wouldn't catch the exception here - it's there to be a signal that something is very wrong.

Could we instead ask the SessionManager whether it has a session for this UUID and if not continue without a session (which means there won't be a URL etc. when editing it - but this makes sense when the session was erased.

And also add a big comment here explaining why we do that :)

ekager:
So this doesn't seem to crash anymore after adding a lot of safe calls and continuing with the null session, but it does bring you back to the darker engaged URL bar vs the pink unengaged URL bar. Again this only happens if you click the URL bar and the trash at almost the exact same time. Thoughts on the trade offs of possibly adding complexity to change this behavior? 
screenshot_20180419-152209
screenshot_20180419-152218

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -160,43 +162,49 @@ class UrlInputFragment :
...
-        urlView.imeOptions = urlView.imeOptions or ViewUtils.IME_FLAG_NO_PERSONALIZED_LEARNING
+        urlView?.setOnFilterListener(::onFilter)
+        urlView?.imeOptions = urlView.imeOptions or ViewUtils.IME_FLAG_NO_PERSONALIZED_LEARNING
 
-        urlInputContainerView.viewTreeObserver.addOnPreDrawListener(object : ViewTreeObserver.OnPreDrawListener {
-            override fun onPreDraw(): Boolean {
-                urlInputContainerView.viewTreeObserver.removeOnPreDrawListener(this)
+        if (urlInputContainerView != null) {
+            urlInputContainerView.viewTreeObserver.addOnPreDrawListener(object : ViewTreeObserver.OnPreDrawListener {
+                override fun onPreDraw(): Boolean {
```
sputnikci: \[ktlint\] WARNING: \[indent\] Unexpected indentation \(81\) \(it should be multiple of 4\) in column 1

pocmo:
> Thoughts on the trade offs of possibly adding complexity to change this behavior?

If it doesn't crash and still works as expected then I'm good with landing this and not add any more complexity.

If we have time to investigate: Instead of trying to fix UrlInputFragment I wonder if we could prevent getting into this situation. Is the order of events important? Is it first erase and then clicking the toolbar? Can we ignore clicks on the toolbar if the session is already erased? Anyhow, if this gets to complex too then let's just land this here and we do not crash anymore.

PR #2551, opened 2018-04-27
Add correct placeholder in lo strings ✌🏻

ekager:
Let's try this


boek:
I think we just have to open a PR against mozilla-l10n/focus-android-l10n. Or this will get overwritten by @MickeyMoz. Gonna merge this in to see if it fixes TaskCluster though before we do it

ekager:
Merged let's see how master looks and then decide the l10n repo strategy

pocmo:
I wonder how this placeholder got in? MickeyMoz has been refusing to open PRs because of this. :)

ekager:
Ahh. Okay should we revert and then just open an issue or tell l10n?

ekager:
Oh just saw #2557 do we need to revert still?

pocmo:
A new import should fix the issue. I'll trigger MickeyMoz so that you do not need to do it manually. :)

PR #2565, opened 2018-04-30
Request Desktop Site Closes #822

### LINKED ISSUE #822: Request Desktop Site Feature
There is no option to switch to the desktop version of the website.

I would like to work on this feature if possible.

ekager:
Hardcoded the desktop UA for now and just looked up a popular one. Let me know if this should be changed.


app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -183,4 +183,32 @@ public static String stripScheme(@Nullable String url) {
 
         return url.substring(start);
     }
+
+    public static String stripSchemeAndSubDomain(String url) {
```
sputnikci: \[PMD\] INFO: The method 'stripSchemeAndSubDomain\(String\)' has a NCSS line count of 20\.

sputnikci: \[PMD\] INFO: The method 'stripSchemeAndSubDomain\(String\)' has a cyclomatic complexity of 7\.

app/src/main/java/org/mozilla/focus/utils/Settings.kt
```diff
@@ -49,6 +49,15 @@ class Settings private constructor(context: Context) {
...
 
+    fun shouldRequestDesktopSite(): Boolean =
+            preferences.getBoolean(
+                getPreferenceKey(R.string.pref_key_performance_request_desktop_site),
+                false)
+
+    fun setShouldRequestDesktopSite(shouldRequestDesktopSite: Boolean) =
+            preferences.edit().putBoolean(getPreferenceKey(R.string.pref_key_performance_request_desktop_site),
+                shouldRequestDesktopSite).apply()
+
```
pocmo: What is this setting for? I assume this should be something we just save in the session for this "tab" (like the state of the tracking switch) and no global setting?

app/src/main/res/layout/request_desktop_check_menu_item.xml
```diff
@@ -0,0 +1,41 @@
...
+        android:id="@+id/check_menu_item_checkbox"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center_vertical|end"
+        android:checked="true"
+        android:clickable="true"
+        android:ellipsize="end"
+        android:focusable="true"
+        android:gravity="center"
+        android:paddingEnd="8dp" />
```
pocmo: We should set the text on the checkbox.

* Otherwise screenreaders won't be able to announce what this checkbox is for.
* Clicking on the text will automatically toggle the checkbox

I'm not sure whether we can change the orientation of the checkbox though. I think by default it is (checkbox - text).

app/src/webview/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -142,6 +142,17 @@ public static void disableBlocking(WebSettings settings, WebView webView) {
         CookieManager.getInstance().setAcceptCookie(true);
     }
 
+    public static void requestDesktopSite(WebSettings settings) {
+        settings.setUserAgentString("Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.113 Safari/537.36");
```
pocmo: It would be nice if we can construct the desktop UA from the WebView UA like we do for the default UA. At least the Chrome version would be good to keep in sync so that we do not confuse scripts that try to detect features based on the version.

@miketaylr may have some input too!

app/src/main/java/org/mozilla/focus/utils/Settings.kt
ekager: The specs detailed it not "carrying back" but "carry foward" as a global setting see #822

pocmo: Oh, I missed that part. I'll comment on the issue. I think this is pretty confusing.

ekager:
Changed behavior to stay within a session and also be saved when adding a homescreen shortcut.

app/src/main/java/org/mozilla/focus/session/SessionManager.java
```diff
@@ -253,11 +254,12 @@ private void createSession(Context context, Source source, SafeIntent intent, St
         addSession(session);
     }
 
-    private void createSession(Context context, Source source, SafeIntent intent, String url, boolean blockingEnabled) {
+    private void createSession(Context context, Source source, SafeIntent intent, String url, boolean blockingEnabled, boolean requestDesktop) {
```
sputnikci: \[PMD\] INFO: Avoid unused private methods such as 'createSession\(Context,Source,SafeIntent,String,boolean\)'\.

PR #2577, opened 2018-05-02
Update custom URL hint. Fixes #2504

### LINKED ISSUE #2504: Replace example.com with mozilla.org
When adding a custom URL you are given the following hint

> Example: example.com

which is a little confusing. We could be more clear by using an actual domain:

> Example: mozilla.org

image

### What / Requirements
- [ ] Replace the contents of the string

Hint reads more clearly.

daivat15:
Fixes [#2504](https://github.com/mozilla-mobile/focus-android/issues/2504).

I have left the previous example string(preference_autocomplete_add_example) as the translations for the new string(preference_autocomplete_add_example2) have not been done yet. Once they are completed, the references of the old string resource can be replaced and the string removed.

### DIFF
--- a/app/src/main/res/layout/fragment_autocomplete_add_domain.xml
+++ b/app/src/main/res/layout/fragment_autocomplete_add_domain.xml
@@ -32,7 +32,7 @@
     <TextView
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:text="@string/preference_autocomplete_add_example"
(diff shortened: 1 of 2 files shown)


boek:
Thanks @daivat15! Could you also update the fragment to use the new string? Then we'll get this merged 👍

daivat15:
I'll do that. I'll remove the old string and its translations too then.

PR #2590, opened 2018-05-07
Settings Reorganization Closes #2357

### LINKED ISSUE #2357: Reorder “Privacy” and “Performance” menu items within Settings

* Focus already has many Settings options. This makes it hard for users to navigate and understand exactly where to find which options.
* Focus’ goal as a high-privacy, no-distraction browsing means that it continue to develop more privacy features: more Settings options. A home for all privacy and security features will make these features easier to access for users who need it (which we think is a significant percentage of our user base).

### What/Requirements

* A Settings page containing main categories
* A Settings sub-page where Focus’ privacy/security/blocking features (our unique selling points) are located

* When each Settings category is broken down clearly
* When features within the privacy/security/blocking sub page is organised in a logical order, that will scale as Focus develops more features

### Background

Focus is planning to expand its suite of privacy options further by disabling JavaScript (#907) and maybe blocking even more things in upcoming versions.

To anticipate these options, we should reorganise what we already have now.

#### High-level things to consider

* Should the Settings UI have one unified heading where “Privacy” and “Performance” toggles is grouped together?
* We’ve proven that privacy is attractive to users and have clear benefits. Will performance dilute our strong privacy message? If so, maybe just call this section “Privacy”

#### Implementation-level things to consider

* One benefit of grouping under one header: we will only have to explain how things work once. Let’s make sure to say something about both privacy and performance.
* A potential sort order: from the least to the most potentially destructive in terms of breaking sites. This may not be entirely clear, though.

ekager:
If we want this in 5.1 we will have to export strings ASAP

### DIFF
--- a/app/src/main/java/org/mozilla/focus/settings/SearchSettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/settings/SearchSettingsFragment.kt
@@ -42,7 +42,8 @@ class SearchSettingsFragment : BaseSettingsFragment(),
                 navigateToFragment(InstalledSearchEnginesSettingsFragment())
                 TelemetryWrapper.openSearchSettingsEvent()
             }
-            resources.getString(R.string.pref_key_screen_autocomplete) -> navigateToFragment(AutocompleteSettingsFragment())
+            resources.getString(R.string.pref_key_screen_autocomplete) ->
+                navigateToFragment(AutocompleteSettingsFragment())
         }
         return super.onPreferenceTreeClick(preferenceScreen, preference)
     }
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/mozilla/focus/settings/SearchSettingsFragment.kt
```diff
@@ -0,0 +1,61 @@
...
+        super.onPause()
+    }
+
+    override fun onPreferenceTreeClick(preferenceScreen: PreferenceScreen?, preference: Preference?): Boolean {
+        when (preference?.key) {
+            resources.getString(R.string.pref_key_search_engine) -> run {
+                navigateToFragment(InstalledSearchEnginesSettingsFragment())
+                TelemetryWrapper.openSearchSettingsEvent()
+            }
+            resources.getString(R.string.pref_key_screen_autocomplete) ->
```
sputnikci: \[detekt\] INFO: \[style/MaxLineLength\] Line detected that is longer than the defined maximum line length in the code style\.

PR #2591, opened 2018-05-07
DO NOT MERGE Turn off Sentry

ekager:


Sdaswani:
@boek this is our rollback strategy if we can't get a sentry crash to report.

PR #2616, opened 2018-05-14
Replace fragment for first run Closes #2615

### LINKED ISSUE #2615: When using a screenreader, elements behind first run fragment are usable
Breakout from #2553 
### First Time Use ###

The first time use UI is superimposed on the main interface which is not appropriately hidden and remains "visible" to TalkBack. I hear "Search or enter address, Edit box" when the app first opens

I hear Search or enter address, Edit box when the app first opens

Even explore by touch seems to fall through, so I can reach every element under the page.

Explore by touch to an invisible item

I can fully interact with the main interface items and browse the web without dismissing the first time use screen.

Browsing huffington post while first time use UI is visible

ekager:


app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.java
```diff
@@ -142,7 +142,7 @@ private void finishFirstrun() {
 
         fragmentManager
                 .beginTransaction()
-                .remove(this)
+                .replace(R.id.container, UrlInputFragment.createWithoutSession(), UrlInputFragment.FRAGMENT_TAG)
```
pocmo: This will break the flow if the app was opened with an Intent, e.g. ACTION_VIEW or a custom tab. Maybe that's okay?

PR #2628, opened 2018-05-17
Handle First Run A11y Closes #2615

### LINKED ISSUE #2615: When using a screenreader, elements behind first run fragment are usable

ekager:
Take 2

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.java
@@ -147,6 +147,13 @@ public class FirstrunFragment extends Fragment implements View.OnClickListener {
 
         final UrlInputFragment inputFragment = (UrlInputFragment) fragmentManager.findFragmentByTag(UrlInputFragment.FRAGMENT_TAG);
         if (inputFragment != null) {
+            View view = inputFragment.getView();
+            if (view != null) {
+                View rootView = view.getRootView();
+                if (rootView != null) {
+                    rootView.setImportantForAccessibility(View.IMPORTANT_FOR_ACCESSIBILITY_AUTO);
+                }
+            }
             inputFragment.showKeyboard();
         }
     }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.java
```diff
@@ -147,6 +147,13 @@ private void finishFirstrun() {
 
         final UrlInputFragment inputFragment = (UrlInputFragment) fragmentManager.findFragmentByTag(UrlInputFragment.FRAGMENT_TAG);
         if (inputFragment != null) {
+            View view = inputFragment.getView();
+            if (view != null) {
+                View rootView = view.getRootView();
+                if (rootView != null) {
+                    rootView.setImportantForAccessibility(View.IMPORTANT_FOR_ACCESSIBILITY_AUTO);
+                }
```
pocmo: Does this work?

* Shouldn't the root view be the root of the activity and therefore be the same for all fragments? Meaning, doesn't this disable accessibility for all views?
* If Focus was opened from a link then I guess there will be a BrowserFragment and no InputFragment?

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -246,6 +246,9 @@ class UrlInputFragment :
             if (!Settings.getInstance(it).shouldShowFirstrun()) {
                 // Only show keyboard if we are not displaying the first run tour on top.
                 showKeyboard()
+            } else {
+                // If we are showing first run tour, make sure accessibility is turned off.
+                view?.rootView?.importantForAccessibility = View.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS
             }
```
pocmo: I wonder if we could temporarily detach the other fragment (browser or input) and then reattach it once first run is done:
https://developer.android.com/reference/android/app/FragmentTransaction.html#detach(android.app.Fragment)

Issue #2631, opened 2018-05-20
About pages are shown as insecure


mueller-ma:
When I visit about:blank the connection is shown as insecure. As the about: pages are locally they aren't insecure.

PR #2644, opened 2018-05-22
Fix localized contents shown as insecure

RomanWuattier:
Localized contents are local pages. As they aren't insecure, the lock
icon should be displayed.

Close: #2631


boek:
Hey @RomanWuattier, thanks for the PR! I'll try to get to it as soon as I can :)

PR #2645, opened 2018-05-22
For #i967 - Adds a string for find in page

boek:


pocmo:
@boek What about accessibility strings for the UI? :)

boek:
@pocmo I didn't think about that until Android Studio reminded me while building out the UI :(. There will be a PR with everything today!

PR #2650, opened 2018-05-23
Upload the release version of the apk to Nimbledroid, instead of the debug build

npark-mozilla:


npark-mozilla:
@pocmo dbolter and snorp wants to use the release build to avoid potential debug-only errors like strictMode, and I suppose they want the app as close as the production version.

pocmo:
@npark-mozilla Strict mode should only log in debug builds.

Only note that release builds will send telemetry, crash reports etc. :)

PR #2692, opened 2018-06-01
Send Mobile Metrics Data from GV Closes #2347

### LINKED ISSUE #2347: Send Histogram Data for Load Times

ekager:


app/src/main/java/org/mozilla/focus/fragment/WebFragment.java
```diff
@@ -107,11 +109,25 @@ public void onPause() {
...
         webView.onPause();
 
         super.onPause();
     }
 
+
+    private boolean dayPassedSinceLastUpload() {
+        final long lastPing = PreferenceManager.getDefaultSharedPreferences(getContext()).getLong("lastPing", 0);
+        // Make sure a minimum of 1 day has passed since we collected data
+        return System.currentTimeMillis() >= (lastPing + (24*60*60*1000));
```
sputnikci: \[Checkstyle\] ERROR: '\*' is not preceded with whitespace\.

sputnikci: \[Checkstyle\] ERROR: '\*' is not followed by whitespace\.

sputnikci: \[Checkstyle\] ERROR: '\*' is not preceded with whitespace\.

sputnikci: \[Checkstyle\] ERROR: '\*' is not followed by whitespace\.

sputnikci: \[Checkstyle\] ERROR: '\*' is not preceded with whitespace\.

sputnikci: \[Checkstyle\] ERROR: '\*' is not followed by whitespace\.

app/src/main/java/org/mozilla/focus/fragment/WebFragment.java
```diff
@@ -107,11 +110,25 @@ public void onPause() {
             webView.saveWebViewState(session);
         }
 
+        if (AppConstants.isGeckoBuild() && dayPassedSinceLastUpload()) {
+            // Record new edited date
+            PreferenceManager.getDefaultSharedPreferences(getContext()).edit()
+                    .putLong("lastPing", System.currentTimeMillis()).apply();
```
boek: Should we pull this key into a constant?

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -288,6 +290,15 @@ object TelemetryWrapper {
         averageTime = 0.0
     }
 
+    private var shouldSendMobileMetrics: Boolean = false
```
boek: Do we usually define the type in Kotlin? Or let it be inferred?

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -303,10 +314,18 @@ object TelemetryWrapper {
...
-        TelemetryHolder.get()
-                .queuePing(TelemetryCorePingBuilder.TYPE)
-                .queuePing(TelemetryEventPingBuilder.TYPE)
-                .scheduleUpload()
+        if (shouldSendMobileMetrics) {
+            TelemetryHolder.get()
+                    .queuePing(TelemetryCorePingBuilder.TYPE)
+                    .queuePing(TelemetryEventPingBuilder.TYPE)
+                    .queuePing(TelemetryMobileMetricsPingBuilder.TYPE)
+                    .scheduleUpload()
```
boek: This might be able to be simplified with `.also`. Let me know what you think!

```
 TelemetryHolder.get()
                    .queuePing(TelemetryCorePingBuilder.TYPE)
                    .queuePing(TelemetryEventPingBuilder.TYPE)
                    .also { if (shouldSendMobileMetrics) queuePing(TelemetryMobileMetricsPingBuilder.TYPE) }
                    .scheduleUpload()
```

ekager:
Before we merge, I still want @pocmo's review to make sure I am satisfying the requirements of this new ping type.

app/src/gecko/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -446,6 +450,25 @@ public void loadData(String baseURL, String data, String mimeType, String encodi
...
+        public void getTelemetrySnapshots() {
+            final GeckoResponse<GeckoBundle> response = new GeckoResponse<GeckoBundle>() {
+                @Override
+                public void respond(GeckoBundle value) {
+                    if (value != null) {
+                        try {
+                            final JSONObject jsonData = value.toJSONObject();
+                            TelemetryWrapper.addMobileMetricsPing(jsonData);
+                        } catch (JSONException e) {
+                            e.printStackTrace();
```
pocmo: nit: Let's either log explicitly using "Log..." or if we think this should never happen then let's wrap the exception in an AssertionError and rethrow.

app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
```diff
@@ -215,7 +215,7 @@ open class AutocompleteListFragment : Fragment() {
             }
         }
 
-        override fun onViewRecycled(holder: RecyclerView.ViewHolder?) {
+        override fun onViewRecycled(holder: RecyclerView.ViewHolder) {
             if (holder is DomainViewHolder) {
                 holder.checkBoxView.setOnCheckedChangeListener(null)
             }
```
pocmo: The changes in this file seem to be unrelated? :) But they are in a separate commit, so no problem :)

app/src/main/java/org/mozilla/focus/fragment/WebFragment.java
```diff
@@ -107,11 +111,25 @@ public void onPause() {
...
 
         super.onPause();
     }
 
+
+    private boolean dayPassedSinceLastUpload() {
+        final long lastPing = PreferenceManager.getDefaultSharedPreferences(getContext()).getLong(lastMobileMetricsPing, 0);
+        // Make sure a minimum of 1 day has passed since we collected data
+        return System.currentTimeMillis() >= (lastPing + dayInMilliseconds);
+    }
```
pocmo: On desktop we check for "midnight locally has passed" - or in other words: If it's at least the next day for this user. You could just save YYYYMMDD and check if the value for today is larger.

app/src/main/java/org/mozilla/focus/fragment/WebFragment.java
```diff
@@ -107,11 +111,25 @@ public void onPause() {
             webView.saveWebViewState(session);
         }
 
+        if (AppConstants.isGeckoBuild() && dayPassedSinceLastUpload()) {
+            // Record new edited date
+            PreferenceManager.getDefaultSharedPreferences(getContext()).edit()
+                    .putLong(lastMobileMetricsPing, System.currentTimeMillis()).apply();
+            webView.getTelemetrySnapshots();
```
pocmo: The naming here is confusing: get***() implies that this method returns something. We could either make it return the JSON data and we construct the ping here or we rename it to something like sendTelemetrySnapshots()

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -288,6 +290,15 @@ object TelemetryWrapper {
...
     }
 
+    private var shouldSendMobileMetrics = false
+
+    @JvmStatic
+    fun addMobileMetricsPing(mobileMetrics: JSONObject) {
+        TelemetryHolder.get().addPingBuilder(TelemetryMobileMetricsPingBuilder(mobileMetrics,
+                TelemetryHolder.get().configuration))
+        shouldSendMobileMetrics = true
+    }
```
pocmo: Can't we just always add the ping builder above? We just do not need to queue a ping always.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -306,6 +317,7 @@ object TelemetryWrapper {
         TelemetryHolder.get()
                 .queuePing(TelemetryCorePingBuilder.TYPE)
                 .queuePing(TelemetryEventPingBuilder.TYPE)
+                .also { if (shouldSendMobileMetrics) it.queuePing(TelemetryMobileMetricsPingBuilder.TYPE) }
                 .scheduleUpload()
```
pocmo: I would expect the decision about whether to send a ping or not to be made here (including all the timestamp code) instead of somewhere outside and then passed into this class.

It would be really nice if we could collect the data here too. But I assume this is hard because of the asynchronous API? Alternatively we could collect the data from startSession() and queue the ping right there.

app/src/webview/java/org/mozilla/focus/webview/SystemWebView.java
```diff
@@ -223,6 +223,11 @@ public void loadData(String baseURL, String data, String mimeType, String encodi
         loadDataWithBaseURL(baseURL, data, mimeType, encoding, historyURL);
     }
 
+    @Override
+    public void getTelemetrySnapshots() {
+        // Do nothing, for GeckoView
```
pocmo: Let's throw something like IllegalStateException here so that we notice if something goes wrong and we start calling this for WebView.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
ekager: The TelemetryMobileMetricsPingBuilder needs the JSONObject passed in when it's created + added and I'm not sure where is the right place to get the reference to GeckoView that I need to get the JSONObject. That's why I originally did it in WebFragment so I would have the IWebView reference there.

PR #2717, opened 2018-06-07
Fixes #2666 - Presents locale names using the locale name

### LINKED ISSUE #2666: [l10n] Some locales missing from language list
Seems like there are at least two locales missing from the builds (both release and Nightly): "zam" and tsz" although they are listed in [locales.py](https://github.com/mozilla-mobile/focus-android/blob/master/tools/l10n/locales.py).
Maybe there are more, but I was only able to identify these two for now. 

Since they are both "lesser" known languages that Android doesn't offer, I'm guessing this may have something to do (it seems like others made it to the language list though).

(CC @pocmo @boek)

boek:


ekager:
Assuming they're supposed to be lower case, LGTM

boek:
@ekager verifying with @Delphine now

Delphine:
Good catch! They should be uppercase. thanks!

PR #2729, opened 2018-06-11
Update GV 06.11.18

ekager:


app/src/gecko/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -283,6 +283,13 @@ public void onExternalResponse(GeckoSession session, GeckoSession.WebResponseInf
                     }
                 }
 
+                @Override
+                public void onCrash(GeckoSession session) {
+                    Log.i(TAG, "Crashed, reopening session");
+                    session.open(geckoRuntime);
+                    session.loadUri(currentUrl);
+                }
```
pocmo: Oh, is that save to do?

It would be awesome to send a custom event to Sentry (and maybe telemetry) here!

ekager: New method, found the implementation in their example - https://dxr.mozilla.org/mozilla-central/source/mobile/android/geckoview_example/src/main/java/org/mozilla/geckoview_example/GeckoViewActivity.java#405

pocmo: Nice! Let's merge this and just file an issue for telemetry/sentry? :)

ekager:
Opened #2731 for adding telemetry here. We will need a data form too

PR #2761, opened 2018-06-19
Add Sentry capture to GeckoView onCrash Closes #2731

### LINKED ISSUE #2731: Add Telemetry/Sentry to GeckoView onCrash
We want to send custom telemetry/Sentry pings to know more about Gecko session crash rates.

### What / Requirements
Set up a Telemetry event to capture a Gecko crash.
Set up a Sentry event to capture a Gecko crash.

Can see pings from Gecko session crashes.

ekager:


app/src/gecko/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -302,6 +303,7 @@ public void onExternalResponse(GeckoSession session, GeckoSession.WebResponseInf
                 @Override
                 public void onCrash(GeckoSession session) {
                     Log.i(TAG, "Crashed, opening new session");
+                    Sentry.capture("GeckoSession crashed, opening new session");
```
pocmo: Let's move all calls to Sentry into SentryWrapper so that we have a single point where we can control what gets send when.

Issue #2741, opened 2018-06-14
Disable telemetry for "distributions" by default


pocmo:
Firefox for Android has a complex way to be customized by partners via Distribution files:
https://wiki.mozilla.org/Mobile/Distribution_Files

Eventually we may want to have something similar in Focus. But we want to do some things differently this time. And for now we only need a way to control whether telemetry is enabled by default.

in Firefox for Android distributions can be: Baked into the APK, downloaded after installation (google play referrer) or loaded from a system path. All those have different advantages/disadvantages.

What we want to do with Focus is: The device manufacturer sets a specific read-only system property in the system image from which we can read the distribution id. At a later time we will use this ID to download a distribution file. For now we are just using the existence of the property as an indicator that we are running on a partner device.

The name of the property will depend on the package name:

```
ro.<package name>.distribution.id

ro.org.mozilla.focus.distribution.id
ro.org.mozilla.klar.distribution.id
```

For now: Whenever this property exists we disable telemetry by default (Not overriding the user choice though). At a later time we will download the matching distribution which will contain customizations options like disabling telemetry.

PR #2775, opened 2018-06-22
For #2741

boek:


app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -739,4 +742,11 @@ object TelemetryWrapper {
     fun addSearchEngineLearnMoreEvent() {
         TelemetryEvent.create(Category.ACTION, Method.CLICK, Object.ADD_SEARCH_ENGINE_LEARN_MORE).queue()
     }
+
+    private fun isDistributedDevice(): Boolean {
```
pocmo: tiny nit: We may not want to disable this for all distribution, so I wonder if we could make the method name more explicit... something like isDeviceWithTelemetryDisabled()?

boek: Sounds good to me

pocmo:
@boek Are you taking care of merging this into master too?

boek:
Yep! I'll put it into master and 6.0 release branch

PR #2799, opened 2018-06-25
Coordinator behavior + find in page margins Closes #2776 + Closes #2725

### LINKED ISSUE #2725: When found word is behind the Find in page text bar, it's not visible
### Steps to reproduce
- Find a word that occurs multiple times, and the non-first occurrence is hidden behind the Find in page text bar
- tap down button to highlight the word that is hidden behind the text bar

### Expected behavior
- Page scrolls, and the word is highlighted and visible

### Actual behavior
- Page does not scroll, and the word stays covered

### Device information
- Pixel 2, API 27

### LINKED ISSUE #2776: Viewport issue on cnn.com
### Steps to reproduce

- Install the Beta with GeckoView
- Open cnn.com

### Description

CNN displays this cookie notice now. That is not the bug though - the bug here is that the cookie notice is not properly anchored to the bottom of the screen. The *I Accept* button is therefore not visible until you start scrolling.

I have attached two screenshots that show the initial page load and what happens when i scroll the page a little bit.

Not verified with WebView Focus, but I can confirm that this works properly on Focus iOS.

### Device information

- Beta as of June 22nd.
- Moto G (2nd Generation) / Android 6.0

device-2018-06-22-122412
device-2018-06-22-122452

ekager:


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -293,14 +299,25 @@ public void onTextChanged(CharSequence s, int start, int before, int count) {
                     }
                 }
         );
+        findInPageQuery.setOnClickListener(this);
+        findInPageQuery.setOnEditorActionListener(new TextView.OnEditorActionListener() {
+            @Override
+            public boolean onEditorAction(TextView textView, int i, KeyEvent keyEvent) {
+                if (i == EditorInfo.IME_ACTION_DONE) {
```
pocmo: Let's use something that is more descriptive than ``i``. The interface of OnEditorActionListener uses `actionId`.

Issue #2801, opened 2018-06-26
Use 62 beta builds of GeckoView


pocmo:
(At the time of writing this issue there's no beta 62 release of GeckoView yet!)

Finding the right URL for a Beta release is a bit more complicated currently.

* First we will have to look at the tags on the `mozilla-beta` repository:
https://hg.mozilla.org/releases/mozilla-beta/tags

* Find the latest beta release tag. They look like this: ``FENNEC_61_0b13_RELEASE`` (With Fennec we build GeckoView too; This tag is for 61.0 Beta 13).

* Click on the changeset ID to find the matching changeset/commit ID (shortened). For the example tag above that's ``5b5e620e33ee``.

* Now go to the list of revisions to find the matching changeset (It should start with the ID you got from the tag, remember the full ID!):
https://tools.taskcluster.net/index/gecko.v2.mozilla-beta.revision

* Click through until you find the artifacts (revision -> mobile -> android-api-16-opt).

* Remember the version of the artifacts. In the example above that's: 61.0.20180611134439

Now you have all the values you will need to construct the repository URL and artifact names:

URLs (replace REVISION with the full/long ID):
* ARM: "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-beta.revision.REVISION.mobile.android-api-16-opt/artifacts/public/android/maven"
* x86: "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-beta.revision.REVISION.mobile.android-x86-opt/artifacts/public/android/maven"
* aarch64: "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-beta.revision.REVISION.mobile.android-aarch64-opt/artifacts/public/android/maven"

Artifacts (insert version number):
* ARM: "org.mozilla:geckoview-beta-armeabi-v7a:VERSION"
* x86: "org.mozilla:geckoview-beta-x86:VERSION"
* aarch64: "org.mozilla:geckoview-beta-arm64-v8a:VERSION"

pocmo:
There's no 62 release build yet. But it should appear today/tomorrow.

pocmo:
FENNEC_62_0b3_RELEASE is now available.

PR #2803, opened 2018-06-26
Update GV 06.26.18

ekager:


pocmo:
63.0 is too far :) See #2801

ekager:
Oh oops so set in my ways, thanks @pocmo

ekager:
Updated gradle with the new beta versions. How often will we update these now?

PR #2809, opened 2018-06-27
Sentry initialization now from FocusApplication and not MainApplicati…

### LINKED ISSUE #2791: Initialize Sentry from FocusApplication
Currently we initialize Sentry from MainActivity.onCreate(). A lot of other code runs before that and if this code crashes then we do not see it in Sentry.

We should initialize Sentry from FocusApplication.onCreate(). That's the earliest point where we have a `Context` object.

agbanagba:
…on -  Closes #2791

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -18,7 +18,7 @@
 
 ### Device information
 
-* Fire TV device: ?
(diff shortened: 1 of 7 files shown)


pocmo:
There are a lot of unrelated things in this PR. Can you try to rebase your branch (against master) and squash your changes into a single commit? :)

PR #2817, opened 2018-06-27
updated build-tools, gradles, and a bunch of dependencies

TacoTheDank:


TacoTheDank:
i was an idiot and got 27.1.1 confused with 27.0.3, as buildtools has 27.0.3 while the compat libraries go to 27.1.1

Issue #2790, opened 2018-06-25
Expose build flavor in Sentry


pocmo:
When seeing a crash in Sentry it's helpful to know what build flavor this is (e.g. klarGeckoAarch64). We can get that value from BuildConfig.FLAVOR.

RomanWuattier:
Hey, can I work on this one?
I didn't use Sentry yet. Should I record a breadcrumb in the current context and build a message containing the current flavor?

pocmo:
Yeah, looking at crashes on Sentry a tag seems to be a good way to expose that!

RomanWuattier:
Hey, the PR add a `BUILD_FLAVOR` tag to future events. If a lower camel case tag or another name fits better, let me know.

PR #2832, opened 2018-06-30
Expose build flavor in Sentry

RomanWuattier:
Add a tag `build_flavor` containing the current build flavor to future Sentry events.

Close: #2790


pocmo:
That's how the existing tags look like:

screenshot_2018-07-02 remoteserviceexception context startforegroundservice did not then call service startforeground

> If a lower camel case tag or another name fits better, let me know.

Yeah, let's go with `build_flavor`.

PR #2845, opened 2018-07-03
Send GV label when reporting webcompat issue Closes #2837

### LINKED ISSUE #2837: Send `label=browser-firefox-focus` when reporting an issue on webcompat.com
Currently, if people report bugs via the "Report Site Issue" feature, we can't distinguish between Fennec and Focus GeckoView, because we send the same UA string.

However, webcompat.com supports adding additional labels to an issue via a `label` param, so we should toss `label=browser-firefox-focus` into `REPORT_SITE_ISSUE_URL`:

https://github.com/mozilla-mobile/focus-android/blob/c169cf6b186b9ddb93de970368dc112a29a621d9/app/src/main/java/org/mozilla/focus/utils/SupportUtils.java

I don't know if it's possible to know if we're building w/ GeckoView, if so, something like `browser-focus-geckoview` would be even better. Is that doable @ekager?

Server side companion bug: https://github.com/webcompat/webcompat.com/issues/2537

ekager:
Hey @miketaylr is this label added correctly at the end of the link? We only show the report site menu item for GeckoView builds now

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -1129,8 +1129,8 @@ public class BrowserFragment extends WebFragment implements View.OnClickListener
 
             case R.id.report_site_issue:
                 SessionManager.getInstance()
-                        .createSession(Source.MENU, SupportUtils.REPORT_SITE_ISSUE_URL
-                                .replace("%s", getUrl()));
+                        .createSession(Source.MENU,
+                                String.format(SupportUtils.REPORT_SITE_ISSUE_URL, getUrl()));
                 TelemetryWrapper.reportSiteIssueEvent();
                 break;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -1128,7 +1128,9 @@ public void onClick(View view) {
                 break;
 
             case R.id.report_site_issue:
-                SessionManager.getInstance().createSession(Source.MENU, SupportUtils.REPORT_SITE_ISSUE_URL.concat(getUrl()));
+                SessionManager.getInstance()
+                        .createSession(Source.MENU, SupportUtils.REPORT_SITE_ISSUE_URL
+                                .replace("%s", getUrl()));
```
pocmo: You may want to use [String.format()](https://developer.android.com/reference/java/lang/String.html#format(java.lang.String,%20java.lang.Object...)) for those replacements.

miketaylr:
@ekager yeap, looks great, thanks!

PR #2863, opened 2018-07-09
Issue #2854: Allow searches including a domain name

### LINKED ISSUE #2854: Can't do searches including a domain name
### Steps to reproduce

Search for "reddit.com chess"

### Expected behavior

Same as on desktop: default search engine result for those 2 keywords.

### Actual behavior

Goes to reddit.com, drops second part.

### Device information

* Android device: Pixel 2 XL
* Focus version: 5.0

KrzysztofBarszcz:
Closes #2854 
As we don't want to go to a domain when the query contains its address and other words, it should be sufficient if we just check if the query is a URL


boek:
Hey @KrzysztofBarszcz, thanks for picking that up!

PR #2888, opened 2018-07-12
Add timeout to countdown latch in save state

ekager:
Could possibly help ANRs with Nightly, but might lose some state, in which case we load the desired URL


pocmo:
I filed this issue and will bring this up with the GV team.

Issue #2792, opened 2018-06-25
Make engine choice a runtime option


pocmo:
We want to switch from WebView to GeckoView soon. In order to roll out GeckoView gradually, while monitoring telemetry and chrashes, we want to be able to ship with both engines and decide at runtime which one to use. This will allow us to do a staged rollout and A/B test the engine.

Currently the choice of engine is a compile time option. For this to work we would have to make it a runtime option.

For that we need to:
* Dissolve "engine" build flavor and move code into two separate packages in the `main` source set.
* Create a single WebViewProvider instance that returns either GeckoWebView or SystemWebView at runtime.
* Make sure that this decision is only made once on app start and does not change while the app is running

colintheshots:
I'm not entirely sure I'll dissolve the build flavor. I think it makes most sense to either have a default renderer or make that flavor dimension set the default renderer using a simple config setting that can later be overridden. I'd migrate all of the renderer flavor-specific code into the main project and just have shared preferences.

colintheshots:
My plan is to migrate all renderer flavor-specific code as mentioned and then choose the renderer based on a shared preference. I would allow run-time switching by changing the preference and invoking ProcessPhoenix, which is ideal for this kind of systemic change to the app and its state.

pocmo:
> I'm not entirely sure I'll dissolve the build flavor.

One less flavor dimension would speed up our automated builds and get us PR feedback faster. So I'd dissolve it as soon as possible. :)

> I would allow run-time switching by changing the preference and invoking ProcessPhoenix, which is ideal for this kind of systemic change to the app and its state.

Note that the engine choice is not something we want to expose to the user. Instead we'll decide what engine to use on app start based on the A/B test configuration. So I think switch at runtime (by restarting the app) won't be needed for now.

colintheshots:
@pocmo and I met this morning and reached an agreement about the best design for this feature. I'll keep the extra flavor dimension so developers can test with different default renderers easily. We may never need ProcessPhoenix for end users to switch live from A to B, but it might be useful to allow devs and QA to switch back and forth with a hidden setting.

FYI, Universal webview targets will be disappearing in the newly switchable Focus renderer versions. This is because we'll be including GeckoView in all APKs. The number of build variants will explode, so it would be wise to specify flavors when requesting builds.

Sdaswani:
@colintheshots to break down.

colintheshots:
# Focus
[...]

PR #2892, opened 2018-07-12
Issue #2792 - Make GeckoView vs WebView selection possible by alterin…

colintheshots:
…g shared preferences and restarting app


app/build.gradle
```diff
@@ -310,19 +292,18 @@ dependencies {
...
     focusImplementation 'com.google.android.gms:play-services-analytics:10.2.6' // Required by Adjust
 
-    // What I actually want to define here is geckoArmCompile and geckoX86Compile. But gradle doesn't
-    // allow me to define dependencies for such a flavor combination. However as we are not building
-    // WebView + (ARM / x86) flavor combinations this should be good enough for now.
     armImplementation "org.mozilla:geckoview-beta-armeabi-v7a:${geckoview_version}"
     x86Implementation "org.mozilla:geckoview-beta-x86:${geckoview_version}"
     aarch64Implementation "org.mozilla:geckoview-beta-arm64-v8a:${geckoview_version}"
 
+    implementation 'com.jakewharton:process-phoenix:2.0.0'
```
pocmo: This patch doesn't seem to use this library yet?

app/src/gecko/java/org/mozilla/focus/web/Config.kt
```diff
@@ -0,0 +1,5 @@
+package org.mozilla.focus.web
```
pocmo: nit: license header here and in the other new files

```Kotlin
```

app/src/main/java/org/mozilla/focus/web/WebViewProvider.kt
```diff
@@ -0,0 +1,83 @@
...
+    var engine : IWebViewProvider? = null
+
+    fun writeEnginePref(context: Context, newEngineName: String = engineName) {
+        requireNotNull(context)
+        val prefs = PreferenceManager.getDefaultSharedPreferences(context)
+        prefs.edit().putString(ENGINE_PREF_STRING_KEY, newEngineName).apply()
+    }
+
+    fun readEnginePref(context: Context) {
+        requireNotNull(context)
```
pocmo: Can you help me understand requireNotNull here? Won't the Kotlin compiler add a null check already because Context is not nullable?

app/src/main/java/org/mozilla/focus/FocusApplication.kt
```diff
@@ -27,6 +28,8 @@ class FocusApplication : LocaleAwareApplication() {
         super.onCreate()
 
         PreferenceManager.setDefaultValues(this, R.xml.settings, false)
+        WebViewProvider.readEnginePref(this)
+        WebViewProvider.writeEnginePref(this)
```
pocmo: I wonder why we immediately write here? Don't we only need to do that if the value was changed, either by some internal QA thingy or from the A/B config later?

app/src/main/java/org/mozilla/focus/web/WebViewProvider.kt
```diff
@@ -0,0 +1,83 @@
...
+        when (name) {
+            WEBVIEW.name -> engine = ClassicWebViewProvider()
+            GECKO.name -> engine = GeckoWebViewProvider()
+        }
+    }
+
+    fun preload(context: Context) {
+        requireNotNull(context)
+        requireNotNull(engine)
+        engine!!.preload(context)
```
pocmo: Also here I do not understand the usage of `requireNotNull` and `!!`. Wouldn't just using `!!` have the same result? Or using the return value of `requireNotNull`?

app/build.gradle
colintheshots: Yes, I split the work into two PRs. The part that restarts the app isn't required yet.

app/src/main/java/org/mozilla/focus/web/WebViewProvider.kt
colintheshots: Since engine is nullable, I either need to use !!. or ?. to make the call safe. It does seem redundant in this case. I can remove the precondition.

app/src/main/java/org/mozilla/focus/FocusApplication.kt
colintheshots: I was persisting the current state of the engine on startup to verify that the engine was sticky if I set the preference. The read is required. I can remove the write in this case.

app/src/main/java/org/mozilla/focus/web/WebViewProvider.kt
colintheshots: I was checking preconditions from the Java callers. It's quite possible this is redundant. I can check the bytecode.

app/src/gecko/java/org/mozilla/focus/web/Config.kt
colintheshots: I'll add the license header.

app/src/main/java/org/mozilla/focus/web/WebViewProvider.kt
colintheshots: Checked the bytecode. It was redundant, but the compiler is smart enough to check only once. I'l remove the extra checks.

PR #2895, opened 2018-07-13
Fix test failures in AddtoHSTest.java - Closes #2873

### LINKED ISSUE #2873: AddtoHSTest.java is failing
### Steps to reproduce
automated tests in AddtoHSTest.java (SearchTermShortcutTest in particular) fail frequently on all devices

### Expected behavior
Tests should successfully create and find desktop shortcuts

### Actual behavior
Each desktop shortcut links to a temporary mock server and port, i.e. http://localhost:12345.  When several tests run back-to-back, there can be more than one shortcut with different ports, but the test will only grab the first one available (which may be a zombie link to a dead port).  

Also, the test is missing a proper swipe to ensure that it lands on the correct screen with the shortcuts (and not the home screen, for example).

### Device information

* Android device: all
* Focus version: 6.2

rpappalax:
r? @npark-mozilla

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/activity/AddtoHSTest.java
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/AddtoHSTest.java
@@ -37,6 +37,8 @@ import static org.mozilla.focus.helpers.TestHelper.webPageLoadwaitingTime;
 public class AddtoHSTest {
     private static final String TEST_PATH = "/";
     private MockWebServer webServer;
+    private int webServerPort;
+    private String webServerBookmarkName;
 
     @Rule
     public ActivityTestRule<MainActivity> mActivityTestRule  = new ActivityTestRule<MainActivity>(MainActivity.class) {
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/org/mozilla/focus/activity/AddtoHSTest.java
```diff
@@ -53,14 +55,16 @@ protected void beforeActivityLaunched() {
...
+            webServerBookmarkName = "localhost_" + Integer.toString(webServerPort);
 
             try {
                 webServer.enqueue(new MockResponse()
                         .setBody(TestHelper.readTestAsset("plain_test.html")));
                 webServer.enqueue(new MockResponse()
                         .setBody(TestHelper.readTestAsset("plain_test.html")));
 
-                webServer.start();
+                //webServer.start();
```
npark-mozilla: @rpappalax so is webserver starting somewhere else, since this is commented out?

npark-mozilla:
Should also squash to a single commit as well

PR #2906, opened 2018-07-13
Send Histogram Data for Load Times Closes #2891

### LINKED ISSUE #2891: Send Histogram Data for Load Times
Per a Slack discussion, we should assemble a histogram and report it through the event `extras` for WV + GV. We should report an array of e.g. 200 values, where each one is 100ms or so. Last value in the array is everything > 20k ms. We're just recording the number of values in each bucket, not the exact values.

ekager:
This will measure the time between the last redirect and the page finished callback, and only send a value if it matches the last known URL. Seems much better than the last loading listener implementation.


app/src/gecko/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -349,7 +349,7 @@ public void onPageStop(GeckoSession session, boolean success) {
                                 isSecure = true;
                             }
 
-                            callback.onProgress(100);
+                            callback.onProgress(99);
```
pocmo: Is this left over from a previous patch or is this needed?

app/src/main/java/org/mozilla/focus/session/SessionCallbackProxy.java
```diff
@@ -124,4 +142,19 @@ public void onExitFullScreen() {
...
+            final long elapsedLoad = endTime - startLoadTime;
+            Log.i(LOG_TAG, elapsedLoad + " - elapsed load");
+            // Even internal pages take longer than 30 ms to load, let's not send any loads faster than this
+            if (elapsedLoad > MIN_LOAD_TIME && !UrlUtils.isLocalizedContent(urlLoading)) {
+                Log.i(LOG_TAG, "Sent load to histogram");
+                TelemetryWrapper.addLoadToHistogram(elapsedLoad);
+            }
+        }
+    }
+
```
pocmo: It feels a bit dirty stuffing all this into `SessionCallbackProxy`- It's primary use is to update `Session` from the IWebView callbacks.
Can we move this into a separate class (like before) and make it subscribe to "loading" and "URL changed" events?

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -320,11 +325,11 @@ object TelemetryWrapper {
...
-        if (numLoads > 0) {
-            TelemetryEvent.create(Category.ACTION, Method.FOREGROUND, Object.BROWSER)
-                    .extra(Extra.AVERAGE, averageTime.toString()).queue()
-            resetAverageLoad()
-        }
+        TelemetryEvent.create(Category.ACTION, Method.FOREGROUND, Object.BROWSER)
+                .extra(Extra.HISTOGRAM, Arrays.toString(histogram)).queue()
+
+        // Clear histogram after queueing it
+        histogram = IntArray(HISTOGRAM_SIZE)
```
pocmo: Can you verify with the telemetry team that it's no problem for them to re-use the same event for the histogram? We would have events with the old and new extras format in the database and maybe it would be easier for them to just look at a different event that only contains the new histogram extras.

app/src/gecko/java/org/mozilla/focus/web/WebViewProvider.java
ekager: We set WebView progress to 99 before calling pageFinished, so I think we should do the same here.

ekager: you're right we don't need this since the max we can set it to in onProgress is 99 anyway.

app/src/main/java/org/mozilla/focus/observer/LoadTimeObserver.kt
```diff
@@ -0,0 +1,57 @@
...
+        })
+        session.url.observe(fragment, object : NonNullObserver<String>() {
+            public override fun onValueChanged(t: String) {
+                if ((urlLoading != null && urlLoading != t) || urlLoading == null) {
+                    startLoadTime = SystemClock.elapsedRealtime()
+                    Log.i(LOG_TAG, "zerdatime $startLoadTime - url changed to $t, new page load start")
+                    urlLoading = t
+                }
+            }
+        })
```
sputnikci: \[detekt\] INFO: \[style/NewLineAtEndOfFile\] Checks whether files end with a line separator\.

app/src/main/java/org/mozilla/focus/observer/LoadTimeObserver.kt
```diff
@@ -0,0 +1,57 @@
...
+        session.loading.observe(fragment, object : NonNullObserver<Boolean>() {
+            public override fun onValueChanged(t: Boolean) {
+                if (t) {
+                    if ((urlLoading != null && urlLoading != session.url.value) || urlLoading == null) {
+                        urlLoading = session.url.value
+                        startLoadTime = SystemClock.elapsedRealtime()
+                        Log.i(LOG_TAG, "zerdatime $startLoadTime - page load $urlLoading start")
+                    }
+                } else {
+                    // Progress of 99 means the page completed loading and wasn't interrupted.
```
sputnikci: \[detekt\] INFO: \[style/MagicNumber\] Report magic numbers\. Magic number is a numeric literal that is not defined as a constant and hence it's unclear what the purpose of this number is\. It's better to declare such numbers as constants and give them a proper name\. By default, \-1, 0, 1, and 2 are not considered to be magic numbers\.

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
```diff
@@ -40,6 +40,7 @@ import org.mozilla.telemetry.storage.FileTelemetryStorage
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
+import java.util.Arrays
```
sputnikci: \[ktlint\] WARNING: \[no\-unused\-imports\] Unused import in column 1

app/src/main/java/org/mozilla/focus/telemetry/TelemetryWrapper.kt
ekager: Talked to Frank and he told me to send the keys + values as a JSON String and approved the output.

app/src/main/java/org/mozilla/focus/observer/LoadTimeObserver.kt
```diff
@@ -0,0 +1,58 @@
...
+
+import android.os.SystemClock
+import android.support.v4.app.Fragment
+import android.util.Log
+import org.mozilla.focus.architecture.NonNullObserver
+import org.mozilla.focus.session.Session
+import org.mozilla.focus.telemetry.TelemetryWrapper
+import org.mozilla.focus.utils.UrlUtils
+
+object LoadTimeObserver {
```
colintheshots: I'm not entirely sure the change to a static observer object is a good one. You're now passing in the fragment to the static method, which could leave a reference around to leak memory. I don't think it'll leak for very long as written though.

ekager:
Done!

Issue #2883, opened 2018-07-11
Notes Experiments page doesn't load


yoasif:
### Steps to reproduce

Navigate to https://testpilot.firefox.com/experiments/notes

### Expected behavior

Page loads. 

### Actual behavior

Background and an image seems to load, but no other page content appears. 

Firefox Klar (stable from F-Droid) loads this with no issue, as does Fennec. Issue appears with default tracker blocklist as well as when disabling from main menu. 

![signal-attachment-2018-07-11-162643](https://user-images.githubusercontent.com/152493/42597749-d2e71544-8527-11e8-83d2-1ecc9ffa7029.png)

### Device information

* Android device: LG-H810, Android 5.1
* Focus version: 6.2 (Build #221921206 Emoji image of 'lizard' 62.0.20180709172241)

bbinto:
Let's figure out the WHY, and discuss after what the next steps are.

cpeterso:
Since this bug only happens with GeckoView and a portion of the page in the screenshot appears to be rendered, this problem might be related to [GeckoView rendering bug 1472356](https://bugzilla.mozilla.org/show_bug.cgi?id=1472356). We have a potential workaround (disabling SkiaGL on Android) in [Firefox bug 1468801](https://bugzilla.mozilla.org/show_bug.cgi?id=1468801).

vesta0:
Can we verify that this has been fixed?

sv-ohorvath:
Still reproducing on Nightly 8.0.4 & Release 8.0.5 on GeckoView.

lobontiumira:
Verified as fixed on the 8.16.0 Focus build with Samsung Galaxy Note 8 (Android 9), and Sony Xperia Z5 Premium (Android 7.1.1).

PR #2907, opened 2018-07-14
GV Turn off TP + Cookie Blocking with Menu Toggle Closes #2894

### LINKED ISSUE #2894: Toggling Tracking Protection in the main menu has no effect with GeckoView
### Steps to reproduce

1. Visit a site that has visible tracking content (see, for example, https://github.com/webcompat/web-bugs/issues/17717)
2. Turn off tracking protection using the toggle in Focus's main menu

### Expected behavior

- The tracking content is visible

### Actual behavior

- The tracking content is blocked

### Device information

I'm *am* seeing this behavior on Focus 6.2 (Build 221931205 :lizard: 62.0.20180709172241).
I'm *am not* seeing this behavior on Klar 4.2 (Build #19), from F-Droid.

ekager:


app/src/gecko/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -185,8 +185,10 @@ public void setBlockingEnabled(boolean enabled) {
                 applyAppSettings();
             } else {
                 if (geckoSession != null) {
```
pocmo: I wonder why we do the `geckoSession != null` check here. Inside the block we do not use `geckoSession` and if it would be null then `geckoSession.getSettings()` at the beginning of the method would fail already.

app/src/gecko/java/org/mozilla/focus/web/WebViewProvider.java
```diff
@@ -185,8 +185,10 @@ public void setBlockingEnabled(boolean enabled) {
...
+                    geckoRuntime.getSettings().setTrackingProtectionCategories(0);
                     geckoRuntime.getSettings().setJavaScriptEnabled(true);
                     geckoRuntime.getSettings().setWebFontsEnabled(true);
+                    geckoRuntime.getSettings().setCookieBehavior(GeckoRuntimeSettings.COOKIE_ACCEPT_ALL);
```
pocmo: Doesn't this disable all those features globally? Will we re-enable them if a user switches tabs?

ekager: Yes we reenable if a user switches tabs, setBlockingEnabled is part of tab initialization

app/src/gecko/java/org/mozilla/focus/web/WebViewProvider.java
ekager: Yeah that makes sense

boek:
Once this lands lets see if it fixes #2883

PR #2939, opened 2018-07-19
Fix AddToHomescreenTest in  >=API 26 

npark-mozilla:
Re-reference 'Add automatically' button in the Add to Home screen dialog


npark-mozilla:
@rpappalax, I noticed the caption of the dialog was changed, causing the test to fail on newer version of android. I made a patch, could you take a look? thanks!

PR #2955, opened 2018-07-23
Override icon for Nightly track, Closes #2818

### LINKED ISSUE #2818: Override icon for Nightly track
It would be helpful to override the icon for Nightly builds so that users can easily see that they are on the Nightly track.

~I do not want to add a new release type to build.gradle because this will slow down everything (more variants to build) for no reason but a different icon. We just got rid of the Beta type and want to remove the Coverage type too: #2313~

colintheshots:
Add a nightly build track defaulting to geckoview with new icons and remove the webview/geckoview flavor dimension. This affects QA automation and builds. @npark-mozilla


app/build.gradle
```diff
@@ -45,14 +45,18 @@ android {
         debug {
             applicationIdSuffix ".debug"
         }
+        nightly {
+            initWith debug
+            applicationIdSuffix ".nightly"
+        }
```
pocmo: Let's remove the suffix. On Google Play the builds across all release tracks need to have the same package name.

pocmo:
Can you make sure to run `ImageOptim` over the icons (not critical for Nightly though)? Often there's still room for optimization after we get PNGs.

app/build.gradle
colintheshots: Will do.

colintheshots:
The PNGs were already pretty well optimized. FileOptimizer managed to save only 6,872 bytes or 5%.

I'll try to get the scripts fixed and re-submit.

colintheshots:
I'll also try to figure out why the code coverage dropped. It clearly stopped running some Webview tests.

npark-mozilla:
Codecov drops occasionally for some odd reason - if it's a red herring we can also look into increasing the ignore threshold.

colintheshots:
Re-submitted with the isGeckoBuild default fix, more compressed icons, and updated build scripts. @pocmo

PR #2956, opened 2018-07-23
Update to GV 62.0b10

ekager:


ekager:
Apparently my save state implementation shouldn't work (but does somehow?)... Should be in a background thread but that is blocked on https://bugzilla.mozilla.org/show_bug.cgi?id=1474618

Issue #2953, opened 2018-07-23
UI tests for "Gecko X86" builds fail with missing APK error


pocmo:
Gecko X86:

> Uploading [app/build/outputs/apk/klarGeckoX86/debug/app-klar-gecko-x86-debug.apk] to Firebase Test Lab...
ERROR: (gcloud.firebase.test.android.run) [app/build/outputs/apk/klarGeckoX86/debug/app-klar-gecko-x86-debug.apk] not found or not accessible

https://tools.taskcluster.net/groups/T6TLRUMYQUGPS91KoXow0g/tasks/UIv3jHO9Te64aAIqCsyxIA/runs/0/logs/public%2Flogs%2Flive.log

pocmo:
@npark-mozilla @colintheshots

kfontecilla:
@npark-mozilla Could you please double-check and close if fixed?

Issue #2952, opened 2018-07-23
Update nimbledroid to use new APKs


pocmo:
```
<=============> 100% EXECUTING [9m 6s]<-------------> 0% WAITINGTraceback (most recent call last):
  File "tools/taskcluster/upload_apk_nimbledroid.py", line 39, in <module>
    focus_file = {'apk': open('app/build/outputs/apk/focusWebviewUniversal/release/app-focus-webview-universal-release-unsigned.apk')}
IOError: [Errno 2] No such file or directory: 'app/build/outputs/apk/focusWebviewUniversal/release/app-focus-webview-universal-release-unsigned.apk'
[taskcluster 2018-07-22 23:17:56.850Z] === Task Finished ===
[taskcluster 2018-07-22 23:17:56.850Z] Unsuccessful task run with exit code: 1 completed in 565.524 seconds
```

https://tools.taskcluster.net/groups/T6TLRUMYQUGPS91KoXow0g/tasks/ENd5yfzhTOuGMDtRRwIh-w/runs/0/logs/public%2Flogs%2Flive.log

pocmo:
@npark-mozilla @colintheshots

npark-mozilla:
Also need to notifiy nimbledroid about our new app id, if this was changed

npark-mozilla:
If the package name 'org.mozilla.focus' got changed, it's a big deal for nimbledroid measurement, but I don't think that's the case here. Let me know if I'm mistaken

PR #2958, opened 2018-07-23
remove ARM compile and upload on taskcluster X86 build task

npark-mozilla:


npark-mozilla:
to address #2953

npark-mozilla:
Also addresses #2952

npark-mozilla:
@pocmo will remove the test for the simulator.  Firebase always ran UI tests on both sim and device for a while - thinking that there might be something that requires testing on both. if not, then we'll stop the run on the sim.

pocmo:
@npark-mozilla Ah, I see! If there are different architectures x86/ARM used on Firebase then you may need the different builds to be able to install them.

npark-mozilla:
@pocmo wait, let me revert above comment - one other reason why it runs on both sim and device is that the tablet (nexus 9) is only available in simulator form in Google firebase.  so I think we need both run, on both webview/geckoview test.

PR #2987, opened 2018-07-25
Update Histogram to send multiple extras

ekager:
Have to update telem library tomorrow as well to have the new limit raised for max extras :)


ekager:
This is ready for review and merge and then we'll have to cut a new dot release. I'll also update telemetry wiki.

PR #3020, opened 2018-07-30
Closes #2852 - on back pressed check if full screen and close

### LINKED ISSUE #2852: Unable to scroll after watching fullscreen video
This may be a GV bug, but we should spend a few hours seeing if it's on Focus.

https://bugzilla.mozilla.org/show_bug.cgi?id=1471340

ekager:


ekager:
True, I realized that it probably makes more sense to also exit full screen when we pause browser fragment anyway. It seems to follow other browser's behavior now, plus we don't care if we lose the state if we close fullscreen anyway. :)

PR #3045, opened 2018-08-02
Create and configure A/B test for selecting between GeckoView and Web…

### LINKED ISSUE #2916: Create and configure A/B test for selecting between GeckoView and WebView
As a user, I would like to use the newest, stable renderer with the best new features.
As a developer, I would like to control the rollout of the new rendering engine to ensure success.

### What / Requirements
The renderer selection feature #2915 will be connected to Fretboard for handling this A/B test experiment. An experiment will be properly configured to toggle users between A and B.

When I can toggle users from classic WebView to GeckoView using our A/B experiments infrastructure, this feature is done.

colintheshots:
…View. Closes #2916


app/build.gradle
```diff
@@ -234,6 +234,7 @@ dependencies {
     implementation 'com.android.support:preference-v7:27.1.1'
     implementation 'com.jakewharton:process-phoenix:2.0.0'
     debugImplementation 'com.facebook.stetho:stetho:1.5.0'
+    implementation "org.mozilla.components:fretboard:0.16"
```
pocmo: Above is a block of other components that uses "$mozilla_components_version" to keep the version number in sync.

app/src/debug/java/org/mozilla/focus/web/Config.kt
```diff
@@ -6,4 +6,5 @@ package org.mozilla.focus.web
 
 object Config {
     const val DEFAULT_NEW_RENDERER = false
+    const val EXPERIMENT_DESCRIPTOR_GECKOVIEW_ENGINE = "4eb60ebe-6bbc-4c26-8221-e446114a3813"
```
pocmo: Urk! This must have slipped when I reviewed the library. The descriptors should use the experiment name and not the ID. The ID is something Kinto generates and not really in our control.
I guess for this iteration we will need to ship as-is but I'll file an issue to get this fixed in a future release of Fretboard.

app/src/main/java/org/mozilla/focus/menu/browser/BrowserMenuAdapter.kt
```diff
@@ -111,7 +111,8 @@ class BrowserMenuAdapter(
                     R.drawable.ic_settings_menu_item))
         }
 
-        if (AppConstants.isGeckoBuild(context)) {
+        if (AppConstants.isGeckoBuild) {
+            items.add(MenuItem.RequestDesktopCheck)
```
pocmo: Oh, how did this item get in here? Is this a related change?

app/src/main/java/org/mozilla/focus/settings/ExperimentsSettingsFragment.kt
```diff
@@ -9,23 +9,37 @@ import android.os.Bundle
...
 import org.mozilla.focus.R
 import org.mozilla.focus.web.Config
 import org.mozilla.focus.web.ENGINE_PREF_STRING_KEY
 
+const val EXPERIMENTS_JSON_FILENAME = "experiments.json"
+const val EXPERIMENTS_BASE_URL = "https://settings.prod.mozaws.net/v1"
+const val EXPERIMENTS_BUCKET_NAME = "main"
+const val EXPERIMENTS_COLLECTION_NAME = "focus-experiments"
+
+val experimentDescriptor = ExperimentDescriptor(Config.EXPERIMENT_DESCRIPTOR_GECKOVIEW_ENGINE)
```
pocmo: nit: Over time we will probably create a bunch of descriptors that we use in various places of the app. Should we centralized them somewhere? Maybe in an `Experiments` object? `Experiments.GECKOVIEW` ?

app/src/main/java/org/mozilla/focus/settings/ExperimentsSettingsFragment.kt
```diff
@@ -9,23 +9,37 @@ import android.os.Bundle
...
         addPreferencesFromResource(R.xml.experiments_settings)
-        val enginePref: SwitchPreference? = preferenceManager
-                .findPreference(ENGINE_PREF_STRING_KEY) as SwitchPreference?
-        enginePref?.isChecked = preferenceManager.sharedPreferences
-                .getBoolean(ENGINE_PREF_STRING_KEY, Config.DEFAULT_NEW_RENDERER)
+        enginePreference = preferenceManager
+                .findPreference(ENGINE_PREF_STRING_KEY) as SwitchPreference
+        application = activity?.application as FocusApplication
+        fretboard = application.fretboard
+        enginePreference.isChecked = fretboard.isInExperiment(application, experimentDescriptor)
```
pocmo: nit: Maybe we could make Fretboard (or a an isInExperiments method?) available via an extension method?

Maybe something like:

```Kotlin
val Context.application: FocusApplication
    get() = applicationContext as FocusApplication

val Context.fretboard: Fretboard
    get() = application.fretboard
```

Or maybe additionally:

```Kotlin
fun Context.isInExperiment(descriptor: ExperimentDescriptor): Boolean =
    fretboard.isInExperiment(this, experimentDescriptor)
```

What do you think?

app/src/main/java/org/mozilla/focus/settings/ExperimentsSettingsFragment.kt
```diff
@@ -37,6 +51,7 @@ class ExperimentsSettingsFragment : PreferenceFragmentCompat(), SharedPreference
         super.onPause()
         preferenceScreen?.sharedPreferences?.unregisterOnSharedPreferenceChangeListener(this)
```
pocmo: Do we still need the pref handling code in here? determineEngine() only seems to use the override?

app/src/main/java/org/mozilla/focus/utils/AppConstants.kt
```diff
@@ -0,0 +1,33 @@
...
+    val isGeckoBuild: Boolean
+        get() = (WebViewProvider.engine == null && Config.DEFAULT_NEW_RENDERER) ||
+                (WebViewProvider.engine?.javaClass?.simpleName == GeckoWebViewProvider::class.java.simpleName)
+
+    val isDevBuild: Boolean
+        get() = !isReleaseBuild
+
+    fun supportsDownloadingFiles(): Boolean {
+        return true
+    }
```
pocmo: Ah, this is nice now! :)

app/src/test/java/org/mozilla/focus/search/SearchEngineWriterTest.kt
```diff
@@ -5,14 +5,16 @@
...
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
-import org.junit.Assert.assertEquals
 
 @RunWith(RobolectricTestRunner::class)
 class SearchEngineWriterTest {
 
+    @Ignore // FIXME Broken test
     @Test
```
pocmo: Hu? Maybe broken on Windows machines because it tries to read a local file from a path we build together assuming it's a Unix system? The test seems to pass on master. Let's not disable it. :) But we could file a separate issue to make it use `pathSeparator` or whatever is the reason for this broken test on your machine.

app/src/main/java/org/mozilla/focus/FocusApplication.kt
```diff
@@ -60,10 +72,21 @@ class FocusApplication : LocaleAwareApplication() {
         customTabSessions.observeForever(TelemetrySessionObserver())
     }
 
+    fun loadExperiments() {
+        val experimentsFile = File(filesDir, EXPERIMENTS_JSON_FILENAME)
+        val experimentSource = KintoExperimentSource(
+                EXPERIMENTS_BASE_URL, EXPERIMENTS_BUCKET_NAME, EXPERIMENTS_COLLECTION_NAME)
+        fretboard = Fretboard(experimentSource, FlatFileExperimentStorage(experimentsFile))
+        launch(CommonPool) {
+            fretboard.loadExperiments() // load current settings from disk
```
mcomella: Do we want to put this on an IO context?

Context: the CommonPool has a limited number of threads so doing several blocking calls (I'm assuming loading settings from disk is blocking) can delay others from using the CommonPool who probably don't expect their execution to be delayed. See https://mail.mozilla.org/pipermail/sync-dev/2018-July/001667.html for a better explanation.

colintheshots: LOL. I made the same complaint earlier about another PR. I think we should create our own thread pool on startup.

app/src/test/java/org/mozilla/focus/search/SearchEngineWriterTest.kt
colintheshots: Yes, this test broke my ability to commit because the unit test failed the pre-commit hook. So it's more than cosmetic. A broken test is a broken test.

app/src/main/java/org/mozilla/focus/settings/ExperimentsSettingsFragment.kt
colintheshots: Oh, right. This is no longer needed.

app/src/main/java/org/mozilla/focus/settings/ExperimentsSettingsFragment.kt
colintheshots: That works for me.

app/src/main/java/org/mozilla/focus/menu/browser/BrowserMenuAdapter.kt
colintheshots: I'm fairly certain it came from a merge conflict. I can pull it.

app/src/debug/java/org/mozilla/focus/web/Config.kt
colintheshots: Well, if we use the name instead, we need some way to specify the environment as well.

app/build.gradle
colintheshots: I'll use it.

app/src/main/java/org/mozilla/focus/FocusApplication.kt
colintheshots: Technically, there is no IO CoroutineDispatcher. That's from RxJava. I just created my own.

app/src/test/java/org/mozilla/focus/search/SearchEngineWriterTest.kt
pocmo: On the command line you can bypass the hook for pushing with "--no-verify"

colintheshots:
I'm going to hold off on the JobSchedulerSyncScheduler for now. I'd prefer to wait for WorkManager to do the job properly. The code will wait until the second app restart to change engines until then. Even with background jobs, we must wait for a real app restart unless we plan to clobber the user's session. Google just released WorkManager -alpha6 and I believe the final version will be well out in time for Android Dev Summit.

colintheshots:
I can't merge now until someone approves my changes to Kinto production. The code depends upon the name changes.

pocmo:
✅  I approved your Kinto changes.

PR #3051, opened 2018-08-02
Closes #3049: Biometric crashes & bug fixes

### LINKED ISSUE #3049: App crashes when using app with no fingerprints setup
## Steps to reproduce
- open app
- put in url, hit submit

### Device information

Sim (without biometric capabilities?)

sblatz:
This PR fixes a crash caused by an edge case of un-enrolling all fingerprints with fingerprint hardware available.

It also fixes a few other bugs such as the settings option not disappearing when fingerprint authentication is not available.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
@@ -67,6 +67,7 @@ import org.mozilla.focus.architecture.NonNullObserver;
 import org.mozilla.focus.autocomplete.AutocompleteQuickAddPopup;
 import org.mozilla.focus.biometrics.BiometricAuthenticationDialogFragment;
 import org.mozilla.focus.biometrics.BiometricAuthenticationHandler;
+import org.mozilla.focus.biometrics.Biometrics;
 import org.mozilla.focus.broadcastreceiver.DownloadBroadcastReceiver;
 import org.mozilla.focus.customtabs.CustomTabConfig;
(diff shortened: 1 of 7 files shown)


app/src/main/res/layout/biometric_prompt_dialog_content.xml
```diff
@@ -26,15 +26,15 @@
         android:gravity="center">
 
         <ImageView
-            android:id="@+id/fingerprint_icon"
+            android:id="@+id/fingerprintIcon"
```
boek: nit: I would keep these as snake_case to be consistent. Another perk that I've noticed is that it's easier to tell in the Kotlin files that the variables are generated ones from the xml

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
```diff
@@ -894,13 +901,25 @@ public void onStatusBarHeightFetched(int statusBarHeight) {
...
             displayBiometricPromptIfNeeded();
         } else {
             getView().setAlpha(1);
         }
     }
 
+    // Handles the edge case of a user removing all enrolled prints while auth was enabled
+    private void checkBiometricStillValid() {
+        // Disable biometrics if the user is no longer eligible due to un-enrolling fingerprints:
+        if (!Biometrics.INSTANCE.hasFingerprintHardware(getContext())) {
```
boek: I would probably move this behavior to the Biometrics class and then call it in MainActivity

PR #3053, opened 2018-08-03
Build and target SDK 28 (9.0) Closes #2909 

### LINKED ISSUE #2909: Build and target SDK 28 (9.0)

* It will be a requirement for using new OS features
* The longer we stay behind the harder it gets to update
* We want to update to 28 in the components soon. To avoid conflicts we want to update the apps too.

### What / Requirements

* [ ] Update build.gradle files
* [ ] Build and run tests to see if there are issues that need to be fixed
* [ ] Is there a 28 support library release or does switching to SDK 28 mean we need to switch to the androidx.* support libraries (many of them seem to still have alpha version numbers).

ekager:
Sorry this is a lot, most of the changes are updating to use support library.


colintheshots:
Do not merge this! This might be more serious than I thought. WebViews targeting API 28 do not allow accessing non-SSL URLs. This means most URLs fail outright, even if they're only going to redirect to an https URL. I haven't found a way to toggle allowing insecure URLs back on quite yet.

colintheshots:
False alert. It looks like adding android:usesCleartextTraffic="true" to the manifest works around it. I had tried almost everything involving the network security config XML.

PR #3032, opened 2018-07-31
Closes #3024 - Updates Icons to new SVGs

ekager:

Issue #3050, opened 2018-08-02
New delete icon uses evenOdd vector fillType 


ekager:
evenOdd fillType is not supported by older versions of Android. @brampitoyo I don't know much about SVGs is it possible to get an icon without this fillType?

ekager:
Maybe this could help - https://medium.com/thoughts-overflow/the-mystery-of-the-disappearing-holes-a-gripping-tale-of-using-svg-in-android-442f6035a452

ekager:
It looks like a few of the new icons optimized for mobile include the even-odd fillType we'll need new icons for those as well

brampitoyo:
@ekager I will see if reversing the path order and changing the fill type to nonzero will fix our problem. Thanks for investigating!

brampitoyo:
[focus-android-icons-nonzero.zip](http://f.cl.ly/items/2y1M033a2x032d1L2623/focus-android-icons-nonzero.zip)

I was able to fix all the affected icons:

* delete
* device-desktop
* internet
* language
* lock
* settings
* tracking-protection
* tracking-protection-disabled

The fixed SVGs don’t contain any `fillType` property (not even `nonzero`). This should minimise our problems going forward.

PR #3054, opened 2018-08-03
Update Menu SVG and remove Request Desktop Icon (cuts off) Closes #3032

ekager:


colintheshots:
I think @pocmo would point out that new icons belong in android-components. Otherwise, this looks fine.

ekager:
Added some of the new icons to Android components, but see #3050 for why we need some new icons in the next week or so

PR #3073, opened 2018-08-06
Fix manual A/B switcher so it can access Fretboard

colintheshots:


app/src/main/java/org/mozilla/focus/settings/ExperimentsSettingsFragment.kt
```diff
@@ -22,9 +21,7 @@ class ExperimentsSettingsFragment : PreferenceFragmentCompat(), SharedPreference
     }
 
     private var rendererPreferenceChanged = false
-    private lateinit var fretboard: Fretboard
```
boek: Were these ever initialized?

colintheshots: Yes, the old code got ripped out that initialized it. I added extension methods, but the code here was referencing this old var by mistake.

PR #3090, opened 2018-08-08
Fixes #3046: Frozen URL bar

### LINKED ISSUE #3046: URL Bar freeze
Probably a regression from URL Bar UI (and animation) changes 

## Steps to reproduce
Unclear. I've saw this once yesterday and someone contacted me on Twitter to report the same, but no reliable STR that I know of.

1. Be browsing and press erase button

### Expected behavior
URL Bar clears and you're able to enter a new address.

### Actual behavior
URL Bar freezes with last address you visited and you can't interact with it. Have to remove Focus from recent apps for this to be fixed. 
38242773_10155955539714865_8613791000730533888_n

### Device information

* Android device: pixel 2
* Focus version: Nightly

sblatz:
I am unable to produce the frozen URL bug after this patch. However, I want to confirm there are no security concerns with keeping the previous fragment on the backStack. There may be a cleaner solution that exists, but after several hours of working on it, this was the only one I could figure out. If this solution is insufficient perhaps someone else should try a crack at it, as I'm unsure what else to try.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.java
@@ -240,6 +240,7 @@ public class MainActivity extends LocaleAwareAppCompatActivity {
 
         transaction
                 .replace(R.id.container, UrlInputFragment.createWithoutSession(), UrlInputFragment.FRAGMENT_TAG)
+                .addToBackStack("browserFragment")
                 .commit();
     }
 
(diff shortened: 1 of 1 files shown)


colintheshots:
The fix is worse than the disease here. The main issue is that Session deletion can be interrupted, not the symptom that it leaves us in a bad state. We're neither fixing the root cause nor the reason the user is likely to be upset. Our app claims to destroy all evidence of the session, so it's not okay to leave it running in memory.

PR #3091, opened 2018-08-08
Update gradle to 4.4 and plugin to 3.1.3

colintheshots:


app/build.gradle
```diff
@@ -110,16 +110,6 @@ android {
         klarX86Debug.root = 'src/klarDebug'
         klarX86Debug.root = 'src/klarDebug'
         klarX86Debug.root = 'src/klarDebug'
-
-        // UI test
-        androidTestWebviewCompile {
```
boek: Is it okay to remove these without replacing them?

colintheshots: These sourceSets were there to handle separate AndroidTest instrumentation roots for Webview and GeckoView. No-Jun already migrated all of those tests to the androidTest directory, where they belong in a default Android app root. He handled the dependency on each rendering engine by adding code that ignores the test if it's for the other engine.

Issue #3037, opened 2018-08-01
Telemetry in Nightly builds broken?


pocmo:
I think the merge of 2eae0a236057bf4cc525d6819ed9fd378ab72c25 may have broken sending telemetry in Nightly builds (nightly != release).

I think `isDevBuild()` should continue to check for the dev build type (because that's the only dev type) and if we do not want to send telemetry for other types (like coverage) then we should do that explicitly in TelemetryWrapper and not change `isDevBuild()` to do something else.

PR #3104, opened 2018-08-11
For #3037 - Reverts change to isDevBuild constant

boek:


app/src/main/java/org/mozilla/focus/utils/AppConstants.kt
```diff
@@ -12,6 +12,7 @@ import org.mozilla.focus.web.WebViewProvider
 
 object AppConstants {
     private const val BUILD_TYPE_RELEASE = "release"
+    private val BUILD_TYPE_DEBUG = "debug"
```
colintheshots: Make this const for inlining.

PR #3110, opened 2018-08-12
Addresses nits in l10n review

boek:


boek:
Going to land this since they all string related to export so l10n can review.

boek:
mozilla-l10n/focus-android-l10n#39

PR #3117, opened 2018-08-13
fix "Add to home screen" force close on fresh installation

SubhrajyotiSen:
When trying to invoke getString() inside onSharedPreferenceChanged(), the Fragment is not attached to a context as shown by the error log
    java.lang.IllegalStateException: Fragment UrlInputFragment{2bbadca} not attached to a context.

using conext explicitly to invoke getString() resolves this


colintheshots:
Virtually the same fix already landed from another PR. The merge conflict shows that we already have a fix in place. Thanks for fixing it though.

SubhrajyotiSen:
The other PR seemed like a version bump. That's why I didn't check it properly. My bad

PR #3135, opened 2018-08-15
Fix multi-architecture versionCode numbering

colintheshots:
GeckoView builds appended a prefix for the versionCode of 1, 2, or 3. This meant ARM and ARM64 builds could not be deployed to Google Play as the build number would decrease. This fix should start version numbering with the latest '3' prefix and numbers that only go upwards.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -242,19 +242,16 @@ android.applicationVariants.all { variant ->
         // The Google Play Store does not allow multiple APKs for the same app that all have the
         // same version code. Therefore we need to have different version codes for our ARM and x86
         // builds.
-        // Our generated version code has a length of 8 (See tools/gradle/versionCode.gradle).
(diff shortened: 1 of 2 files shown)


app/build.gradle
```diff
@@ -242,19 +242,16 @@ android.applicationVariants.all { variant ->
...
         if (variant.flavorName.contains("X86")) {
-            versionCode = versionCode + (3 * multiplier)
+            versionCode = versionCode + 2
         } else if (variant.flavorName.contains("Aarch64")) {
-            versionCode = versionCode + (2 * multiplier)
-        } else if (variant.flavorName.contains("Arm")) {
-            versionCode = versionCode + (1 * multiplier)
-        }
+            versionCode = versionCode + 1
+        } else // variant.flavorName.contains("Arm")) use generated version code
```
pocmo: This lonely `else` .. is that valid Groovy code?

colintheshots: Oops, I'll fix that.

PR #3136, opened 2018-08-15
Update Home Icon Closes #3130

### LINKED ISSUE #3130: Home SVG uses even-odd fillType
Sorry @brampitoyo one more that I found! The home icon currently uses the evenOdd fill type, could we get a new icon for this one too? :)

ekager:


app/src/main/res/drawable/ic_home.xml
```diff
@@ -1,14 +1,12 @@
-<?xml version="1.0" encoding="utf-8"?>
```
boek: Do we not want this at the top of the drawable files?

ekager: Oops we do; updating

ekager:
Also added licenses to new files imported last week without them

PR #3145, opened 2018-08-16
Fixes #3085 - Nests search suggestions in a scrollview

### LINKED ISSUE #3085: The first search suggestion box on Android tablet covers over all the other suggestions
### Steps to reproduce

* Turn search suggestions on
* Enter any keyword
* Scroll down

### Expected behavior

All search suggestions should appear on top of each other.

### Actual behavior

The first search suggestion box covers all the other suggestions.

screenshot_1533689189

boek:


boek:
@colintheshots potentially, what would that do?

colintheshots:
fillViewport allows child view to expand to the height of the parent, if needed. This matters when the child view is shorter than expected. I'd try with and without the setting using only one search suggestion in the recyclerview and see which works better.
http://www.curious-creature.com/2010/08/15/scrollviews-handy-trick/

boek:
@colintheshots Hmm, I tried with one/two results and many results toggling it on and off. I didn't notice any difference (I'm still not sure exactly what I'm looking for). so I'm going to merge this for now :)

PR #3154, opened 2018-08-17
Closes #2947: Change settings title from to "Settings menu"

### LINKED ISSUE #2947: [A11y] Settings menu, the title doesn't say "menu" or "list"
## Steps to reproduce
1. Activate TalkBack.
2. Open a webpage.
3. Open the main menu.
4. Open the Settings menu.

### Expected behavior
The Settings menu title should also say that it's a "menu" or "list".

### Actual behavior
The reader says the "Settings" title only.

### Device information

* Android device: 
LG Nexus 5 (Android 6.0.1)
Google Pixel (Android P beta)
* Focus version: 6.3

vvengers:
I accidentally changed the gradle version in the first commit to 3.1.4. I changed this back to 3.1.3 in the second commit together with the intended change.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -48,7 +48,7 @@
 
     <!-- Menu item: Shows information about features and changes that are new to this version -->
     <string name="menu_whats_new">What\'s New</string>
-    <string name="menu_settings">Settings</string>
+    <string name="menu_settings">Settings menu</string>
     <string name="menu_about">About</string>
     <string name="menu_help">Help</string>
     <string name="menu_rights">Your Rights</string>
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.4'
+        classpath 'com.android.tools.build:gradle:3.1.3'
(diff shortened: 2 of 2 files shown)


ilayaraja97:
[comment to this pull request](https://github.com/mozilla-mobile/focus-android/issue/2947#issuecomment-413929832)

colintheshots:
There aren't a lot of good options here. Either we change the title for all users or we add a subtitle. The latter doesn't seem as bad to me. Maybe see if we can come up with a good subtitle string.

colintheshots:
@brampitoyo What do you think? We can't add a contentDescription to a fragment title, but changing "Settings" to "Settings Menu" for all users seems a bit much. Should we add subtitles to help Talkback users?

brampitoyo:
> …changing "Settings" to "Settings Menu" for all users seems a bit much.

Totally agreed. Visually showing “Settings Menu” is unacceptable for multiple reasons, chief of which is violating user expectation that this page – and this page in every other app on the user’s phone – is called “Settings”, not “Settings Menu”.

Furthermore, the item inside the main menu is called “Settings”, and the title of the page should reflect that. It makes sense, when using TalkBack, for the system to also say “Settings”.

(Try using TalkBack and opening the Android Settings app. TalkBack will call the title of this page “Settings”.)

PR #3184, opened 2018-08-21
DO NOT MERGE New tabs prototype

colintheshots:
This is an early UX exploration of how tabs might be redone to give users full control over their tabs.


colintheshots:
This seems to be mostly working as intended. There may be memory issues, but I managed to load more than ten pretty heavy sessions at once.

It's good enough in that sense, but the loading status hangs if I exit to the switcher fragment and return. I need one more fix to restore the page load progress.

colintheshots:
So there are three remaining issues I'm aware of:

1. Jumping to the tab switcher before the page has obtained DNS leaves the tab in a bad state. The progress bar is stuck at 5% and the tab switcher shows the page title as about:blank. The page does not load when you return.
2. Jumping to the tab switcher after DNS completes, but long before the page has finished loading leaves the progress bar stuck even though the page loads completely. Also, it may never stay loaded, but may reload each time the user jumps between the switcher and the page.
3. Switching too rapidly between tabs with autoplaying video currently crashes the app. Youtube reliably crashes when swiping too quickly back and forth.

pocmo:
@colintheshots I like your tabs prototype! The initial design of tabs in Focus was intentionally sparse - but with all the feedback we can come up with something better now that better suits the user's use cases.

I would like to encourage you to try building the new tabs tray as a component. We have some building blocks in the a-c repository already. Those may be a good starting point:

* `concept-tabstray` includes the interfaces that describe a tabs tray. Other components using a tabs tray only know about this *concept* component so that they should be able to work with any implementation of it.

* `browser-tabstray` is an implementation of the concept / interface. It's not nice and it didn't follow any UX spec. It's just a basic boring implementation to test with tabs in our sample apps. Your tabs tray implementation could replace **this** implementation - or you could land it as a separate component like  `browser-tabstray-pager` or something like that.

* `feature-tabs` is the *feature* component that connects everything. Here you'll find the presenters etc.. In a nutshell this component will subscribe to state changes and call the corresponding methods on a `concept-tabstray` implementation to update the tabs tray.

* `samples-browser` is a simple browser app using the components above and others. It's good for testing the components. Whenever needed we create other sample apps in the repository.

Some reasons why I think this would be a great candidate for a component:

* The only real dependency that `browser-tabstray` currently has is `browser-session` (It obviously needs to know about the state). This component is about to land in Focus and will replace all the legacy session code (this already happened in Firefox for Fire TV recently). With that we will have the foundation to "drop in" tabs tray implementations like yours (+ `feature-tabs` to connect things).

* You working towards an **actual** tabs tray implementation will help us to improve and harden the `concept-tabstray` interfaces needed to describe a tabs tray.

* Your tabs tray component can be used in other apps and/or be the foundation for the tabs tray we are going to need for Fenix and others.

Let us know if we can help with componentizing this :)

Sdaswani:
yes would love to see this as a component @colintheshots!

colintheshots:
I will consider this branch for creating an Android component.

PR #3189, opened 2018-08-22
Upgrade components, Gradle, Kotlin, Support  ✌

colintheshots:


ekager:
Looks like we need to match their support library versions

ekager:
Maybe we should update to 0.20 now? Looks like taskcluster issues might be from locale imports

PR #3206, opened 2018-08-23
Save/Restore GV with tab switching

ekager:


ekager:
This doesn't work reliably because onRestoreInstanceState is not always called, like when the screen is turned off + then on. I need to find a reliable lifecycle call to know when we are switching tabs or when the app is actually being backgrounded for this way to work. CC @pocmo  @colintheshots any ideas?

ekager:
The onCreateView code is only for non GeckoView; I thought moving it to onResume for GV would give restoring more time to happen.

colintheshots:
Please do address the duplicate code. It shouldn't be in both lifecycle callbacks. If it did belong, we wouldn't want identical code verbatim in both places.

PR #3219, opened 2018-08-24
No issue: add CODEOWNERS file.

mcomella:
I created a new APT subteam to manage this because we're totally going
to forget to update this file and maybe less so the codeowners team.

Based on https://help.github.com/articles/about-codeowners/


boek:
perfect!

PR #3223, opened 2018-08-27
add tip screenshot for 7.0 

npark-mozilla:
Update l10n screenshot code to:
add the search suggestion help balloon
add the main screen tip
use sleep() instead of taking dummy screenshots (hopefully we can replace this with idlingresource definition soon)


boek:
Looks like there is a linting error, once that is fixed I think we'll be gtg 👍

Issue #3190, opened 2018-08-22
Convert Java classes to Kotlin to make browser-session integration easier


pocmo:
This is a prerequisite for #3179 and will dramatically reduce the patch size and make the integration easier.

It would be helpful if we could convert (and clean up) the following files:

Activities:
* [x] app/src/main/java/org/mozilla/focus/activity/EraseShortcutActivity.java
* [x]  app/src/main/java/org/mozilla/focus/activity/MainActivity.java
* [x]  app/src/main/java/org/mozilla/focus/activity/TextActionActivity.java

Fragments:
* [x]  app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.java
* [x]  app/src/main/java/org/mozilla/focus/fragment/WebFragment.java
* [x]  app/src/main/java/org/mozilla/focus/session/ui/SessionsSheetFragment.java

WebView:
* [x]  app/src/main/java/org/mozilla/focus/webview/SystemWebView.java

Others:
* [x]  app/src/main/java/org/mozilla/focus/menu/browser/RequestDesktopCheckItemViewHolder.java
* [x]  app/src/main/java/org/mozilla/focus/menu/context/WebContextMenu.java
* [x]  app/src/main/java/org/mozilla/focus/session/SessionNotificationService.java
* [x]  app/src/main/java/org/mozilla/focus/session/ui/EraseViewHolder.java
* [x]  app/src/main/java/org/mozilla/focus/session/ui/SessionsAdapter.java
* [x]  app/src/main/java/org/mozilla/focus/utils/SupportUtils.java

pocmo:
CC @boek @colintheshots

ekager:
Closing, they all look complete to me

pocmo:
Yes, thanks! :)

PR #3225, opened 2018-08-27
Issue #3190: Convert activities from Java to Kotlin.

pocmo:
Started to work a bit on #3190 to unblock the component integration.

Those classes were mostly converted by Android Studio. I went over it and cleaned up some warnings.


app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -0,0 +1,300 @@
...
+
+    override fun onCreateView(name: String, context: Context, attrs: AttributeSet): View? {
+        return if (name == IWebView::class.java.name) {
+            // Inject our implementation of IWebView from the WebViewProvider.
+            WebViewProvider.create(this, attrs)
+        } else super.onCreateView(name, context, attrs)
+    }
+
+    override fun onBackPressed() {
+        val fragmentManager = supportFragmentManager
```
colintheshots: This feels redundant.

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -0,0 +1,300 @@
...
+
+        val sessionsSheetFragment = fragmentManager.findFragmentByTag(
+            SessionsSheetFragment.FRAGMENT_TAG) as SessionsSheetFragment?
+        if (sessionsSheetFragment != null &&
```
colintheshots: We use this pattern three times in a row, reusing nearly identical code. It seems we could take advantage of Kotlin's lambdas and non-local return at label syntax to simplify this.

app/src/main/java/org/mozilla/focus/activity/TextActionActivity.kt
```diff
@@ -0,0 +1,45 @@
...
+
+        val searchTextCharSequence = intent.getCharSequenceExtra(Intent.EXTRA_PROCESS_TEXT)
+        val searchText = if (searchTextCharSequence != null) {
+            searchTextCharSequence.toString()
+        } else {
+            ""
+        }
+        val searchUrl = UrlUtils.createSearchUrl(this, searchText)
+
+        val searchIntent = Intent(this, MainActivity::class.java)
```
colintheshots: Use a let() block to initialize the Intent. We should really have a Kotlin Intent builder, but let's not worry about it. It seems likely KTX will come out with something soon.

PR #3226, opened 2018-08-27
Issue #3190: Convert fragments from Java to Kotlin.

pocmo:
Started to work a bit on #3190 to unblock the component integration.

I first converted the files to Kotlin from Android Studio and then cleaned it up. `BrowserFragment` was converted with a bunch of errors and gazillion of warnings. There's more that can be made nicer now. However I would like to get this in first.


pocmo:
I need to rebase this after #3225 lands to avoid conflicts.

app/src/main/java/org/mozilla/focus/fragment/WebFragment.kt
```diff
@@ -0,0 +1,135 @@
...
+        return view
+    }
+
+    override fun applyLocale() {
+        val context = context
+        val localeManager = LocaleManager.getInstance()
+        if (!localeManager.isMirroringSystemLocale(context)) {
+            val currentLocale = localeManager.getCurrentLocale(context)
+            Locale.setDefault(currentLocale)
+            val resources = context!!.resources
```
colintheshots: Instead of crashing when context isn't set, should we just use the current/default locale?

app/src/main/java/org/mozilla/focus/session/ui/SessionsSheetFragment.kt
```diff
@@ -0,0 +1,138 @@
...
+        cardView!!.viewTreeObserver.addOnPreDrawListener(object : OneShotOnPreDrawListener(cardView!!) {
+            override fun onPreDraw(view: View) {
+                playAnimation(false)
+            }
+        })
+
+        val sessionsAdapter = SessionsAdapter(this)
+        SessionManager.getInstance().sessions.observe(this, sessionsAdapter)
+
+        val sessionView = view.findViewById<RecyclerView>(R.id.sessions)
```
colintheshots: Use a let() block here to init sessionView.

app/src/main/java/org/mozilla/focus/session/ui/SessionsSheetFragment.kt
```diff
@@ -0,0 +1,138 @@
...
+
+        return view
+    }
+
+    @Suppress("ComplexMethod")
+    private fun playAnimation(reverse: Boolean): Animator {
+        isAnimating = true
+
+        val offset = resources.getDimensionPixelSize(R.dimen.floating_action_button_size) / 2
+        val cx = cardView!!.measuredWidth - offset
```
colintheshots: If the fragment becomes detached during the animation, won't this crash?

app/src/main/java/org/mozilla/focus/session/ui/SessionsSheetFragment.kt
```diff
@@ -0,0 +1,138 @@
...
+        val cy = cardView!!.measuredHeight - offset
+
+        // The final radius is the diagonal of the card view -> sqrt(w^2 + h^2)
+        val fullRadius = Math.sqrt(
+            Math.pow(cardView!!.width.toDouble(), 2.0) + Math.pow(cardView!!.height.toDouble(), 2.0)
+        ).toFloat()
+
+        val sheetAnimator = ViewAnimationUtils.createCircularReveal(
```
colintheshots: Use a let() block to setup sheetAnimator.

app/src/main/java/org/mozilla/focus/session/ui/SessionsSheetFragment.kt
pocmo: I think it won't crash here as the `cardView` reference is never set to null - I made this more obviously by using `lateinit` in the updated patch. But I agree that this animation is pointlessly keeping a reference and "completing" the animation even though the fragment may no longer be attached - leaking the views temporarily.

pocmo:
Thank you @colintheshots. I updated+rebased the PR and addressed your review commits.

PR #3230, opened 2018-08-27
Add localeToNameMap

yfdyh000:
Allow to override the native name of locales.

Override the "中文 (中国)" to "中文 (中国大陆)" based on the Mozilla Simplified Chinese Community Discussion, moderate geopolitical issues for localizations.

Not compiled and tested.


Delphine:
Thanks for flagging me. If this is the communities' choice let's go by this, then.
I can't approve the code part though, sorry :/

yfdyh000:
@boek I am worried about the correctness of this code now.

`locale` seems to be an locale object, `containsKey` and `get` may not work.
`this.tag.equals` may be correct, but `String tag` doesn't seem to be used, and why it exists with `locale` at the same time.
`String toString` has been override so worthless.

PR #3257, opened 2018-08-30
Use vanilla Jacoco and re-enable Codecov

### LINKED ISSUE #3029: codecov is not aware of Kotlin
e.g. Kotlin files don't appear in the codecov results directories: https://codecov.io/gh/mozilla-mobile/firefox-tv/tree/master/app/src/main/java/org/mozilla/focus/browser

Our Jacoco android plugin doesn't support Kotlin apparently – https://github.com/arturdm/jacoco-android-gradle-plugin/issues/37 – but codecov does ([supported languages](https://docs.codecov.io/docs), [kotlin example](https://github.com/codecov/example-kotlin), [android example](https://github.com/codecov/example-android), [gradle example](https://github.com/codecov/example-gradle)) We can:
- Look in this issue for solutions
- Look for people who've forked and improved it the jacoco android plugin
- Write our own build code to run codecov (or make a PR upstream to the plugin)

colintheshots:
Fixes #3029

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
 script:
   - ./gradlew -Pcoverage --no-daemon clean assembleFocusX86Nightly
   - echo "--" > .adjust_token
-  - ./gradlew -Pcoverage --no-daemon jacocoTestFocusX86NightlyUnitTestReport
+  - ./gradlew -Pcoverage --no-daemon jacocoFocusX86NightlyTestReport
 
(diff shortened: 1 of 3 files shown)


colintheshots:
It's not entirely clear if this will resolve all of the Kotlin problems. It seems Kotlin-generated getters and setters from Kotlin properties may still be counted for code coverage. However, most issues should improve.

PR #3269, opened 2018-08-31
Issue #3190 Convert SystemWebView to Kotlin

ekager:


app/src/main/java/org/mozilla/focus/webview/SystemWebView.kt
```diff
@@ -0,0 +1,330 @@
...
+
+    override fun onResume() {
+        super.onResume()
+        resumeTimers()
+    }
+
+    override fun restoreWebViewState(session: Session) {
+        val stateData = session.webViewState
+
+        val backForwardList: WebBackForwardList? = if (stateData != null)
```
boek: kotlinify: `stateData?.let { super.restoreState(it) }`

app/src/main/java/org/mozilla/focus/webview/SystemWebView.kt
```diff
@@ -0,0 +1,330 @@
...
+    override fun setBlockingEnabled(enabled: Boolean) {
+        client.isBlockingEnabled = enabled
+        if (enabled) {
+            WebViewProvider.applyAppSettings(context, settings, this)
+        } else {
+            WebViewProvider.disableBlocking(settings, this)
+        }
+
+        if (callback != null) {
+            callback!!.onBlockingStateChanged(enabled)
```
boek: Should we just do a safe call? `callback?.onBlock..`

app/src/main/java/org/mozilla/focus/webview/SystemWebView.kt
```diff
@@ -0,0 +1,330 @@
...
+
+    override fun setRequestDesktop(shouldRequestDesktop: Boolean) {
+        if (shouldRequestDesktop) {
+            WebViewProvider.requestDesktopSite(settings)
+        } else {
+            WebViewProvider.requestMobileSite(context, settings)
+        }
+
+        if (callback != null) {
+            callback!!.onRequestDesktopStateChanged(shouldRequestDesktop)
```
boek: same as above

app/src/main/java/org/mozilla/focus/webview/SystemWebView.kt
```diff
@@ -0,0 +1,330 @@
...
+        return object : WebChromeClient() {
+            override fun onProgressChanged(view: WebView, newProgress: Int) {
+                if (callback != null) {
+                    // This is the earliest point where we might be able to confirm a redirected
+                    // URL: we don't necessarily get a shouldInterceptRequest() after a redirect,
+                    // so we can only check the updated url in onProgressChanges(), or in onPageFinished()
+                    // (which is even later).
+                    val viewURL = view.url
+                    if (!UrlUtils.isInternalErrorURL(viewURL) && viewURL != null) {
+                        callback!!.onURLChanged(viewURL)
```
boek: Do we still need the `!!` if we're inside a null check?

We could also potentially flatten this if we use the null check to early return.

app/src/main/java/org/mozilla/focus/webview/SystemWebView.kt
```diff
@@ -0,0 +1,330 @@
...
+            }
+
+            override fun onShowCustomView(
+                view: View,
+                webviewCallback: WebChromeClient.CustomViewCallback
+            ) {
+                val fullscreenCallback =
+                    IWebView.FullscreenCallback { webviewCallback.onCustomViewHidden() }
+
+                callback!!.onEnterFullScreen(fullscreenCallback, view)
```
boek: Do we always expect callback to not be null here?

app/src/main/java/org/mozilla/focus/webview/SystemWebView.kt
```diff
@@ -0,0 +1,330 @@
...
+                val download = Download(
+                    url,
+                    userAgent,
+                    contentDisposition,
+                    mimetype,
+                    contentLength,
+                    Environment.DIRECTORY_DOWNLOADS,
+                    null
+                )
+                callback!!.onDownloadStart(download)
```
boek: Do we still need the `!!` inside a null check?

boek:
Also looks like detektcheck doesn't like how many functions are in this class

Might have to suppress:
```
 TooManyFunctions - 21/10 - [SystemWebView] at focus-android/app/src/main/java/org/mozilla/focus/webview/SystemWebView.kt:47:1
```

app/src/main/java/org/mozilla/focus/webview/SystemWebView.kt
colintheshots: So this is something we'll see a lot, no doubt. Because callback is a var, Kotlin wisely surmises that code it doesn't know about might change its value to null from another thread. The simplest solution is to use callback?.let {} so the value inside that let block is no longer mutable.

app/src/main/java/org/mozilla/focus/webview/SystemWebView.kt
colintheshots: About the !! inside the null check. This is something we'll see a lot, no doubt. Because callback is a var, Kotlin wisely surmises that code it doesn't know about might change its value to null from another thread. The simplest solution is to use callback?.let {} so the value inside that let block is no longer mutable.

PR #3271, opened 2018-08-31
Consider going directly to website when clicking on a URL in suggestions

### LINKED ISSUE #3232: Consider going directly to website when clicking on a URL in search suggestions
Users may type in websites and get search suggestions completing to that website (e.g. type "facebook" and "www.facebook.com" will show up in the search suggestions). However, clicking them will do a search for that website (e.g. https://duckduckgo.com/?q=http%3A%2F%2Fwww.facebook.com%2F&t=hi&ia=web). They probably wanted to go directly to the website: maybe we should do that for them?

There may be privacy/security risks here in taking that extra step though.

colintheshots:
Fixes #3232


colintheshots:
Question about the implementation: The first result in search suggestions shows the search engine logo before the query. Should this still perform a search or should it go directly to the URL? Perhaps we should change the logo if we believe it's a URL?

boek:
So the first row w/ icon is to specifically search using the search engine. It is there even without search suggestions.

It really makes more sense in the autocomplete world where you want to search for `face` but autocomplete is trying to take you to `facebook.com`.

This is probably a good question for @brampitoyo

brampitoyo:
@boek @colintheshots 

The search engine favicon is there specifically to indicate that the page you’re about to open is hosted by Google, Yahoo, DuckDuckGo, etc.

We only show this favicon once to avoid the risk of being overwhelmed and desensitised by seeing the same icons repeated down the line.

But if the search suggestion specifically shows a URL, then it does make sense for us to go straight to the URL. What icon should we show in that case? The link icon – which I think we already have an SVG for.

An example search results that contain a URL in the middle:

    [G favicon] facebook
                facebook login
                facebook messenger
    [link.svg]  www.facebook.com
    [G favicon] facebook app

Example search results that contain a URL on top:

    [link.svg]  www.twitter.com
    [G favicon] twitter login
                twitter search
                twitter fortnite
                twitter app

colintheshots:
@brampitoyo The problem is that in the case of Google search suggestions and a URL typed in, we show the URL twice. Once with a search favicon and once without. This leads to obvious confusion. I changed my code to make the first version link to the search page for the query.

When we say, URL, I should specify that we're not requiring a schema. Hence, any text containing a period or colon, but no space, qualifies. We don't know positively that it's a URL. It could be a proper name like will.i.am, for example.

colintheshots:
Should I change the code to hide the first row when we detect a schema?

colintheshots:
What I'm doing here in the latest code is both. The above PR makes the first row with the search icon always search. The second row, which when using Google suggests the same text, does not show a search icon and goes directly to the URL if it looks like a URL. I feel this is a reasonable compromise.

The first row, which is hardcoded and always appears even when the network is slow, will always search.

brampitoyo:
> The first row, which is hardcoded and always appears even when the network is slow, will always search.

> The second row, which when using Google suggests the same text if it looks like a URL, does not show a search icon and goes directly to the URL if it looks like a URL.

This behaviour sounds reasonable to me.

Just one small request: will it be possible to show a link icon in cases where we make this “educated guess”, so the user knows that the link goes directly to a site (not a search results page)?

colintheshots:
Yes, it's possible. I'll have to do a bit of re-working on the PR.

colintheshots:
Here's the new version @brampitoyo 
screenshot from 2018-09-10 17-41-57

colintheshots:
Here's the new version @brampitoyo 
screenshot from 2018-09-10 17-45-14

ekager:
So we are accepting that this will sometimes lead to a nonexistent site like "facebook.comac" above?

colintheshots:
I believe it should be acceptable. I typed in a garbage URL. Facebook.om doesn't exist.

brampitoyo:
@ekager wrote:

> So we are accepting that this will sometimes lead to a nonexistent site like "facebook.comac" above?

Ideally, search queries that don’t fit a proper URL pattern shouldn’t be labelled a link. But is there an easy way for us to reliably detect these false positives?

I’m hoping that Google will demote their search ranking enough, so that there’s less chance of users scrolling down, seeing it, and tapping on it.

Does this assumption sound reasonable?

-----

@colintheshots The screenshot looks good, but I’ve got one concern.

My original design was to only **show the search provider icon once or twice**. Why? I was hoping that Google won’t show more than one URLs.

Ideally, like this:



So it’s clear: `twitter cma` and everything below it are search queries, and there’s only one URL recommendation.

However, your screenshot showed that Google may recommend many URLs. As a consequence, we’ll need to label every single link with an icon, so the user knows where it will lead.

But as a result, the few search queries we show are “lost” in a sea of URL recommendations!



Ok. Maybe one way to solve this problem is by giving each search query a search provider icon:



This is all well and good – except for the case when the search autocomplete contains no URL. Then every query is plastered with a search provider icon.

on 2x

Is there a way out of this problem, that will show as few icons as possible, while keeping the clarity of which item will lead to a link, and which will lead to a query?

brampitoyo:
One way to solve the icon saturation problem is by **making URLs distinct, and de-label search results**. This way, search results won’t saturate the field with their search provider (or magnifying glass) icons.

Of course, we then have to assume two things:

1. Users know which search engine is currently being used, even when there’s no search provider icon
2. Users know that tapping on a plain, unlabelled field will lead to a search results page (and not a URL)





-----

Another way to solve this problem is through **reordering and clear visual separation of different types of recommendations**.





This solution has no downsides, but would require us to sort search query by type. Is it possible to do?

colintheshots:
Here's the new version based on our discussion.
screenshot from 2018-09-12 20-55-23

PR #3277, opened 2018-09-01
Closes #3254: show checkbox while removing search engine 

### LINKED ISSUE #3254: Remove search engines doesn't work
## Steps to reproduce
1. Go to Settings>Search>Default search engine.
2. Tap the 3 dots button and select Remove.
3. Select the search engines you want to remove.

### Expected behavior
A checkbox is displayed for each engine to tap the ones you want to remove.

### Actual behavior
The checkboxes aren't displayed and can't select any of the installed engines.

### Device information
Google Pixel (Android 9)
LG Nexus 5 (Android 6.0.1)

* Focus version: 7.0

SubhrajyotiSen:
Also convert MultiselectSearchEngineListPreference to kotlin to override itemResId


app/src/main/java/org/mozilla/focus/search/MultiselectSearchEngineListPreference.kt
```diff
@@ -0,0 +1,75 @@
...
+        defaultButton.isActivated = true
+    }
+
+    // Whenever an engine is checked or unchecked, we notify the menu
+    private fun bindEngineCheckboxesToMenu() {
+        for (i in 0 until searchEngineGroup!!.childCount) {
+            val engineButton = searchEngineGroup!!.getChildAt(i) as CompoundButton
+            engineButton.setOnCheckedChangeListener { _, _ ->
+                val context = context
+                if (context is ContextThemeWrapper) {
```
colintheshots: I feel like this should be a global utility extension function on Context. We've repeated this pattern a lot in the code. Maybe Context.asActivity()?

SubhrajyotiSen: I found 2 such patterns. But how will the function actually work?
I am unable to understand how the utility function will handle the casting and after returning also execute the needed function call on the casted context

colintheshots: Here's the simplest version. Put this utility function in a util class and use it.

`fun Context.asActivity() = (this as? ContextThemeWrapper)?.baseContext as? Activity ?: this as? Activity`

`context?.asActivity()?.invalidateOptionsMenu()`

colintheshots: Using https://kotlinlang.org/docs/reference/typecasts.html#safe-nullable-cast-operator

SubhrajyotiSen:
@colintheshots  I have updated the PR using the extension function

PR #3280, opened 2018-09-02
No issue: Convert SettingsFragment to Kotlin

SubhrajyotiSen:


SubhrajyotiSen:
` ForbiddenComment - [onCreatePreferences] at focus-android/app/src/main/java/org/mozilla/focus/settings/SettingsFragment.kt:25:9`

The comment is compulsory right?

colintheshots:
Detekt is saying that TODO, FIXME comments cannot be checked in because they usually mean you missed something in your PR. You can always suppress these kinds of warnings if they aren't appropriate. @Suppress("ForbiddenComment")... https://github.com/arturbosch/detekt/blob/master/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/ForbiddenComment.kt

SubhrajyotiSen:
Since the TODO is for future reference with respect to an issue, suppressing it should be fine right?

colintheshots:
Yes, I believe that's fine. If that passes TaskCluster, I'll approve and override the codecov requirement, since this was simply converting existing code.

PR #3284, opened 2018-09-02
No issue: Redundant keyboard opening removed

EmmanuelMess:
`UrlInputFragment.onStart()` already calls `showKeyboard()`, `FirstrunFragment.finishFirstrn()`'s call is redundant, am I missing something?
Also, I didn't create an issue, but can if you feel it necessary.

All Robolectric test run OK. Device tests: 31/42 tests failed and 1 was not run, but they don't seem relevant to this PR. Should I share the logs?


EmmanuelMess:
Maybe also remove checking code from here, as `onStart()` will only get called after the fragment becomes visible:

https://github.com/mozilla-mobile/focus-android/blob/b12326b2f4eab052e0682292ec3ced30cb16ab82/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt#L341-L346

EmmanuelMess:
> Could you make the second change you mentioned?
> Thanks for your help!

No problem, done!

PR #3333, opened 2018-09-06
No issue: Suspend Media When Inactive GV

ekager:


ekager:
Note: I think this is actually not working in GV, but I want to add the call anyway so that when it's fixed we have it 👍

PR #3340, opened 2018-09-06
Don't automatically open new tabs Closes #3300

### LINKED ISSUE #3300: Make the "Open link in new tab" to just open a new tab, not also navigate to it
The current implementation opens the link and takes the user directly to the newly opened tab. This can be annoying if the user wants to open multiple tabs and then navigate to each of them separately. 

The action should just open a tab and let the user say on the current page. Maybe a toggle could be implemented for the option in Setting. That way both functionalities still exist.

ekager:
@brampitoyo I think this is almost done. 

The Setting:
screenshot_20180906-153448

The Snackbar:
screenshot_20180909-122929

I had a few questions about the Setting:
1. There are 0 other checkboxes in settings and it looked a bit strange so I made it a switch, but I can change it if you want. 
2. Should there be a title + summary for the setting, or just "When you open link in a new tab, switch to it immediately". Currently it's just that String. 
3. I assumed that this setting would go in Mozilla Settings because it's not privacy/search related, was that correct? 

I can send you a build to test it out if you like, let me know! I'm assuming this is going into the next release, so no rush

### DIFF
--- a/app/src/main/java/org/mozilla/focus/session/SessionManager.java
+++ b/app/src/main/java/org/mozilla/focus/session/SessionManager.java
@@ -4,18 +4,14 @@
 
 package org.mozilla.focus.session;
 
-import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.annotation.VisibleForTesting;
-import android.support.design.widget.Snackbar;
 import android.text.TextUtils;
-import android.view.View;
 
-import org.mozilla.focus.R;
 import org.mozilla.focus.architecture.NonNullLiveData;
 import org.mozilla.focus.architecture.NonNullMutableLiveData;
 import org.mozilla.focus.customtabs.CustomTabConfig;
@@ -23,7 +19,6 @@ import org.mozilla.focus.shortcut.HomeScreen;
 import org.mozilla.focus.utils.AppConstants;
 import org.mozilla.focus.utils.Settings;
 import org.mozilla.focus.utils.UrlUtils;
-import org.mozilla.focus.utils.ViewUtils;
 import org.mozilla.focus.web.GeckoWebViewProvider;
 import org.mozilla.geckoview.GeckoSession;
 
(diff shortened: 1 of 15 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -721,4 +721,11 @@ be temporary, and you can try again later.</li>
     <!-- Tip displayed on home view explaining how to add a custom autocomplete URL
         Argument 1 is the app name (Firefox Focus/Firefox Klar/etc). -->
     <string name="tip_disable_tips">Turn off tips on the %1$s start screen</string>
+
+    <!-- Displayed in a snackbar when a user opens a new tab, gives action to switch to the newly opened tab -->
+    <string name="new_tab_opened_snackbar">New Tab Opened</string>
```
EmmanuelMess: Small nitpick, but shouldn't this be "New tab opened"?

app/src/main/java/org/mozilla/focus/session/SessionManager.java
```diff
@@ -282,6 +293,28 @@ private void addSession(Session session) {
         }
     }
 
+    private void addNewTabSession(final Session session, Context context) {
+        if (currentSessionUUID == null || Settings.getInstance(context).shouldOpenNewTabs()) {
+            currentSessionUUID = session.getUUID();
+        } else {
+            // Show Snackbar that allow users to switch to tab they just opened
+            final Snackbar snackbar = ViewUtils.getBrandedSnackbar(
```
EmmanuelMess: Shouldn't this UI logic be in some UI logic related class, `SessionManager` simply deals with sessions and the currently opened sessions. Maybe in a fragment or `MainActivity`?

app/src/main/res/values/strings.xml
ekager: Oops thanks for the catch!

ekager:
New General category with tabs setting
screenshot_20180910-101127
screenshot_20180910-101131

colintheshots:
Looks like TaskCluster is upset about unused imports.

ekager:
Fixed and squashed!

PR #3339, opened 2018-09-06
Fixes #3155: Update UI tests to reflect search suggestion changes.

rpappalax:
Fixes #3155

PR #3342, opened 2018-09-06
Update the test to use the new ID

npark-mozilla:


npark-mozilla:
it's covered by #3339

PR #3352, opened 2018-09-07
ScreenshotTest is no longer dependent on search suggestion dialog acceptance

npark-mozilla:


ekager:
checkStyle is failing on TC

Issue #2940, opened 2018-07-19
WebViewDataTest.java is failing, and it may be that we're not deleting cache properly.


nojunpark:
## Steps to reproduce
- Run WebViewDataTest.java (https://github.com/mozilla-mobile/focus-android/blob/master/app/src/androidTestWebview/java/org/mozilla/focus/privacy/WebViewDataTest.java) on a real device 
 
### Expected behavior
- test should pass
### Actual behavior
- test fails, with following error on both Pixel 2 and Moto G5:
`
java.lang.AssertionError: File '000003.log' should not contain any traces of browser session (localhost, path=/data/user/0/org.mozilla.focus.debug/app_webview/Local Storage/leveldb/000003.log)
at org.junit.Assert.fail(Assert.java:88)
at org.junit.Assert.assertTrue(Assert.java:41)
at org.junit.Assert.assertFalse(Assert.java:64)
at org.mozilla.focus.privacy.WebViewDataTest.assertNoTraces(WebViewDataTest.java:263)
at org.mozilla.focus.privacy.WebViewDataTest.assertNoTraces(WebViewDataTest.java:252)
at org.mozilla.focus.privacy.WebViewDataTest.assertNoTraces(WebViewDataTest.java:252)
at org.mozilla.focus.privacy.WebViewDataTest.assertNoTraces(WebViewDataTest.java:252)
at org.mozilla.focus.privacy.WebViewDataTest.DeleteWebViewDataTest(WebViewDataTest.java:227)
at java.lang.reflect.Method.invoke(Native Method)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
at android.support.test.internal.runner.junit4.statement.RunAfters.evaluate(RunAfters.java:61)
at android.support.test.rule.ActivityTestRule$ActivityStatement.evaluate(ActivityTestRule.java:433)
`
### Device information
Pixel 2, Moto G5
* Android device: API 27, API 24
* Focus version: master branch

npark-mozilla:
@pocmo could you take a look at this please? if this is really the case, it's something that we probably need to fix soon.

PR #3354, opened 2018-09-07
Disable webview test that permanently fails

npark-mozilla:
see https://github.com/mozilla-mobile/focus-android/issues/2940


npark-mozilla:
@boek hmm, the codecov here is again frozen - if you can force merge this as well it will be appreciated.

PR #3358, opened 2018-09-07
No issue: Fix displayed URL for internal pages in GV

ekager:


app/src/main/java/org/mozilla/focus/web/GeckoWebViewProvider.kt
```diff
@@ -589,10 +589,10 @@ class GeckoWebViewProvider : IWebViewProvider {
             encoding: String,
             historyURL: String
         ) {
-            geckoSession.loadData(data.toByteArray(Charsets.UTF_8), mimeType, baseURL)
-            currentUrl = baseURL
-            isLoadingInternalUrl = currentUrl == LocalizedContent.URL_RIGHTS || currentUrl ==
+            isLoadingInternalUrl = historyURL == LocalizedContent.URL_RIGHTS || historyURL ==
```
colintheshots: This could get tedious if we add to the list of internal URLs. It would be better to have a constant list somewhere. Personally, I'd use a Kotlin list contains check like:
`val isLoadingInternalUrl = historyURL in setOf(LocalizedContent.URL_RIGHTS, LocalizedContent.URL_ABOUT)`
...where the setOf() could refer to a constant set.

PR #3361, opened 2018-09-09
Fixes #2989: Remove show badge notification channel

### LINKED ISSUE #2989: App Shortcut and Notification contains the same action
Currently, Firefox Focus shows a very useful app shortcut on API level >25: Erase. This shortcut allows users to erase their browsing history in the background, without needing to open Firefox Focus.

Additionally, Focus also shows a permanent notification. When the notification itself is tapped, it performs an erase. This notification also contains two additional actions: Open, and Erase and Open.

But some versions of Android (I’m running 8.1) shows notifications _on the bottom_ of app shortcuts. As a result, we get an awkward doubling of functionality: two shortcuts that do the same thing, but are located in two different places and shown differently:

I think we can do a few things to solve this doubling problem:

1. On API level >25, don’t show permanent notification at all. If the user wants to erase, they can use the app shortcut
2. On API level >25, when the app shortcut menu is invoked, our notification should be prevented from being displayed

I favour solution 2, as this gives users multiple ways to Erase from outside the app.

EmmanuelMess:
This is a shot in the dark, I cannot test if this works, because [it seems to be broken on emulators](https://stackoverflow.com/a/44528600/3124150) and I don't have an Android O phone to test on.
I am posting this PR because, even if this patch doesn't work, it might help lead to the correct solution.

Fixes #2989

### DIFF
--- a/app/src/main/java/org/mozilla/focus/session/SessionNotificationService.kt
+++ b/app/src/main/java/org/mozilla/focus/session/SessionNotificationService.kt
@@ -129,7 +129,7 @@ class SessionNotificationService : Service() {
         channel.description = notificationChannelDescription
         channel.enableLights(false)
         channel.enableVibration(false)
-        channel.setShowBadge(true)
+        channel.setShowBadge(false)
 
         notificationManager.createNotificationChannel(channel)
     }


colintheshots:
This change only happens at notification channel creation time, which means this fix won't affect existing users on upgrade. If we think this is a good fix, we need some way to conditionally re-create the notification channel with the new setting.

EmmanuelMess:
> See whether there's a way to re-create the notification channel only if a fix hasn't yet been applied.

It doesn't seem like Android will let us modify preferences the user might have set for the channel. Even if the user hasn't set them.

What I might be able to do is recreate the channel with another id and copy the preferences, changing "showBadge" to false (no matter what). But I think that is overkill.

Issue #3179, opened 2018-08-21
Use browser-session component for browser state


pocmo:
We recently replaced all the session related code in Firefox for Fire TV with the `browser-session` component ([Issue](https://github.com/mozilla-mobile/firefox-tv/issues/929)). Now we are close to having all the features that are needed to do the same in Focus/Klar.

pocmo:
Depends on #3190.

PR #3367, opened 2018-09-10
Issue #3179: Use browser-session component for browser state.

pocmo:
This brings in our first **core** component (`browser-session`). The AC team will work on getting `browser-engine` feature complete to replace the WebView/GeckoView abstraction layer with that. With both components landed we have the foundation to land future feature components.

This migration introduces some risk. I tested this locally and couldn't find any major issues. However please decide whether it's safe to land with regards to your A/B test rollout. If I understood correctly you are release from a release branch and new things could potentially land on *master*?


pocmo:
This is almost a 1:1 replacement for now. We should be able to make things a bit nicer in future iterations and when landing `browser-engine`.

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -65,10 +66,10 @@ open class MainActivity : LocaleAwareAppCompatActivity() {
             TelemetryWrapper.openFromIconEvent()
         }
 
-        sessionManager.handleIntent(this, intent, savedInstanceState)
-
         registerSessionObserver()
 
+        //intentProcessor.handleIntent(this, intent, savedInstanceState)
```
colintheshots: Remove comment?

app/src/main/java/org/mozilla/focus/session/ui/SessionsAdapter.kt
```diff
@@ -65,8 +64,24 @@ class SessionsAdapter/* package */ internal constructor(private val fragment: Se
...
+    override fun onSessionSelected(session: Session) {
+        onUpdate(fragment.requireComponents.sessionManager.sessions)
+    }
+
+    override fun onAllSessionsRemoved() {
+        onUpdate(fragment.requireComponents.sessionManager.sessions)
+    }
+
+    private fun onUpdate(sessions: List<Session>) {
+        this.sessions = sessions
```
colintheshots: In general, it would be better to have DiffUtil calculate the diff and update the adapter when possible. There are non-trivial performance and animation differences with notifyDataSetChanged().

Sdaswani:
@pocmo what's the regression testing plan for integrating this code? It seems to have a lot of surface area so want to make sure we are thinking about regressions. Also do we have documentation on how we add new hooks into engines (e.g., GV) going forward?

pocmo:
> @pocmo what's the regression testing plan for integrating this code? It seems to have a lot of surface area so want to make sure we are thinking about regressions.

That's a good question. It makes sense to have a thorough regression test after this lands (cc @npark-mozilla). I did some smoke testing and have this build as my default browser to avoid/find major issues.

> Also do we have documentation on how we add new hooks into engines (e.g., GV) going forward?

Not yet. This PR does not replace the engine yet (= adding `browser-engine`). This would be the next step after this PR lands. On our side we need to make sure we have all the features for that. Ideally we start the integration once we are confident in GV rolling out and the WebView builds fading out.

npark-mozilla:
@colintheshots  so this won't affect #3431 implementation?

Sdaswani:
Note to @vesta0 this is landing in 8.0.

pocmo:
> Please do not merge this until we have sign off from @npark-mozilla @pocmo @st3fan that this has been QA-verified. It’s early enough in the 8.0 release cycle but it seems to be lacking QA.

If we don't merge this then those patches will rot again because of a bunch of new merge conflicts. Don't we do QA in master?

npark-mozilla:
@sv-ohorvath Hi Oana, could you try out this PR on your phone soon and run a quick sanity check to see whether there are any issues?  Thanks!

Sdaswani:
@sv-ohorvath @npark-mozilla i would suggest a full regression test give this is a large refactor.

npark-mozilla:
@Sdaswani if we're doing full regression check, wouldn't it make sense to do it once it's in master as @pocmo suggests?

sv-ohorvath:
I'll start working on this today, but it will only be done by tomorrow EOD. Would it be necessary to test both gecko & webview builds? Please let me know, thanks!

Sdaswani:
Why would that make more sense @npark-mozilla ? If we merge to master, then do a regression test, wouldn't those regressions have made it onto master? Suppose there are many? Also since it involves a component, it's better to suss out those issues with the component prior to merge. Doing the regression test now will make it certain that it's the refactor that has caused the regression (since that is what this PR is about).

npark-mozilla:
If we find issues before merging the PR, what I've usually seen is that PR of this size will take a while before ironing out regressions AND properly rebasing to master.  It is a big change, so if we do see regression, it makes it easier to find them in master and submit fixes to master asap. I'm okay with having regressions in master - we always had/have one, and we have dealt with them. 

I know we cut nightly too from master, but that also works out, since nightly users would find bugs that we might have missed.  If we go the route of testing large PRs before merging to master, then we'd have to apply that to every PR from now on- and on long run that'll slow down the dev process, since devs won't be able to merge their PR before we check them.  @pocmo thoughts?

npark-mozilla:
I forgot to say this: but for PR of this size, we'll do additional full regression check as soon as it's merged to master (upon request that is), then after code freeze, we do a final one, so there's barely any difference to QA work.

pocmo:
My reasoning for getting it in now/early:

* I would like to get it into the hands of Nightly users as soon as possible.
* A bunch of other things have landed in master and may have caused regressions. I can't be responsible for all regressions in this PR. We can fix all regressions in master but it's going to be complicated to separate regressions between master regressions and PR regressions.
* Master is a moving target (especially if we fix regressions in two places). If we do the regression test first then by the time it is done master may have changed: We will have to update the PR again, requiring yet another regression test? Updating the PR may not be trivial if we wait too long.

st3fan:
@Sdaswani / @vesta0 We are committed to deal with regressions quickly by issuing followup patches before the end of this sprint (October 5th). We will work closely together with the team and QA to address and re-verify found issues.

Sdaswani:
Thanks @st3fan for this commitment: https://github.com/mozilla-mobile/focus-android/pull/3367#issuecomment-424742902

I'm OK merging it but I think @vesta0 needs to understand product value better. Is there a need to do this work now? [It doesn't sound like there are features](https://github.com/mozilla-mobile/focus-android/pull/3367#issue-214271730) that Focus can take advantage of yet?

sv-ohorvath:
Found 4 new regressions on the PR, but we'll resume testing once these are fixed and once a decision is made if this will be merged on Master.
https://github.com/mozilla-mobile/focus-android/issues/3485
https://github.com/mozilla-mobile/focus-android/issues/3484
https://github.com/mozilla-mobile/focus-android/issues/3486
https://github.com/mozilla-mobile/focus-android/issues/3487

pocmo:
Thank you @sv-ohorvath and @bsurd!

I added patches fixing #3485, #3484 and #3486 to the PR. #3487 seems to be unrelated and I commented on the issue about what's causing this (for the APT team to pick up).

I'll rebase the PR now - looks like there's a new conflict.

pocmo:
> I'm OK merging it but I think @vesta0 needs to understand product value better. Is there a need to do this work now? It doesn't sound like there are features that Focus can take advantage of yet?

Why now: We have been working on getting our core components (`browser-session` & `browser-engine`) into a state where we can start integrating them into our existing apps. This will get the browser core of all our apps on a shared implementation. We reached feature parity with Focus for the `browser-session` component and are ready to integrate it now. Waiting longer will mean that both diverge again and we will have to catch up again.

No features: Yet! This is the foundation for shared feature components. If we want to build feature components then this is the first step to be able to build features that work across our apps. So features are coming too - but not with this initial PR. :)

Sdaswani:
Yes, thank you so much @sv-ohorvath and @bsurd! I'm ok with this merging now, and I recommend to @vesta0 that we go ahead and merge it.

npark-mozilla:
Let's merge this PR in master, and we'll continue testing it.

sv-ohorvath:
Resumed testing on the latest Nightly build. But due to the size of the test suite, I'll be able to finish testing only by Monday EOD. I'll keep everyone updated with the results.

sv-ohorvath:
Got a bit delayed, but here are the final results:
Ran the full regression tests on builds running both WebView and GeckoView .

Issues found on Nightly: 
https://github.com/mozilla-mobile/focus-android/issues/3485 - fixed
https://github.com/mozilla-mobile/focus-android/issues/3484 - fixed
https://github.com/mozilla-mobile/focus-android/issues/3486 - fixed
https://github.com/mozilla-mobile/focus-android/issues/3487 - fixed 
https://github.com/mozilla-mobile/focus-android/issues/3507 - fixed
https://github.com/mozilla-mobile/focus-android/issues/3505 - open

PR #3388, opened 2018-09-12
Taskcluster: Use json-e to generate tasks

JohanLorenzo:
json-e is a way to templatize JSON (therefore YAML) data. taskcluster-github now exposes it: https://docs.taskcluster.net/docs/reference/integrations/taskcluster-github/docs/taskcluster-yml-v1. Migrating to this format enables extra-security checks when signing APKs. In summary, signing tasks are run on high-security workers called scriptworker https://github.com/mozilla-releng/scriptworker. These workers trace the task definition back to `.taskcluster.yml` in the repo. This way, it knows if someone try to spin a release that doesn't come from the repo.


JohanLorenzo:
We should also test what happens when a "push to master" occur **and** when a new release is shipped.

I should also look into the nightly hook

PR #3390, opened 2018-09-12
Closes #3370 Do not Init Sentry for Unsupported Versions

### LINKED ISSUE #3370: Consider Filtering Out Sentry Pings From Unsupported Devices
When investigating #2903, it looks like this was a crash exclusive to versions lower than what Focus supports. It's possible to download our APKs on unsupported versions not through GP, and now that we're not doing Crash Reporting through GP, we're seeing these crashes that aren't prudent for us to fix. 

Maybe we should not send crash data for unsupported versions of Focus.

ekager:
This could help a bit at least?

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
@@ -8,6 +8,7 @@ import android.arch.lifecycle.Observer
 import android.arch.lifecycle.ViewModelProviders
 import android.content.Context
 import android.content.Intent
+import android.os.Build
 import android.os.Bundle
 import android.preference.PreferenceManager
 import android.util.AttributeSet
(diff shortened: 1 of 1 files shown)


colintheshots:
I had to dig a few levels deep to ensure this will work. It seems Sentry is designed to auto-init if you don't call init(). However, it will init with a null Dsn, which will then attempt to lookup a Dsn. This checks JNDI, System environment variables, and Java properties before giving up and using "noop://localhost?async=false". Documentation suggests that if all lookups fail, Sentry operations will do nothing. So I think this works.

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -44,10 +46,13 @@ open class MainActivity : LocaleAwareAppCompatActivity() {
     protected open val currentSessionForActivity: Session
         get() = sessionManager.currentSession
 
+    @SuppressLint("ObsoleteSdkInt")
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        SentryWrapper.init(this)
+        if (Build.VERSION.SDK_INT > Build.VERSION_CODES.KITKAT) {
```
colintheshots: >= LOLLIPOP? Build.VERSION_CODES.KITKAT is 19. This is a bit of a hack as it won't likely stay current with our latest, but it's better than nothing.

Issue #3372, opened 2018-09-11
[l10n] Sign-off on locales for v7


flodolo:
* Total number of locales that can ship with Focus Android v7: 91 locales
* 7 new locales can ship: Aymara (ay), Galician (gl), Marathi (mr), Punjabi (pa-IN), Náhuat Pipil (ppl), K'iche' (quc), Sundanese (su).
* Incomplete locales needing to be only in screenshots section (ADDITIONAL_SCREENSHOT_LOCALES in locales.py): - (see notes)

Notes: 
* zh-HK was added in Pontoon, but hasn't added any translation, so it's not in the repo yet. Leaving out of the screenshots list of locales.
* @boek There are two locales ready to ship but that don't have screenshots: Huastec (hus), Yucatec Maya (yua). Any idea, from a build perspective, how safe it would be to add them? I've double checked screenshots for all new locales, and couldn't find any issue besides some truncation but I'm not sure if these 2 languages could contain syntax errors at this point. If it's not safe, they need to be added to the screenshots locale.

List of locales that can ship (in alphabetical order):
af
am
an
anp
ar
ast
ay
az
bg
bn-BD
bn-IN
bs
ca
cak
cs
cy
da
de
dsb
el
en-CA
eo
es-AR
es-CL
es-ES
es-MX
eu
fa
fi
fr
fy-NL
ga-IE
gl
gu-IN
he
hi-IN
hr
hsb
hu
hy-AM
ia
id
it
ixl
ja
ka
kab
kk
ko
kw
lo
lt
meh
mix
mr
ms
my
nb-NO
ne-NP
nl
nn-NO
oc
pa-IN
pai
pl
ppl
pt-BR
quc
quy
ro
ru
sk
sl
sq
sr
su
sv-SE
ta
te
th
tr
trs
tsz
tt
uk
ur
vi
wo
zam
zh-CN
zh-TW

cc @Delphine

boek:
Hey @flodolo,

I added the 7 new locales and we now have screenshots for Huastec (hus), Yucatec Maya (yua).
https://npark-mozilla.github.io/index.html#yua-phoneScreenshots
https://npark-mozilla.github.io/index.html#hus-phoneScreenshots

If these look good to you I can move those to shipping locales as well.

flodolo:
> If these look good to you I can move those to shipping locales as well.

I think we can do that. Again, some truncations, but nothing major.

boek:
[...]

PR #3399, opened 2018-09-13
For #3372 - Adds two new release locales

boek:


tools/l10n/locales.py
```diff
@@ -108,6 +109,7 @@
 	"ur",
 	"vi",
 	"wo",
+	"yua"
```
ekager: Missing a comma here

PR #3400, opened 2018-09-13
For #3372 - Import strings

boek:


boek:
Closing because we now have imports!

PR #3409, opened 2018-09-13
Fixes #815: Add open in app in WebContextMenu

### LINKED ISSUE #815: Longpress on link menu: Open link with app
An "open with ... app" longpress menu item for some links like for example YouTube, Playstore, or Wikipedia links would be great. Similar to the one in the Firefox for Android app.
That way one wouldn't have to open the initial website in another browser to then be able to open a specific link in a specific app.

EmmanuelMess:
Fixes #815 

@colintheshots
I'm having problems with the argument count of `WebContextMenu.setupMenuForHitTarget()` 7 out a maximum 5. I feel the correct solution is to make `WebContextMenu` not an object, because the other two solutions are:
* to convert arguments to fields (possibly ending up with a memory leak) or
* passing objects with more than one parameter and separating them inside the function (adding unnecessary clutter).

But I feel that converting `WebContextMenu` to a class overextends the scope of this PR. What should I do?

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/activity/MultitaskingTest.java
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/MultitaskingTest.java
@@ -5,12 +5,14 @@
 package org.mozilla.focus.activity;
 
 import android.content.Context;
-import android.os.SystemClock;
 import android.preference.PreferenceManager;
 import android.support.test.InstrumentationRegistry;
+import android.support.test.espresso.IdlingRegistry;
 import android.support.test.espresso.web.webdriver.Locator;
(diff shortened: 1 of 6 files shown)


colintheshots:
Getting rid of one parameter is easy. Add this to ContextUtils:
```fun Context.asFragmentActivity() = (this as? ContextThemeWrapper)?.baseContext as? FragmentActivity
        ?: this as? FragmentActivity```

The second parameter would require calling getAppDataForLink() inside setupMenuForHitTarget() as you already have all the values you need internally. I tested it and this works.

EmmanuelMess:
Thanks!

PR #3431, opened 2018-09-17
Make GV tests use GV for Klar, WV for Focus

### LINKED ISSUE #3335: Need a way to make Geckoview/webview default for UI tests
Right now Klar Debug build variants use webview as default, and we cannot run UI tests on Geckoview, since they will run on debug only.  None of the UI tests are running on Geckoview at this moment I believe.

### What / Requirements
We need a way to make geckoview default, in order to run ui tests on geckoview variant, if we consider running UI tests on Geckoview a must.

@colintheshots thoughts?

colintheshots:
Fixes #3335


colintheshots:
This maybe isn't enough yet. I seems like the web engine is already picked before it runs beforeActivityLaunched().

colintheshots:
@npark-mozilla Please re-review this. The new solution seems to work well.

npark-mozilla:
This fixes #3335 :)

npark-mozilla:
@rpappalax FYI

Issue #2896, opened 2018-07-13
Just-in time information and one-tap action – Default Browser


Sdaswani:
### Why/User Benefit/User Problem
- Allow users to quickly set Focus as their default browser

### What / Requirements
- Provide intent sheet on a specific trigger (screen tip and/or settings page)

### Acceptance Criteria (how do I know when I’m done?)
- Intent sheet showing up and working
- User can set Focus as their default browser without leaving the app

## Success criteria/Telemetry
- Probe to track how many times/people initiated the intent sheet and selected Focus successfully

--------------
Per https://github.com/mozilla-mobile/focus-android/issues/2816#issuecomment-402697895 @brampitoyo should think of a nice contextual hint to ask users to set default browser.

brampitoyo:
![set-default-browser 2x](https://user-images.githubusercontent.com/51007/43119678-79508ea4-8f6b-11e8-810f-126dcf851ff3.png)

Sdaswani:
Cool @brampitoyo - when do we show this toast / tip?

brampitoyo:
@Sdaswani I think that the menu item should show up only when Focus isn’t set a default browser, but the toast and modal dialogue should show up whenever the menu item is tapped.

Sdaswani:
@brampitoyo this toast only shows up in the setting screen right?

brampitoyo:
Yes.

Sdaswani:
Looking at the data, out of 400 million foregrounds, settings has only be accessed some 2 millions times. Not very many people will ever see this tip. https://sql.telemetry.mozilla.org/dashboard/android-focus

kfontecilla:
@brampitoyo Is there an update here as well? See Susheel's last comment. Thx

brampitoyo:
Hi @kfontecilla 

“Set default browser” can be a start screen tip that will lead to a one-tap action:

![set default browser 2x](https://user-images.githubusercontent.com/51007/43692078-5416194e-9978-11e8-811f-82711705bc16.png)

This solves the problem that Susheel outlined.

However, labelling the Settings menu item clearly (the initial idea up above) is still a good idea. It won’t hurt us to do it in parallel to the start screen tip.

brampitoyo:
@Sdaswani Yes. My current plan is to simply underline the deep link, which would apply to both iOS and Android.

(Obviously, if the deep link leads to SUMO, we want to say something along the lines of “Learn More”).

bbinto:
[...]

PR #3437, opened 2018-09-18
Closes 2896: Just in time default browser

sblatz:
New behavior can be found [here](https://github.com/mozilla-mobile/focus-android/issues/2896#issuecomment-422407921).


sblatz:
Good point. I brought this up in the original ticket!

PR #3438, opened 2018-09-18
Fixes #2966 quick add autocomplete

### LINKED ISSUE #2966: One-tap action – Add Custom URL
Per https://github.com/mozilla-mobile/focus-android/issues/2816#issuecomment-407289032

add-custom-url 2x

boek:
image


colintheshots:
Is this using Bram's latest design now?

sblatz:
Another quick question for @boek: should adding a url in the quick toggle enable the "Autocomplete" toggle in settings for Custom URL's? It currently doesn't but this also seems confusing from a user perspective--I just added a url to autocomplete and now it's not autocompleting?

(Again, maybe this should be a follow up ticket or perhaps we could ask both of these in the original)

Issue #3443, opened 2018-09-19
Figure out Why GV A/B Experiment Isn't Rolling Out As Planned


Sdaswani:
@colintheshots is investigating if Fretboard distributes experiments incorrectly (e.g., we ask it to distribute it to 50% of users, and it only distributes it to 20%).

colintheshots:
Based on the results of my Fretboard distribution test, I believe Fretboard is bucketing correctly. I think the issue is either in how the core ping is getting sent or in the data query somehow. https://github.com/mozilla-mobile/android-components/blob/8ba3e0addaf38e2dea1377232eecd18d16e80171/components/service/fretboard/src/test/java/mozilla/components/service/fretboard/FretboardTest.kt

boek:
Looks like it's working now!

PR #3450, opened 2018-09-19
Use best effort to load experiments on initial startup

colintheshots:
Related to #3443.


boek:
@colintheshots Does this need to go into 7.0?

colintheshots:
I think it should. My only concern about this code is what happens if the network is being slow and the user is fast. Will we have users attempting to browse on bad internet and crashing the browser because the engine isn't loaded. We use engine!!. unsafe calls to crash if the engine isn't set yet in the current code. Maybe I should come up with a second PR to handle slow engine startup?

boek:
That is a good point, I don't think we would want to ship something where the app would crash in that case.

PR #3452, opened 2018-09-20
No issue: disable codecov patch checks.

mcomella:
Status checks always fail because of this:

image

So I think it'd be a good thing to do. We're doing this on TV too.

I want to have a separate discussion about the best way to improve our test culture (assuming there are problems with it) so I don't think this will have an impact.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -13,7 +13,10 @@ coverage:
        if_not_found: success
     patch:
       default:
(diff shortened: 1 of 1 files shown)


mcomella:
After the Android meeting, we should probably set this to some minimal threshold so it creates discussions.

PR #3462, opened 2018-09-21
Use media type if downloaded file lacks extension

### LINKED ISSUE #2743: Files are saved in .bin format
When i try to download files (especially images) the files are saved in .bin format.
For example,
When i try to download image from Twitter 
the download link refer the file format as .jpg but it will be saved as .bin format.
img-20180615-wa0023
img-20180615-wa0024

gaul:
Previously Focus did not consider the Content-Type header when
downloading URLs without extensions, yielding a .bin suffix and
preventing opening the file via the Snackbar.  Fixes #2743.


gaul:
I am not sure how I could add unit tests since this relies on `DownloadManager` interactions.  Any suggestions?

ekager:
Hi @gaul could you move any of this code into a utility method in an extension of DownloadUtils (or use the methods already there, it looks like there's a new content disposition method we don't use)?

gaul:
@ekager Thanks for the suggestion!  `guessFileName` has some weird behavior but if mozilla-mobile/android-components#839 merges I can simplify this pull request.

ekager:
Great, let's see what they think next week! Thanks for looking into this! :)

boek:
Hey @gaul, the latest android-components has landed in master 👍

gaul:
@boek Thanks for the update!  I will respin this pull request this weekend.

gaul:
@ekager I addressed your `guessFileName` suggestion.

ekager:
Hi @gaul ,thanks for this! The File and File2 bit is a bit confusing, does the changed DownloadUtils.guessFileName method still require creating two files and comparing them even if you pass in the mime type when creating the file? Could you just help me understand why we need both? :)

gaul:
@ekager Unfortunately `DownloadManager` does not give us the control we want: selecting the file name during the response.  `BrowserFragment.queueDownload` gives the original name "foo.bin" based on the URL then equeues the request.  When receiving the response containing the Content-Type, `DownloadManager` notes this header but continues to use the original name.  `DownloadBroadcastReceiver.onReceive` receives the completion event then Focus has the opportunity to rename the file name with the Content-Type to "foo.jpg".

ekager:
Thanks! So, in that case, should we move this code to send the correct file name in `queueDownload`? Would that save the file correctly and give us the correct name also in `DownloadBroadcastReceiver.onReceive`? You're saying that the system `DownloadManager` can read the Content-Type correctly even without the changing of the file name, but is it being saved with an incorrect name as well?

ekager:
Hey @colintheshots , we're still getting reviews saying files save as .bin when we don't know the extension, and we've been working on a fix here. Do you have any feedback on the solution?

PR #3469, opened 2018-09-24
Issue #3443: Figure out Why GV A/B Experiment Isn't Rolling Out

colintheshots:


colintheshots:
It looks like the timeout on the disk read wasn't blocking properly. Removing it means the disk read always happens properly. The network request could happen async on a background thread safely as long as the disk read is 100% reliable.

PR #3471, opened 2018-09-24
Fix type error in locale lookup

gaul:
Found via error-prone.  Follows on to
mozilla-mobile/focus-android#3230.


gaul:
@yfdyh000 Please review.

yfdyh000:
This fix looks good, I have been looking for someone to fix it.
I can't verify it due to I can't get a build environment.

colintheshots:
@gaul @yfdyh000 Is this working correctly in production? It seems like the Map contains "zh-CN". "CN" is the country portion though, right? I've got a Sentry crash issue that appears related to the LocaleListPreference that I cannot reproduce on a western Pixel 2XL.

yfdyh000:
@colintheshots I haven't seen a working build yet, so I doubt it about its works.

PR #3490, opened 2018-09-26
Fixes typo and Fixes #3479: Quick add autocomplete URL does not enable autocomplete.

### LINKED ISSUE #3479: Quick add autocomplete URL does not enable autocomplete
## Steps to reproduce
1. Navigate to a website
2. Click URL bar
3. "Add link to autocomplete"
4. Type in part of the website name

### Expected behavior
The URL autocompletes (i.e. the setting is enabled when I choose to add a URL to autocomplete)

### Actual behavior
The URL does not autocomplete (unless you go into the settings to manually enable it)

### Device information

* Android device: Moto G4
* Focus version: 8.0 build 11

sblatz:


boek:
Ahah good catch!! 

Instead of enabling it every time we add a URL we should just enable it by default. My reasoning for this is that if I disable my URL list it should stay disabled until I re-enable it. It's a weird case but I think it makes more sense.

boek:
@sblatz we might need a defaultValue here as well: https://github.com/mozilla-mobile/focus-android/blob/36b900bdbc994290def3f3ac28f2c5bcafecf991/app/src/main/res/xml/autocomplete.xml#L20

sblatz:
Yep, just added that!

Issue #3487, opened 2018-09-26
"Show homescreen tips" doesn't work - Issue #3179 regression


sv-ohorvath:
## Steps to reproduce
1. Open the Settings menu and go to the Mozilla about menu.
2. Check if the "Show homescreen tips" toggle is on by default.
3. Turn the toggle on.
4. Return to the homescreen, load a page and then press the back button.

### Expected behavior
 "Show homescreen tips" toggle is on by default. Homescreen tips should be displayed.

### Actual behavior
 "Show homescreen tips" toggle is off. No tips displayed on the homescreen.

### Device information
* Android device: Nokia 6 (Android 7.1.1), Huawei P9 Lite (Android 6);
* Focus version: Custom build with changes from #3367 applied. Not reproducing on the Nightly build.
* Issue reproducible for both GeckoView and WebView.

pocmo:
This seems to be unrelated to the PR.

Homescreen tips are hidden behind an A/B experiment. Local builds seem to use the "release experiment" which is disabled. Nightly uses a separate A/B experiment and there it is enabled.

We may want to hide the tips settings toggle if the home screen tips experiment is disabled for the installation? But that's unrelated to the PR and not my decision (@sblatz @vesta0 @Sdaswani).

sv-ohorvath:
The toggle isn't displayed anymore on the dev build. Works ok on Nightly. Thanks!

PR #3516, opened 2018-09-28
For #3487 - Fixes the A/B test logic for homescreen tips

boek:


Sdaswani:
Deferring to @sblatz for a review.

Issue #3566, opened 2018-10-04
[Meta] Biometrics hotfixes


boek:
- [ ] RuntimeException on createKey when user doesn't have lockscreen enabled

boek:
https://sentry.prod.mozaws.net/operations/focus-android/issues/4781874/
RuntimeException: Unable to resume activity {org.mozilla.focus/org.mozilla.focus.activity.CustomTabActivity}: java.security.InvalidAlgorithmParameterException: java.lang.IllegalStateException: Secure lock screen must be enabled to create keys requiring user authentication

sblatz:
Likely involves #3542.

PR #3567, opened 2018-10-04
for #3566 - Fixes Biometric crashing on some devices

boek:


app/src/main/java/org/mozilla/focus/biometrics/Biometrics.kt
```diff
@@ -6,10 +6,15 @@ package org.mozilla.focus.biometrics
 
 import android.content.Context
 import android.support.v4.hardware.fingerprint.FingerprintManagerCompat
+import android.app.KeyguardManager
 
 object Biometrics {
     fun hasFingerprintHardware(context: Context): Boolean {
         val fingerprintManager = FingerprintManagerCompat.from(context)
-        return fingerprintManager.isHardwareDetected && fingerprintManager.hasEnrolledFingerprints()
+        val keyguardManager = context.getSystemService(Context.KEYGUARD_SERVICE) as KeyguardManager
```
Baron-Severin: This was a pretty weird edge case, it's probably worth adding a comment explaining why we need this check.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -755,8 +755,13 @@ class BrowserFragment : WebFragment(), LifecycleObserver, View.OnClickListener,
             Settings.getInstance(requireContext()).shouldUseBiometrics() &&
             Biometrics.hasFingerprintHardware(requireContext())
         ) {
+            if (biometricController == null) {
+                biometricController = BiometricAuthenticationHandler(context!!)
```
Baron-Severin: The code that instantiates this object in `onCreate` does a few additional checks that look important.  Maybe move that code here (so it gets run when foregrounding the app) instead of duplicating it.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -755,8 +755,13 @@ class BrowserFragment : WebFragment(), LifecycleObserver, View.OnClickListener,
             Settings.getInstance(requireContext()).shouldUseBiometrics() &&
             Biometrics.hasFingerprintHardware(requireContext())
```
Baron-Severin: This looks like it's supposed to exactly match another conditional statement found in `onPause`, but there's no clear connection between them.  Suggest breaking them out into a shared function, so we never change one but forget the other.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -755,8 +755,13 @@ class BrowserFragment : WebFragment(), LifecycleObserver, View.OnClickListener,
             Settings.getInstance(requireContext()).shouldUseBiometrics() &&
```
Baron-Severin: (I can't drop a comment so far away from changed code.  Pretend this is on line 190)

If we split-screened between our app and settings, would `onResume` be called when moving focus between the two?  If not, since we've seen that `shouldUseBiometrics` can change during a session, we're opening ourselves up to a potential NPE on 190.  Suggest a null check there.   We're just turning off a listener, so we shouldn't have to do anything else to handle the case where the controller is null.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
Baron-Severin: My mistake, the code in `onCreate` actually has fewer checks.  Do you know any reason why this class would need to be initialized early?  If not, probably worth pulling that initialization out of `onCreate`.

boek:
Thanks for the review @Baron-Severin! Made some changes. Ready for another look 👍

Issue #2901, opened 2018-07-13
Sentry: RemoteServiceException: Context.startForegroundService() did not then call Service.startForeground()


cpeterso:
https://sentry.prod.mozaws.net/operations/focus-android/issues/4326368/

```
RemoteServiceException: Context.startForegroundService() did not then call Service.startForeground()
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1870)
    at android.os.Handler.dispatchMessage(Handler.java:105)
    at android.os.Looper.loop(Looper.java:164)
    at android.app.ActivityThread.main(ActivityThread.java:6798)
    at java.lang.reflect.Method.invoke(Method.java)
...
(2 additional frame(s) were not displayed)

Context.startForegroundService() did not then call Service.startForeground()
```

This is one of the Focus top crshes, with 349 crashes from 86 users.

colintheshots:
In general, this common exception occurs when a phone is being sluggish. Android gives the app five seconds to start the foreground service and it didn't happen in time. This could be strictly due to a slow phone with an old processor and less memory requiring frequent garbage collections. More likely, the issue is caused by ANRs, which can simply be more significant on some phones than others. This is likely related to other ANR issues already reported.

pocmo:
This looks interesting and could be causing this:
https://issuetracker.google.com/issues/76112072

Not sure how to work around that though..

pocmo:
With this being our top crash I assume this is not a P5 anymore (CC @Sdaswani @vesta0 @boek).

pocmo:
![screenshot_2018-10-05 anrs crashes - firefox focus the privacy browser - google play console](https://user-images.githubusercontent.com/89638/46526163-c1abb600-c88d-11e8-935c-c5e2b5aee737.png)

Something in 7.0.x has definitely made this more likely to occur.

One guess is the "wait a bit for fretboard to update" workaround that blocks the UI thread and may lead to `onStartCommand` being called too late (CC @colintheshots)?

pocmo:
> https://issuetracker.google.com/issues/76112072

Calling `stopService()` before the Service had time to get to `startForeground()` is definitely triggering this exception 🤔:

```
E  android.app.RemoteServiceException: Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{e8f676b u0 org.mozilla.focus.debug/org.mozilla.focus.session.SessionNotificationService}
[...]

PR #3576, opened 2018-10-05
For #2901 - Posts starting the foreground service

boek:


app/src/main/java/org/mozilla/focus/session/SessionNotificationService.kt
```diff
@@ -150,17 +151,24 @@ class SessionNotificationService : Service() {
             val intent = Intent(context, SessionNotificationService::class.java)
             intent.action = ACTION_START
 
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                context.startForegroundService(intent)
-            } else {
-                context.startService(intent)
-            }
+            // Post starting the service to the end of the main thread to give it the best
```
mcomella: nit: "to give it the best chance to complete in time."

In time? What issues do we see if we don't let it complete? Elaborate! This issue is serious/complex enough that'd I'd include an issue number.

You may also want to mention this is a speculative fix and that some processing still occurs afterwards.

app/src/main/java/org/mozilla/focus/session/SessionNotificationService.kt
```diff
@@ -150,17 +151,24 @@ class SessionNotificationService : Service() {
...
+                    context.startService(intent)
+                }
+            })
         }
 
         internal fun stop(context: Context) {
             val intent = Intent(context, SessionNotificationService::class.java)
 
-            context.stopService(intent)
+            // Since we're posting start, we should also post stop
```
mcomella: nit: You should explain why that is: to ensure they're run in the order they were called in, you have to put them in the same sequential queue.

app/src/main/java/org/mozilla/focus/session/SessionNotificationService.kt
```diff
@@ -150,17 +151,24 @@ class SessionNotificationService : Service() {
...
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                context.startForegroundService(intent)
-            } else {
-                context.startService(intent)
-            }
+            // Post starting the service to the end of the main thread to give it the best
+            // chance to complete in time.
+            ThreadUtils.postToMainThread(Runnable {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                    context.startForegroundService(intent)
```
mcomella: Can you check what happens when this is called when the app is in the background? e.g. post to main thread with a delay of ten seconds, open app, press home button, and wait for foreground service notification to start and if it doesn't show up, we have a problem.

I'm concerned this may crash if called when the app is stopped. Possible solution: use LiveData/ViewModel because LiveData doesn't call its observers unless the app is started.

boek: @mcomella 

Good idea. After checking the notification service shows up after 10 seconds and it works as intended.

app/src/main/java/org/mozilla/focus/session/SessionNotificationService.kt
```diff
@@ -150,17 +151,26 @@ class SessionNotificationService : Service() {
...
+                    context.startService(intent)
+                }
+            })
         }
 
         internal fun stop(context: Context) {
             val intent = Intent(context, SessionNotificationService::class.java)
 
-            context.stopService(intent)
+            // We want to make sure we always call `stopService` after `stopService. So we're giving
```
mcomella: nit: second -> `startService`; don't forget the second tick.

mcomella: super nit: this is acceptable as is but I'd be even more explicit, "So we're putting these actions on the same sequential run queue."

Without being explicit (to be fair, there is some redundancy in doing so), it's hard to know what the "same treatment" was intended to be without looking at both code paths, especially if this code changes much.

app/src/main/java/org/mozilla/focus/session/SessionNotificationService.kt
```diff
@@ -150,17 +151,26 @@ class SessionNotificationService : Service() {
             val intent = Intent(context, SessionNotificationService::class.java)
             intent.action = ACTION_START
 
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                context.startForegroundService(intent)
-            } else {
-                context.startService(intent)
-            }
+            // For #2901: The application is crashing due to the service not calling `startForeground`
```
mcomella: super-nits:
- "not calling `startForeground` before it times out"
- "speculative fix to decrease the time between starting the service": how? How does it work? I think it's important to say, "We're running after blocking UI thread methods like FocusApplication.onCreate (when clicking links) that could have delayed the `startForeground` invocation.`

Or similar, this wording isn't the best: late on a Friday. :)

PR #3599, opened 2018-10-09
Focus TLD in long URL

ekager:
In reference to https://bugzilla.mozilla.org/show_bug.cgi?id=1445758


app/src/main/java/org/mozilla/focus/FocusApplication.kt
```diff
@@ -72,6 +73,9 @@ class FocusApplication : LocaleAwareApplication() {
 
             TelemetryWrapper.init(this@FocusApplication)
             AdjustHelper.setupAdjustIfNeeded(this@FocusApplication)
+            launch {
+                TLDExtractor.init(this@FocusApplication)
+            }
```
pocmo: Afaik `launch()` without a coroutine context will run on the current context - which would be the main thread here -> Blocking main thread for reading files -> ANR.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -1336,6 +1337,19 @@ class BrowserFragment : WebFragment(), LifecycleObserver, View.OnClickListener,
...
+            if (urlView != null) {
+                val paint = urlView!!.paint
+                val charWidth = paint.measureText("a", 0, 1)
+                val urlViewWidth = urlView!!.measuredWidth
+                val numChars = urlViewWidth / charWidth
+                if (url.length > numChars) {
+                    val index = UrlUtils.getIndexOfTLD(url)
+                    urlView?.bringPointIntoView(index)
+                }
+            }
```
pocmo: Doing the `null` check and still having to use `!!` is not super nice. One of the extension methods (maybe `let` or `apply`) may make that nicer.

Another alternative may be "assigning and early return":
```Kotlin
val view = urlView ?: return

// And then use `view`
```

app/src/main/java/org/mozilla/focus/session/IntentProcessor.kt
```diff
@@ -98,18 +98,18 @@ class IntentProcessor(
                     return null
                 }
 
-                return if (!UrlUtils.isUrl(dataString)) {
+                return if (!UrlUtils.isUrl(dataString!!)) {
```
pocmo: Note that this now throws/crashes if there's an ACTION_SEND intent without EXTRA_TEXT provided. Not sure if this will actually happen in the real world though.

app/src/main/java/org/mozilla/focus/utils/TLDExtractor.kt
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Comparator
+import java.util.regex.Pattern
+
+/**
+ * Finds the Top Level (or Secondary Level) Domain using a public list of TLD domains: http://publicsuffix.org/list/
+ * Initialize as soon as possible in application.
+ * Based on:
+ *  http://www.supermind.org/blog/1078/extracting-second-level-domains-and-top-level-domains-tld-from-a-url-in-java
+ **/
+object TLDExtractor {
```
pocmo: This is something that would be nice in a component and we haven an issue for that already:
https://github.com/mozilla-mobile/android-components/issues/95

app/src/main/java/org/mozilla/focus/utils/TLDExtractor.kt
```diff
@@ -0,0 +1,102 @@
...
+import java.util.ArrayList
+import java.util.Collections
+import java.util.Comparator
+import java.util.regex.Pattern
+
+/**
+ * Finds the Top Level (or Secondary Level) Domain using a public list of TLD domains: http://publicsuffix.org/list/
+ * Initialize as soon as possible in application.
+ * Based on:
+ *  http://www.supermind.org/blog/1078/extracting-second-level-domains-and-top-level-domains-tld-from-a-url-in-java
```
pocmo: Note that we have such code in Fennec already:
https://searchfox.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/util/publicsuffix/PublicSuffix.java

The code in Fennec was based on Guava and uses a minimal list and can do lookups quite efficiently.

app/src/main/java/org/mozilla/focus/utils/TLDExtractor.kt
```diff
@@ -0,0 +1,102 @@
...
+                        }
+                    }
+                }
+            }
+            Collections.sort(terms, StringLengthComparator)
+            for (t in terms) add(t)
+            compile()
+            bufferedReader.close()
+            inputStream.close()
+            inputStreamReader.close()
```
pocmo: This code is very Java-y and doesn't use any of the Kotlin idoms to read lines nicely from streams.

app/src/main/java/org/mozilla/focus/utils/TLDExtractor.kt
```diff
@@ -0,0 +1,102 @@
...
+        } catch (e: IOException) {
+            throw IllegalStateException(e)
+        }
```
pocmo: Crashing the app if we can't read the list is quite aggresive. I wonder if we want to continue without the list instead.

app/src/main/java/org/mozilla/focus/utils/UrlUtils.kt
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    fun getIndexOfTLD(url: String): Int {
+        val tld = getTopLevelDomain(url) ?: ""
+        return url.indexOf(tld, 0, true) + tld.length
+    }
+
+    fun getTopLevelDomain(url: String): String? {
+        return TLDExtractor.extractTLD(URL(url).host)
+    }
```
pocmo: nit: Those methods seem to deal with the "public suffix" which is not the top level domain in all cases. For example `blogspot.com` is a public suffix (because users can register a domain under that suffix) while the TLD is just "com".

app/src/test/java/org/mozilla/focus/utils/UrlUtilsTest.kt
```diff
@@ -0,0 +1,173 @@
...
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mozilla.focus.browser.LocalizedContent
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class UrlUtilsTest {
```
pocmo: You converted the test file to Kotlin but it seems like there are no tests that cover the new methods? :)

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -1336,6 +1337,19 @@ class BrowserFragment : WebFragment(), LifecycleObserver, View.OnClickListener,
...
 
+        fun bringURLTLDIntoView() {
+            if (urlView != null) {
+                val paint = urlView!!.paint
+                val charWidth = paint.measureText("a", 0, 1)
+                val urlViewWidth = urlView!!.measuredWidth
+                val numChars = urlViewWidth / charWidth
+                if (url.length > numChars) {
+                    val index = UrlUtils.getIndexOfTLD(url)
+                    urlView?.bringPointIntoView(index)
```
pocmo: Note that for URLs we load Gecko already does the extraction internally. In Fennec we get the base domain (public suffix + 1) from Gecko whenever something loads. It may be possible to ask the GeckoView team to extend the API that provides the URL.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -1324,6 +1324,7 @@ class BrowserFragment : WebFragment(), LifecycleObserver, View.OnClickListener,
                 swipeRefresh!!.isRefreshing = false
 
                 updateSecurityIcon(session)
+                bringURLTLDIntoView()
```
pocmo: Does this work well with URLs that do not contain any domains, e.g. URLs with IP addresses?

Also: Since we are doing that on the main thread. Is that matcher efficient enough to not block unnecssarily?

ekager:
@pocmo , I agree that this should be in Components, when I started working on it I didn't realize it would be so large or that Fennec had a reusable solution. Is this feature in AC's roadmap?

PR #3613, opened 2018-10-10
Update gradle

sblatz:
Generated by running the following command (per @mcomella's suggestion): ```./gradlew wrapper --distribution-type=all --gradle-version=4.10.2```


ekager:
I was also getting this error locally and it looks like it can be fixed by updating spotbugs.

here: classpath "gradle.plugin.com.github.spotbugs:spotbugs-gradle-plugin:1.6.4"
and here: ext.spotbugs_version = '3.1.4'

sblatz:
@ekager I updated those values! :)

PR #3639, opened 2018-10-15
Closes #3638: implement parallel UI test execution

### LINKED ISSUE #3638: Parallelize UI test execution
Currently our UI test execution is slow due to serial execution and monitoring results which require a painful drilldown into TaskCluster logs and Firebase.  For these reasons, we restrict our UI test execution to master only, which further removes developers from monitoring/fixing test failures. Implementing the Flank library in our CI will enable us to shard test execution and will output a variety of test results locally.  There are pending issues to consolidate the test configurations and also be able to push issues directly into GCP storage.

### What / Requirements
Update /tools/taskcluster scripts to execute using flank jar (rather than directly from Firebase API)

Tests execute in parallel.
NOTE:  email notification using taskcluster notify will be separate PR

rpappalax:
Closes #3638


colintheshots:
I'll let No-Jun review this. As a practical matter though, shouldn't we be using the Android Test Orchestrator and having it clear data between tests? This seems especially important when altering parallelism.

https://developer.android.com/training/testing/junit-runner#using-android-test-orchestrator

rpappalax:
> I'll let No-Jun review this. As a practical matter though, shouldn't we be using the Android Test Orchestrator and having it clear data between tests? This seems especially important when altering parallelism.

Whenever we launch a test in Firebase, each device is factory reset at the beginning so we start clean.  But, as you mention, if we were to run more than 1 test per device, then yes, we would probably benefit from something like Orchestrator, if that can help prevent dirty data between tests.  

However,  our plan going forward was to only run 1 test per device (in parallel) to keep the test run times as low as possible and also help us to better isolate errors, thus test data would be isolated.  If (on the off chance) we ever decided we still wanted to double up on tests, flank also works with Orchestrator apparently.

rpappalax:
@colintheshots it looks like using Orchestrator is standard practice.  Since we're not using it yet anyway and this will at least get parallelism in place, we'll go ahead with this and add Orchestrator next (in a separate PR).

PR #3645, opened 2018-10-16
Closes #2471 - Use A-C Error Pages for GV 

### LINKED ISSUE #2471: Use Android Components ErrorPages

ekager:


jonalmeida:
\o/

Context in #https://github.com/mozilla-mobile/android-components/issues/1056#issuecomment-430088229

FYI: We don't have any translations for the error pages yet, but there might be collisions with resources names if these get translated in Focus' project.

colintheshots:
@jonalmeida If there's a chance of collision, it should be as simple as adding a library-specific prefix to the resources. We should have error page translations. I'm not sure we can close the issue if translations aren't possible.

jonalmeida:
@colintheshots we are adding prefixes in a future release, but with this PR I assume translations for Focus would be done sooner.

Nothing breaks, just a bit of duplicate work. :)

ekager:
IMO built in GV error pages are not great anyway, I think this would be fine esp for Nightly. Do I have to do anything to pull the string resources into ours to be translated?

ekager:
Are the internal GV error pages localized? The white screen? 
If not, this doesn't really present a regression perse.

pocmo:
> So I did a bit of research and I believe library string resources won't be included in the localization file. It's up to the library to have them localized.

Correct. The current L10N tooling only picks up files from this repository.

We are working on getting translations into our repository. In the meantime the workaround is to use the same string ids in this repository and override them with your own translations.

> Are the internal GV error pages localized? The white screen?

No. GV intentionally doesn't ship any translations.

>  If not, this doesn't really present a regression perse.

Yeah, technically this seems to be true :)

ekager:
I think we should merge this because it's still an improvement from GV's current error page situation and not a translation regression. A-C team is working on getting l10n working and translations for these strings. Should be a few weeks and then all we will have to do is update our A-C version. Thanks all! :)

PR #3653, opened 2018-10-16
Remove fenix artifacts exposition.

MihaiTabara:


JohanLorenzo:
Thanks for spotting this! That's an overlook of mine in https://github.com/mozilla-mobile/focus-android/pull/3388

PR #3649, opened 2018-10-16
Fix Queue args after taskcluster 5.0.0 breaking API release.

MihaiTabara:
As @JohanLorenzo mentioned, most likely the Docker image bumped [here](https://github.com/mozilla-mobile/focus-android/pull/3621/files#diff-ac0229d1171c0983b27003af4c7441a5R23) introduced the `taskcluster-client.py` bump as well to `5.0.0` which had breaking API changes within its last release (https://pypi.org/project/taskcluster/). We faced the same issues in `android-components` and we fixed them [here](https://github.com/mozilla-mobile/android-components/pull/993/files). 

If we can confirm the Dockerfile bumped the `taskcluster` I think we're safe to merge this and fix the CI taskcluster issues.

PR #3656, opened 2018-10-16
Fix API url for taskcluster-client calls.

MihaiTabara:
Follow-up PR fix after https://github.com/mozilla-mobile/focus-android/pull/3649, https://github.com/mozilla-mobile/focus-android/pull/3653 and https://github.com/mozilla-mobile/focus-android/pull/3654


MihaiTabara:
CC: the task timed-out, I think we need to specify the `.net` though. What do you think?

PR #3658, opened 2018-10-16
Closes #3630 - Adds pre-push hook and update read me

### LINKED ISSUE #3630: Add recommended pre-push hook to quicken feedback loops
Developers time is valuable; it's ideal to get the quickest feedback turn-around time as possible. One way to do that is to make sure the tests and static analysis tools run locally before it hits gitHub and has to run on slow CI. We can a recommended pre-push hook, which all the devs have to manually set up (it's one line!), so that they hit these problems locally before they hit the server.

We have this in firefox-tv for a while: https://github.com/mozilla-mobile/firefox-tv/blob/master/quality/pre-push-recommended.sh

And document it in our README: https://github.com/mozilla-mobile/firefox-tv#pre-push-hooks

It's up to your team if you want to implement it but it's been working great for us so far!

### What / Requirements
- Add pre-push hook
- Update readme saying how to use it
- Send out email asking people to use it

ekager:


quality/pre-push-recommended.sh
```diff
@@ -0,0 +1,27 @@
...
+# `ln -s ../../quality/pre-push-recommended.sh .git/hooks/pre-push`
+
+# Descriptions for each gradle task below can be found in the
+# output of `./gradlew tasks`.
+./gradlew -q \
+        checkstyle \
+        ktlint \
+        pmd \
+        detektCheck \
+        assembleFocusX86DebugAndroidTest \
```
colintheshots: So we build the UI tests and then only run the unit tests?

ekager: Talked to @mcomella and they do it this way in ffftv so that they can see that the UI tests at least compile before someone pushes a PR

PR #3669, opened 2018-10-17
Fixes #3607 - Check if activity is finishing before showing prompt

### LINKED ISSUE #3607: BadTokenExceptionorg.mozilla.focus.gecko.GeckoViewPrompt in onAlert
```
BadTokenException

Unable to add window -- token android.os.BinderProxy@d5bf30f is not valid; is your activity running?

    org.mozilla.focus.gecko.GeckoViewPrompt in onAlert at line 119
    Called from: org.mozilla.geckoview.GeckoSession in handlePromptEvent
```

ekager:


ekager:
No clear STR but from some searching I think we should also check if the activity is finishing before launching dialogs

PR #3671, opened 2018-10-17
Moves the mobile metrics ping builder

boek:


boek:
Also added some tests 👍

app/src/main/java/org/mozilla/focus/web/GeckoWebViewProvider.kt
```diff
@@ -635,16 +636,21 @@ class GeckoWebViewProvider : IWebViewProvider {
...
         }
 
-        private fun sendTelemetrySnapshots() {
+        private fun storeTelemetrySnapshots() {
+            val storage = MobileMetricsPingStorage(context)
+            if (!storage.shouldStoreMetrics()) return
+
             geckoRuntime!!.telemetry.getSnapshots(true).then({ value ->
-                if (value != null) {
+                launch(IO) {
```
colintheshots: Just a comment. Right now, I created a CoroutineDispatcher called "IO" because the CommonPool can be dangerous and cause deadlocks on some devices.

```
val IO: CoroutineDispatcher by lazy {
    Executors.newCachedThreadPool().asCoroutineDispatcher()
}
```

However, the coroutine standard library adds its own Dispatchers.IO in a later version. We should switch from mine to the library's version as soon we're upgraded enough. I should've used a more distinct name.

boek: Sounds good to me

PR #3680, opened 2018-10-18
Add l10n.toml for l10n testing with compare-locales, and automation to run it

Pike:
On PRs, this just runs validation after the unit tests.
On master, this triggers another task in the graph w/out dependencies, either in nor out.
We might want to replace check_translations.py eventually, but not yet.

I wonder at which point we should have a task graph for PRs, too.


colintheshots:
It looks like this breaks the build...

```
pip install compare-locales
/bin/bash: 5.0: No such file or directory
```

colintheshots:
It sure looks like it's treating the versions as STDIN and STDOUT for the UNIX shell because that's what the syntax means.

Pike:
Sorry, I had uncommitted changes locally, and missed out on the quoting.

Builds green now.

PR #3694, opened 2018-10-20
Revert 3687 fix3624

ekager:
Found a less complicated fix

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -67,7 +67,7 @@
         </activity>
 
         <activity android:name=".activity.MainActivity"
-            android:launchMode="singleTop"
+            android:launchMode="singleTask"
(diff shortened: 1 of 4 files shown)


colintheshots:
Is there a reason why we're reverting a second time?

ekager:
Making SettingsActivity's parent MainActivity fixes the original issue without the complexity of this fix.

ekager:
And introduced new regressions like #3697

colintheshots:
> Making SettingsActivity's parent MainActivity fixes the original issue without the complexity of this fix.

I'm not seeing that the fix does anything. Opening the launcher icon still erases the SettingsActivity with singleTask and the parent activity set the MainActivity. ParentActivityName, from my past experience, only helps with Up navigation.

colintheshots:
> And introduced new regressions like #3697

It's a one-liner edge case. There's a fix in the PRs.

colintheshots:
Per my conversation with @ekager, it seems her phone doesn't exhibit the original issue so the proposed fix isn't viable. Closing.

Issue #3556, opened 2018-10-02
WebviewdataTest.java is failing on Geckoview as well, retaining cache after page is cleared


npark-mozilla:
## Steps to reproduce
Run WebViewDataTest.java on Geckoview.  Use this PR (https://github.com/mozilla-mobile/focus-android/pull/3555) 

### Expected behavior
The test should pass

### Actual behavior
Test fails, after page is cleared, the directory contains the page cache
Also, the test shows that `service-worker.js` page is never requested by Focus when loading the test html page.

It is similar to the issue #2940 , but now happening on Geckoview as well. Perhaps it's Focus issue?

### Device information
* Android device: Nexus 5X simulator, API 27
* Focus version: commit 0747ec1f2a133a78a9cae7de732e7078924c8b8c

npark-mozilla:
![screen shot 2018-10-02 at 4 29 42 pm](https://user-images.githubusercontent.com/6930774/46375494-947ac000-c661-11e8-90aa-660d7810c9b6.png)

Sdaswani:
@boek please move this to the 8.0 sprint if you dont' expect to get it done today.

PR #3706, opened 2018-10-22
For #3556 - Fixes WebViewDataTest to work with GeckoView

boek:


app/src/androidTest/java/org/mozilla/focus/privacy/WebViewDataTest.java
```diff
@@ -54,15 +54,16 @@
     private static final String LOGTAG = "WebViewDataTest";
 
     /**
-     * A list of files Android Studio will inject into the data dir if you:
+     * A list of files Android Studio will inject into the data dir if you:x`
```
npark-mozilla: typo?

app/src/androidTest/assets/test.html
```diff
@@ -28,13 +28,10 @@ <h2>Service worker</h2>
 <script type="text/javascript">
         // Add a key value pair to local storage. This is required by WebViewDataTest which asserts
         // that the local storage directory should exist
```
npark-mozilla: Can we add a comment saying that this test is for now geckoview only?

boek: @npark-mozilla the test should still work with WebView :) Or is it GV only because of that assertion?

npark-mozilla: Wait, I think I'm misunderstanding.  Thought GV doesn't have service workers in private mode, so I thought you removed them, but it is appended. How does it work exactly now?  (I thought it won't work on WV because service worker.js is disabled)

boek: service workers were disabled on GV with privacy mode. Switched to just loading the JS dynamically with Javascript which gives us similar behavior for the tests 👍

PR #3718, opened 2018-10-23
Fix zh-CN locale correction

colintheshots:
I noticed this issue while addressing a crash spotted in LocaleListPreference. The locale to name map is not correctly matching by the locale tag.


colintheshots:
@YFdyh000 Does this seem a reasonable change to make the code work?

PR #3755, opened 2018-10-25
Closes #3744 - Create SessionManager Extension to Close GeckoSessions

### LINKED ISSUE #3744: Regression GeckoSessions not closed when all sessions removed
Possible source of some session leakage in Nightly from #3367, this code was removed and I don't think it was added back in. When multiple tabs are open and all cleared, sessions will be left open: 
https://github.com/mozilla-mobile/focus-android/pull/3367/files#diff-0dc0ec4838f44e70af7c119fb92b53e3L384

cc @pocmo where do you think a good place to clean up the geckosessions is now? I can't use the SessionObserver because I need to access the saved session bundles and by the time the callbacks are invoked, the sessions are destroyed.

ekager:


ekager:
Think @pocmo mentioned in the ticket #3744 this would be in the engine component but unclear if that's going to be integrated into Focus.

ekager:
Oops accidentally closed this, will merge when tasks complete again 😨

pocmo:
> This is good work. It should ultimately bubble up to the SessionManager component eventually. We have an Engine interface that doesn't describe how to close and clean up sessions and Focus is using a DummyEngine anyway.

Yeah, this will happen automatically when using the  `engine` component. The linking of a `Session` and an `EngineSession` (aka `GeckoSession`) will be done by the components and they will automatically unlink and close the `EngineSession` once the `Session` is removed from the `SessionManager`. I briefly looked at whether we could re-use the functionality here, but this would mean that you end up writing your own implementation of `browser-engine-gecko`.

PR #3778, opened 2018-10-26
Closes #3754 - Only Select Normal Browsing Sessions

### LINKED ISSUE #3754: Clearing normal history then shows custom tab
## Steps to reproduce
1. Open a link in a custom tab and leave it open
2. Open Focus and open a link, and then erase

### Expected behavior
Normal browsing tab is erased, custom tab remains

### Actual behavior
Normal browsing tab is erased then custom tab is displayed (with custom tab UI) in Focus

### Device information

* Android device: Pixel 2
* Focus version: Nightly

ekager:
This fixes the problem I was seeing, but not sure if we will ever need to `showBrowserScreenForCurrentSession` on selection of a custom tab? I couldn't think of a case where we would have to since custom tabs don't have multitasking. Maybe someone from A-C can sanity check this fix?

### DIFF
--- a/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
+++ b/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
@@ -104,7 +104,9 @@ open class MainActivity : LocaleAwareAppCompatActivity() {
     private fun registerSessionObserver() {
         components.sessionManager.register(object : SessionManager.Observer {
             override fun onSessionSelected(session: Session) {
-                showBrowserScreenForCurrentSession()
+                if (!session.isCustomTabSession()) {
+                    showBrowserScreenForCurrentSession()
+                }
             }
 
             override fun onAllSessionsRemoved() {


app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -104,7 +104,9 @@ open class MainActivity : LocaleAwareAppCompatActivity() {
     private fun registerSessionObserver() {
         components.sessionManager.register(object : SessionManager.Observer {
             override fun onSessionSelected(session: Session) {
-                showBrowserScreenForCurrentSession()
+                if (!session.isCustomTabSession()) {
+                    showBrowserScreenForCurrentSession()
+                }
```
pocmo: This is a bug in `SessionManager`. A custom tab should never get selected in the first place.

pocmo:
This will cause other issues. With that patch the UI and `SessionManager` will be out of sync.

ekager:
Ah true, so we should revert this and wait for the AC fix?

pocmo:
Good question. I think you'll still see this issue if you leave the activity (so that it gets destroyed) and the next time you go back.

We moved this into our new sprint (0.30). Is there a Focus release planned soon - so that a dot release would be required?

PR #3788, opened 2018-10-29
No Issue: converted HomeScreen.java to kotlin

Hackertronix:


ekager:
Thanks for your contribution! Unfortunately, the branch has gone stale as we've been changing lots of files, so I have to close this. Generally, we do not merge straight conversions unless there is an issue open to do so. Feel free to look for other bugs to pick up

PR #3798, opened 2018-10-29
For mozilla-l10n/focus-android-l10n#42 - Fixes comments for l10n

boek:


colintheshots:
Approved because we're allowing the extra translations until we re-import strings.

Issue #3830, opened 2018-11-02
New homescreen tip to inform user about the new allowlist feature


vesta0:
For this feature: https://github.com/mozilla-mobile/focus-android/issues/957

### Why/User Benefit/User Problem

As a user (new/existing) I would like to know how to add afrequently visited site to my allowlist so the site doesn't break by tracker blocking

### What / Requirements

Add a new homescreen tip on how to add a site to the allowlist, to link to new Sumo page (TBD)

### Acceptance Criteria (how do I know when I’m done?)

-Users will see a home screen tip informing them on how to use this feature:

When a user navigates to a site and disables the "trackers Blocked" toggle, the site will automatically get added to the allowlist, and tracking protection will remain off as long as the user is on that site. Once the user navigates away to another site, tracking protection is re-enabled.

vesta0:
List of existing tips: https://github.com/mozilla-mobile/focus-android/wiki/Homescreen-Tips

(we may want to modify the existing tip: "Site behaving unexpectedly? Try turning off Tracking Protection" or add a new tip)

BrianNJones:
@vesta0 
Trusted site? Allowlist disables Tracking Protection for sites you know and trust.

vesta0:
Thank you so much @BrianNJones ! And as you mentioned on Slack we will link the "Allowlist disables Tracking Protection" part to the Sumo article. 

@jonisavage could you please put a Sumo article together for this new Allowlist feature? Beta will be available on Nov 13 so would be great to have it ready by then!

jonisavage:
Sure thing, here's the direct link: https://support.mozilla.org/kb/add-trusted-websites-your-allow-list

But please use this link in your code: https://support.mozilla.org/1/mobile/%VERSION%/%OS%/%LOCALE%/focus-android-allowlist

vesta0:
Thank you @jonisavage ! Just one thing - the screenshot you have is from iOS whereas we have only built this feature for Android for now. Let me know if you need any more info before making adjustments.

sv-ohorvath:
Verified on Nightly the new tip and SUMO link.

PR #3832, opened 2018-11-02
For #3830 - Adds a tip for Allowlist

boek:


app/src/main/java/org/mozilla/focus/tips/TipManager.kt
```diff
@@ -33,6 +34,13 @@ class Tip(val id: Int, val text: String, val shouldDisplay: () -> Boolean, val d
         private const val FORCE_SHOW_DISABLE_TIPS_LAUNCH_COUNT = 2
         private const val FORCE_SHOW_DISABLE_TIPS_INTERVAL = 30
 
+        fun createAllowlistTip(context: Context): Tip {
+            val id = tip_explain_allowlist
+            val name = context.resources.getString(id)
+
+            return Tip(id, name, { true })
```
boek: We'll have to remember to figure this out once we build out the allowlist feature

boek: #3833 so we don't forget

PR #3846, opened 2018-11-05
Closes #3509 Closes #3510 Closes #3511 Exception List Setting

### LINKED ISSUE #3509: Exception list settings UI
image

### LINKED ISSUE #3510: Exception list manager
We'll need something to manage the list of exceptions for #957

One thing to think about is we're starting to have quite a few areas where we're managing a list of things in preferences. (Autocomplete, exceptions, and blocklists) I think we could build a generic collection so we're not duplicating that work every time.

### LINKED ISSUE #3511: Change behavior of tracking protection toggle
This is a piece of #957
---

We're going to want to change the behavior of the tracking protection toggle from being based on the session to now follow the new exception list.

When visiting a website we will check the list and if it matches turn off tracking protection. If a user toggles for a given website it should reflect the state in the exception list.

ekager:
To test, I've been adding this link to the Exceptions List:
https://mozilla.github.io/tracking-test/disconnect.html

Unfortunately, because of the GV TP bug, I haven't been able to verify with GV yet.

Some remaining questions I have:
1. Is saving the host of the site as the exception correct? 
2. Do you have any suggestions about loading the exception list in onURLChanged since this will probably be called a lot but it's the only way to catch redirects for the right hosts to catch as exceptions - should we save a local list instead and notify changes?


vesta0:
@ekager @boek yes let's add a probe to indicate when user removes an item from their allowlist, to inform us on whether users switch off the toggle without intending to add that site to their allowlist (or change their minds) and how often that happens.

PR #3871, opened 2018-11-10
Closes #3858 - Add Telemetry for Exceptions List

### LINKED ISSUE #3858: Add Telemetry Event for When User Removes Item From Allowlist
As part of #3510 let's add a new telemetry event for when the user removes a site from their exception list to inform us on whether users switch off the toggle without intending to add that site to their allowlist (or change their minds) and how often that happens. 
Maybe we should also add a probe for the user opening the Exceptions List page in Settings? That way we can compare the users that add sites to the allowlist.
cc @vesta0

ekager:


ekager:
# Request for data collection review form

**All questions are mandatory. You must receive review from a data steward peer on your responses to these questions before shipping new data collection.**

1) **What questions will you answer with this data?**

* Do users switch off the toggle without intending to add that site to their allowlist (or change their minds) and how often that happens.
* Are users aware and using the exceptions list?
* Do users edit their exceptions lists? 

2) **Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements?**

* We are changing the blocking toggle behavior and want to know how users will interact with the new behavior of adding the domain to the exceptions list.

3) **What alternative methods did you consider to answer these questions? Why were they not sufficient?**

N/A

4) **Can current instrumentation answer these questions?**

* We are reusing telemetry on the blocking toggle as an indicator of when the user adds a site to the exceptions list. For the other events, we cannot reuse because they are new features.

5) **List all proposed measurements and indicate the category of data collection for each measurement, using the Firefox [data collection categories](https://wiki.mozilla.org/Firefox/Data_Collection) on the found on the Mozilla wiki.**

* Category 2.

**Events**

| Event                                    | category | method     | object | extra  |
|------------------------------------------|----------|------------|--------|--------|
| Open Exceptions Setting | action   | open | allowlist |    |
| Remove Exceptions Domains | action   | remove | allowlist |  total*  |
| Remove All Exceptions Domains | action   | remove_all | allowlist |  total*  |

* - number of domains removed

6) **How long will this data be collected?**

* @mmccorks will permanently monitor this data.

7) **What populations will you measure?**

* All release, beta, and nightly users with telemetry enabled.

8) **Please provide a general description of how you will analyze this data.**

* Re-dash queries + dashboards

9) **Where do you intend to share the results of your analysis?**

* Only on re-dash and with mobile teams.

ekager:
Waiting for data review from @liuche

PR #3880, opened 2018-11-12
Fixes #3853 - Dont clear the cache when cleaning up geckoview

### LINKED ISSUE #3853: Don't clear cache when using GeckoView
Currently, Focus clears the cache directory when cleaning up a session, even when using GeckoView. This results in a 3s startup time penalty on the Pixel 2 because Gecko has to unpack libraries again. I recommend we not do this :)

boek:
This was originally added to fix WebViewDataTest in #3706 

I took it out to get the test to fail again and it's passing. Maybe the newer version of GeckoView fixed a bug?


ekager:
Looks like there are unused imports now @boek

colintheshots:
I pushed the unused imports fix

PR #3881, opened 2018-11-12
Fixes #3853:  Don't clear cache when using GeckoView

### LINKED ISSUE #3853: Don't clear cache when using GeckoView

colintheshots:


colintheshots:
Duplicate

PR #3849, opened 2018-11-06
No issue: Update AC to 0.30.0

ekager:

PR #3891, opened 2018-11-13
Fixes #3849: Update to Android Components 0.31 and Kotlin 1.3.0

colintheshots:


boek:
Looks like some tests also need to be updated?

PR #3935, opened 2018-11-18
No issue: Update to JUnit 5 and Gradle 4.7

kasecato:


colintheshots:
Thanks for your contribution! I made the change directly as I cannot confirm the binary checksum easily.

PR #3936, opened 2018-11-18
Add "New tab" option to browser menu

rutvora:
Found it much needed by me and my friends atleast. Considering there was a "open in new tab" option, why not this?


colintheshots:
Thanks for your contribution! Our team wants to improve multitasking in general with a substantive fix that won't clutter an already too-expansive menu. I understand the immediate desire for a fix, but we need to consider the user experience. We have a branch we're testing in a customer study so we can produce a solution that addresses their chief concerns. You can view that progress here: https://github.com/colintheshots/focus-android/tree/newTabsV2

PR #3989, opened 2018-11-30
Fix IRC notification (broken build) and run master graphs on every br…

JohanLorenzo:
…anch of the main repo

Fixes https://tools.taskcluster.net/groups/Rl-7lqy_Tuyhyw-EkZxP7w/tasks/KxpkitluS4KPEQVjOAlnUA/runs/3/logs/public%2Flogs%2Flive.log#L658

r? @rpappalax

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -17,8 +17,7 @@ tasks:
       deadline: {$fromNow: '2 hours'}
       provisionerId: aws-provisioner-v1
       workerType: github-worker
-      routes:
-        - 'notify.irc-channel.#android-ci.on-any'
+      routes: []
       scopes: []
       payload:
         maxRunTime: 7200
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -59,7 +58,7 @@ tasks:
 ###############################################################################
   - $if: 'tasks_for == "github-push"'
     then:
-      $if: 'event.ref == "refs/heads/master"'
+      $if: 'event.repository.fork == false'
```
JohanLorenzo: This event is documented at https://developer.github.com/v3/activity/events/types/#pushevent

.taskcluster.yml
```diff
@@ -17,8 +17,7 @@ tasks:
       deadline: {$fromNow: '2 hours'}
       provisionerId: aws-provisioner-v1
       workerType: github-worker
-      routes:
```
JohanLorenzo: Do you want notification on pull requests too, @rpappalax ?

rpappalax: @JohanLorenzo  for now, we just want additional notifications whenever the UI tests complete.  I added that in to the python script here: https://github.com/mozilla-mobile/focus-android/blob/master/tools/taskcluster/schedule-master-build.py#L101  hopefully the scopes allow notify from python client.

JohanLorenzo: Okay, since this type of job doesn't use `schedule-master-build.py`, we don't need to expose this scope on PRs.

rpappalax:
@boek does this require focus-codeowner approval (since part of build process)?

PR #4043, opened 2018-12-18
Fixes #4041: Removes long label from erase&open

### LINKED ISSUE #4041: Extra characters in "Erase and open" in the app shortcut
## Steps to reproduce
1. Long tap the Focus homescreen icon.
### Expected behavior
"Erase and open" & "Erase browsing history" options are displayed.
### Actual behavior
There are some extra characters in "Erase and open".
screenshot_20181218-154037
### Device information

* Android device: 
OnePlus 5T (Android 8.0), Nokia 6 (Android 7.1.1)
* Focus version: 8.0.3 Beta & Nightly.

sblatz:
Since we're creating these labels entirely in XML, there isn't an easy way for us to insert the name of the app into the long label (which is resulting in the placeholder text still appearing for end users). I think just using the short label is sufficient since users will know which app icon they're taking the shortcut action on.

### DIFF
--- a/app/src/focusBeta/res/xml-v25/shortcuts.xml
+++ b/app/src/focusBeta/res/xml-v25/shortcuts.xml
@@ -21,7 +21,7 @@
       android:enabled="true"
       android:icon="@drawable/ic_shortcut_erase"
       android:shortcutShortLabel="@string/shortcut_erase_and_open_short_label"
-      android:shortcutLongLabel="@string/shortcut_erase_and_open_long_label">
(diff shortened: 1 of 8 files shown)


colintheshots:
TaskCluster is failing due to 64 ExtraTranslation errors for the removed string.

sblatz:
@colintheshots do I need to fix something or is it expected that it fails with a string change like this?
I could keep the deprecated string around if that makes it not fail.

Issue #4044, opened 2018-12-18
MalformedURLException from org.mozilla.focus.telemetry.TelemetryWrapper.addLoadToHistogram 


cpeterso:
```
Caused by: java.net.MalformedURLException: 
   at java.net.URL.<init> (URL.java:596)
   at java.net.URL.<init> (URL.java:486)
   at java.net.URL.<init> (URL.java:435)
   at org.mozilla.focus.telemetry.TelemetryWrapper.addLoadToHistogram (TelemetryWrapper.kt:336)
   at org.mozilla.focus.observer.LoadTimeObserver$addObservers$1.onLoadingStateChanged (LoadTimeObserver.kt:53)
   at mozilla.components.browser.session.Session$$special$$inlined$observable$4$lambda$1.invoke (Session.kt:159)
   at mozilla.components.browser.session.Session$$special$$inlined$observable$4$lambda$1.invoke (Session.kt:22)
   at mozilla.components.support.base.observer.ObserverRegistry.notifyObservers (ObserverRegistry.kt:101)
   at mozilla.components.browser.session.Session.notifyObservers (Session.kt)
   at mozilla.components.browser.session.Session.notifyObservers (Session.kt:297)
   at mozilla.components.browser.session.Session.access$notifyObservers (Session.kt:22)
   at mozilla.components.browser.session.Session$$special$$inlined$observable$4.afterChange (Delegates.kt:72)
   at kotlin.properties.ObservableProperty.setValue (ObservableProperty.kt:41)
   at mozilla.components.browser.session.Session.setLoading (Session.kt)
   at org.mozilla.focus.session.SessionCallbackProxy.onPageFinished (SessionCallbackProxy.kt:38)
   at org.mozilla.focus.webview.FocusWebViewClient.onPageFinished (FocusWebViewClient.java:221)
   at jY.c (PG:230)
   at zn.handleMessage (PG:72)
   at android.os.Handler.dispatchMessage (Handler.java:102)
   at android.os.Looper.loop (Looper.java:154)
   at android.app.ActivityThread.main (ActivityThread.java:6320)
```

from this Focus 8.0 crash report:

[...]

PR #4070, opened 2018-12-27
Handle malformed URLs without crashing

colintheshots:
Handle malformed URLs without crashing. Use clientID for Fretboard.


colintheshots:
Fixes #4044

Issue #4075, opened 2018-12-30
Convert Java classes to Kotlin


nikit19:
## Steps to reproduce
There are several classes in the repository written in Java, I would like to convert them to Kotlin. I would like to start by converting the util classes.
### Expected behavior
All the Java classes should be converted to Kotlin

I would like to work on this issue.

devanand:
I would like to work on this issue. Could I take it?

Aman199825:
I would like to work on this issue.May I ?

kamaleshpati:
i would like to work on it..can i?

shangeethsivan:
@colintheshots @vesta0 I have raised a PR for this can you please check it out. ?

prajwalgowdahg:
i would like to work on this

losdc:
I would like to work on this.Can I?

bhushan-borole:
Hello @nikit19 
Is this issue still available to be worked upon?
I would like to go for it.
Thank you

luisgrcia:
Hi :)
I think you know my question.
Is this issue still available to work on it?
Have a nice day 😄

luisgrcia:
> Hi :)
> I think you know my question.
> Is this issue still available to work on it?
> Have a nice day 😄

#4632

kanika-narang:
Hi, 

I am interested in contributing, is this task still open. 

Thanks and have a nice day.

gurleennsidhuu:
Hi. Can I work on this issue?

yashj6294:
Hi,
I would like to work on this issue

PR #4094, opened 2019-01-05
Issue #4075: Ported the StatusBarUtils class from Java to Kotlin

HansJesse:
First commit to this project: Hoping to start contributing to Mozilla! Please guide me if my pull or git process is incorrect. Started small to make sure I do things properly first.


app/src/main/java/org/mozilla/focus/utils/StatusBarUtils.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    companion object {
+        private var STATUS_BAR_HEIGHT = -1
+
+        fun getStatusBarHeight(view: View?, listener: (Int) -> Unit) {
+            if (StatusBarUtils.STATUS_BAR_HEIGHT > 0) {
+                listener(STATUS_BAR_HEIGHT)
+            }
+
+            view?.let {
```
colintheshots: This is super close. This would work as written, but it'd be cleaner Kotlin code if you simplified a bit. view?.let{} surrounding view.setOnApplyWindowInsetsListener should be identical to just view?.setOnApplyWindowInsetsListener with the safe call operator. https://kotlinlang.org/docs/reference/null-safety.html#safe-calls

HansJesse: Oh looking at it again, I agree completely. That is simpler/cleaner. I made the change, tested it on my device, and pushed it up. Thanks!

PR #4110, opened 2019-01-09
Closes #4105 Separate Erase All Fun For Just Erase Icon

### LINKED ISSUE #4105: Pressing the Back button erases all tabs
## Steps to reproduce
1. Go to wikipedia.com
2. Long press a link and open it in a new tab.
3. Switch to the 2nd tab, then press the back button.
### Expected behavior
The 2nd tab is closed, but the first remains open.
### Actual behavior
Both tabs are deleted.
### Device information

* Android device: Pixel (Android 9), HTC Desire 820 (Android 6.0.1)
* Focus version: Nightly 8.0.4 & debug local build v 8.0.5, both GV & WV

ekager:
When I moved the erase button to the toolbar, I changed the erase function to erase all sessions to match the new behavior, but missed that erase() needs to be used in other cases to just erase the current session. This adds a function to eraseAll() in the case of the new toolbar erase button. 

I can separate the talkback/cleanup code out into its own function, but @boek I wanted your eyes on it to check if any of the crash reporter code should be using eraseAll instead of erase?


colintheshots:
@boek Code freeze is this Friday. Is this change good?

PR #4111, opened 2019-01-10
Closes #3921 - Show Snackbar after page load after toggling block toggle

### LINKED ISSUE #3921: Let user know about the new behaviour for trackers blocked/allowlist toggle
As a user, I need to know that disabling tracking protection for a site, will add the site to my allowlist permanently (until I turn it back on) and from then on everytime I navigate to this site, tracking protection will be disabled.

As a user, I need to know that turning the tracking protection toggle back on, will remove the site from my allowlist. 

### What / Requirements
Add messaging (could be just in time messaging) to tracking protection toggle to communicate adding/removing sites from allowlist.

Toggle switch behaviour is communicated to the user.

ekager:


ekager:
@boek this has 4 strings for 9.0

PR #4126, opened 2019-01-11
Adds comments to resolve mozilla-l10n/focus-android-l10n#45

boek:


boek:
This will fail CI until we export/import strings

Issue #4095, opened 2019-01-06
convert widget classes to Kotlin


nikit19:
## Steps to reproduce
There are several classes in the widget directory written in Java, I would like to convert them to Kotlin. 

### Expected behavior
All the Java classes should be converted to Kotlin

I would like to work on this issue.

PR #4138, opened 2019-01-15
convert widget classes to Kotlin

nikit19:
#4095


nikit19:
@colintheshots PR updated. Please review

PR #4144, opened 2019-01-16
Update Gradle, spotbugs, Kotlin

colintheshots:


colintheshots:
D'oh! There's new lints to deal with. I'll fix.

colintheshots:
This is super old now. I'll re-do it later if we need it.

PR #4147, opened 2019-01-16
Fixes #4125: Update to GeckoView Beta 65

### LINKED ISSUE #4125: Upgrade to GV 65

colintheshots:


jonalmeida:
LGTM 👍 

Do verify if the `IMAGE_SRC` is correctly retrieved from an image that is also clickable, i.e. image src is the image link and uri is the `<a>` tag's `href` value. I can't recall off-hand if those values were switched.

colintheshots:
I will exclude support-v4 from GV Beta to handle the TaskCluster failure.

Issue #4078, opened 2019-01-01
Convert tests written in Java to Kotlin


nikit19:
## Steps to reproduce
There are several tests written in Java.
### Expected behavior
They should be converted to Kotlin

ninad458:
This one should be closed I guess

PR #4150, opened 2019-01-17
convert tests written in Java to Kotlin

nikit19:
#4078


app/src/androidTest/java/org/mozilla/focus/activity/AccessAboutAndYourRightsPagesTest.kt
```diff
@@ -0,0 +1,129 @@
...
+import org.hamcrest.Matchers.containsString
+import org.mozilla.focus.fragment.FirstrunFragment.Companion.FIRSTRUN_PREF
+import org.mozilla.focus.helpers.EspressoHelper.openMenu
+
+// This test visits each page and checks whether some essential elements are being displayed
+// https://testrail.stage.mozaws.net/index.php?/cases/view/81664
+@RunWith(AndroidJUnit4::class)
+class AccessAboutAndYourRightsPagesTest {
+
+    @Rule
```
mkykadir: I get error with Kotlin tests that states
`...runners.rules.ValidationError: The @Rule 'mActivityTestRule' must be public`
and test cannot start but after replacing @Rule with @get:Rule Kotlin tests can start and run.

PR #4151, opened 2019-01-17
Automation: Resolves issue with multiline string concatenation

mitchhentges:


mitchhentges:
[Successful staging build](https://tools.taskcluster.net/groups/QU_AENBFQx2mr4WWEQ2cOA). The push failure is due to a known change in dep cert [1517851](https://bugzilla.mozilla.org/show_bug.cgi?id=1517851)

Issue #4184, opened 2019-01-25
Unable to run Focus on Emulators


RotBolt:
Whenever I try to build and run app on emulator, I get this

`[INSTALL_FAILED_NO_MATCHING_ABIS: Failed to extract native libraries, res=-113]`

INSTALL_FAILED_NO_MATCHING_ABIS is when you are trying to install an app that has native libraries and it doesn't have a native library for your cpu architecture.

Also Focus is an browser application, it doesn't have any hardware related functionality, so it should run on emulator.

lobontiumira:
I was able to run the Focus 8.13.1 (build 1191.0a1-20210624093849) on an emulator.
@RotBolt could you, please, retry?

PR #4185, opened 2019-01-25
Add splits for CPU architectures to support Emulators

KaenDagger:
Running on emulator results in 
`[INSTALL_FAILED_NO_MATCHING_ABIS: Failed to extract native libraries, res=-113]`
referencing #4184 

if we add splits for different architectures (in build.gradle), it would be able to run emulators


colintheshots:
Emulators are working fine on the latest master. This would need a rebase anyway. Closing.

PR #4202, opened 2019-02-07
Update to use GV 65

droeh:
Adds stubs for new API, updates various constants, works around API changes, etc.


colintheshots:
Code failed the style check for line length. Be sure to use the pre-push hook commands to run detekt and ktlint.

        MaxLineLength - [GeckoWebViewProvider.kt] at focus-android/app/src/main/java/org/mozilla/focus/web/GeckoWebViewProvider.kt:354:126

colintheshots:
It's failing on mixed support library versions because of the palette dependency. I know I already fixed this in master, so look there.

https://tools.taskcluster.net/groups/KxAriBv-RY6P5ZGfPHUBIw/tasks/KxAriBv-RY6P5ZGfPHUBIw/runs/0/logs/public%2Flogs%2Flive.log#L3506

droeh:
> It's failing on mixed support library versions because of the palette dependency. I know I already fixed this in master, so look there.

Thanks, think I found it.

PR #4203, opened 2019-02-08
testWebView folder classes converted to Kotlin

noloman:
Converted the tests in the `testWebView` folder to Kotlin


noloman:
can somebody point out why `Taskcluster` is failing here please?

Issue #4214, opened 2019-02-26
Lack of GV65 crash reports -- possible incorrect version reported?


droeh:
Opening an issue for this based on Slack conversation -- we're not seeing any crash reports for GV 65 despite 8.0.8 using it; we're still seeing significant crash reports for GV 64. It's possible that we're misreporting the GV version number, or (worse) that we're not reporting GV 65 crashes at all. It's not clear yet if this is a Focus problem or a GV CrashReporter problem.

lobontiumira:
I will close this issue, since it is no longer up to date.

PR #4221, opened 2019-03-01
Update android components

droeh:
This updates the AC version to 0.32.2, should fix issue #4214 hopefully -- but it's not clear how to test CrashReporter locally, as BuildConfig.SENTRY_TOKEN is an empty string for local builds. #


colintheshots:
We just updated to a later version of A-C. Sorry we didn't get to this sooner!

PR #4265, opened 2019-05-01
Update to AC 0.51.0 and GV Nightly 4/29/2019.

droeh:
Fairly minor changes this time.


colintheshots:
ProGuard appears to be blowing up. Please see if we need additional -dontwarn rules.

PR #4284, opened 2019-05-28
Update to AC 0.53.0, GV 68 5/20/19.

droeh:
Updates AC and GV versions and replaces android support library with androidx.


app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
```diff
@@ -119,7 +119,7 @@ open class AutocompleteListFragment : Fragment(), CoroutineScope {
             inflater.inflate(R.layout.fragment_autocomplete_customdomains, container, false)
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
-        domainList.layoutManager = LinearLayoutManager(activity, LinearLayoutManager.VERTICAL, false)
+        domainList.layoutManager = LinearLayoutManager(activity, RecyclerView.VERTICAL, false)
```
colintheshots: Is this the correct VERTICAL constant?

PR #4330, opened 2019-08-14
No issue: Update disconnect blocklists

colintheshots:


app/build.gradle
```diff
@@ -306,6 +306,11 @@ task buildBlocklists(type:Copy) {
     // Android can't handle dashes in the filename, so we need to rename:
     rename 'disconnect-blacklist.json', 'blocklist.json'
     rename 'disconnect-entitylist.json', 'entitylist.json'
+    rename 'adGuard-blacklist.json', 'adguard_blacklist.json'
+    rename 'easyList-blacklist.json', 'easylist_blacklist.json'
+    rename 'easyPrivacy-blacklist.json', 'easyprivacy_blacklist.json'
+    rename 'fanBoyAnnoyance-blacklist.json', 'fanboyannoyance_blacklist.json'
+    rename 'fanBoySocial-blacklist.json', 'fanboysocial_blacklist.json'
     // google_mapping.json already has an expected name
```
MihaiTabara: There’s some:
```
/opt/focus-android/app/src/main/res/raw/social-tracking-protection-blacklist.json: Error: '-' is not a valid file-based resource name character: File-based resource names must contain only lowercase a-z, 0-9, or underscore[0K
```
which makes me think that `shavar-prod-lists/social-tracking-protection-blacklist.json` might need renaming.

I wonder if we're missing a rename of `social-tracking-protection-blacklist.json` -> `social_tracking_protection_blacklist.json` here too.

colintheshots: I updated the PR to rename all of the blocklist filenames with a simple replacement formula.

PR #4424, opened 2019-11-06
Revert "Updates taskcluster python package to no longer be locked to <5"

JohanLorenzo:
Reverts mozilla-mobile/focus-android#4409

This patch causes failures like https://tools.taskcluster.net/groups/EnCf1saBRCqX7vK-2_EQig/tasks/EnCf1saBRCqX7vK-2_EQig/runs/0/logs/public%2Flogs%2Flive_backing.log, which prevents v8.0.24 from being shipped. Here's why:

a. While modified `tools/docker/Dockerfile` hasn't been published to dockerhub.
b. Therefore, the decision task is calling `taskcluster.Queue({ 'baseUrl': 'http://taskcluster/queue/v1' })` but the taskcluster python client (baked in the docker image) is still at version 4. 
c. This client doesn't error out when an unknown key/value is passed (`baseUrl` has been supported since version 5)
d. Because it doesn't error out, the taskcluster python client opens a connection to the default taskcluster instance, which is `https://taskcluter.net` (note http**s** and **.net**). It should have used the proxy `http://taskcluster/` which does contain the scopes the decision task is given. Instead, the default instance doesn't have any scopes attached (which is expected).

This is why the decision task cannot spin a single task because it doesn't have any single scope.

The right fix would be to rebuild the docker image, but I've kept getting errors, yesterday. Thus, let's unblock the release by reverting the offending patch.

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -5,7 +5,7 @@
 # Inspired by:
 # https://hub.docker.com/r/runmymind/docker-android-sdk/~/dockerfile/
 
-FROM ubuntu:18.04
+FROM ubuntu:17.10
 
 MAINTAINER Sebastian Kaspari "skaspari@mozilla.com"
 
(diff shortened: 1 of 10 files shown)


jonalmeida:
> The changes look good. It seems TaskCluster failed because it's using the wrong target for UI tests and attempting to run WebView tests, an engine which no longer exists in Focus.

Wouldn't this still fail when we try to build again? (hopefully not)

PR #4426, opened 2019-11-07
Updates automation to use new provisioner and workers

mitchhentges:
With the taskcluster instance migration, we've updated the names of some of the workers. This PR modernizes `focus-android` accordingly

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -15,8 +15,8 @@ tasks:
     then:
       created: {$fromNow: ''}
       deadline: {$fromNow: '2 hours'}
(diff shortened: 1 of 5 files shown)


.taskcluster.yml
```diff
@@ -204,22 +204,22 @@ tasks:
         decision_task_id: {$eval: as_slugid("decision_task")}
         expires_in: {$fromNow: '1 year'}
         repository: https://github.com/mozilla-mobile/focus-android
-        scheduler_id: focus-nightly-sched
```
JohanLorenzo: I'm glad to see this scheduler_id go away \o/

PR #4432, opened 2019-11-08
Consistently use mobile-level-{level} scheduler

mitchhentges:
Some tasks were still using `taskcluster-github` as the scheduler


MihaiTabara:
Piggy-backed these changes in https://github.com/mozilla-mobile/focus-android/pull/4433/ so that I respin the Taskcluster CI.

PR #4434, opened 2019-11-12
Resolves permissions issue for master-push CI

mitchhentges:
This issue seemed weird to me, since `queue:create-task:highest:mobile-3/*` should be sufficient to perform `queue:create-task:mobile-3/b-linux`. In fact, if you have that `highest` scope and perform the physical act of creating a task on `b-linux`, it works! However, if you have that `highest` scope to create a task that needs the other scope (the one without `highest`), it'll fail!

[This is because Taskcluster evaluates scopes differently: when checking if you have a scope, it checks literally. When actually creating the task, it uses a "scope expression".](https://mozilla.slack.com/archives/GEZ4ZJCRM/p1573580169031500)

You can work around this by having the task `assume` from the role, rather than just declaring the scopes it wants

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -73,11 +73,13 @@ tasks:
         routes:
           - 'notify.irc-channel.#android-ci.on-any'
         scopes:
-          - queue:create-task:mobile-3/b-linux
-          - queue:scheduler-id:mobile-level-3
-          - secrets:get:project/focus/firebase
-          - secrets:get:project/focus/nimbledroid
-          - queue:route:notify.irc-channel.#android-ci.on-any
+          $let:
+            short_head_branch:
+              $if: 'event.ref[:11] == "refs/heads/"'
+              then: {$eval: 'event.ref[11:]'}
+              else: ${event.ref}
+          in:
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -73,11 +73,13 @@ tasks:
...
-          - secrets:get:project/focus/firebase
-          - secrets:get:project/focus/nimbledroid
-          - queue:route:notify.irc-channel.#android-ci.on-any
+          $let:
+            short_head_branch:
+              $if: 'event.ref[:11] == "refs/heads/"'
+              then: {$eval: 'event.ref[11:]'}
+              else: ${event.ref}
+          in:
+            - 'assume:repo:focus-android:branch:${short_head_branch}'
```
MihaiTabara: Hm, I'm not sure if this is needed. I think we only need `branch:master`.

mitchhentges: This operates on any `master-push`, and the command checks out `event.after`, so I think that it's more general than just `master`

MihaiTabara: Sounds good 👍

PR #4447, opened 2019-12-19
Bug 1602861 - Add empty .cron.yml to allow the general cron task to run

JohanLorenzo:
r? @tomprince


JohanLorenzo:
@colintheshots I notice the required check isn't set to `FirefoxCI (pull_request)` but `Firefox CI (push) instead. The latter won't work on PRs, it's just for pushes. Can you change this Github config? Otherwise, I don't think anybody will be able to land PRs without forcing them with admin privileges 

image

PR #4478, opened 2020-03-09
Update to AC 35.0.0 and GV 74.0.20200305014626

jonalmeida:
Includes API upgrades and a desktop mode viewport fix that was missing
from the previous upgrade.

Minor version bump to signify a GV version upgrade. This resets it
back to the old versioning system instead of using the patch version.


jonalmeida:
We need a new GV build, so holding of this PR.

PR #4491, opened 2020-03-19
Resolves #2787: Let the user choose where to save files

### LINKED ISSUE #2787: Let the user choose where to save files

Instead of saving pictures in the Download folder, Focus creates a new folder called Pictures. This way WhatsApp can't find the images and I have to find it manually.

### What / Requirements

An option that let the user choose where to save the downloads with the download folder as the default location.

--> when it's possible to save everything from the web anywhere.

AyanamiSean:
Resolves #2787,
We add a new activity (directory chooser) when the users tries to download a file so that the user can select the directory to save the file. We replace the download object in onFinishDownLoadDialog of BrowwserFragement by a new download object whose directory is updated to the path that the user chose. The path is getting from onActivityResult and we store the path inside the Path object we created.


AyanamiSean:
We also found a problem, there will be two copies of files downloaded, one is under the directory that the user chose, the other one is under the Download directory, seems like the one under Download is a shortcut of the other one, since if we delete either file, the other one is also gone.

PR #4492, opened 2020-03-26
Resolves #3808: Consider search shortcuts in the URL bar

### LINKED ISSUE #3808: Consider search shortcuts in the URL bar

There are many kinds/types/categories of things that I search on the web: recipes, movies, song lyrics, products I want to buy, etc. When I am looking for products, for example, I like to use Amazon’s search function. It gives me a better search result than Google can, and I can then buy the product right from the website.

This applies to other topics. For research, I prefer Wikipedia. For lyrics, I prefer Genius. For movies, I prefer IMDb.

I still prefer Google for my everyday searches, but I also want a faster way to quickly search on these specific websites if I have specific queries

### What / Requirements

The ability to specify _where_ I want to search, without changing my default search engine.

The user can get to their answers faster than if they’re using Google, Yahoo or Bing.

### Resources

DuckDuckGo has [!Bangs](https://duckduckgo.com/bang)

Firefox has [search shortcuts](https://venturebeat.com/wp-content/uploads/2018/10/firefox-63-search-shortcut.jpg?resize=1024%2C512&strip=all), and we can probably follow our own approach of using the `@` character. The metaphor is obvious: just as you `@` somebody to mention them in a conversation, you `@` a site to mention them in a search. It’s like saying “@amazon, I want to search for Harry Potter”.

mathu97:
Resolves #3808 

This is an implementation for search shortcuts via the URL bar. It essentially mimics the way firefox desktop allows for search shortcuts using the `@` character in the URL bar. 

Demo

Looking for some feedback, so we could maybe add this feature into Focus!


mathu97:
@brampitoyo, would love to get your feedback on this!

brampitoyo:
@mathu97 I was thinking, why require the user to type the `@` key before giving them the search engine? Can we show those search shortcuts without them having to type anything?

In other words, perhaps the design we currently have on Firefox Preview, might be interesting to try on Firefox Focus? What do you think?

13ABEL:
We considered that route, but there were some UX issues we had. Perhaps you can help?

Preview opens with a "home" page that displays your current tabs, shortcuts, etc. You have to press the search option on the bottom to open the search fragment which displays the search suggestions.

Focus opens with the search fragment by default and displays a short snippet in the screen. It's only when text is entered that the search suggestions will be displayed. We're currently not sure what the optimal UX would be to display the search options

Preview home screen:
91443275_143410840433276_1120421160660500480_n

Search screen on Preview:


Home screen on Focus:
90817461_271558937206594_6440591085721878528_n

Search screen on Focus:
91397980_510523472973435_4594580572436692992_n

app/src/main/java/org/mozilla/focus/utils/UrlUtils.java
```diff
@@ -17,8 +17,52 @@
...
+    public static String[] splitShortcutFromQuery(String query) {
+        String[] splitQuery = query.split(" ", 2);
+        if (splitQuery.length == 2 && splitQuery[0].startsWith("@")){
+            return splitQuery;
+        }
+
+        return null;
+    }
+
+    public static SearchEngine getSearchEngine(Context context, String query) {
```
13ABEL, line 55: should change query to suggestion text
Will do this morning

brampitoyo:
What if we make the search shortcuts:

1. Non-blocking. The homescreen tips will still show up, but so will search shortcuts.
2. Re-accessible. If, after you type a query, you’d like to modify the engine.

So when Focus starts, show a small list of search engine:



- If you start typing straight away, your default search provider will be used.
- If you tap a different search engine, the icon will change and the small menu will go away.

If you’d like to change a search engine in the middle of typing a keyword, you can do so easily by tapping the search provider icon again. A small downwards arrow provides an affordance..



The mockup isn’t perfect and doesn’t reflect Focus UI today, but I hope it can be an inspiration?

13ABEL:
We're currently a bit tied up with exams hence the lack of updates but we'll see if we can pick it up again after they're completed :)

ekager:
Hey everyone! Thank you so much for wanting to help out on Focus and this feature looks really cool! Unfortunately, we can't accept feature PRs to Focus at this time BUT we think we could use your help over on Fenix/Android-Components to implement a similar feature (and hopefully a lot of the work could carry-over) https://github.com/mozilla-mobile/fenix/issues/794

If ya'll are interested, we'd love to have you over on those projects whenever you have some time! Find us on Matrix if you want to chat https://wiki.mozilla.org/Matrix
https://github.com/mozilla-mobile/fenix#getting-involved

Good luck on exams!!

PR #4513, opened 2020-05-07
Update to AC 40.0.0 and GV 76.0.20200429185419

jonalmeida:


st3fan:
Thank you for picking this up. Are you also able to turn this into a release for QA?

jonalmeida:
> Thank you for picking this up. Are you also able to turn this into a release for QA?

Yes, I can manage to do that.

PR #4520, opened 2020-05-22
Update Android Gradle Plugin to 3.6.3

rciovati:
This PR updates the Android Gradle Plugin to version 3.6.3 (see the [release notes](https://developer.android.com/studio/releases/gradle-plugin#3-6-0)). 

This new version, in addition to build speed improvements, changes the way native libs are packaged, hopefully reducing download size and app install size:

Screenshot 2020-05-22 at 12 36 52

I did a small benchmark using [apkanalyzer](https://developer.android.com/studio/command-line/apkanalyzer) and it seems that the difference is not significant, but I'm not sure how accurate is the tool compared to the optimizations performed by Google Play in terms of compression.

| type | AGP 3.5 | AGP 3.6 |
| --- | --- | --- |
| download size | 54906074 | 54824394 |
| file size | 58154903 | 58068947 | 

(this one is for the `aarch64` variant).

---

In order to upgrade I had to upgrade Gradle as well as the old version is not supported by the new plugin.
I took the change to remove kapt as it's not used and cleaup some Kotlin flags that for some reasons were throwing a warning with the new plugin.


build.gradle
```diff
@@ -4,8 +4,7 @@ buildscript {
     ext.architecture_components_version = '2.1.0'
     ext.support_libraries_version = '1.0.0'
     ext.espresso_version = '3.1.0-alpha4'
-    ext.kotlin_version = '1.3.61'
-    ext.coroutines_version = '1.3.0'
+    ext.kotlin_version = '1.3.72'
```
rciovati, line 7: the latest kotlin version was already brought transitively

app/src/main/java/org/mozilla/focus/utils/Debounce.kt
```diff
@@ -22,8 +21,7 @@ import kotlinx.coroutines.launch
  * @param channel the channel to debounce
  * @return a throttled channel
  */
-@ObsoleteCoroutinesApi
-@ExperimentalCoroutinesApi
+@OptIn(ExperimentalCoroutinesApi::class)
```
rciovati, line 24: Using `@OptIn(ExperimentalCoroutinesApi::class)` here instead of `@ExperimentalCoroutinesApi` so that you don't have to propagate it in the entire call chain.

build.gradle
```diff
@@ -15,8 +14,8 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.3'
-        classpath 'de.mannodermaus.gradle.plugins:android-junit5:1.4.2.1'
+        classpath 'com.android.tools.build:gradle:3.6.3'
+        classpath 'de.mannodermaus.gradle.plugins:android-junit5:1.6.2.0'
```
rciovati, line 18: the old version is not compatible with AGP 3.6

PR #4533, opened 2020-06-03
Update to AC 44.0.1 and GV 77.0.20200602142841

jonalmeida:


jonalmeida:
It seems like we need to fix our latest a-c release. Once that's merged, we can merge https://github.com/mozilla-mobile/focus-android/pull/4523 and bump the Focus version.

Issue #4532, opened 2020-06-03
Add feature-prompts component


pocmo:
To replace custom prompts code.

pocmo:
I assume we cannot easily make the engine work with the custom prompts code in Focus. So we may need to integrate the component?

PR #4562, opened 2020-06-14
Issue #4532: Add feature-prompts component

jonalmeida:
Depends on a newer a-c version that includes this fix: https://github.com/mozilla-mobile/android-components/pull/7379

We can still merge it since it isn't a breaking API and upgrade the a-c version after our next release.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -11,6 +11,8 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" /> <!-- Used by sentry. -->
     <uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
+    <!-- Needed to prompt the user to give permission for camera access -->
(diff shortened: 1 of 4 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -11,6 +11,8 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" /> <!-- Used by sentry. -->
     <uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
+    <!-- Needed to prompt the user to give permission for camera access -->
```
jonalmeida, line 14: nit: Needed to prompt the user to give permission when uploading images from the camera.

pocmo: I wonder if we can make this an optional feature in the component? Giving camera permission to Focus may look odd? Maybe file a follow-up? 🤔

jonalmeida: Yeah, that's a good point. I filed https://github.com/mozilla-mobile/android-components/issues/7399 to investigate using a runtime permission instead.

PR #4568, opened 2020-06-17
Issue #4553: Integrate feature-downloads component.

pocmo:
This replaces the downloads code in Focus with the `feature-donwloads` component.

The download is working, the notification is showing and once finished a snackbar appears in the application. However the snackbar shows `null` as the filename (and has a file path with the file name `null` too) and can't open the file. I compared the implementation with Fenix and can't really see a difference that could cause that. @jonalmeida @Amejia481 Do you have any idea why that may happen?


Amejia481:
@pocmo  the null filename was a regression from https://github.com/mozilla-mobile/android-components/commit/c59ff6f1597c5f81ed80ca902d1b767675b631d6, it was fixed on https://github.com/mozilla-mobile/fenix/issues/9046 (AC 45.0.0). If the issue persist please let me know :)

pocmo:
Ah, great, thank you, in that case I'll try updating to the latest AC version here. :)

pocmo:
Yep, that fixed all problems. Updated the PR. :)

PR #4570, opened 2020-06-17
Get rid of (some) view references in BrowserFragment

pocmo:


app/src/main/java/org/mozilla/focus/exceptions/ExceptionsListFragment.kt
```diff
@@ -118,9 +119,7 @@ open class ExceptionsListFragment : Fragment(), CoroutineScope {
         removeAllExceptions.setOnClickListener {
             requireComponents.trackingProtectionUseCases.removeAllExceptions()
 
-            // val domains = ExceptionDomains.load(context!!)
-            // TelemetryWrapper.removeAllExceptionDomains(domains.size)
-            // ExceptionDomains.remove(context!!, domains)
+            TelemetryWrapper.removeAllExceptionDomains()
```
jonalmeida, line 122:
```suggestion
            val exceptions = (exceptionList.adapter as DomainListAdapter).selection()
            TelemetryWrapper.removeAllExceptionDomains(exceptions.count)
```

A bit hacky, but similar to the `ExceptionsRemoveFragment`, we could try to use the adapter size to submit in event.

pocmo: I like that! I'll do that in a follow-up. :)

PR #4576, opened 2020-06-22
Continue with migration to engine component

pocmo:


app/src/main/java/org/mozilla/focus/browser/integration/FullScreenIntegration.kt
```diff
@@ -0,0 +1,104 @@
...
+        feature.start()
+    }
+
+    override fun stop() {
+        feature.stop()
+    }
+
+    private fun fullScreenChanged(enabled: Boolean) {
+        if (enabled) {
+            toolbarView.visibility = View.GONE
```
jonalmeida, line 46: Setting the `toolbarView` to `GONE` shows a grey bar at the top of the screen and `setExpanded` no longer works.

Screen Shot 2020-06-22 at 10 02 25 PM


I think we need to just call `setExpanded` to hide the toolbar, and then disable the toolbar behaviour when in fullscreen mode.

Not sure if you had a chance to take a look at this patch that does this: https://github.com/jonalmeida/focus-android/commit/6ef407097535917e73a22d64d29c08aad33632ed

Maybe there are other ways though?

app/src/main/java/org/mozilla/focus/browser/integration/FullScreenIntegration.kt
```diff
@@ -0,0 +1,104 @@
...
+
+            toolbarView.setExpanded(false, true)
+            statusBar.visibility = View.GONE
+
+            engineView.setDynamicToolbarMaxHeight(0)
```
jonalmeida, line 51: We should remove this line (and in the else-branch) as well since it only helps when we have a toolbar at the bottom of the screen. Otherwise, it causes weird web content bugs with bottom toolbars like this one on youtube:

Screen Shot 2020-06-22 at 10 28 22 PM

PR #4591, opened 2020-07-08
Use setPrimaryClip instead of accessing primaryClip directly

### LINKED ISSUE #4590: Deprecated use of primaryClip property of ClipboardManager
## Steps to reproduce
Build with platforms;30 and buildtools;30.0.0
### Expected behavior
Build without error
### Actual behavior
Build breaks due to deprecated use of primaryCli property of ClipboardManager
### Device information
nil
* Android device: ?
* Focus version: ?

fgndev:
Using Android platform 30 and build-tools version 30.0.0 the build breaks with a deprecation message as the primaryClip property is set directly instead of using the setPrimaryClip method
fixes #4590


jonalmeida:
Thanks for the PR! This was fixed as part of the upgrade to 29. Will close this for now.

PR #4649, opened 2020-11-09
Update to Android Components 63.0.0 and GeckoView 82

jonalmeida:
- The `service-telemetry` component is now removed so we pin the
 version to the last released version. I opted for pinning instead of
 removing the component until we know what we want to remove/replace it
 with in the future. We have already disabled the GV integration as well
 and pinning just makes the diff smaller.
 - Added the GeckoSession to the `SessionExtensions` hashmap now that
 GeckoSession is no longer Parcelable and Focus relies on holding the
 in-memory instance of the session when using the 'open-in' feature.

The questionable part here is holding a weak ref to the GeckoSession which we were previously putting into the `Bundle` and ignoring the `about:blank` loading in the `GeckoWebViewProvider`.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -164,7 +164,7 @@ dependencies {
     implementation "org.mozilla.components:feature-customtabs:$mozilla_components_version"
     implementation "org.mozilla.components:lib-crash:$mozilla_components_version"
     implementation "org.mozilla.components:lib-fetch-httpurlconnection:$mozilla_components_version"
-    implementation "org.mozilla.components:service-telemetry:$mozilla_components_version"
+    implementation "org.mozilla.components:service-telemetry:$mozilla_components_version_telemetry"
     implementation "org.mozilla.components:service-fretboard:$mozilla_components_version", {
         exclude group: 'android.arch.work', module: 'work-runtime'
     }
--- a/app/src/main/java/org/mozilla/focus/ext/Session.kt
+++ b/app/src/main/java/org/mozilla/focus/ext/Session.kt
@@ -6,6 +6,7 @@ package org.mozilla.focus.ext
 
 import android.os.Bundle
 import mozilla.components.browser.session.Session
+import org.mozilla.geckoview.GeckoSession
 import java.util.WeakHashMap
 
 // Extension methods on the Session class. This is used for additional session data that is not part
@@ -25,9 +26,23 @@ private fun getOrPutExtension(session: Session): SessionExtension {
 }
 
 private class SessionExtension {
+    var savedGeckoSession: GeckoSession? = null
     var savedWebViewState: Bundle? = null
     var shouldRequestDesktopSite: Boolean = false
 }
+
+/**
+ * Saving the Gecko Session attached to the component session. We need this now because we can no
+ * longer serialize a gecko session into a [Bundle] as we did before.
+ *
+ * Temporary solution until we can use the browser-engine component.
+ *
+ * Component upstream issue:
+ * https://github.com/mozilla-mobile/android-components/issues/408
+ */
+var Session.savedGeckoSession: GeckoSession?
+    get() = getOrPutExtension(this).savedGeckoSession
+    set(value) { getOrPutExtension(this).savedGeckoSession = value }
 
 /**
  * Saving the state attached ot a session.
--- a/app/src/main/java/org/mozilla/focus/web/GeckoWebViewProvider.kt
+++ b/app/src/main/java/org/mozilla/focus/web/GeckoWebViewProvider.kt
@@ -622,9 +622,10 @@ class GeckoWebViewProvider : IWebViewProvider {
                     geckoSession.open(geckoRuntime!!)
                 }
                 setSession(geckoSession)
-            } else {
+            } else if (restored) {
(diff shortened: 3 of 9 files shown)


jonalmeida:
Will update to GV 83 when that version is available with the latest AC version.

app/src/main/java/org/mozilla/focus/ext/Session.kt
```diff
@@ -25,10 +26,24 @@ private fun getOrPutExtension(session: Session): SessionExtension {
...
 }
 
+/**
+ * Saving the Gecko Session attached to the component session. We need this now because we can no
+ * longer serialize a gecko session into a [Bundle] as we did before.
+ *
+ * Temporary solution until we can use the browser-engine component.
+ *
+ * Component upstream issue:
+ * https://github.com/mozilla-mobile/android-components/issues/408
```
pocmo, line 41: I hope we can soon :(

app/src/main/java/org/mozilla/focus/web/GeckoWebViewProvider.kt
```diff
@@ -597,7 +605,7 @@ class GeckoWebViewProvider : IWebViewProvider {
 
         override fun restoreWebViewState(session: Session) {
             val stateData = session.savedWebViewState!!
-            val savedSession = stateData.getParcelable<GeckoSession>(GECKO_SESSION)!!
+            val savedSession = session.savedGeckoSession!!
```
pocmo, line 608: Does this also work with "don't keep activities"? 🤔

app/src/main/java/org/mozilla/focus/web/GeckoWebViewProvider.kt
```diff
@@ -636,8 +644,8 @@ class GeckoWebViewProvider : IWebViewProvider {
         }
 
         override fun saveWebViewState(session: Session) {
+            session.savedGeckoSession = geckoSession
             val sessionBundle = Bundle()
-            sessionBundle.putParcelable(GECKO_SESSION, geckoSession)
```
csadilek, line 640: Might be missing something, but wouldn't we have to listen to `onSessionStateChange` and still add the `SessionState.toString` to the Bundle and restore it later? Otherwise we may lose the session on pause/resume...

jonalmeida: From what I read [here](https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/GeckoSession.java#2458-2479), the `SessionState` only stores the history, form and scroll data for the session. Focus wants to hold on to state temporarily when clicking the 'Open in Focus' in custom tabs and when putting the app into the background. Adding the whole gecko session to the `SessionExtensions` seemed to be a better solution to hold the session without needing to add/restore the `SessionState`.

The existing way for this is with the `SessionExtensions` where the session was stored in a bundle. I added the `savedGeckoSession` item in the `SessionExtensions` without the serializing.

Sebastian raised a concern about that above about losing this session when doing this that I have to address which may change this.

app/src/main/java/org/mozilla/focus/ext/Session.kt
jonalmeida: 🤞

app/src/main/java/org/mozilla/focus/web/GeckoWebViewProvider.kt
jonalmeida: Hmm.. in a Custom Tab this fails gracefully by showing the Focus home screen, but on the main browser it doesn't. Looking into this..

jonalmeida: I tested this latest change over the week~end~, and with "don't keep activities" there is something wacky that happens when restoring the session where the session is correctly restored first, but our `NestedGeckoView` is created after and it holds a default state which then overwrites what the UI sees.

I'd say this is more reason to hop onto the AC Engine now instead of trying to fix this.

app/build.gradle
```diff
@@ -164,7 +164,7 @@ dependencies {
     implementation "org.mozilla.components:feature-customtabs:$mozilla_components_version"
     implementation "org.mozilla.components:lib-crash:$mozilla_components_version"
     implementation "org.mozilla.components:lib-fetch-httpurlconnection:$mozilla_components_version"
-    implementation "org.mozilla.components:service-telemetry:$mozilla_components_version"
+    implementation "org.mozilla.components:service-telemetry:$mozilla_components_version_telemetry"
```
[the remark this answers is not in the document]
jonalmeida: I didn't see any dependencies upgrades because of this one.

jonalmeida (commented):
I've updated the PR that fixes the restore state when the activity is re-created, but it is admittedly fickle we could definitely argue that landing the AC engine upgrades is more valuable.

jonalmeida:
If the CI build is fine, I'd like to land this as-is and ask QA if they can have a round of testing on it.

jonalmeida:
Force merging because Travis CI is disabled so this will be pending forever.

PR #4660, opened 2020-11-30
Update Dockerfile to pull in newer SDK

st3fan:
This patch to the `Dockerfile` pulls in the latest Android SDK and patches things up a bit so that the image still builds.

I tested this by running the following command:

```
docker build -t mozilla-mobile/focus-android:test .
```

@pocmo can you help to push this to the Docker Hub? Do we still have that ability?

Talking about Docker Hub - it is unclear which version of the image correspond to what branch/revision in Git. I think we should probably just create this as `1.5`.

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -11,6 +11,11 @@ MAINTAINER Sebastian Kaspari "skaspari@mozilla.com"
 
 # -- System -----------------------------------------------------------------------------
 
+# Because this image is based on a very old Ubuntu, we have to grab the index for
+# an alternative source.
+RUN sed -i -e 's/archive\.ubuntu\.com/old-releases\.ubuntu\.com/' /etc/apt/sources.list
+RUN sed -i -e 's/security\.ubuntu\.com/old-releases\.ubuntu\.com/' /etc/apt/sources.list
+
 RUN apt-get update -qq
 
 RUN apt-get install -y openjdk-8-jdk \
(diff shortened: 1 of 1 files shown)


tools/docker/Dockerfile
```diff
@@ -47,12 +55,14 @@ RUN echo y | android update sdk --no-ui --all --filter platform-tools | grep 'pa
 RUN echo y | android update sdk --no-ui --all --filter android-26 | grep 'package installed'
 RUN echo y | android update sdk --no-ui --all --filter android-27 | grep 'package installed'
 RUN echo y | android update sdk --no-ui --all --filter android-28 | grep 'package installed'
+RUN echo y | android update sdk --no-ui --all --filter android-29 | grep 'package installed'
```
pocmo, line 58: We should clean that up at some point to create a smaller image and with that faster builds. In the past we always had to support the current and next SDK while we migrate over... but now we have a bunch of unused ones here.

In fact in our other repos we do not install any SDK manually anymore (and accept only the licenses) because the Gradle step later in this script will on itself figure out what needs to be downloaded nowadays.

pocmo (approved):

pocmo:
> @pocmo can you help to push this to the Docker Hub? Do we still have that ability?

Yeah, we have a Docker Hub `mozillamobile` org. Ping me on Slack with your username and I can add you. I can also try to build and upload here today. But I actually do not have Docker installed on that machine anymore. :)

> Talking about Docker Hub - it is unclear which version of the image correspond to what branch/revision in Git. I think we should probably just create this as `1.5`.

In `.taskcluster.yml` you'll find multiple `image` sections that reference the image to be used for specific tasks, e.g.:
https://github.com/mozilla-mobile/focus-android/blob/master/.taskcluster.yml#L25

pocmo:
Looks like parts of it are still documented here:
https://github.com/mozilla-mobile/shared-docs/blob/master/android/docker_guide.md

pocmo:
Alright, building locally now...

pocmo:
I ran into a build error twice. If it succeeded for you then maybe let's just upload your build later today. :)

PR #4675, opened 2020-12-09
Bump version to 8.11.0

jonalmeida:


jonalmeida:
Version bump, doesn't require CI to pass.

PR #4690, opened 2021-01-19
Version bump to 8.12.0

jonalmeida:
Using version 8.12.0 since this is a major release.


jonalmeida:
Force merging as well since this is just a version bump.

PR #4709, opened 2021-03-12
Add configuration for enabling mergify

pocmo:


jonalmeida (approved):

jonalmeida:
Somewhat related: Do we need the `FirefoxCI (pull_request)`? It looks like it does only a build task, but the `FirefoxCI (push)` does linting, UI tests, etc.

jonalmeida:
Oh hrmm.. one looks like it's for contributors and the other is for members. Still duplicate work for us right now.

pocmo:
> Somewhat related: Do we need the `FirefoxCI (pull_request)`? It looks like it does only a build task, but the `FirefoxCI (push)` does linting, UI tests, etc.

The pull request doesn't do that? Then we definitely should change that and make that task do all the things too. "Push" is only for pushes to *this* repository.

pocmo:
It looks like the one task does all the things here:
https://github.com/mozilla-mobile/focus-android/blob/master/.taskcluster.yml#L37

PR #4711, opened 2021-03-12
Update Kotlin and Coroutine version

jonalmeida:
Keeping this in sync with Android Components.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -135,7 +135,10 @@ android {
 }
 
 tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
-        kotlinOptions.allWarningsAsErrors = true
+        kotlinOptions {
+            allWarningsAsErrors = true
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/browser/binding/AbstractBinding.kt
```diff
@@ -21,6 +22,7 @@ abstract class AbstractBinding(
 ) : LifecycleAwareFeature {
     private var scope: CoroutineScope? = null
 
+    @OptIn(ExperimentalCoroutinesApi::class)
```
pocmo, line 25: Alternatively we could also globally optin to the API and wouldn't need to add those annotations everywhere:
https://github.com/mozilla-mobile/android-components/blob/b266ac3d16e15cabc969cfe33c9cf965f557e257/components/feature/prompts/build.gradle#L25-L29

pocmo (approved):
👍

PR #4712, opened 2021-03-12
BrowserFragment: Start to replace Session usage with browser-state

pocmo:
Slowly starting to replace `Session` and `SessionManager` usage in Focus.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
@@ -34,8 +34,8 @@ import com.google.android.material.snackbar.Snackbar
 import kotlinx.android.synthetic.main.fragment_browser.*
 import mozilla.components.browser.session.Session
 import mozilla.components.browser.session.SessionManager
-import mozilla.components.browser.state.selector.findTab
 import mozilla.components.browser.state.selector.findTabOrCustomTab
+import mozilla.components.browser.state.state.CustomTabConfig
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -383,7 +393,7 @@ class BrowserFragment :
     }
 
     private fun initialiseCustomTabUi(view: View) {
-        val customTabConfig = session.customTabConfig!!
+        val customTabConfig = tab.ifCustomTab()!!.config
```
jonalmeida, line 396:
```suggestion
        val customTabConfig: CustomTabConfig = tab.ifCustomTab()?.config ?: return
```

nit: early return, maybe?

jonalmeida (approved):
✅

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
pocmo: Yeah, this whole code is not nice. At this point we already know that it's a custom tab, so having to deal with nullability here is annoying. Hm, actually, maybe I can just pass in the `CustomTabConfig` if the check happens outside, I'll try that and update the PR.

jonalmeida: It shows up now, so I fixed the conflict in the PR and we can merge it.

PR #4715, opened 2021-03-15
Replace SessionManager usage with calling use cases.

pocmo:
Not all of them yet... but a good start. :)

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
@@ -61,7 +61,6 @@ import org.mozilla.focus.utils.StatusBarUtils
 import org.mozilla.focus.utils.SupportUtils
 import org.mozilla.focus.utils.UrlUtils
 import org.mozilla.focus.utils.ViewUtils
-import org.mozilla.focus.utils.createTab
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -760,12 +759,15 @@ class UrlInputFragment :
                 .remove(this)
                 .commit()
         } else {
-            val session = createTab(url, source = SessionState.Source.USER_ENTERED)
+            val tabId = requireComponents.tabsUseCases.addPrivateTab(
```
jonalmeida, line 762: If we use `addPrivateTab` here then we won't be able to apply/persist tracking protection exceptions? Maybe the `createTab` function or an alternative should go here?

jonalmeida (approved):
Looks good! One question though related temporary TP exception solution.

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
pocmo: In the next PR I am changing `createTab()` to set  `private = true` and I would like to step by step get rid of `createTab()` and replace it by the use cases. This will temporarily break persisted TP exceptions until we get the GV patch, but it will get us closer to the state we want to be in.

PR #4716, opened 2021-03-15
Replace NotificationSessionObserver with Middleware.

pocmo:


jonalmeida (approved):
✅ 

Future perf: calling `startService` for all the actions we're filtering in the middleware may have a perf impact.

From: https://developer.android.com/reference/android/content/Context#startService(android.content.Intent)

> Note: Each call to startService() results in significant work done by the system to manage service lifecycle surrounding the processing of the intent, which can take multiple milliseconds of CPU time. Due to this cost, startService() should not be used for frequent intent delivery to a service, and only for scheduling significant work. Use bound services for high frequency calls.

PR #4722, opened 2021-03-16
Migrate tab sheet to use BrowserStore instead of SessionManager.

pocmo:

### DIFF
--- a/app/src/main/java/org/mozilla/focus/session/ui/TabViewHolder.kt
+++ b/app/src/main/java/org/mozilla/focus/session/ui/TabViewHolder.kt
@@ -6,40 +6,39 @@ package org.mozilla.focus.session.ui
 
 import android.animation.Animator
 import android.animation.AnimatorListenerAdapter
-import androidx.recyclerview.widget.RecyclerView
 import android.view.View
 import android.widget.TextView
-import mozilla.components.browser.session.Session
+import androidx.recyclerview.widget.RecyclerView
+import mozilla.components.browser.state.state.TabSessionState
 import org.mozilla.focus.R
 import org.mozilla.focus.ext.beautifyUrl
 import org.mozilla.focus.ext.requireComponents
 import org.mozilla.focus.telemetry.TelemetryWrapper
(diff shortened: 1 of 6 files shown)


pocmo:
```
> Could not resolve all files for configuration ':app:lintClassPath'.
   > Could not download groovy-all.jar (org.codehaus.groovy:groovy-all:2.4.15)
      > Could not get resource 'https://jcenter.bintray.com/org/codehaus/groovy/groovy-all/2.4.15/groovy-all-2.4.15.jar'.
         > Could not GET 'https://jcenter.bintray.com/org/codehaus/groovy/groovy-all/2.4.15/groovy-all-2.4.15.jar'.
            > Connection reset
```

Let's try again...

app/src/main/java/org/mozilla/focus/session/ui/TabViewHolder.kt
```diff
@@ -6,40 +6,39 @@ package org.mozilla.focus.session.ui
...
     private val textView: TextView
 ) : RecyclerView.ViewHolder(textView), View.OnClickListener {
     companion object {
-        @JvmField
-        internal val LAYOUT_ID = R.layout.item_session
+        internal const val LAYOUT_ID = R.layout.item_session
     }
 
-    private var sessionReference: WeakReference<Session> = WeakReference<Session>(null)
+    private var tabReference: WeakReference<TabSessionState> = WeakReference<TabSessionState>(null)
```
jonalmeida, line 27: nit: TabSessionState probably doesn't need to be a weakref any more.

jonalmeida (approved):
📑 ✅ ✨

app/src/main/java/org/mozilla/focus/session/ui/TabViewHolder.kt
pocmo: Oh, yeah, right. I didn't even process that code. Looks like we only keep the tab around to select it later, and we can do that with just the tab ID now, so no need to keep more than that. :)

PR #4725, opened 2021-03-16
Close #4723: Use CustomTabsService from Components

### LINKED ISSUE #4723: Add feature-customtabs component
We can move more of Session Manager out of the activity at the same time.

jonalmeida:


app/src/main/java/org/mozilla/focus/activity/CustomTabActivity.kt
```diff
@@ -15,12 +18,15 @@ import org.mozilla.focus.ext.components
 class CustomTabActivity : MainActivity() {
     private lateinit var customTabId: String
 
-    private val customTabSession: Session by lazy {
-        components.sessionManager.findSessionById(customTabId)
+    private val customTabSession: SessionState by lazy {
+        components.store.state.findTabOrCustomTab(customTabId)
```
pocmo, line 22: Afaik this should be `findCustomTab()` and then we could use the `CustomTabSessionState` type? 🤔

app/src/main/java/org/mozilla/focus/customtabs/CustomTabsService.kt
```diff
@@ -0,0 +1,11 @@
+package org.mozilla.focus.customtabs
```
pocmo, line 1: nit: license header

pocmo (approved):

pocmo:
I just saw that:
* I had added `findCustomTab()` to a local branch too.
* You are on PTO for the rest of the week

So I'll:
* Merge as-is
* Open a PR with follow-up changes since my changes will cause a conflict anyways :)

app/src/main/java/org/mozilla/focus/activity/CustomTabActivity.kt
jonalmeida: I tried that initially, but when we use 'Open in Focus' we crash because the tab is no longer a custom tab.

jonalmeida: We could have started to use the `migrate` use case, but this started opening up work in MainActivity and thought we might have conflicts if you had started working on that already.

pocmo: I'll put up a patch for `CustomTabsUseCases.remove()` it should only remove actual custom tabs. It currently can also remove normal tabs, but only because it still uses `SessionManager` under the hood.

Issue #4540, opened 2020-06-04
Migration branch: Verify tracking protection is working as expected


pocmo:
~~https://mozilla.github.io/tracking-test/etp-test.html~~

From the Github README, we should use this page instead: https://senglehardt.com/test/trackingprotection/test_pages/

Amejia481:
While testing, I noticed that every time I changed the cookie setting, the preference `network.cookie.cookieBehavior` on `about:config` was not getting updated until I restarted the app, I put a PR for addressing the issue https://github.com/mozilla-mobile/focus-android/pull/4727.

Amejia481:
Apart from that, I checked and tested the tracking protection settings and it's working as expected, great job 👏🏽 .

There are other privacy feature that we could include as future improvements: 

- **Fingerprinting**: (https://github.com/mozilla-mobile/focus-android/issues/4550)
- **Cryptominers**: (https://github.com/mozilla-mobile/focus-android/issues/4549)
- [**CookiePurging**](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Privacy/Redirect_tracking_protection): Fenix has checkbook for it (Redirect Trackers) defaults to `true` and desktop enables it without any UI. To alter it use [select(cookiePurging: Boolean = false)](https://github.com/mozilla-mobile/android-components/blob/90450d2f0cfb2e35a5af0e5f2a904fbf4446337a/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt#L422).
- [**MOZILLA_SOCIAL**](https://github.com/mozilla-mobile/android-components/blob/90450d2f0cfb2e35a5af0e5f2a904fbf4446337a/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt#L360): You could add it as part of the social toggle, as it adds some extra domains that are not in the [SOCIAL](https://github.com/mozilla-mobile/android-components/blob/90450d2f0cfb2e35a5af0e5f2a904fbf4446337a/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt#L336) category, it was created to augment the social category.

PR #4727, opened 2021-03-17
For issue #4540: Update tracking protection policy when cookie setting change

Amejia481:
When we change the cookie setting, users are not able to see changes applied until they restart the app, with this fix every time users change the cookie setting they will be reflected instantly.


pocmo (approved):
Thank you! 👍

PR #4733, opened 2021-03-18
Removing SessionManager and Session references

pocmo:
Kind of work in progress. Haven't replaced all of them yet. Mostly tried to do only 1:1 replacements for now. But that sometimes feels weird and we may want to do some architectural changes eventually.


pocmo:
Pushed some more commits. With that `SessionManager` and `Session` should no longer be in use outside of the `Components` class.

jonalmeida (approved):
This honestly looks great! 👏 

While testing out the PR, I did discover an [unrelated bug](https://github.com/mozilla-mobile/focus-android/issues/4763) too.

PR #4754, opened 2021-03-22
Remove architecture product flavor and use APK split instead. And run…

### LINKED ISSUE #4543: Replace architecture specific build variants with split APK configuration

pocmo:
With this patch we no longer need to have different product flavors for the different CPU architectures and instead use APK splitting supported by the Gradle plugin. This also makes our tests run on Firebase again, they do not pass, but they run. :)

Fixes #4543.


tools/taskcluster/flank.yml
```diff
@@ -34,15 +33,13 @@ gcloud:
...
-      version: 26 
+    - model: blueline
+      version: 28
 
 flank:
+  project: moz-fx-mobile-firebase-testlab
+
   # test shards - the amount of groups to split the test suite into
   # set to -1 to use one shard per test.
   testShards: -1
```
AaronMT, line 45: ~20 tests or so, feel free to nudge this to 4 shards (5 tests per shard) here while the file is touched

pocmo: 👍 Updated it to `4`.

tools/taskcluster/flank.yml
```diff
@@ -34,18 +33,16 @@ gcloud:
     - /sdcard/screenshots
   performance-metrics: true
 
-  # test-targets will be pulled dynamically from package.  DO NOT MODIFY
-  test-targets:
-TEST_TARGETS
-
   device:
-    - model: sailfish 
-      version: 26
```
sv-ohorvath, line 43: Probably we'd also want num-flaky-test-attempts: 1, to repeat the test if it is failing intermittently, once we start running the tests in PRs

pocmo: Sounds good. Do you want me to update this in this PR? I hope we can land this soon and then iterate on the config as needed.

sv-ohorvath: Either way is fine by me. I can add it when I'm adding tests targets back

sv-ohorvath: Just saw that blueline is a physical device on Firebase's device list, and it might increase tests execution time by a lot, compared to a virtual device. Did you pick that for a reason or can I also change that, if it proves to be too slow?

tools/docker/Dockerfile
```diff
@@ -57,8 +57,8 @@ RUN git clone https://github.com/mozilla-mobile/focus-android.git
 
 # Build project and run gradle tasks once to pull all dependencies
 WORKDIR /opt/focus-android
-RUN ./gradlew assembleFocusAarch64Debug \
-  && ./gradlew testFocusAarch64DebugUnitTest \
+RUN ./gradlew assembleFocusDebug \
```
jonalmeida, line 60: Do we need to rebuild the Docker file for this?

jonalmeida (approved):
✅ LGTM

jonalmeida:
Guessing we're ignoring the failing UI tests with #295 filed for now.

PR #4767, opened 2021-03-24
Prevent non-private tabs and requests

pocmo:


jonalmeida (approved):
That's a pretty cool idea! ✨

PR #4769, opened 2021-03-25
Fix and add new Search UI tests. Add working UI tests back to flank.yml

sv-ohorvath:
Fixed search tests.
Added some test robots to organize shared code.
Disabled failing tests, left old UI tests that are working enabled, configured the flank file to keep them running on branches.
Changed the target device to virtual for lower costs (~a quarter of the cost for psychical devices).


tools/taskcluster/flank.yml
```diff
@@ -33,19 +37,23 @@ gcloud:
...
+    - model: Pixel2
+      version: 29
 
 flank:
   project: moz-fx-mobile-firebase-testlab
 
   # test shards - the amount of groups to split the test suite into
   # set to -1 to use one shard per test.
-  testShards: 4
+  max-test-shards: 50
```
AaronMT, line 53: I think we should use 4, right? ~20 tests currently, grouped into 4 shards, 5 tests per shard.

AaronMT (approved):
This is excellent 👍🏼 

One change on the Flank config above recommended

pocmo (approved):

tools/taskcluster/flank.yml
sv-ohorvath: @AaronMT I wanted to change that now, because as I add more tests, I don't want to update this file again. I'm not super-familiar with these, as I understand it: splits tests into the max num of shards (when we have 50 tests = 1 per shard), but if we don't have 50 tests, it won't create the shards not needed. I copied the max number from Fenix. Please correct me if I'm wrong.

sv-ohorvath: Discussed with Aaron on Slack, we'll stick to 4 shards for now.

PR #4775, opened 2021-03-29
Fix static shortcut title.

NightlyNexus:
This was showing "%1$s" to users on launchers that display the long title for the "Erase and open" action.

The string is used statically here: https://github.com/mozilla-mobile/focus-android/blob/b3160fa7df9c43bb25fb9840dad138fddb45e6e6/app/src/focusDebug/res/xml-v25/shortcuts.xml#L24

Screenshot


app/src/main/res/values-tr/strings.xml
```diff
@@ -30,7 +30,7 @@
   <string name="shortcut_erase_short_label">Sil</string>
   <string name="shortcut_erase_long_label">Gezinti geçmişini sil</string>
   <string name="shortcut_erase_and_open_short_label">Sil ve aç</string>
-  <string name="shortcut_erase_and_open_long_label">Sil ve %1$s’u aç</string>
```
NightlyNexus, line 33: Might be worth having a Turkish speaker check this. I'm not sure what the `’u` does.

pocmo (commented):
Hey! Thank you for your help fixing this issue. There's a specific process we need to follow to make this work with our translations system, which unfortunately means some changes in your PR are not necessary. :(

If we change a string then we only need to change it in `res/values/`. In addition to that we need to give it a new name. This will make it show up in our translation tooling as a string that needs changes in other languages. If there's no other logical name for the string then it's fine to just append `2`.

The translators will work on Pontoon: https://pontoon.mozilla.org/projects/ and the other string files will get updated automatically.

NightlyNexus:
That's great, thank you. I should have looked into how you do translations. I'll make the change shortly.

NightlyNexus:
Updated

pocmo (commented):
Thank you! 👍 

Our sync with Pontoon is currently not working and so I am holding back landing this. But I hope we can resolve this soon. :)

pocmo:
Okay, I think we are ready to land this.

- We need to resolve a conflict in `res/values/strings.xml`
- We wouldn't need to land the changes in `res/values-*/strings.xml` since those will come from the translators.

NightlyNexus:
I finally rebased and updated for whenever you have a minute to look.

pocmo:
Thank you!

We are close to a release and may not be able to land it right now - since there may not be enough time to translate it.

CC @jonalmeida

jonalmeida:
We've cut our 93 release branch, so we can safely review and land this for it to be picked up for translation as part of the next release cycle.

pocmo (approved):
Let's go!

PR #4776, opened 2021-03-29
Prototype: Use app-wide store for app state and use it for navigation.

pocmo:
This PR prototypes using an app-wide `Store` to keep the app's state independently from fragments/activities and use it for navigation.

We currently do have some bugs that are caused by the different fragments needing to coordinate global states like "the app is locked" and failing to pass that from one fragment to the other causes bugs like the app unlocking. Centralizing the state and how we change the screen based on that, makes it easier to debug and change this logic.

Using the store we can:
* Remember what state we are in and if needed "navigate" to display this state
* Can continue to use our different fragments and transactions without changing everything
* Centralize fragment transactions in one place
* Enhance it to keep more state, either screen-specific state or app-wide state
* Are future proof: We can render the state with fragments, but in the future we can call into `@Composable` functions instead of performing fragment transactions.

This PR doesn't cover everything yet (e.g. locking), but is far enough to use all basic functions of the app.


pocmo:
Added some more commits:
* Separating `MainActivity` and `CustomTabActivity`. Not extending `MainActivity` makes the code easier to follow.
* Changing the biometric code to use the new `AppStore`. Now locking the app is a global state that doesn't depend on the `BrowserFragment`.

app/src/main/java/org/mozilla/focus/state/AppAction.kt
```diff
@@ -0,0 +1,67 @@
...
+        val height: Int
+    ) : AppAction()
+
+    /**
+     * All tabs have been removed.
+     */
+    object NoTabs : AppAction()
+
+    /**
+     * The user finished edition the URL of the tab with the given [tabId].
```
jonalmeida, line 37: nit: typo 
```suggestion
     * The user finished editing the URL of the tab with the given [tabId].
```

app/src/main/java/org/mozilla/focus/activity/CustomTabActivity.kt
```diff
@@ -4,40 +4,69 @@
...
 import mozilla.components.support.utils.SafeIntent
+import org.mozilla.focus.R
 import org.mozilla.focus.ext.components
+import org.mozilla.focus.fragment.BrowserFragment
+import org.mozilla.focus.locale.LocaleAwareAppCompatActivity
 
 /**
  * The main entry point for "custom tabs" opened by third-party apps.
  */
-class CustomTabActivity : MainActivity() {
```
jonalmeida, line 16: Nice! 👏

app/src/main/java/org/mozilla/focus/state/AppState.kt
```diff
@@ -0,0 +1,69 @@
...
+     * @property tabId The ID of the displayed tab.
+     * @property showTabs Whether to show the tabs tray.
+     */
+    data class Browser(
+        val tabId: String,
+        val showTabs: Boolean
+    ) : Screen() {
+        // Whenever the showTabs property changes we want to treat this as a new screen and force
+        // a navigation.
+        override val id: String = "${super.id}_$showTabs}"
```
jonalmeida, line 51: This is an interesting way to change the ID. 😄

jonalmeida (approved):
I did a first pass of it and some testing, so far everything looks great!

PR #4717, opened 2021-03-15
Close #4528: Remove HttpUrlConnection and OkHttp usages

jonalmeida:
The `SearchSuggestionsFetcher` can be removed in the future when we decide to use the awesomebar feature, for now, this removes all the other client usages from production code.

Haven't been able to run tests on the patch because of some local gradle errors:

<details>

```
[Robolectric] org.mozilla.focus.settings.SearchEngineValidationTest.URL timing out is not valid: sdk=28; resources=BINARY
Downloading from maven 
Downloading: org/robolectric/android-all/Q-robolectric-5415296/android-all-Q-robolectric-5415296.pom from repository sonatype at https://oss.sonatype.org/content/groups/public/
Error transferring file: java.security.NoSuchAlgorithmException: Error constructing implementation (algorithm: Default, provider: SunJSSE, class: sun.security.ssl.SSLContextImpl$DefaultSSLContext)
[WARNING] Unable to get resource 'org.robolectric:android-all:pom:Q-robolectric-5415296' from repository sonatype (https://oss.sonatype.org/content/groups/public/): Error transferring file: java.security.NoSuchAlgorithmException: Error constructing implementation (algorithm: Default, provider: SunJSSE, class: sun.security.ssl.SSLContextImpl$DefaultSSLContext)
Downloading: org/robolectric/android-all/Q-robolectric-5415296/android-all-Q-robolectric-5415296.pom from repository central at http://repo1.maven.org/maven2
Error transferring file: Server returned HTTP response code: 501 for URL: http://repo1.maven.org/maven2/org/robolectric/android-all/Q-robolectric-5415296/android-all-Q-robolectric-5415296.pom
[WARNING] Unable to get resource 'org.robolectric:android-all:pom:Q-robolectric-5415296' from repository central (http://repo1.maven.org/maven2): Error transferring file: Server returned HTTP response code: 501 for URL: http://repo1.maven.org/maven2/org/robolectric/android-all/Q-robolectric-5415296/android-all-Q-robolectric-5415296.pom
Downloading: org/robolectric/android-all/Q-robolectric-5415296/android-all-Q-robolectric-5415296.jar from repository sonatype at https://oss.sonatype.org/content/groups/public/
Error transferring file: java.security.NoSuchAlgorithmException: Error constructing implementation (algorithm: Default, provider: SunJSSE, class: sun.security.ssl.SSLContextImpl$DefaultSSLContext)
[...]

PR #4783, opened 2021-04-01
Closes #4761: Integrate feature-search component.

### LINKED ISSUE #4739: Search engines can't be removed
## Steps to reproduce
1. Go to Settings>Search>SearchEngine;
2. In 'Installed search engines' dialog, click more options button and select Remove.
3. Select search engines to remove. -> click "Remove".

### Expected behavior
The search engine selected is removed from the list.

### Actual behavior
Selected search engines can't be removed.

### Device information
* Android device: Samsung Note 10 (Android 11), Google Pixel (Android 10)
* Focus version: dev build from master - #1186.0.1- 20210310152336

### LINKED ISSUE #4761: Replace custom search code with feature-search component
(This needs further investigation / prototyping to break down further)

pocmo:
This patch replaces the custom search code in Focus with our `feature-search` component.

Closes #4739
Closes #4761


pocmo:
For some reason detekt is crashing with a `ArrayIndexOutOfBoundsException`. I am seeing the same on my "app state prototype" branch. I'll try to update detekt on a separate branch and hopefully that will fix it.

app/src/main/java/org/mozilla/focus/search/SearchEngineWriter.kt
```diff
@@ -1,75 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
```
jonalmeida, line 1: Needs test to be deleted for CI to pass.

jonalmeida (commented):
The changes look good to me, but I ran into an issue.

While adding a new search engine it would fail with `SocketTimeoutException` when doing a network test of the search engine [here](https://github.com/mozilla-mobile/focus-android/blob/master/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.kt#L207).

This seemed related to the changes in https://github.com/mozilla-mobile/focus-android/pull/4717, however when I tried the same URL on a master build, it works fine and I don't know why. 🤔 

#### Steps to reproduce
1. Add a new search engine `https://www.google.com/search?q=%s`.
2. Click 'Save'.

pocmo:
I get an "unknown" error from GV in that situation. After that I also see a background crash and I can't browse websites. It works if I browse a page before going to settings. So I assume that in this situation GV gets initialized wrong somehow. I'll continue to investigate.

pocmo:
> So I assume that in this situation GV gets initialized wrong somehow. I'll continue to investigate.

This doesn't always work either. Interestingly `https://example.org/?q=%s` always works for me. So I wonder if this is somehow Google specific?

pocmo:
That seems to be the case... if I fake the `User-Agent` then it works...

pocmo:
@jonalmeida Can you take another look? Since it's only Google I think we can move this to a new issue and since Google is already in the list, I am thinking it has a super low priority probably.

jonalmeida (approved):
Looks great! All my custom search engines were migrated too. 🎉

PR #4786, opened 2021-04-01
Resolves #4785: Explicitly set extractNativeLibs=true in app manifest

### LINKED ISSUE #4785: Explicitly set extractNativeLibs=true in Focus manifest
Similarly to https://github.com/mozilla-mobile/fenix/issues/18381, I'd like to land this in Focus as a speculative crash fix.

dblohm7:
This is a speculative GV crash fix, identical to what was landed in Fenix.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -25,6 +25,7 @@
     
     <application
         android:allowBackup="false"
+        android:extractNativeLibs="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:supportsRtl="true"


jonalmeida:
@dblohm7 since this is a speculative fix that you would want to monitor the results, did you want a specific Focus release for it or will our next release on merge day (~2021-04-19) be fine? We need to land this patch on the `release` branch while our main branch is going through some rapid development that isn't ready to go out yet.

Alternatively, we can see if the issue is fixed on Fenix since that change should be in nightly already.

dblohm7:
Next release is fine. I'm still monitoring the Fenix results but so far it's been too soon to tell.

pocmo (approved):
✅ since it is already shipping in Fenix.

PR #4789, opened 2021-04-07
MainActivityFirstrunTestRule: Skip first run screen.

pocmo:
The state of `AppStore` is determined on application start and therefore before `beforeActivityLaunched()` runs. So if we are in the `FirstRun` screen state then we skip that. We still set the preference additionally for further app launches.


pocmo:
There are still failures on Firebase, even though that worked locally. I'll investigate.

pocmo:
Oh, heh, now this broke `FirstRunDialogueTest` :D - I'll put up a fix for that.

pocmo:
On Firebase I see failures in:
* `FirstRunDialogueTest`
* `openDownloadFileTest`
* `downloadNotificationTest`
* `cancelDownloadTest`

pocmo:
>     * `FirstRunDialogueTest`

I pushed a fix for this one.

>     * `openDownloadFileTest`
>     * `downloadNotificationTest`
>     * `cancelDownloadTest`

Those pass for me locally. Let's see if they still fail in the next push.

pocmo:
> This one hangs for me - while showing the notification tray. But I have been testing on a Pixel 4a. Let's see how it behaves in automation.

Ah, this only happened because there was another notification in my tabs tray. Once that was gone the test passed too.

pocmo:
Still the three download tests failing in automation.

sv-ohorvath (approved):
LGTM, let's see if the checks pass.

PR #4800, opened 2021-04-13
Update to GeckoView 88.0.20210412175251 with version bump to 8.15.0

jonalmeida:


jonalmeida:
Did a quick sanity check; looks good.

Issue #4746, opened 2021-03-19
about:crashcontent doesn't work


lobontiumira:
## Steps to reproduce
1. Open a page.
2. In the same URL bar go to about:crashcontent.

### Expected behavior
The app doesn't close. A "Tab crashed" page is displayed in-app.

### Actual behavior
The page turns blank. The URL displays the address from the page from step 1.

### Device information
* Android device: Google Pixel (Android 10), Samsung Note 10 (Android 11)
* Focus version: build from master (#1186.0.1- 20210310152336)
* Not reproducible on Foxus 8.14.0 with Google Pixel 2 (Android 9)

https://user-images.githubusercontent.com/40694158/111788384-af405e00-88c8-11eb-8251-6f281dfed896.mp4

PR #4810, opened 2021-04-19
Issue #4746: Move CrashReporter initialization to Components class.

### LINKED ISSUE #4748: Crash system dialog not displayed
## Steps to reproduce
1. Open focus and trigger the app to crash using this URL about:crashparent.

### Expected behavior
App crashed and a dialog is displayed.

### Actual behavior
The dialog is not displayed.

### Device information
* Android device: Samsung Note 10 (Android 11), Google Pixel (Android 10)
* Focus version: dev build from master - 1186.0.1- 20210310152336

pocmo:
The main problem for testing was that the crash reporter was disabled when telemetry was disabled, which is always the case for debug builds. This isn't needed since actually reporting is opt-in via the prompt. Now the behavior is similar to Fenix. In addition to that I got rid of the unnecessary `CrashReporterWrapper.

Closes #4748.


jonalmeida (approved):
Thank you! I didn't catch that. 😄

PR #4812, opened 2021-04-20
Fixed and re-enabled ShareWebsiteTest.kt UI test

sv-ohorvath:
Ran the test on Firebase 20 times and all passed.


isabelrios (approved):
LGTM, Thanks!

PR #4813, opened 2021-04-20
Remove old survey code.

pocmo:
This came up in the first L10N export: `strings.xml` contained translated strings (`tip_take_survey_es`, ..). Since this old survey isn't shown to users anymore, this patch removes the code and the strings.


jonalmeida (approved):
Verified with product that we don't need this any more. 👍

PR #4814, opened 2021-04-20
Remove unused shavar blocklist

jonalmeida:
We only needed this blocklist for `WebView` based Focus that we no longer use, so I think we can safely get rid of this now. 🙂


jonalmeida:
JSON files be like..

Screen Shot 2021-04-20 at 12 00 45 PM

pocmo (approved):
🔥

PR #4829, opened 2021-04-26
Switch from master to main branch and remove unused code.

pocmo:
Afaik this is all that is needed in the repo. After this lands I can change the repo and then let's see.... :)

I came across a bunch of old unused stuff and I just deleted it in this PR (old l10n scripts, blocklists, ..)


app/proguard-rules.pro
```diff
@@ -15,33 +15,6 @@
...
--dontwarn org.acra.ErrorReporter
-
-# If your project uses WebView with JS, uncomment the following
-# and specify the fully qualified class name to the JavaScript interface
-# class:
-#-keepclassmembers class fqcn.of.javascript.interface.for.webview {
-#   public *;
-#}
-
-# Uncomment this to preserve the line number information for
```
jonalmeida, line 36: This looks useful for debugging crash logs. I wonder why we don't use it in Focus?

jonalmeida (approved):
Nice! 👏 

I think the only thing left is to switch the default and branch protection settings on the project?

I'm going to leave this for you land at your convenience.

app/proguard-rules.pro
pocmo: Interesting. I'll move this to a new issue to investigate. I assume that this may cause the class files and then the app to larger in size?

pocmo: https://github.com/mozilla-mobile/focus-android/issues/4836

pocmo:
> I think the only thing left is to switch the default and branch protection settings on the project?

Yeah, GitHub now has a migration tool for this, it will move the protection settings and PR targets all too.

I'll merge now and invoke that migration tool. Let's see what happens. 😆

Issue #3624, opened 2018-10-12
Application restarts when interrupted


sylvestre:
Reported by Samsung

<Defect>
The application restarts when interrupted by HOME KEY

<Procedure>
1. Execute the application > More menu > 'Settings' > 'Privacy & Security' > Check screen
2. HOME KEY > Re-execute app > Check screen

<Expected Result>
The application should show the previously watched screen after interruption.
<Test Device Info>
Detected Device: SM-N950F, SW Version: N950FXXU5CRHA

sylvestre:
the video:
http://sylvestre.ledru.info/moz/1_restart.mp4
the app has been removed from the Samsung galaxy apps

boek:
This one is proving hard to fix without a device to reproduce. Maybe @ekager you can get someone to bring a Samsung device?

ekager:
This happens without a Samsung device @boek. To reproduce just open settings while browsing, press home on device, and go back to Focus. Settings Activity is killed because of our singleTask launch mode I believe. Removing this fixes this specific problem, but allows multiple instances (See #3636 for a "fix" and my comments from testing).

npark-mozilla:
@sv-ohorvath Oana, could you look into creating additional test cases for this feature?   @colintheshots , if you mark it with the label 'QAReady', we'll take it as it needs extra attention.  We'll ping you when we need additional information.  Thanks!

sv-ohorvath:
Wrote some test cases, please review. If you think that I should add anything else, please let me know: 
https://testrail.stage.mozaws.net/index.php?/cases/view/210271
Verified this issue on Nightly and found the following bugs:
Custom tabs is broken #3674
Make default browser triggers a blank screen instead of returning to the settings menu #3675
GeckoView: Links sent from other apps go to about:blank #3676
WV: Links opened from other apps don't bring the app to foreground #3677

npark-mozilla:
@sv-ohorvath test case looks good to me! and since we have those new open bugs, we can mark this issue with QAVerified flag.

PR #4830, opened 2021-04-26
Issue #4342: MainActivity: Use 'singleTask'.

### LINKED ISSUE #4342: The app crashes by tapping on 'Open and Erase' button when you are on Settings menu
## Steps to reproduce
1. Open Firefox Focus;
2. Load a webpage;
3. Enter on the Settings menu;
4. Press the home button;
5. Pull down the notification tray and tap on the 'Erase and Open' button;

### Expected behavior
The initial screen of Focus is displayed, with the previous webpage deleted;

### Actual behavior
The app crashes when you 'Erase and Open' from the Settings Menu;

### Device information
* Android device: Google Pixel (Android Q), Samsung Galaxy S8 (Android 9);
* Focus version: 8.0.16 GV 69.0
*  [pushlog](https://github.com/mozilla-mobile/focus-android/files/3533572/crash.focus.txt)

pocmo:
This fixes #4342 and is in sync with how we use the flag in Fenix and Reference Browser. However switching the flag, will cause #3624 to come back. However I think the right fix for #3624 is to turn Focus into a real single-activity app - like Fenix. This also makes sense in preparation for Jetpack Compose. I'll file a follow-up to investigate getting rid of the separate activity for settings.


jonalmeida (approved):
✅ 

Should we also add clear top on the [erase button](https://searchfox.org/mozilla-mobile/source/focus-android/app/src/main/java/org/mozilla/focus/session/SessionNotificationService.kt#52) in the notification too?

pocmo:
> ✅
> 
> Should we also add clear top on the [erase button](https://searchfox.org/mozilla-mobile/source/focus-android/app/src/main/java/org/mozilla/focus/session/SessionNotificationService.kt#52) in the notification too?

Good point. I'll test that.

pocmo:
The erase action doesn't seem affected.

PR #4833, opened 2021-04-27
Re-enabled AccessAboutAndYourRightsPagesTest.kt. Renamed it to MozillaSupportPagesTest.kt

sv-ohorvath:
Ran tests on Firebase 30 times and passed.


isabelrios (approved):
LGTM!

Issue #4839, opened 2021-04-29
UI test failure - ErrorPagestTest#noNetworkConnectionErrorPageTest on API 29


sv-ohorvath:
### Firebase Test Run:
https://console.firebase.google.com/u/1/project/moz-fx-mobile-firebase-testlab/testlab/histories/bh.2b4ca2c7fdb2f0f0/matrices/7803009004200227047
### Stacktrace:
`java.lang.RuntimeException: java.util.concurrent.ExecutionException: androidx.test.espresso.IdlingResourceTimeoutException: Wait for [javaClass] to become idle timed out
at androidx.test.espresso.Espresso.onIdle(Espresso.java:342)
at org.mozilla.focus.helpers.TestHelper.setNetworkEnabled(TestHelper.kt:132)
at org.mozilla.focus.activity.ErrorPagesTest.tearDown(ErrorPagesTest.kt:37)`

The test is **failing on FTL API 29,** but working fine on API 28. Also, **works fine in an AS emulator with API 29.**
The failure indicates that the network connection wasn't established in the expected amount of time, causing a timeout.
Disabling to have the UI test task be green again until I can find a solution.

A similar test (identical code for NetworkIdlingResources and enabling/disabling the connection) will fail on Fenix too if ran on API 29: https://console.firebase.google.com/u/1/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/7732391941108455461/executions/bs.49e82ab7c9525a76/testcases/1/test-cases

Logs indicate a device connectivity issue:
`Calling main entry com.android.commands.svc.Svc
04-29 05:19:25.621: V/SettingsProvider(1503): Notifying for 0: content://settings/global/mobile_data
04-29 05:19:25.624: D/AndroidRuntime(9241): Shutting down VM
04-29 05:19:25.625: D/MultiSimSettingController(1799): onUserDataEnabled
04-29 05:19:25.625: D/MultiSimSettingController(1799): setUserDataEnabledForGroup subId 1 enable true
04-29 05:19:25.646: I/netd(1329): isAlive() -> {"true"} <0.00ms>
04-29 05:19:25.664: W/app_process(9241): Thread attaching while runtime is shutting down: Binder:9241_1
04-29 05:19:25.664: I/AndroidRuntime(9241): NOTE: attach of thread 'Binder:9241_1' failed
[...]

PR #4840, opened 2021-04-29
For #4839: Disable noNetworkConnectionErrorPageTest causing failures …

sv-ohorvath:
…on master

This test fails to setup the network connection back on, causing other failures. Works only on API 28, not 29.


isabelrios (approved):
LGTM, thanks for taking care of this

PR #4854, opened 2021-05-04
Remove code for handling configuration changes.

pocmo:
I think we can actually get rid of this code for manually handling configuration changes now (and I wanted to do this for Fenix for a while). I think nowadays we can just let Android recreate the activity and our code will reattach the GeckoSession to the view and everything should be fine. 🤔


jonalmeida:
Hmm.. looking at the [usages here](https://searchfox.org/mozilla-mobile/search?q=onConfigurationChanged&path=focus-android), the `UrlInputFragment` is on the MainActivity and needs to know orientation updates which seems important.

Maybe we can reduce to just `orientation`?

pocmo:
> Hmm.. looking at the [usages here](https://searchfox.org/mozilla-mobile/search?q=onConfigurationChanged&path=focus-android), the `UrlInputFragment` is on the MainActivity and needs to know orientation updates which seems important.

I think this was just one of those annoying workarounds we had to add *because* we were handling configurations manually? I'll double check.

jonalmeida:
> > Hmm.. looking at the [usages here](https://searchfox.org/mozilla-mobile/search?q=onConfigurationChanged&path=focus-android), the `UrlInputFragment` is on the MainActivity and needs to know orientation updates which seems important.
> 
> I think this was just one of those annoying workarounds we had to add _because_ we were handling configurations manually? I'll double check.

I did a quick test, we're no longer showing the menu on rotation with this change, but that's also consistent with Fenix, so let's merge it for now.

jonalmeida (approved):

PR #4865, opened 2021-05-06
Complete browser-state migration and remove browser-session dependency.

pocmo:
This is the matching branch for the changes in A-C: https://github.com/mozilla-mobile/android-components/pull/10212

With that Focus will no longer depend on `browser-session` and will only use `browser-state`.

The change in A-C will not land until the beginning of a new Fenix Nightly cycle.


Amejia481 (approved):
✅

PR #4866, opened 2021-05-07
[WIP] fixed nits in focus strings

Delphine:
Talking with @mathjazz tomorrow and going over this


app/src/main/res/values/strings.xml
```diff
@@ -50,7 +50,7 @@
     <string name="menu_add_to_home_screen">Add to Home screen</string>
 
     <!-- Menu item: Shows information about features and changes that are new to this version -->
-    <string name="menu_whats_new">What\'s New</string>
+    <string name="menu_whats_new">What's New</string>
```
pocmo, line 53: Should this also be `’` instead of `'`, like https://github.com/mozilla-mobile/fenix/blob/master/app/src/main/res/values/strings.xml#L134

mathjazz: Yes.

Delphine: Good catch! (for once, I'm not the one chasing people around about this 😂 )

pocmo (approved):

PR #4871, opened 2021-05-07
Clean up strings.

pocmo:
Removing a bunch of unused strings. Since we switched to a bunch of Android Components those strings are now unused in Focus. Getting rid of them will make the string import into Pontoon simpler. :)


jonalmeida (approved):
🔥

pocmo:
`3154 errors` because of `ExtraTranslation` 😂. We suppress that in Fenix. I'll port a patch to Focus here.

Issue #4731, opened 2021-03-18
"Find in Page" search box is still displayed after new links are opened


sflorean:
## Steps to reproduce
1. Open a url containing text -> open menu, and tap 'Find in Page'.
2. Enter a word to search (that exists in more than one place).
3. Open another link/links.

### Expected behavior
"Find in Page" search box is not visible after the new links are opened.

### Actual behavior
"Find in Page" search box is still displayed.

### Notes
* The searchbox color is white.
* Not reproducible on Focus - Play Store.

### Device information
* Android device: Samsung Note 10(Android 11), LG G7 Fit (Android 8.1)
* Focus version: dev build # 1186.0.1- 20210310152336

pocmo:
The same seems to happen in Reference Browser and it looks like we do not have any specific code handling this in the feature. I am not really sure yet why it works that way in Fenix. That may be a side effect of something else.

gouri-panda:
@pocmo Maybe in-migration from `SessionManger` to `BrowserStore`, they forgot to add the feature. I'm interested in working on this issue. Can I work on it?

pocmo:
@gouri-panda Yes, you can work on this. But let's discuss some options here, before writing code. :)

My first thought is whether `EngineObserver` should remove `findResults` for the current page on navigation and then the presenter should detect that and notify the app to do whatever it needs to do (e.g. hide the bar).

* https://github.com/mozilla-mobile/android-components/blob/master/components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt#L63
* https://github.com/mozilla-mobile/android-components/blob/master/components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/internal/FindInPagePresenter.kt#L33-L40
* https://github.com/mozilla-mobile/android-components/blob/master/components/feature/findinpage/src/main/java/mozilla/components/feature/findinpage/FindInPageFeature.kt#L68

pocmo:
I noticed that we already clear the results when the loading state changes:
https://github.com/mozilla-mobile/android-components/blob/master/components/browser/session/src/main/java/mozilla/components/browser/session/engine/EngineObserver.kt#L135

Would we want to close the bar in that case? 🤔

gouri-panda:
@pocmo I'm not so familiar with the `BrowserState` code base. How things work before `BrowserStore` i mean how `SessionManager` handles if the URL changed in session? Thanks!

mcarare:
[...]

PR #4884, opened 2021-05-12
For #4731: Close find in page when navigating to another page.

mcarare:
Not sure about the naming conventions if this approach is used elsewhere in Focus.
We can also use this to update other components when navigating.


pocmo (commented):
I would like to see this as part of the feature in A-C.

I also wonder if we should quickly sync with UX/product, since this is not the behavior on desktop: The "find in page" bar will be kept open on navigation until explicitly closed.

mcarare:
Closing this. Will implement this in A-C when we get some UX feedback on the matter.

PR #4899, opened 2021-05-22
Migrate tp exceptions

Amejia481:


Amejia481:
This depends on https://github.com/mozilla-mobile/focus-android/pull/4865 as there are multiple breaking changes that are coming from AC

app/src/main/java/org/mozilla/focus/FocusApplication.kt
```diff
@@ -38,6 +39,8 @@ open class FocusApplication : LocaleAwareApplication(), CoroutineScope {
 
     override fun onCreate() {
         super.onCreate()
+        val exceptionsMigrator = EngineProvider.provideTrackingProtectionMigrator(this)
+        exceptionsMigrator.start(this)
```
pocmo, line 43: Let's move this after the crash reporter and log setup, so that we can see logs and crashes from it. :)

I also think this needs to go into the `isMainProcess()` block to only run it in the main process and not in every gecko child process.

app/src/main/java/org/mozilla/focus/exceptions/GeckoExceptionsMigrator.kt
```diff
@@ -0,0 +1,71 @@
...
+     */
+    private fun fetchExceptions(context: Context): List<String> {
+        return (preferences(context)
+                .getString(KEY_DOMAINS, "") ?: "")
+                .split(SEPARATOR)
+                .filter { it.isNotEmpty() }
+
+    }
+
+    private fun isOver(context: Context) = preferences(context).contains(PREFERENCE_NAME)
```
pocmo, line 49: Does this need to be `KEY_DOMAINS`? 🤔

app/src/main/java/org/mozilla/focus/exceptions/GeckoExceptionsMigrator.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    private fun preferences(context: Context): SharedPreferences =
+            context.getSharedPreferences(PREFERENCE_NAME, Context.MODE_PRIVATE)
+
+    private fun migrate(uri: String) {
+        val privateMode = true
```
pocmo, line 55: nit: I can we can just put `true` into the function call below and do not need the `privateMode` reference?

build.gradle
```diff
@@ -11,6 +11,11 @@ buildscript {
     ext.mozilla_components_version_telemetry = '57.0.9'
 
     repositories {
+        maven {
+            name "Mozilla TC"
+            url "https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/Na_wzkiXSVCbHp6ZiksMuQ/runs/0/artifacts/public/build/maven"
+        }
+        mavenLocal()
```
pocmo, line 18: Now we no longer need those, right? :)

app/src/main/java/org/mozilla/focus/exceptions/GeckoExceptionsMigrator.kt
Amejia481: Yeah, 🙃  thanks.

app/src/main/java/org/mozilla/focus/exceptions/GeckoExceptionsMigrator.kt
Amejia481: Yes, 🙃  thanks

app/src/main/java/org/mozilla/focus/exceptions/GeckoExceptionsMigrator.kt
Amejia481: Yes, I just wanted to be explicit about what it's this parameter about, as this is java api we can't use named parameters,

pocmo (commented):
Looks good. Once we have the matching A-C upgrade then I'll test the migration on a device too.

Amejia481:
Thanks!

pocmo (approved):
I verified that tracking protection exceptions get migrated. 

Unfortunately GeckoView doesn't seem to persist them. However that's unrelated to this patch here.

Issue #4880, opened 2021-05-11
Long tap context menu is different


ebalazs-sv:
## Steps to reproduce
1. Launch Focus and visit i.e. wikipedia.org;
2. Long tap a link;
3. Observe the context menu;
### Expected behavior
The context menu is:
![rsz_2screenshot_20210511-152018](https://user-images.githubusercontent.com/48799384/117814268-c10d0480-b26c-11eb-8290-7f682160191e.png)

### Actual behavior
The context menu is:
![rsz_screenshot_20210511-152007](https://user-images.githubusercontent.com/48799384/117814273-c2d6c800-b26c-11eb-9051-74ab1c5e20c9.png)
### Notes
This is also reproducible for long tap on cnn.com, list variations.

* Android device: Pixel 2 (Android 9);
* Focus version: #1190 dev build from main;
* Not reproducible on Focus 8.15.4;

gouri-panda:
@pocmo I want to work on this issue ;)

gouri-panda:
In `ContextMenuFragment` we should add a method to change the background. Should I create a PR In `AC`?

pocmo:
I think we should be able to control this through the app theme?

ionutbedregeanu:
I would like to work on this issue

PR #4932, opened 2021-06-10
Update alert dialog style in AppTheme

ionutbedregeanu:
For #4880 

A custom alert dialog style was already implemented but wasn't set in AppTheme

### Video

https://user-images.githubusercontent.com/11731652/121487472-c9c33880-c9da-11eb-9588-e447bbd5e131.mp4


jonalmeida (approved):
✨

jonalmeida:
Noting that the change doesn't match identically to what was there before, but this is a good start. Adding the divider and custom styling for the URL is something we can do in a follow-up too.

Issue #4924, opened 2021-06-09
Splash: Update Splash screen design


amedyne:
- Add the shield icon (Tracking Protection) to the left of the URL bar 
- Update Focus Icon
- Keyboard

Implement Dark Mode. 

https://www.figma.com/file/2V6wgf9fRFdYfLaf5c3oPj/MVP---iOS%2FAndroid---Product-UI?node-id=804%3A11

ionutbedregeanu:
All ticket changes are implemented behind a feature flag(FeatureFlags.isMvp).
FeatureFlags.isMvp = true -> mvp changes are visible
FeatureFlags.isMvp = false -> mvp changes are hidden

lobontiumira:
Hi all!
I've verified with Google Pixel (Android 10), on the latest Debug build 8.13.1 (1191.0a1-20210705095222) that contains the latest modifications, and here are my findings:
* the search field has no ETP icon when there is no page loaded,
* the ETP icon is not as in Figma,
* there is no space between the search field and the three-dot menu, as in Figma,
* there is no dark mode by default,
* the "What's new" option is not moved to "Settings":

![Untitled](https://user-images.githubusercontent.com/40694158/126646956-5a6e18cd-d40d-4018-806a-20f5dfebf5f8.png)

@pocmo I understood you can help us with more details regarding the search bar design, as it will be like it is now in Fenix.
I'll move the issue back to "In Development" until further notice.
Thank you!

ionutbedregeanu:
Dark mode background its implemented by [4923](https://github.com/mozilla-mobile/focus-android/issues/4923)

ionutbedregeanu:
Meanwhile almost all changes mentioned by @lobontiumira were made, except the logo update.
I will make  a new pr for that

lobontiumira:
Verified on the latest debug build from main 8/24 with Lenovo Tab M10 (Android 10), and the TP icon has the new UI, as in Figma. Tapping on it, displays the tracking protection panel:

![TP icon](https://user-images.githubusercontent.com/40694158/130610750-47a2613e-1620-4081-a66a-377d82821227.png)

@amedyne in Settings -> Privacy & Security the section is "Tracking Protection", but in the tracking protection panel is displayed as "Enhanced Tracking Protection". Should I address this in #4931 ?
Thank you!

amedyne:
[...]

PR #4948, opened 2021-06-16
Update splash screen design

ionutbedregeanu:
For #4924 

- Add the shield icon (Tracking Protection) to the left of the URL bar

- Update Focus logo

- Move tips section to the bottom of the screen

- Force app to use dark mode theme


cadeyrn:
Only the drawables in `focus` were replaced. I guess `klar` should be updated as well. 😉

pocmo:
@ionutbedregeanu We somehow need to move this behind a feature flag, since we do not want this to get out with the next release just yet.

Let's hold back on the toolbar changes for now (shield icon). @jonalmeida and I talked yesterday and we want to try to integrate the `browser-toolbar` component from Android Components (the same we use in Fenix) and then we wouldn't need to do those changes in the custom, legacy toolbar here in Focus (Prototype of integration: #4951).

ionutbedregeanu:
For now we don't have Klar logo resource so I will create a new ticket for that and remove WIP from this ticket

app/src/main/java/org/mozilla/focus/browser/LocalizedContent.kt
```diff
@@ -45,7 +45,7 @@ object LocalizedContent {
         substitutionMap["%about-version%"] = aboutVersion
         val aboutContent = resources.getString(R.string.about_content, appName, learnMoreURL)
         substitutionMap["%about-content%"] = aboutContent
-        val wordmark = HtmlLoader.loadPngAsDataURI(context, R.drawable.wordmark)
+        val wordmark = HtmlLoader.loadPngAsDataURI(context, R.drawable.mvp_wordmark)
```
jonalmeida, line 48: Naming nit: Once we've released, the term "MVP" will be redundant and we'd have to make changes around the code base to correct it after the fact. I know it's not ideal, but maybe `wordmark2`?
```suggestion
        val wordmark = HtmlLoader.loadPngAsDataURI(context, R.drawable.wordmark2)
```

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -277,8 +278,15 @@ class UrlInputFragment :
...
     }
 
-    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?):
-        View? = inflater.inflate(R.layout.fragment_urlinput, container, false)
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? = if (MvpFeatureManager.isEnabled) {
+        inflater.inflate(R.layout.mvp_fragment_urlinput, container, false)
```
jonalmeida, line 286: Same nit as above:
```suggestion
        inflater.inflate(R.layout.fragment_urlinput2, container, false)
```

app/src/main/res/layout/focus_preference.xml
```diff
@@ -60,7 +60,6 @@
             android:layout_below="@android:id/title"
             android:layout_alignStart="@android:id/title"
             android:textAppearance="?android:attr/textAppearanceListItemSecondary"
-            android:textColor="?android:attr/textColorSecondary"
```
jonalmeida, line 63: Was this in error?

app/src/main/res/values/styles.xml
```diff
@@ -3,15 +3,17 @@
...
+    <style name="AppTheme" parent="Theme.MaterialComponents.DayNight.NoActionBar.Bridge">
         <item name="colorPrimary">@color/colorPrimary</item>
         <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
         <item name="colorAccent">@color/colorAccent</item>
-
         <item name="android:windowContentTransitions">true</item>
-
         <item name="popupMenuStyle">@style/PopupMenu</item>
 
+        <!-- Use the same colors for both themes until light theme it will be implemented -->
```
jonalmeida, line 13: 👍

jonalmeida (approved):
Looks good! Left a few nits and questions, but I think we can land this after that.

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -42,6 +44,10 @@ open class MainActivity : LocaleAwareAppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        if (MvpFeatureManager.isEnabled) {
+            AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_YES)
```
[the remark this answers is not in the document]
jonalmeida: Ah, I see the answer I'm looking for in the message of the previous commit. Let's add a similar note here too.

> Because the app theme was changed to DayNight it automatically load dark
> colors for some attributes when the app is in light mode but for now we
>  don't want that.
> Those attributes will be updated at light theme implementation

AaronMT:
(If you want to rebase your open PRs too, I recently landed some Github Action checks and the UI tests have some fixes recently on them)

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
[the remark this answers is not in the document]
ionutbedregeanu: Considering the last commit I think I don't need this anymore, so can be removed. I will do some tests to check

app/src/main/res/layout/focus_preference.xml
ionutbedregeanu: Is redundant, but I will revert this change because is somehow out of scope

PR #4951, opened 2021-06-16
Integrate browser-toolbar component.

pocmo:
This is a work-in-progress version of a `browser-toolbar` integration. It's not feature complete yet, but it does allow using the browser (editing & displaying).


pocmo:
@jonalmeida Okay, I think this is ready to review :) For the TP icon to fully work, we need #4960 and the latest A-C Nightly.

app/src/main/java/org/mozilla/focus/browser/binding/BlockingThemeBinding.kt
```diff
@@ -29,7 +30,7 @@ class BlockingThemeBinding(
     private val isCustomTab: Boolean,
     private val statusBar: View,
     private val urlBar: View,
-    private val blockView: View
+    private val toolbar: BrowserToolbar
```
jonalmeida, line 33: unused?

app/src/main/java/org/mozilla/focus/browser/integration/NavigationButtonsIntegration.kt
```diff
@@ -0,0 +1,148 @@
+package org.mozilla.focus.browser.integration
```
jonalmeida, line 1: nit: license

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -0,0 +1,90 @@
...
+                store,
+                toolbar,
+                sessionId = customTabId,
+                useCases = customTabsUseCases,
+                closeListener = { fragment.closeCustomTab() }
+            )
+        }
+
+        if (HardwareUtils.isTablet(context)) {
+            navigationButtonsIntegration = NavigationButtonsIntegration(
```
jonalmeida, line 67: Nice!

app/src/main/java/org/mozilla/focus/searchsuggestions/SearchSuggestionsViewModel.kt
```diff
@@ -107,6 +106,8 @@ class SearchSuggestionsViewModel(application: Application) : AndroidViewModel(ap
             State.Disabled(givePrompt)
         }
 
+        Log.w("SKDBG", " * State(${hashCode()}) = $state")
```
jonalmeida, line 109: Probably meant to be removed? :)

jonalmeida (approved):
This looks great!

Two things I observed:
 - The progress bar doesn't show up when the toolbar is hidden, but that should be easy to do in a follow-up.
 - We may break password managers that use accessibility services to read the url from the toolbar based on the ID of the view.

app/src/main/java/org/mozilla/focus/searchsuggestions/SearchSuggestionsViewModel.kt
pocmo: 🤣

pocmo:
> * The progress bar doesn't show up when the toolbar is hidden, but that should be easy to do in a follow-up.

Yeah, I looked at Fenix and over there we keep the toolbar visible during loads (and in other situations), we should probably do the same in Focus.

> * We may break password managers that use accessibility services to read the url from the toolbar based on the ID of the view.

Possible, but I think there's no way around that.

Issue #4922, opened 2021-06-08
Create a Focus megazord for native components


jonalmeida:
We're going to be using Glean and Nimbus in Focus. Today, we can use those components separately, but do we **need** to create one?

rfk:
I expect that it should be safe to use Glean and Nimbus in Focus without creating a special Focus megazord, and instead using the default setup that we ship with Fenix. The only observable downside should be that your app ships some unrelated application-services Rust code that it doesn't need.

I'm motivated to use this as an opportunity to revamp some of how we managing megazords for Android consumers, so please flag me if/when this downside becomes a problem for you and we can figure out a good path forward.

For my context: will Focus use GeckoView at all? (I ask because GeckoView+Glean has its own interesting megazording thing going on as well).

jonalmeida:
> For my context: will Focus use GeckoView at all? (I ask because GeckoView+Glean has its own interesting megazording thing going on as well).

Focus uses GeckoView exclusively. 🙂 

I do recall reading the doc about this new dependency structure, although I didn't see a strict timeline for it, so assumed we should move forward with a megazord-like solution for the time being IF that is a low-effort solution.

rfk:
> assumed we should move forward with a megazord-like solution for the time being IF that is a low-effort solution.

👍🏻 to moving forward in this manner, yes.

jonalmeida:
@jhugman and I spoke about this in relation to https://github.com/mozilla-mobile/focus-android/pull/5890.

With Focus, being smaller/faster is part of the product image, so if we can reduce the size this would be good as we integrate more native components in the future.

The discussion that rfk hinted at above is still ongoing, however James said it is leaning more towards an app-specific megazord, so there is more reason to go in this direction.

iorgamgabriel:
A smaller megazord for Focus would probably be nice to have even after the migration.

jonalmeida:
A megazord is useful to de-duplicate native dependencies in our rust components. It would be useful to compare what is the file-size impact with and without a megazord before we decide to invest time into setting one up.

I'm not sure how to do that exactly, but maybe someone on the app services team could assist us?

csadilek:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1802009

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #4952, opened 2021-06-16
Setup and initialize Nimbus experiments

jonalmeida:
Taken from Fenix, I think this is all that's needed today to get started with Nimbus as far as the client code is concerned.

cc: @jhugman


jonalmeida:
Removed the gitignore that prevents adding json files to git without a force add: https://github.com/mozilla-mobile/focus-android/pull/4953

app/src/main/java/org/mozilla/focus/experiments/NimbusSetup.kt
```diff
@@ -0,0 +1,97 @@
...
+        if (crashReporter.enabled) {
+            crashReporter.submitCaughtException(e)
+        } else {
+            Logger.error("Failed to initialize Nimbus", e)
+        }
+        NimbusDisabled()
+    }
+
+fun getNimbusAppName(): String {
+    return if (BuildConfig.FLAVOR.contains("focus")) {
```
jonalmeida, line 92: Pre-emptively assuming this will be the app names given what I saw see is there for iOS right now.

rfk:
Noting for mid-term planning purposes: this new dependency means you're going to be pulling in *all* of the application-services compiled Rust code. At some point we'll want to have a conversation about making Focus its own built artifact with just the subset of Rust code that you need.

jonalmeida:
> Noting for mid-term planning purposes: this new dependency means you're going to be pulling in _all_ of the application-services compiled Rust code. At some point we'll want to have a conversation about making Focus its own built artifact with just the subset of Rust code that you need.

One step ahead! https://github.com/mozilla-mobile/focus-android/issues/4922

app/src/main/java/org/mozilla/focus/experiments/NimbusSetup.kt
```diff
@@ -0,0 +1,97 @@
...
+        "focus-android"
+    } else {
+        "klar-android"
+    }
```
travis79, line 96: These are likely to be the correct names, but I think that `_` rather than `-` will be used.  See here: https://github.com/mozilla/probe-scraper/blob/edd24dda40487f0274da3182a58baf8d02f0582e/repositories.yaml#L499

travis79: Meaning, I expect `focus_android` and `klar_android`.

travis79 (changes requested):
One minor naming change for the app name, but other than that, this looks good to me.  FYI, @jhugman is out for a bit, so I don't know if you want to wait and block on his review or not.

app/src/main/java/org/mozilla/focus/experiments/NimbusSetup.kt
```diff
@@ -0,0 +1,97 @@
...
+        // and
+        // https://github.com/mozilla/probe-scraper/blob/master/repositories.yaml
+        val appInfo = NimbusAppInfo(
+            appName = getNimbusAppName(),
+            // Note: Using BuildConfig.BUILD_TYPE is important here so that it matches the value
+            // passed into Glean. `Config.channel.toString()` turned out to be non-deterministic
+            // and would mostly produce the value `Beta` and rarely would produce `beta`.
+            channel = BuildConfig.BUILD_TYPE
+        )
+        Nimbus(context, appInfo, serverSettings).apply {
```
jhugman, line 52: I'd like to see an error reporter passed here: e.g. https://github.com/mozilla-mobile/fenix/pull/20049/files#diff-1e627319338ce89fe0c17b48d5e02d49fcf559ff01e8400a623bbbca1c6f9e73R24-R29

jhugman (approved):
This looks great! I'd like to see an error reporter here, but nothing enough to hold up for another review cycle. 🎉

app/src/main/java/org/mozilla/focus/experiments/NimbusSetup.kt
jonalmeida: We do the same logic in the catch below already by forwarding the exceptions to the crash reporter. Is there something else we're looking to add?

app/src/main/java/org/mozilla/focus/experiments/NimbusSetup.kt
jonalmeida: Oops, underscore is what I meant to have used. Thanks for the catch!

app/src/main/java/org/mozilla/focus/experiments/NimbusSetup.kt
jonalmeida: Ping @jhugman

app/build.gradle
```diff
@@ -395,6 +396,33 @@ def getEnabledLocales() {
...
+
+print("Nimbus endpoint: ")
+android.applicationVariants.all { variant ->
+    def variantName = variant.getName()
+
+    if (!variantName.contains("Debug")) {
+        try {
+            def url = new File("${rootDir}/.nimbus").text.trim()
+            buildConfigField 'String', 'NIMBUS_ENDPOINT', '"' + url + '"'
+            println "(Added from .nimbus file)"
```
pocmo, line 411: Is the releng part of this already done?

app/src/main/java/org/mozilla/focus/experiments/NimbusSetup.kt
```diff
@@ -0,0 +1,97 @@
...
+            NimbusServerSettings(url = Uri.parse(url))
+        } else {
+            null
+        }
+
+        // Global opt out state is stored in Nimbus, and shouldn't be toggled to `true`
+        // from the app unless the user does so from a UI control.
+        // However, the user may have opt-ed out of making experiments already, so
+        // we should respect that setting here.
+        // TODO change this to separate pref if we need a separate UI control.
```
pocmo, line 35: Is this already decided?

pocmo (approved):

PR #4953, opened 2021-06-17
Remove some more leftover webview code

jonalmeida:
The gitignore was needed when we were generating json blocklists in Focus. We don't need that any more.


pocmo (approved):
👍

Issue #4928, opened 2021-06-09
Splash: Implement Splash screen Alerts


amedyne:
https://www.figma.com/file/2V6wgf9fRFdYfLaf5c3oPj/MVP---iOS%2FAndroid---Product-UI?node-id=903%3A4136

ionutbedregeanu:
QA note:
All ticket changes are implemented behind a feature flag(FeatureFlags.isMvp).
FeatureFlags.isMvp = true -> mvp changes are visible
FeatureFlags.isMvp = false -> mvp changes are hidden

lobontiumira:
Hi all,
I've verified with Google Pixel (Android 10), on the latest Debug build 8.13.1 (1191.0a1-20210705095222) that contains the latest modifications, and the "Browsing history cleared" snackbar is displayed after the user closes the opened tab/tabs by tapping the "Delete icon".

<img src="https://user-images.githubusercontent.com/40694158/126648986-ddf40688-54eb-4484-8e4b-479044b3d25d.png" width="300" height="550" />

PR #4956, opened 2021-06-18
Update splash screen alerts design

ionutbedregeanu:
For #4928

Update splash screen alerts design
Update app theme
Create a new colors file resource for dark theme
Create a mvp feature flag
Apply mvp changes only when the feature flag is enabled, otherwise keep the old behavior

Because the app theme was changed to DayNight it automatically load dark colors for some attributes when the app is in light mode but for now we don't want that.
Those attributes will be updated at light theme implementation

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/ViewUtils.kt
+++ b/app/src/main/java/org/mozilla/focus/utils/ViewUtils.kt
@@ -9,11 +9,11 @@ import android.app.Activity
 import android.content.Context
 import android.os.Handler
 import android.os.Looper
+import android.view.Gravity
 import androidx.annotation.StringRes
 import com.google.android.material.snackbar.Snackbar
 import androidx.core.content.ContextCompat
 import androidx.core.view.ViewCompat
-import android.view.Gravity
 import android.view.MenuItem
 import android.view.View
 import android.view.inputmethod.InputMethodManager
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/mozilla/focus/utils/ViewUtils.kt
```diff
@@ -95,12 +95,16 @@ object ViewUtils {
...
         val snackbarView = snackbar.view
-        snackbarView.setBackgroundColor(ContextCompat.getColor(context, R.color.snackbarBackground))
-
         val snackbarTextView = snackbarView.findViewById<View>(R.id.snackbar_text) as TextView
         snackbarTextView.setTextColor(ContextCompat.getColor(context, R.color.snackbarTextColor))
-        snackbarTextView.gravity = Gravity.CENTER
-        snackbarTextView.textAlignment = View.TEXT_ALIGNMENT_CENTER
+        if (MvpFeatureManager.isEnabled) {
+            snackbarView.setBackgroundResource(R.drawable.background_snackbar)
+            snackbarTextView.gravity = Gravity.CENTER_VERTICAL
```
jonalmeida, line 102: The alignment and startPadding seem off. CENTER_VERTICAL doesn't seem to have worked but I haven't looked into why just yet.

This PR:
Screen Shot 2021-07-15 at 6 57 42 PM

Designs:
Screen Shot 2021-07-15 at 6 57 58 PM

jonalmeida (approved):
Some nits, but looks good to land https://github.com/mozilla-mobile/focus-android/pull/4948 first and rebase onto that one before landing this.

Issue #4929, opened 2021-06-10
Splash: Update 3dot Menu in Home/Splash  Screen


amedyne:
Conform to latest UI designs.

PR #4957, opened 2021-06-22
Update home menu design

ionutbedregeanu:
For #4929 

Add an icon to every menu item -> updates on layouts
Add new icons as resources
Update settings icon

I marked this PR as a draft because this menu it will be refactored using AC components


jonalmeida:
I think we can rebase this now that we've landed the BrowserToolbar from AC. 🙂

jonalmeida:
Looking at the designs, the only difference are the added icons, so maybe we can just add that to the existing menu.

jonalmeida:
Opened https://github.com/mozilla-mobile/focus-android/pull/5072 that uses this PR to address the comment above.

Issue #4617, opened 2020-08-28
The trashcan icon is moving with the keyboard when having more than 1 tab open


ebalazs-sv:
## Steps to reproduce
1. Launch Focus and visit google.com;
2. Long tap on Images and choose Open link in new tab;
3. Tap on Google input field;
4. Observe the trashcan icon above the keyboard;
### Expected behavior
The trashcan icon is not displayed above the keyboard.
### Actual behavior
The trashcan icon is displayed above the keyboard.

![rsz_screenshot_20200828-161912](https://user-images.githubusercontent.com/48799384/91565171-75b6fe00-e94a-11ea-95b2-ef998b03e242.png)

### Device information

* Android device: Pixel 2 (Android 9);
* Focus version: Focus 8.7.2 Build ID: 342410345 GV 80.0.20200827204751

Ankita13-code:
Heyy @ebalazs-sv!
This seems as a pretty good first issue. I would like to work on it so can you please assign this issue to me?

gouri-panda:
@pocmo Can I work on this issue?
https://github.com/mozilla-mobile/focus-android/blob/f7de21b82c0be09f4419cc216bc8ad963c7c31bd/app/src/main/java/org/mozilla/focus/widget/ResizableKeyboardViewDelegate.java#L55-L63
It seems if we remove the `viewToHide`'s  setVisibility method. it will work fine. and also we need to add more padding (i.e in `updateBottomPadding`) e.g 25dp to the actual bottom  of the fabicon because some devices have a suggestions bar above the keyboard

lobontiumira:
Verified as fixed on the debug build 8.13.1 (1191.0a1-20210705095222) with Google Pixel (Android 10).

PR #4963, opened 2021-06-23
Hide floating buttons when the keyboard is shown

ionutbedregeanu:
For #4617 

This functionality was already implemented but for a single view and because
we are having two floating buttons(erase and sessions counter) I added a new
method to be able to update the visibility of multiple views

#### Video

https://user-images.githubusercontent.com/11731652/123086344-fee47780-d42b-11eb-812c-ba3fd0c15a1f.mp4

### DIFF
--- a/app/src/main/java/org/mozilla/focus/widget/ResizableKeyboardViewDelegate.java
+++ b/app/src/main/java/org/mozilla/focus/widget/ResizableKeyboardViewDelegate.java
@@ -10,13 +10,13 @@ import android.animation.AnimatorListenerAdapter;
 import android.animation.ValueAnimator;
 import android.content.res.TypedArray;
 import android.graphics.Rect;
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
 import android.util.AttributeSet;
 import android.view.View;
 import android.view.ViewTreeObserver;
-
+import androidx.annotation.NonNull;
 import org.mozilla.focus.R;
+
+import java.util.ArrayList;
 
 /**
  * A helper class to implement a ViewGroup that resizes dynamically (by adding padding to the bottom)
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/mozilla/focus/widget/ResizableKeyboardViewDelegate.java
```diff
@@ -84,7 +76,6 @@ public void onGlobalLayout() {
                 0, 0);
 
         try {
-            idOfViewToHide = styleAttributeArray.getResourceId(R.styleable.ResizableKeyboardViewDelegate_viewToHideWhenActivated, -1);
```
jonalmeida, line 87: I think we could use a typed Array here and pass the list of IDs through the XML (since we already have the functionality, and we just need to allow it to work for multiple views. What do you think?

```java
arrayOfViewsToHide = styleAttributeArray.getTextArray(R.styleable.ResizableKeyboardViewDelegate_viewToHideWhenActivated, -1);

// ...

if (arrayOfViewsToHide != null && !arrayOfViewsToHide.isEmpty()) {
    // loop through array and setVisibility
}
```

Then in an `arrays.xml`:

```xml
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <array name="viewsToHide">
        <item>@id/erase</item>
        <item>@id/tabs</item>
    </array>
</resources>
```

The final usage in the layout file:

```xml
<org.mozilla.focus.widget.ResizableKeyboardCoordinatorLayout
    app:viewToHideWhenActivated="@array/viewsToHide"
    android:layout_marginTop="0dp"
    android:id="@+id/main_content"
```

app/src/main/java/org/mozilla/focus/widget/ResizableKeyboardViewDelegate.java
```diff
@@ -147,4 +150,13 @@ private int calculateDifferenceBetweenHeightAndUsableArea() {
 
         return delegateView.getResources().getDisplayMetrics().heightPixels - rect.bottom;
     }
+
+    private void updateDynamicViewsVisibility(int visibility) {
+        if (!arrayOfViewsToHide.isEmpty()) {
+            for (Integer viewId : arrayOfViewsToHide) {
+                View viewToHide = delegateView.findViewById(viewId);
+                viewToHide.setVisibility(visibility);
```
jonalmeida, line 158: Should we check if this is nullable before calling this?

Kotlin has changed the way I think about views. :)

jonalmeida (approved):
I haven't tested the code, but I assume you have. In which case, LGTM. 🙂

app/src/main/java/org/mozilla/focus/widget/ResizableKeyboardViewDelegate.java
ionutbedregeanu: Yep, we should. Thanks!

PR #4977, opened 2021-07-08
Bring back shared preference exceptions storage.

Amejia481:
As mentioned on https://github.com/mozilla-mobile/focus-android/pull/4899#pullrequestreview-702210325 GV is not persisting TP exceptions so after app restart, as a workaround we need to store then in the previous shared preference store and load them into GV on startup.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/Components.kt
+++ b/app/src/main/java/org/mozilla/focus/Components.kt
@@ -7,6 +7,9 @@ package org.mozilla.focus
 import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
 import mozilla.components.browser.state.engine.EngineMiddleware
 import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.concept.engine.DefaultSettings
@@ -35,7 +38,6 @@ import mozilla.components.service.location.LocationService
 import mozilla.components.service.location.MozillaLocationService
 import org.mozilla.focus.activity.MainActivity
 import org.mozilla.focus.components.EngineProvider
-import org.mozilla.focus.telemetry.GleanMetricsService
 import org.mozilla.focus.downloads.DownloadService
 import org.mozilla.focus.engine.ClientWrapper
 import org.mozilla.focus.engine.LocalizedContentInterceptor
@@ -47,6 +49,7 @@ import org.mozilla.focus.search.SearchMigration
 import org.mozilla.focus.state.AppState
 import org.mozilla.focus.state.AppStore
 import org.mozilla.focus.state.Screen
+import org.mozilla.focus.telemetry.GleanMetricsService
 import org.mozilla.focus.telemetry.TelemetryMiddleware
 import org.mozilla.focus.utils.Settings
 import java.util.Locale
@@ -154,7 +157,10 @@ class Components(
         }
 
         val exceptionsMigrator = EngineProvider.provideTrackingProtectionMigrator(context)
-        exceptionsMigrator.start(context)
+
+        GlobalScope.launch(Dispatchers.IO) {
+            exceptionsMigrator.start(context)
+        }
     }
 }
 
--- a/app/src/main/java/org/mozilla/focus/Components.kt
+++ b/app/src/main/java/org/mozilla/focus/Components.kt
@@ -7,9 +7,6 @@ package org.mozilla.focus
 import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.launch
 import mozilla.components.browser.state.engine.EngineMiddleware
 import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.concept.engine.DefaultSettings
@@ -42,6 +39,7 @@ import org.mozilla.focus.downloads.DownloadService
 import org.mozilla.focus.engine.ClientWrapper
 import org.mozilla.focus.engine.LocalizedContentInterceptor
 import org.mozilla.focus.engine.SanityCheckMiddleware
+import org.mozilla.focus.exceptions.ExceptionMigrationMiddleware
 import org.mozilla.focus.locale.LocaleManager
 import org.mozilla.focus.notification.PrivateNotificationMiddleware
 import org.mozilla.focus.search.SearchFilterMiddleware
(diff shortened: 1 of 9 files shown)


pocmo (commented):
The patch looks good. But when I kill the app then there are not exceptions anymore on the next launch.

Amejia481:
As we are reading the list on MAIN, strict mode is cause the migration to happen and as a result exceptions are not in GV. If I disabled strict mode I'm able to see the exceptions working.

Amejia481:
I think in this case we can to make this part **sync** as we don't want users to load sites before loading the exceptions to GV first.

Amejia481:
I added this commit https://github.com/mozilla-mobile/focus-android/pull/4977/commits/2eb6ce1937927f4a2fcff9495e58911740cf82ee to skip the strict mode policies for the TP migrator.

Amejia481:
Are sure that having the migrator here is always going to get executed on startup? https://github.com/mozilla-mobile/focus-android/blob/75076623d52fbaad5893710ca21b16686552c66f/app/src/main/java/org/mozilla/focus/FocusApplication.kt#L64

Amejia481:
With this approach with have an issue, a Tracking protection exceptions are not persistent on GeckoView, they only
 live while sessions are alive, we have to me sure we reload them when a session ends.
 
 For example, when a user start fresh we load all the exceptions into GV but when the user hit the trash icon the session gets killed and GV understand that previous TP exceptions are not valid anymore, at that point we have to reload the exceptions into GV, what do you think?

app/src/main/java/org/mozilla/focus/Components.kt
```diff
@@ -155,6 +157,7 @@ class Components(
 
         val exceptionsMigrator = EngineProvider.provideTrackingProtectionMigrator(context)
         exceptionsMigrator.start(context)
+        StrictMode.setThreadPolicy(threadPolicy)
```
pocmo, line 160: We could also take this off the main thread, I guess? Reading the shared preference is thread-safe and `setPermission` has an `@AnyThread` annotation. Only the engine needs to be created on the main-thread, but since we are calling `components.engine.warmUp()` in `FocusApplication` before running this migration, that has already happened.

pocmo (commented):
I am somehow still seeing problems, where I add an exception and on a fresh app start there are none. 🤔

Amejia481:
> I am somehow still seeing problems, where I add an exception and on a fresh app start there are none. 🤔

When that happen could you check for the tag `GeckoExceptionsMigrator` on logcat? It's very likely the migrator is not running for this case and the exceptions are not getting loaded on GV, on start up we should always see these entries in logcat, if not the exceptions are not getting loaded at that time.

`GeckoExceptionsMigrator Starting tracking protection exceptions migration`
`GeckoExceptionsMigrator Finished tracking protection exceptions migration`

app/src/main/java/org/mozilla/focus/Components.kt
Amejia481: I'm getting exception from getStorageController() when running out of Main :( 

```
    java.lang.IllegalThreadStateException: Expected thread 2 ("main"), but running on thread 15965 ("DefaultDispatcher-worker-3")
        at org.mozilla.gecko.util.ThreadUtils.assertOnThreadComparison(ThreadUtils.java:110)
        at org.mozilla.gecko.util.ThreadUtils.assertOnThread(ThreadUtils.java:86)
        at org.mozilla.gecko.util.ThreadUtils.assertOnUiThread(ThreadUtils.java:77)
        at org.mozilla.geckoview.GeckoRuntime.getStorageController(GeckoRuntime.java:855)
        at org.mozilla.focus.exceptions.GeckoExceptionsMigrator.migrate(GeckoExceptionsMigrator.kt:56)
        at org.mozilla.focus.exceptions.GeckoExceptionsMigrator.start(GeckoExceptionsMigrator.kt:27)
        at org.mozilla.focus.Components.migrateTrackingProtectionExceptions(Components.kt:158)
        at org.mozilla.focus.FocusApplication$onCreate$1.invokeSuspend(FocusApplication.kt:62)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

Amejia481: I changed a bit the order in which we are getting the `storageController` and just take `start` of the main thread. 

https://github.com/mozilla-mobile/focus-android/pull/4977/files#diff-af021628211199f43ada47c8919e6cb79a1205a19ed9abbe349a2d01108cfd44R161

app/src/main/java/org/mozilla/focus/exceptions/ExceptionMigrationMiddleware.kt
```diff
@@ -0,0 +1,65 @@
...
+        context: MiddlewareContext<BrowserState, BrowserAction>,
+        next: (BrowserAction) -> Unit,
+        action: BrowserAction
+    ) {
+        next(action)
+
+        when (action) {
+            is TabListAction -> checkIfMigrationIsNeeded(context.state)
+            else -> {}
+        }
```
Amejia481, line 41: nit: Maybe `if` as we just want to identify `if(action is TabListAction)` only.

Amejia481 (commented):
This looks good ✅, I can't approve as I summited the PR  😄, but you can approve your changes :)

app/src/main/java/org/mozilla/focus/exceptions/ExceptionMigrationMiddleware.kt
pocmo: Good point, this code went through several iterations with different numbers of actions. :)

pocmo (approved):

app/src/main/java/org/mozilla/focus/exceptions/ExceptionMigrationMiddleware.kt
Amejia481: I just imagine how fun was that process 😄

pocmo:
@Mergifyio backport releases_v91.0

Issue #4923, opened 2021-06-09
Add Noodle Shape as Focus background


amedyne:
Designs refer to here: https://www.figma.com/file/2V6wgf9fRFdYfLaf5c3oPj/MVP---iOS%2FAndroid---Product-UI

ionutbedregeanu:
I want to work on this

ionutbedregeanu:
New design: https://www.figma.com/file/jgsM2hWimZbPmtUJn7IGIB/Focus-Landing-Page-Design-(Kate)?node-id=1648%3A87291

lobontiumira:
Verified on the 93.0 (build 11 GV 92.0a1-20210730093652) with Google Pixel (Android 10):

![noodle](https://user-images.githubusercontent.com/40694158/128199246-43e24775-c54b-4473-b3c4-ccf4adaa5b63.png)

PR #4987, opened 2021-07-13
Update Focus background

ionutbedregeanu:
For #4923

Update Focus background
Figma design: https://www.figma.com/file/jgsM2hWimZbPmtUJn7IGIB/Focus-Landing-Page-Design-(Kate)?node-id=1648%3A87291

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
@@ -312,7 +312,12 @@ class UrlInputFragment :
         if (isOverlay) {
             keyboardLinearLayout?.visibility = View.GONE
         } else {
-            backgroundView?.setBackgroundResource(R.drawable.background_gradient)
+            val backgroundId = if (FeatureFlags.isMvp) {
+                R.drawable.dark_background
+            } else {
+                R.drawable.background_gradient
+            }
+            backgroundView?.setBackgroundResource(backgroundId)
 
             dismissView?.visibility = View.GONE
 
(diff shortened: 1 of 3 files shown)


jonalmeida:
I don't think we should add PNGs for the background. How would they look for different screen sizes? In landscape mode?

Using PNGs is also a large amount of memory that we would need to allocate.

Looking at the existing code, we use [a shape for the current gradient](https://github.com/mozilla-mobile/focus-android/blob/main/app/src/main/res/drawable/background_gradient.xml). Since our new background is much more complex, wonder if we could export the SVG from the figma designs and use that as a background drawable. I haven't tested this out, but it's hunch. Happy to help try it out tomorrow if you run into problems. 🙂

ionutbedregeanu:
> I don't think we should add PNGs for the background. How would they look for different screen sizes? In landscape mode?
> 
> Using PNGs is also a large amount of memory that we would need to allocate.
> 
> Looking at the existing code, we use [a shape for the current gradient](https://github.com/mozilla-mobile/focus-android/blob/main/app/src/main/res/drawable/background_gradient.xml). Since our new background is much more complex, wonder if we could export the SVG from the figma designs and use that as a background drawable. I haven't tested this out, but it's hunch. Happy to help try it out tomorrow if you run into problems. 🙂

@jonalmeida Totally agree with you, but as you said our new background is more complex and I had some vector import errors. That's why I chose to import the background as png and use  a tool for lossless compression(pngcrush)
Maybe I should ask design team for some help

Screenshot 2021-07-13 at 15 50 22

jonalmeida:
Spoke with UX, I think we're probably going to get a new resource for this since vector drawables don't support all those SVG properties.

ionutbedregeanu:
> Spoke with UX, I think we're probably going to get a new resource for this since vector drawables don't support all those SVG properties. Let's punt on this until there's a solution that works for us/UX.

I made a new  update and seems all good

jonalmeida (approved):
Brilliant, looks so good! 🎉 ✨

Issue #4930, opened 2021-06-10
Browser: Update Loading/Progress bar 


amedyne:
### Acceptance Criteria (how do I know when I’m done?)
Conform to new design and colors.

ionutbedregeanu:
QA note:
All ticket changes are implemented behind a feature flag(FeatureFlags.isMvp).
FeatureFlags.isMvp = true -> mvp changes are visible
FeatureFlags.isMvp = false -> mvp changes are hidden

lobontiumira:
Verified on the debug build8.13.1 (1191.0a1-20210705095222) with Google Pixel (Android 10):

![Untitled](https://user-images.githubusercontent.com/40694158/126650226-ab9a9495-ce76-4725-9955-ea87909379a3.png)

PR #4989, opened 2021-07-14
Update loading progress bar gradient

ionutbedregeanu:
For #4930

Update loading progress bar gradient
New color resources added to use a different gradient for progress bar in dark theme

Because the app theme was changed to DayNight it automatically load dark colors for some attributes when the app is in light mode but for now we don't want that.
Those attributes will be updated at light theme implementation

### DIFF
--- a/app/src/main/res/values-night/colors.xml
+++ b/app/src/main/res/values-night/colors.xml
@@ -4,4 +4,7 @@
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <resources>
     <color name="snackbarBackground">#2E255D</color>
+    <color name="colorProgressGradientStart">#FF9059FF</color>
+    <color name="colorProgressGradientCenter">#FFFF4AA2</color>
+    <color name="colorProgressGradientEnd">#FFFFBD4F</color>
 </resources>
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -42,6 +44,10 @@ open class MainActivity : LocaleAwareAppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        if (MvpFeatureManager.isEnabled) {
```
jonalmeida, line 47: Let's rebase this PR since you mentioned we don't need this any more.

jonalmeida (approved):
Apart from a rebase, LGTM! 🙂

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
ionutbedregeanu: From what I noticed on [this pr](https://github.com/mozilla-mobile/focus-android/pull/4956) I'm still constrained to keep this to load dark theme resources when FeatureFlags.isMvp= true

Issue #5004, opened 2021-07-22
Update feature flag "isMvp"


ionutbedregeanu:
### Why/User Benefit/User Problem
It helps QA team to see the mvp changes without manually updating the flag
### What / Requirements
Update const val isMvp = false/true to var isMvp = BuildConfig.BUILD_TYPE.contains("debug")

ionutbedregeanu:
Note for QA: After this patch you should be able to see all mvp changes on DEBUG build

lobontiumira:
Verified on the debug build 8.13.1 (1192.0a1-20210721092353) with Google Pixel (Android 10).

PR #5005, opened 2021-07-22
Update FeaturesFlag.isMvp

ionutbedregeanu:
For #5004

Update const val isMvp = false/true to var isMvp = BuildConfig.BUILD_TYPE.contains("debug")
It helps QA team to see the mvp changes without manually updating the flag

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/FeatureFlags.kt
+++ b/app/src/main/java/org/mozilla/focus/utils/FeatureFlags.kt
@@ -3,9 +3,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 package org.mozilla.focus.utils
 
+import org.mozilla.focus.BuildConfig
+
 /**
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/utils/FeatureFlags.kt
```diff
@@ -3,9 +3,11 @@
...
 package org.mozilla.focus.utils
 
+import org.mozilla.focus.BuildConfig
+
 /**
  * Feature flag for MVP
  */
 object FeatureFlags {
-    const val isMvp = false
+    var isMvp = BuildConfig.BUILD_TYPE.contains("debug")
```
Mugurell, line 12: nit: Wouldn't `val isMvp = BuildConfig.DEBUG` be enough?
Same thing, in a bit shorter format.

Mugurell (approved):

PR #5008, opened 2021-07-23
Update Android Components version

jonalmeida:


jonalmeida:
> Cannot inline bytecode built with JVM target 1.8 into bytecode that is being built with JVM target 1.6. Please specify proper '-jvm-target' option

Probably related to the gradle/Java 11 bump in AC. Hmm..

jonalmeida:
>   /home/runner/work/focus-android/focus-android/service-telemetry/src/main/java/org/mozilla/telemetry/measurement/ExperimentsMapMeasurement.kt:18: Error: Call requires API level 24 (current min is 21): java.util.Map#forEach [NewApi]
          experiments.forEach { entry ->
                      ~~~~~~~
  

This would be true if we were using the Java API, but we're using the Ktolin one which does not have the same API limitations. 🤔

jonalmeida:
I remember now: we had the same problem in AC a few years ago. There is a misunderstanding of which API is used when upgrading to Java 8.

We can be explicit to use the Kotlin one by changing:

```java
collection.forEach { key, value ->
  // apply transformation
}
```

Into:
```kotlin
collection.forEach { (key, value) ->
  // apply transformation
}
```

jonalmeida:
> EDIT 2: It seems like the lint bug fix is to upgrade it to gradle build tools to `4.2.1`, but that requires gradle version and target SDK which is a bigger can of worms than this PR.

I filed https://github.com/mozilla-mobile/focus-android/issues/5009 for this.

Issue #5012, opened 2021-07-27
The tracking protection of a page cannot be enabled back until deleted from Exceptions


lobontiumira:
## Steps to reproduce
1. Open a page.
2. From the three-dot menu, disable the tracking protection - the page reloads with the tracking protection icon cut.
3. From the three-dot menu, enable the tracking protection.

### Expected behavior
The page reloads with the tracking protection icon uncut, and the tracking protection toggle ON.

### Actual behavior
The page reloads with the tracking protection icon still cut, and the tracking protection toggle is still OFF.

### Device information
* Android device: HTC 10 (Android 8), Google Pixel (Android 10)
* Focus version: 8.18.0-rc.1

https://user-images.githubusercontent.com/40694158/127102374-e0f28055-f4c9-4e5d-a026-ae6202de3bdd.mp4

mcarare:
@amedyne This looks like something we want to fix before we ship #4931 because it affects the same functionality we agreed to implement for the ETP toggle.

Amejia481:
I'm taking a look I'll keep the issue posted

mcarare:
@Amejia481 I fixed this issue for the etp panel in this commit: https://github.com/mozilla-mobile/focus-android/pull/5019/commits/3f1cf8f820bd818b80eca5572f081360048fd1be.
It also fixes the list persistence between app runs.  Let me know if there is a better approach.

pocmo:
@mcarare Can you move this to a separate PR, so that we can review it independently? :)

pocmo:
Opened PR #5061 to update A-C, which should come with a fix for this.

lobontiumira:
Verified as fixed on the 91.1.0 build with Google Pixel (Android 10), and OnePlus 5T (Android 10).

PR #5037, opened 2021-07-30
For #5012 - Fix removing exceptions and sync list with engine.

mcarare:


app/src/main/java/org/mozilla/focus/exceptions/ExceptionMigrationMiddleware.kt
```diff
@@ -35,8 +37,10 @@ class ExceptionMigrationMiddleware(
     ) {
         next(action)
 
-        if (action is TabListAction) {
+        if (action is TabListAction || action is ToggleExclusionListAction) {
             checkIfMigrationIsNeeded(context.state)
+        } else if (action is InitAction) {
```
mcarare, line 42: This was added so that the list would be synced if you go directly to settings without opening a tab.

pocmo: Ah, I see. I didn't add it initially to avoid migrating twice, what we would do now if you open a tab?

app/src/main/java/org/mozilla/focus/exceptions/ExceptionMigrationMiddleware.kt
```diff
@@ -35,8 +37,10 @@ class ExceptionMigrationMiddleware(
     ) {
         next(action)
 
-        if (action is TabListAction) {
+        if (action is TabListAction || action is ToggleExclusionListAction) {
```
pocmo, line 40: Why is the toggle action needed here?

app/src/main/java/org/mozilla/focus/ext/SessionState.kt
```diff
@@ -23,3 +25,18 @@ fun SessionState.ifCustomTab(): CustomTabSessionState? {
...
+ * Removes a [TrackingProtectionException] if one is set for the current session.
+ */
+fun SessionState.removeFromExceptions(trackingProtectionUseCases: TrackingProtectionUseCases) {
+    var exceptions: List<TrackingProtectionException>
+    trackingProtectionUseCases.fetchExceptions.invoke {
+        exceptions = it.filter { exception ->
+            exception.url == this.content.url
+        }
+        exceptions.firstOrNull()?.let {
+            trackingProtectionUseCases.removeException(exceptions.first())
```
mcarare, line 39: Not sure why and where if passing the tab `id` the exception is not removed by using the other `removeException` method that has an `id` param. So I used this approach to identify the exception that needs removal. Question: Can this have multiple exceptions return, should we iterate through them instead of just removing the first one?

app/src/main/java/org/mozilla/focus/menu/browser/BlockingItemViewHolder.kt
```diff
@@ -75,7 +77,7 @@ internal class BlockingItemViewHolder(
 
             buttonView.context.components.apply {
                 if (isChecked) {
-                    trackingProtectionUseCases.removeException(fragment.tab.id)
+                    fragment.tab.removeFromExceptions(trackingProtectionUseCases)
```
pocmo, line 80: How's the behavior different from what the use case does? And if that is needed shouldn't the use case do that? CC: @Amejia481

app/src/main/java/org/mozilla/focus/exceptions/ExceptionMigrationMiddleware.kt
mcarare: Maybe we can remove `action is TabListAction` from the first if, so we would not migrate again. We would need this migration just at app start and when toggling etp for a session (ToggleExclusionListAction)

app/src/main/java/org/mozilla/focus/exceptions/ExceptionMigrationMiddleware.kt
mcarare: Without migrating the changes the toggle stays off. Maybe there is a deeper bug in `RemoveExceptionUseCase`?

app/src/main/java/org/mozilla/focus/menu/browser/BlockingItemViewHolder.kt
mcarare: Instead of passing a tab id and identifying the exception to be removed, we identify the exception from the list we fetch.

app/src/main/java/org/mozilla/focus/exceptions/ExceptionMigrationMiddleware.kt
pocmo: Yeah, it should only be needed to migrate once. After that Gecko and our saved list should stay in sync. (Except when the last tab was closed, then we need to re-migrate when the next GeckoSession comes up).

app/src/main/java/org/mozilla/focus/menu/browser/BlockingItemViewHolder.kt
Amejia481, line 80: This looks like it could be an issue on AC/GV, as we should be able to remove the exception using the `taId`

Amejia481: Yeah, completely agree, it looks like this is an AC/GV issue as we should be able to remove the exception by using the `taId`, I'm trying to pinpoint the issue.

mcarare: We use the id in Fenix and it works.

Amejia481: Ok, I found the reason, it looks like [this GV](https://github.com/mozilla-mobile/android-components/blob/dca90e65d62d69db63c7d9547d36f880ecf10770/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/TrackingProtectionExceptionFileStorage.kt#L106) call doesn't work for private sessions. I see the same issue on Fenix on private tabs

Amejia481: Let's put a fix on AC for both

Amejia481: I filed a [ticket for the GV](https://bugzilla.mozilla.org/show_bug.cgi?id=1723280)  team, in the meantime I put temporary fix on [AC](https://github.com/mozilla-mobile/android-components/pull/10728).

app/src/main/java/org/mozilla/focus/ext/SessionState.kt
Amejia481: Do we still need this the AC [fix](https://github.com/mozilla-mobile/android-components/pull/10728)? Is the issue still reproducible?

mcarare: The change is still not in Focus. I will keep this until then, I think we will still need some changes in the migration middleware.

Amejia481: 👍🏽

app/src/main/java/org/mozilla/focus/exceptions/ExceptionMigrationMiddleware.kt
pocmo: We also need it when all `GeckoSession` instances are closed. Once all "private" ones are gone, the private context is gone and the migrated exceptions are lost. That's why we do it before the opening of a new tab.

Issue #5042, opened 2021-08-02
Show total trackers blocked from install


mcarare:
This is a breakout from https://github.com/mozilla-mobile/focus-android/issues/4931 for better tracking and a separate review of the PR.

mcarare:
Closed in https://github.com/mozilla-mobile/focus-android/pull/5049.

PR #5044, opened 2021-08-02
For #5042 - Count total number of blocked trackers.

mcarare:


app/src/main/java/org/mozilla/focus/browser/binding/BlockedTrackersBinding.kt
```diff
@@ -0,0 +1,44 @@
...
+import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifChanged
+
+/**
+ * Binding responsible for updating the total number of tracked blockers since install.
+ */
+class BlockedTrackersBinding(
+    store: BrowserStore,
+    private val tabId: String,
+    private val updateCount: (Int) -> Unit
+) : AbstractBinding<BrowserState>(store) {
```
pocmo, line 23: A `Middleware` may be a better choice and independent from the UI. With the `TrackerBlockedAction` it's also easier to count them. :)

mcarare: I preferred to handle this locally in order to not interfere with release trains for AC and Focus. But If you think using middleware would improve this, I can give it a go tomorrow.

mcarare: I meant I can use the binding locally, I thought you suggested implementing a middleware in AC for this.

pocmo: In theory you could have a middleware on `BrowserStore` (implemented in `focus-android` for now), that counts every `TrackerBlockedAction`, persists the count, and also dispatches an action on `AppStore` (which you can use for reading the state in the UI).

mcarare:
Closed in favor of https://github.com/mozilla-mobile/focus-android/pull/5049.

PR #5049, opened 2021-08-02
For #5042 - Count total number of blocked trackers.

mcarare:


app/src/main/java/org/mozilla/focus/browser/BlockedTrackersMiddleware.kt
```diff
@@ -0,0 +1,48 @@
+package org.mozilla.focus.browser
```
pocmo, line 1: nit: license header

pocmo (approved):
Looks good so far to me.

I still wonder whether we should update a count property in `AppState`/`AppStore`, so that we can observe it live in the UI (which is probably nicer than reading a preference), but that's something we can always add in a later PR. :)

app/src/main/java/org/mozilla/focus/browser/BlockedTrackersMiddleware.kt
mcarare: Thanks for the review!

Issue #5082, opened 2021-08-09
Browser page content doesn't get refreshed with a new page


sv-ohorvath:
## Steps to reproduce
1. Go to a website (e.g. Facebook)
2. Tap the URL bar and type in another URL (twitter.com). Press go
3. Observe the page content after loading.
### Expected behavior
The Twitter webpage should be displayed
### Actual behavior
Still showing the initial webpage (Facebook)
### Device information

* Android device: Pixel 2, Android 9
* Focus version: latest 93.0 debug build from main
* https://github.com/mozilla-mobile/focus-android/issues/4895 started failing with [this commit](https://github.com/mozilla-mobile/focus-android/commit/7d086c0da10371aa012208ccca23c8c737ebed6b)

AndiAJ:
Hi, **verified as fixed** on Focus 93.0 Debug build from main using the following devices:
• Google Pixel 3a (Android 11)
• Huawei Mate 20 Lite (Android 10)
• OnePlus A3 (Android 6.0.1)

►Video
![20210813-163559](https://user-images.githubusercontent.com/51314259/129366132-ee82f9a1-a222-4a12-9004-e049a6148d7b.gif)

Issue #5081, opened 2021-08-09
Buggy typing when replacing an URL


sv-ohorvath:
## Steps to reproduce
1. Open a webpage.
2. Tap the URL bar and type one or 2 characters.
3. Observe the typed chars.
### Expected behavior
Can type in the search bar
### Actual behavior
The first chars are getting deleted.
### Device information

* Android device: Pixel 2, Android 9
* Focus version: latest 93.0 debug build from main

AndiAJ:
Hi, **verified as fixed** on Focus 93.0 Debug build from main using the following devices:
• Google Pixel 3a (Android 11)
• Huawei Mate 20 Lite (Android 10)
• OnePlus A3 (Android 6.0.1)

►Video
![20210813-164125](https://user-images.githubusercontent.com/51314259/129366433-ca5006b5-95a1-414b-9223-867344ba7809.gif)

PR #5085, opened 2021-08-09
For #5081, #5082 - Fix switching to edit mode.

mcarare:
This dispatches EditAction for editing the URL instead of switching toolbar mode.


mcarare:
Left that for future refactoring, but we should pass lambdas instead of the fragment.

jonalmeida (approved):

PR #5092, opened 2021-08-10
Update to Android-Components 92.0.20210809190134

gabrielluong:


app/src/main/java/org/mozilla/focus/session/IntentProcessor.kt
```diff
@@ -84,19 +84,19 @@ class IntentProcessor(
...
                     }
                     intent.hasExtra(TextActionActivity.EXTRA_TEXT_SELECTION) -> createSession(
-                        SessionState.Source.TEXT_SELECTION,
+                        SessionState.Source.Internal.TextSelection,
                         intent,
                         intent.dataString ?: ""
                     )
                     else -> createSession(
-                        SessionState.Source.ACTION_VIEW,
+                        SessionState.Source.External.ActionView(null),
```
gabrielluong, line 99: @grigoryk can you assist on what the proper values we should be passing for the external sources? I am also curious if `null` is acceptable here for Focus.

grigoryk: `null` is fine here, yeah. Basically we'll just be omitting information about the refferrerring app, which is not a problem if the application isn't going to use it. This is also possible in Fenix if it's running on API21.

If you did want to integrate external referrer, you can get at the information about package/category at the activity level: https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt#L106-L127

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -483,7 +483,7 @@ class BrowserFragment :
         } else if (sessionFeature.get()?.onBackPressed() == true) {
             return true
         } else {
-            if (tab.source == SessionState.Source.ACTION_VIEW || tab.isCustomTab()) {
+            if (tab.source is SessionState.Source.External.ActionView || tab.isCustomTab()) {
```
grigoryk, line 486: In Fenix I've changed a similar check to just be `tab.source is SessionState.Source.External`. This covers not just ActionView but also other intents we could be opened with (ActionSend, ActionSearch), and a CustomTab source (assuming it was set correctly). You'd want to verify that `tab.isCustomTab()` will produce the same result as `tab.source is SessionState.Source.External.CustomTab`.

grigoryk (approved):
I didn't see where CustomTab is set as a source. Is that handled by a-c..?

buildSrc/src/main/java/AndroidComponents.kt
```diff
@@ -3,5 +3,5 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 object AndroidComponents {
-    const val VERSION = "92.0.20210803190134"
+    const val VERSION = "92.0.0"
```
jonalmeida, line 6: We don't want to pin `main` to a release version of AC. Similar to Fenix, we want to have release branches that are pinned.

jonalmeida:
> I didn't see where CustomTab is set as a source. Is that handled by a-c..?

Depends what you're looking for, at the entry point of the app, it's [here](https://github.com/mozilla-mobile/focus-android/blob/main/app/src/main/java/org/mozilla/focus/activity/IntentReceiverActivity.kt#L36). At the entry point of the browser, it's [here](https://github.com/mozilla-mobile/focus-android/blob/main/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt#L114).

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
gabrielluong: Changed this to `tab.source is SessionState.Source.External` as recommended. I think `tab.isCustomTab()` is fine as-is since it checks `SessionState` is an instance of `CustomTabSessionState`.

Issue #5102, opened 2021-08-11
`MainDispatcherLoader.<clinit>` unexpectedly takes 209ms to run on COLD MAIN start up


mcomella:
I took a profile of COLD MAIN start up on my Moto G5: https://share.firefox.dev/3iFn6LU

The main thing that appears unnecessarily slow is the `PromptMiddleware.<init>` call: let's investigate!

mcomella:
We see a similar issue in fenix: https://share.firefox.dev/3CPxBEHIt It also does a long, `java.net.URL.openStream` thing as a child of `kotlinx.coroutines.internal.MainDispatcherLoader.<clinit>`. I tried upgrading to the latest stable versions of coroutines and kotlin that jetpack compose would let me do but still no fix.

mcomella:
I looked into the `MainDispatcherLoader.<clinit>` part. Here's the source: https://github.com/Kotlin/kotlinx.coroutines/blob/8c98180f177bbe4b26f1ed9685a9280fea648b9c/kotlinx-coroutines-core/jvm/src/internal/MainDispatchers.kt#L26-L36

It seems like the coroutines library is calling a slow method, `ServiceLoader.load`, which is expected to be optimized by R8. However, if we don't have a recent enough version of R8, I'm guessing this might not be optimized.

mcomella:
It looks like we have [the latest version of the Android Gradle Plugin, v7.0.0](https://developer.android.com/studio/releases/gradle-plugin#7-0-0) so it seems we can't modify the versions to fix this issue. 

Looking at blame of the code I linked above, this coroutine code was written 2 years ago: that implies that maybe the later releases of AGP may have broken this optimization. They also allude to a need for us to manually disable the fast service loader [in the commit that added the code](https://github.com/Kotlin/kotlinx.coroutines/commit/b9b7d828b18b2462817b6ed0c1a56233a4b6fcf4) so maybe there's a configuration issue.

mcomella:
I downgraded AGP to 4.2.0 (and removed compose during a previous debugging attempt) and no longer see the issue: https://share.firefox.dev/3fVvG7H Let's:
- can we just downgrade to AGP 4.2.0 without removing compose and still get the fix?
- does this also fix fenix?
- file an upstream bug against R8
- perhaps document this more broadly (moz perf blog?) so the android community can workaround this issue

mcomella:
[...]

PR #5122, opened 2021-08-12
For #5102: use fast service loader via ProGuard.

mcomella:
For the root cause, see the comments added in the commit.

This improves COLD MAIN time to first frame by 97ms across 10
iterations (medians):
- before: 616.5ms
- after: 519.5ms


jonalmeida (approved):
For my own future memory, we don't need to add the [`ANDROID_DETECTED return true`](https://github.com/mozilla-mobile/fenix/blob/main/app/proguard-rules.pro#L44-L46) bit to proguard because the coroutine library [auto-detects](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/internal/FastServiceLoader.kt#L17) that for us.

mcomella:
> For my own future memory, we don't need to add the [`ANDROID_DETECTED return true`](https://github.com/mozilla-mobile/fenix/blob/main/app/proguard-rules.pro#L44-L46) bit to proguard because the coroutine library [auto-detects](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/internal/FastServiceLoader.kt#L17) that for us.

This might have been a micro-optimization to avoid the `Class.forName` call: I'm not sure how slow these calls are but that call is also used to construct the MainDispatcherFactory in the fast service loader and this impl seems to have a minimal impact so it can't be too slow. I'm fine leaving out this possible optimization and keeping the code simpler.

mcomella:
@Mergifyio backport releases_v92.0

PR #5129, opened 2021-08-13
No issue: Pre-land strings for the design changes.

mcarare:


mcarare:
@gabrielluong Do you prefer to pre-land the new strings for the shortcuts in a separate PR? I can add them to this one, but I  wanted to make sure you don't land duplicates when you put up the PR for the feature.

app/src/main/res/values/strings.xml
```diff
@@ -685,17 +693,21 @@
 
     <!-- Tip displayed on home view explaining how to request the desktop version of a website.
      Argument 1 will be replaced with a line break in the application. -->
-    <string name="tip_request_desktop"><![CDATA[Rather see the full desktop site?%1$s Menu > Request desktop site]]></string>
+    <string name="tip_request_desktop2">Want to see the full version of a site? Switch on Desktop Site from the %1$s menu></string>
```
mcarare, line 696: cc: @Delphine What is the placeholder for? Can you also please check the other strings? Thank you!

app/src/main/res/values/strings.xml
```diff
@@ -715,7 +727,7 @@
     <!-- Preference for exceptions tracking protection -->
     <string name="preference_exceptions">Exceptions</string>
     <!-- Description for the exceptions preference screen-->
-    <string name="preference_exceptions_description">You have disabled Content Blocking for these websites.</string>
+    <string name="preference_exceptions_description2">You have disabled content blocking for these websites.</string>
```
Delphine, line 730: No need to change string ID for this, as each locale handles capitalization differently than English. So, for capitalization changes in English, no need for a new string ID :) Otherwise localizers have to translate  again the entire strings. We usually only change string ID if the sentence really changes in context

app/src/main/res/values/strings.xml
Delphine: I've checked other strings and they look good, thanks for asking!

But concerning your question about the placeholder... I think a dev or UX would be able to reply that for sure.
I really can't remember (and it's good you call it out, because localizers are going to have a hard time understanding it if they don't get context in the comment).

jonalmeida: I think this a bug, the placeholder was probably supposed to be `app_name`, so it would read: "Rather see the full desktop site? Firefox Focus Menu > Request desktop site"

gabrielluong:
> @gabrielluong Do you prefer to pre-land the new strings for the shortcuts in a separate PR? I can add them to this one, but I wanted to make sure you don't land duplicates when you put up the PR for the feature.

Not necessary I have them in a PR.

PR #5135, opened 2021-08-13
Closes #5134: remove unnecessary coroutines proguard rules.

### LINKED ISSUE #5134: Remove unnecessary kotlinx.coroutines proguard rules
In 89cdb96cf305af041ca2ab3131376fdc4e4019df, we added the coroutines lib proguard rules directly to our project. [The docs suggest this config is now bundled with the library](https://github.com/kotlin/kotlinx.coroutines/#r8-and-proguard) so we don't need to add them ourselves:
> R8 and ProGuard rules are bundled into the kotlinx-coroutines-android module. For more details see "Optimization" section for Android.

mcomella:
According to the coroutines docs, these rules are bundled with the
library so we no longer need to add them to our project. These rules
were added in 89cdb96cf305af041ca2ab3131376fdc4e4019df and this commit
is effectively a revert of that commit.

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -108,11 +108,6 @@
 # Kotlinx
 ####################################################################################################
 
--keepnames class kotlinx.coroutines.internal.MainDispatcherFactory {}
--keepnames class kotlinx.coroutines.CoroutineExceptionHandler {}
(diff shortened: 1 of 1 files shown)


jonalmeida (approved):
Thanks for looking into this and cleaning it up!

PR #5136, opened 2021-08-16
For #5001 - Implement Shortcuts for Focus

### LINKED ISSUE #5001: ShortCuts: Implement Shortcuts for Focus
Latest designs: https://www.figma.com/file/KggJlo8pg2T2Q8nn0Ec5c3/Shortcuts?node-id=1836%3A46849

Shortcut tiles are shown in these states:

- New Tab/home/start page. Appears on app startup and after you delete browsing history.

- Search zero screen. Appears while visiting a website, when the address bar is selected –as long as the web address is either unmodified, or erased to blank (hit backspace)

gabrielluong:
Fixes #5001

### DIFF
--- a/app/src/main/java/org/mozilla/focus/Components.kt
+++ b/app/src/main/java/org/mozilla/focus/Components.kt
@@ -32,6 +32,8 @@ import mozilla.components.feature.session.SettingsUseCases
 import mozilla.components.feature.session.TrackingProtectionUseCases
 import mozilla.components.feature.tabs.CustomTabsUseCases
 import mozilla.components.feature.tabs.TabsUseCases
+import mozilla.components.feature.top.sites.PinnedSiteStorage
+import mozilla.components.feature.top.sites.TopSitesUseCases
 import mozilla.components.feature.webcompat.WebCompatFeature
 import mozilla.components.feature.webcompat.reporter.WebCompatReporterFeature
 import mozilla.components.lib.crash.CrashReporter
@@ -59,6 +61,7 @@ import org.mozilla.focus.state.AppStore
 import org.mozilla.focus.state.Screen
 import org.mozilla.focus.telemetry.GleanMetricsService
 import org.mozilla.focus.telemetry.TelemetryMiddleware
+import org.mozilla.focus.topsites.DefaultTopSitesStorage
 import org.mozilla.focus.utils.Settings
 import java.util.Locale
 
@@ -73,7 +76,8 @@ class Components(
     val appStore: AppStore by lazy {
         AppStore(
             AppState(
-                screen = determineInitialScreen(context)
+                screen = determineInitialScreen(context),
+                topSites = emptyList()
             )
         )
     }
@@ -169,6 +173,10 @@ class Components(
     val searchTelemetry: InContentTelemetry by lazy { InContentTelemetry() }
 
     val icons by lazy { BrowserIcons(context, client) }
+
+    val topSitesStorage by lazy { DefaultTopSitesStorage(PinnedSiteStorage(context)) }
+
+    val topSitesUseCases: TopSitesUseCases by lazy { TopSitesUseCases(topSitesStorage) }
 }
 
 private fun determineInitialScreen(context: Context): Screen {
--- a/app/src/main/java/org/mozilla/focus/browser/integration/BrowserMenuController.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/integration/BrowserMenuController.kt
@@ -1,9 +1,18 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
 package org.mozilla.focus.browser.integration
 
+import androidx.annotation.VisibleForTesting
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import mozilla.components.browser.state.selector.findTabOrCustomTabOrSelectedTab
+import mozilla.components.browser.state.state.SessionState
+import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.feature.session.SessionUseCases
+import mozilla.components.feature.top.sites.TopSitesUseCases
 import org.mozilla.focus.menu.ToolbarMenu
 import org.mozilla.focus.state.AppAction
 import org.mozilla.focus.state.AppStore
@@ -13,6 +22,8 @@ import org.mozilla.focus.state.Screen
 class BrowserMenuController(
     private val sessionUseCases: SessionUseCases,
     private val appStore: AppStore,
+    private val store: BrowserStore,
+    private val topSitesUseCases: TopSitesUseCases,
     private val currentTabId: String,
     private val shareCallback: () -> Unit,
     private val requestDesktopCallback: (isChecked: Boolean) -> Unit,
@@ -21,7 +32,14 @@ class BrowserMenuController(
     private val openInCallback: () -> Unit,
     private val openInBrowser: () -> Unit
 ) {
+    @VisibleForTesting
+    private val currentTab: SessionState?
+        get() = store.state.findTabOrCustomTabOrSelectedTab(currentTabId)
 
+    @VisibleForTesting
+    internal var ioScope: CoroutineScope = CoroutineScope(Dispatchers.IO)
+
+    @Suppress("ComplexMethod")
     fun handleMenuInteraction(item: ToolbarMenu.Item) {
         when (item) {
             is ToolbarMenu.Item.Back -> sessionUseCases.goBack(currentTabId)
@@ -30,6 +48,26 @@ class BrowserMenuController(
             is ToolbarMenu.Item.Stop -> sessionUseCases.stopLoading(currentTabId)
             is ToolbarMenu.Item.Share -> shareCallback()
             is ToolbarMenu.Item.FindInPage -> showFindInPageCallback()
+            is ToolbarMenu.Item.AddToShortcuts -> {
+                ioScope.launch {
+                    currentTab?.let { state ->
+                        topSitesUseCases.addPinnedSites(
+                            title = state.content.title,
+                            url = state.content.url
+                        )
+                    }
+                }
+            }
+            is ToolbarMenu.Item.RemoveFromShortcuts -> {
+                ioScope.launch {
+                    currentTab?.let { state ->
+                        appStore.state.topSites.find { it.url == state.content.url }
+                            ?.let { topSite ->
+                                topSitesUseCases.removeTopSites(topSite)
+                            }
+                    }
+                }
+            }
             is ToolbarMenu.Item.RequestDesktop -> requestDesktopCallback(item.isChecked)
             is ToolbarMenu.Item.AddToHomeScreen -> addToHomeScreenCallback()
             is ToolbarMenu.Item.OpenInBrowser -> openInBrowser()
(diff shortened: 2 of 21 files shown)


app/src/main/java/org/mozilla/focus/Components.kt
```diff
@@ -71,9 +74,12 @@ class Components(
     private val clientOverride: Client? = null
 ) {
     val appStore: AppStore by lazy {
-        AppStore(AppState(
-            screen = determineInitialScreen(context)
-        ))
+        AppStore(
+            AppState(
+                screen = determineInitialScreen(context),
+                topSites = topSitesStorage.cachedTopSites
```
pocmo, line 80: nit: Since the app state is a global singleton, I assume topSites will always be an empty list at this point anyways? 🤔

app/src/main/java/org/mozilla/focus/browser/integration/BrowserMenuController.kt
```diff
@@ -21,7 +34,14 @@ class BrowserMenuController(
     private val openInCallback: () -> Unit,
     private val openInBrowser: () -> Unit
 ) {
+    @VisibleForTesting
+    private val currentTab: SessionState?
+        get() = store.state.findTabOrCustomTabOrSelectedTab(currentTabId)
+
+    @VisibleForTesting
+    internal var ioScope: CoroutineScope = CoroutineScope(Dispatchers.IO)
```
pocmo, line 42: It's a bit unfortunate that this scope never gets cancelled (so it's essentially a global scope?). But it looks like this controller has no idea what the lifetime of its menu is.

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -295,12 +302,44 @@ class UrlInputFragment :
         inflater: LayoutInflater,
         container: ViewGroup?,
         savedInstanceState: Bundle?
-    ): View? = if (FeatureFlags.isMvp) {
-        inflater.inflate(R.layout.fragment_urlinput2, container, false)
-    } else {
-        inflater.inflate(R.layout.fragment_urlinput, container, false)
+    ): View? {
+        val view = if (FeatureFlags.isMvp) {
```
pocmo, line 306: nit: This flag is no longer needed, we can get rid of it on `main`.

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -295,12 +302,44 @@ class UrlInputFragment :
...
+            inflater.inflate(R.layout.fragment_urlinput2, container, false)
+        } else {
+            inflater.inflate(R.layout.fragment_urlinput, container, false)
+        }
+
+        val topSites = view.findViewById<ComposeView>(R.id.topSites)
+        topSites.setContent {
```
pocmo, line 313: Does this even work with the old layout where this view doesn't exist? :) But maybe just get rid of the branching based on the flag here.

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -313,6 +352,16 @@ class UrlInputFragment :
...
         )
 
+        topSitesFeature.set(
+            feature = TopSitesFeature(
+                view = DefaultTopSitesView(requireComponents.appStore),
+                storage = requireComponents.topSitesStorage,
+                config = ::getTopSitesConfig
+            ),
+            owner = this,
+            view = view
```
pocmo, line 362: This seems a bit hacky. If I understand it correctly this feature will load top sites and then pass them to the view, and we have a "fake" view implementation that will dispatch it on `AppStore`?

I guess ideally we wouldn't have to go through this `DefaultTopSitesStorage`manually  at all. Dispatch an action to update the state and a middleware updates the storage. But maybe to big of a change right now.

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -768,4 +817,14 @@ class UrlInputFragment :
...
     }
+
+    /**
+     * Returns a [TopSitesConfig] which specifies how many top sites to display.
+     */
+    private fun getTopSitesConfig(): TopSitesConfig {
+        return TopSitesConfig(
+            totalSites = 4,
+            frecencyConfig = null
+        )
```
pocmo, line 828: nit: This looks like something we could just inline? 🤔

app/src/main/java/org/mozilla/focus/home/HomeScreen.kt
```diff
@@ -0,0 +1,84 @@
...
+ * @param onTopSiteClick Invoked when the user clicks on a top site.
+ */
+@Composable
+fun HomeScreen(
+    topSites: List<TopSite>,
+    topSitesMenuItems: List<TopSiteMenuItem>,
+    onTopSiteClick: (TopSite) -> Unit = {}
+) {
+    Column(horizontalAlignment = Alignment.CenterHorizontally) {
+        Spacer(modifier = Modifier.height(24.dp))
```
pocmo, line 33: This is interesting. I guess an alternative could be a top padding modifier on the column? But this looks nice and readable too.

app/src/main/java/org/mozilla/focus/home/HomeScreen.kt
```diff
@@ -0,0 +1,84 @@
...
+            topSites = topSites,
+            menuItems = topSitesMenuItems,
+            onTopSiteClick = onTopSiteClick
+        )
+    }
+}
+
+@Preview(showBackground = true)
+@Composable
+fun DefaultHomeScreen() {
```
pocmo, line 45: Is this used anywhere? The URLs don't seem right. Or is this only for having a preview? Then maybe mark it as `internal`/`private` or with an annotation like `@VisibleForTesting`?

app/src/main/java/org/mozilla/focus/topsites/DefaultTopSitesStorage.kt
```diff
@@ -0,0 +1,89 @@
...
+    private var scope = CoroutineScope(coroutineContext)
+
+    // Cache of the last retrieved top sites
+    var cachedTopSites = listOf<TopSite>()
+
+    override fun addTopSite(title: String, url: String, isDefault: Boolean) {
+        scope.launch {
+            pinnedSitesStorage.addPinnedSite(title, url, isDefault)
+
+            notifyObservers { onStorageUpdated() }
```
pocmo, line 37: (With Jetpack Compose this feature/subscriber mechanism seems super complicated and convoluted for now reason. Maybe that's something we could simplify eventually)

app/src/main/java/org/mozilla/focus/topsites/DefaultTopSitesStorage.kt
```diff
@@ -0,0 +1,89 @@
...
+ */
+class DefaultTopSitesStorage(
+    private val pinnedSitesStorage: PinnedSiteStorage,
+    coroutineContext: CoroutineContext = Dispatchers.IO
+) : TopSitesStorage, Observable<TopSitesStorage.Observer> by ObserverRegistry() {
+
+    private var scope = CoroutineScope(coroutineContext)
+
+    // Cache of the last retrieved top sites
+    var cachedTopSites = listOf<TopSite>()
```
pocmo, line 31: This seems to be read and modified from different threads. Does this need at least `@Volatile`?

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
mcarare: This would crash with the old layout. But as we do not need this anymore on main....

app/src/main/java/org/mozilla/focus/topsites/TopSites.kt
```diff
@@ -0,0 +1,221 @@
...
+    menuItems: List<TopSiteMenuItem>,
+    onTopSiteClick: (TopSite) -> Unit = {}
+) {
+    Row(
+        modifier = Modifier
+            .padding(horizontal = 10.dp)
+            .size(width = 324.dp, height = 84.dp),
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        Row(
```
mcarare, line 58: Is there a reason for having a Row inside another row?

pocmo: https://www.youtube.com/watch?v=6cqbSNERnc0 😂

app/src/main/java/org/mozilla/focus/browser/integration/BrowserMenuController.kt
```diff
@@ -30,6 +50,25 @@ class BrowserMenuController(
...
+            }
+            is ToolbarMenu.Item.RemoveFromShortcuts -> {
+                ioScope.launch {
+                    currentTab?.let { state ->
+                        topSitesStorage.findTopSite(url = state.content.url)?.let { topSite ->
+                            topSitesUseCases.removeTopSites(topSite)
+                        }
+                    }
+                }
+            }
```
mcarare, line 71: These 2 should also trigger a state change. Otherwise, you will still have the same option in the menu for a site you already added.

app/src/main/java/org/mozilla/focus/Components.kt
jonalmeida: Doesn't this also mean we have two caches now? One in the AppStore and one in the top sites storage? Do we need both?

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
pocmo: I just removed it from `main` :)

app/src/main/java/org/mozilla/focus/browser/integration/BrowserMenuController.kt
gabrielluong: The DefaultTopSiteStorage will observe these changes and notify the observers, which will eventually get to our view that will dispatch the state changes.

app/src/main/java/org/mozilla/focus/topsites/DefaultTopSitesStorage.kt
gabrielluong: Removed `cachedTopSites`

app/src/main/java/org/mozilla/focus/Components.kt
gabrielluong: Made AppStore our primary cache.

app/src/main/java/org/mozilla/focus/browser/integration/BrowserMenuController.kt
mcarare: I was talking about the menu that stays the same `hasTopSite` and `canAddTopSite` do not change after these actions.

gabrielluong: I did investigate this, but I don't think it has to do with dispatching the states but rather menu items are cached. This can be a follow up.

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -294,8 +301,40 @@ class UrlInputFragment :
...
+
+        val topSites = view.findViewById<ComposeView>(R.id.topSites)
+        topSites.setContent {
+            val topSitesState = requireComponents.appStore.observeAsComposableState { state -> state.topSites }
+
+            HomeScreen(
+                topSites = topSitesState.value!!,
+                topSitesMenuItems = listOfNotNull(
+                    TopSiteMenuItem(
+                        title = view.context.getString(R.string.remove_top_site),
```
pocmo, line 315: nit: In a ComposeScope you should be able to use `stringResource()`

pocmo (approved):

PR #5139, opened 2021-08-16
Integrate AwesomeBar() composable from compose-awesomebar component.

pocmo:
Additionally needs:
* An updated A-C version (to fix the colors in the awesomebar)
* An update of ktlint to make it understand annotations in function parameters (like we needed for A-C and Fenix)


app/src/main/java/org/mozilla/focus/searchsuggestions/ui/SearchSuggestionsFragment.kt
```diff
@@ -182,110 +158,61 @@ class SearchSuggestionsFragment : Fragment(), CoroutineScope {
...
-                    clickListener)
-        }
+@Composable
+private fun SearchOverlay(
+    viewModel: SearchSuggestionsViewModel
+) {
+    val state = viewModel.state.observeAsState()
+
+    when (state.value) {
+        is State.Disabled -> { /* Not implemented in Compose, still handled by the View implementation */ }
```
MozillaNoah, line 170: Is the intent here to eventually transition the other states over to this Composeable?

MozillaNoah (approved):

app/src/main/java/org/mozilla/focus/searchsuggestions/ui/SearchSuggestionsFragment.kt
pocmo: Yeah, exactly. I didn't want to rewrite all those prompts for now.

app/src/main/java/org/mozilla/focus/searchsuggestions/ui/SearchSuggestionsFragment.kt
```diff
@@ -182,110 +158,61 @@ class SearchSuggestionsFragment : Fragment(), CoroutineScope {
...
+        is State.NoSuggestionsAPI -> { /* Not implemented in Compose, still handled by the View implementation */ }
+        State.ReadyForSuggestions -> SearchSuggestions(viewModel)
+        null -> {}
```
mcarare, line 173: is the `null` branch needed? this needed? `State` has only the 3 "states.

app/src/main/java/org/mozilla/focus/theme/FocusTheme.kt
```diff
@@ -0,0 +1,30 @@
...
+    darkTheme: Boolean = isSystemInDarkTheme(),
+    content: @Composable () -> Unit
+) {
+    MaterialTheme(
+        colors = if (darkTheme) darkColorPalette else lightColorPalette,
+        content = content
+    )
+}
+
+private val darkColorPalette: Colors = darkColors()
```
mcarare, line 27: I think this will add the Material default colors to FocusTheme, not the ones we use. We could add our own colors we already use with something in the lines of :` darkColors (primary = colorResource(id = R.color.primary) )` etc...

mcarare (approved):
LGTM with the added comments.

app/src/main/java/org/mozilla/focus/searchsuggestions/ui/SearchSuggestionsFragment.kt
pocmo: You are right, this shouldn't be needed here. :)

app/src/main/java/org/mozilla/focus/theme/FocusTheme.kt
pocmo: I'll update to the latest A-C version here, which should fix the text color problem. I'll also add the background color here.

jonalmeida:
Rebased and fixed conflicts from my backports that landed in `main`.

pocmo:
I still need to fix a UI test, that doesn't work with the jetpack compose UI since it's looking for views.

pocmo:
I ended up disabling `SearchTest.enableSearchSuggestionOnFirstRunTest()` for now:
* This needs a bit more investigation. Compose code is usually tested in isolation, but here it's mixed into the view tree of the activity. `createAndroidComposeRule()` is supposed to help with that, but it refuses to launch the activity for me.
* It feels less risky: It's pretty obvious if search suggestions would regress and I think we would notice that. So I think we can take some time to rewrite the test.

PR #5140, opened 2021-08-16
Update ktlint

pocmo:
Needed to make ktlint understand annotations in function parameters (like we may use in Compose code, like in https://github.com/mozilla-mobile/focus-android/pull/5139).

Like in Fenix / A-C I also added the `ktlintFormat` to apply the formatting to code and ran it once.


mcarare (approved):
LGTM!

PR #5145, opened 2021-08-16
Add taskgraph

sarah-clements:
This is ready for review now! Note to my team: Aaron and Richard might be pushing some config changes for the ui-tests to try to get them passing (basically, cruft was removed and flank and test files copied over from Fenix but some things need to be tweaked and updated). Shouldn't be a blocker to review since the taskgraph integration part for this pr has been completed, as far as I can tell.


.taskcluster.yml
```diff
@@ -2,6 +2,334 @@ version: 1
 policy:
   pullRequests: public
 tasks:
+    - $let:
+          taskgraph:
+              branch: taskgraph
+              revision: 2b2622598df02bde211d8cedb334b7b22fb883a4
```
ahal, line 8: Please update this to the latest revision:
https://hg.mozilla.org/ci/taskgraph

.taskcluster.yml
```diff
@@ -2,6 +2,334 @@ version: 1
...
+                                      - $if: 'tasks_for == "github-release"'
+                                        then:
+                                            MOBILE_HEAD_TAG: '${event.release.tag_name}'
+                              features:
+                                  taskclusterProxy: true
+                                  chainOfTrust: true
+                              # Note: This task is built server side without the context or tooling that
+                              # exist in tree so we must hard code the hash
+                              image:
+                                  mozillareleases/taskgraph:decision-mobile-682fbaa1ef17e70ddfe3457da3eaf8e776c4a20fe5bfbdbeba0641fd5bceae2a@sha256:bbb2613aaab79d17e590fbd78c072d0643be40fd1237195703f84280ecc3b302
```
ahal, line 240: We'll want to use the latest `decision-mobile` image as well:
https://hub.docker.com/layers/mozillareleases/taskgraph/decision-mobile-0e1cefb4ec46060d53bcd0eb2c0dd296210fa969998af468d693d1c6922efd1d/images/sha256-7cfb913bee636f1ab1477a09f8ad746e1f8c68fab0e2a3e1408c985d2ddc27f7?context=explore

.taskcluster.yml
```diff
@@ -2,6 +2,334 @@ version: 1
...
+                                  - '--task-cwd=/builds/worker/checkouts/src'
+                                  - '--'
+                                  - bash
+                                  - -cx
+                                  - $let:
+                                        extraArgs: {$if: 'tasks_for == "cron"', then: '${cron.quoted_args}', else: ''}
+                                    in:
+                                        $if: 'tasks_for == "action"'
+                                        then: >
+                                            PIP_IGNORE_INSTALLED=0 pip install --user /builds/worker/checkouts/taskgraph &&
```
ahal, line 257: In the `decision-mobile` image, `pip` is Python 2 and `pip3` is Python 3. So please update all these to `pip3`.

taskcluster/focus_android_taskgraph/loader/__init__.py
```diff
@@ -0,0 +1,73 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from __future__ import absolute_import, print_function, unicode_literals
```
ahal, line 5: Since you'll now be running with Python 3, you can update these files here. I like to use the [pyupgrade](https://github.com/asottile/pyupgrade) tool to do it automatically:
```
$ pyupgrade --py36-plus `find taskcluster -name "*.py" -type f`
```

from the root of the repo.

sarah-clements: hrm, should this part use the backticks? find taskcluster -name "*.py" -type f as shown above? I get command not found when I copy and type that verbatim. Also, what exactly is this doing? I took a quick look in the docs but its not clear to me.

ahal: Did you install `pyupgrade`? Otherwise I guess you don't have the `find` command or maybe backticks aren't supported in your shell.

Pyupgrade will automatically re-format your code to use features from the minimum specified python version. So the above will e.g, auto-remove future imports, get rid of `six`, convert `.format()` to `f-strings`, etc. Unfortunately `pyupgrade` only supports modifying a single file a a time, so you need to use shell commands to pass it the list of files to re-format (thus the `find` command).

You could try using `pyupgrade --py36-plus $(find taskcluster -name "*.py" -type f)` instead? If that still doesn't work, try getting the `find` command to work on its own first.

sarah-clements: > You could try using `pyupgrade --py36-plus $(find taskcluster -name "*.py" -type f)` instead? If that still doesn't work, try getting the `find` command to work on its own first.

Aha! That worked.

.taskcluster.yml
```diff
@@ -2,6 +2,334 @@ version: 1
...
+                                            --repository-type="$MOBILE_REPOSITORY_TYPE"
+                                            --tasks-for='${tasks_for}'
+                                            ${extraArgs}
+
+                              artifacts:
+                                  'public':
+                                      type: 'directory'
+                                      path: '/builds/worker/artifacts'
+                                      expires: {$fromNow: '1 year'}
+                                  'public/docker-contexts':
```
sarah-clements, line 288: I'm seeing an error in the decision task at the end of the log stating:
```
[task 2021-08-17T15:42:48.301Z] bash: taskcluster/scripts/decision-install-sdk.sh: No such file or directory
[taskcluster 2021-08-17 15:42:48.532Z] === Task Finished ===
[taskcluster 2021-08-17 15:42:48.640Z] Artifact "public/docker-contexts" not found at "/builds/worker/checkouts/src/docker-contexts"
```

This was copied over from Fenix so is this specific to that repository?

sarah-clements: On line 367 (from commented out config), I see this:
```
artifacts:
  #         public:
  #           type: directory
  #           path: /opt/focus-android/app/build/reports
  #           expires: {$fromNow: '1 week'}
```
Should I be using that path instead for artifacts? @bhearsum please advise.

ahal: I think the real error here is that:
`bash: taskcluster/scripts/decision-install-sdk.sh: No such file or directory`

If I had to guess, this is causing the Decision task to terminate early (but without raising an exception?) before it gets to the point of writing out artifacts.. Then taskcluster notices that we declared an artifact that doesn't exist and complains about it.

I could be wrong though..

bhearsum: That sounds right to me. You'll need to copy in https://github.com/mozilla-mobile/fenix/blob/main/taskcluster/scripts/decision-install-sdk.sh to fix it.

sarah-clements: Two things I noticed: The sdk installation is baked into the dockerfile here: https://github.com/mozilla-mobile/focus-android/pull/5145/files#diff-50d0c8ed55e4948f99160e4e1da1d5bedb8b620b289b724a9eb977fe1e13c458R42

Is the idea that this will be extracted into a separate file in that script instead?

Also, @bhearsum I saw in your VPN pr that you don't install this script in the taskcluster.yml else statement the way fenix does (and what I copied into focus): https://github.com/mozilla-mobile/mozilla-vpn-client/pull/1562/files#diff-a728f7e52d751b98eafa856e45594533339b44f229d7b83f930df335391e7e15R266

I'm curious why, and which is the correct way for Focus? :)

bhearsum: The Dockerfile you've linked to there is the base image for build, test, etc. tasks. The issue you're hitting here is with _decision_ tasks, which run using the Docker image specified at https://github.com/mozilla-mobile/focus-android/pull/5145/files#diff-a728f7e52d751b98eafa856e45594533339b44f229d7b83f930df335391e7e15R240 -- which is why we need to install the SDK there as well. (Decision tasks don't use the same images for reasons I can't totally recall at the moment...)

> Also, @bhearsum I saw in your VPN pr that you don't install this script in the taskcluster.yml else statement the way fenix does (and what I copied into focus): https://github.com/mozilla-mobile/mozilla-vpn-client/pull/1562/files#diff-a728f7e52d751b98eafa856e45594533339b44f229d7b83f930df335391e7e15R266

The reason it's not needed for VPN is because there's nothing the decision task is doing that needs the Android SDK. It may, in the future, when we're doing the Android builds in Taskcluster, but its build system is also significantly different than the browser products, so it's unclear.

Focus's (and the other browsers') build & task scheduling is deeply intertwined with Android and Gradle, so they need both available during the decision task.

sarah-clements: > The Dockerfile you've linked to there is the base image for build, test, etc. tasks. The issue you're hitting here is with _decision_ tasks, which run using the Docker image specified at https://github.com/mozilla-mobile/focus-android/pull/5145/files#diff-a728f7e52d751b98eafa856e45594533339b44f229d7b83f930df335391e7e15R240 -- which is why we need to install the SDK there as well. (Decision tasks don't use the same images for reasons I can't totally recall at the moment...)
> 

hrm, so I moved that dockerfile (that's the only one they have) into `taskcluster/docker/base`. Not sure what file you were linking to in the diff. Was it this? https://github.com/mozilla-mobile/focus-android/pull/5145/files#diff-a1854fd13c43b022659a78b2e7a76c978bcada7d3795c30463ac250520161666R31

> The reason it's not needed for VPN is because there's nothing the decision task is doing that needs the Android SDK. It may, in the future, when we're doing the Android builds in Taskcluster, but its build system is also significantly different than the browser products, so it's unclear.
> 
> Focus's (and the other browsers') build & task scheduling is deeply intertwined with Android and Gradle, so they need both available during the decision task.

I think that makes sense, but this still seemed to work by removing the sdk installation in the else block.

bhearsum: > hrm, so I moved that dockerfile (that's the only one they have) into taskcluster/docker/base. Not sure what file you were linking to in the diff. Was it this? https://github.com/mozilla-mobile/focus-android/pull/5145/files#diff-a1854fd13c43b022659a78b2e7a76c978bcada7d3795c30463ac250520161666R31

I was trying to link to `.taskcluster.yml` actually :). There's a block in there that specifies the docker image to use for the decision task. Let's try this link: https://github.com/mozilla-mobile/focus-android/blob/09d21aaa7e3727b0b540182e052bc81ffd56bfaa/.taskcluster.yml#L239

taskcluster/ci/build/kind.yml
```diff
@@ -0,0 +1,42 @@
...
+        max-run-time: 7200
+        chain-of-trust: true
+
+jobs:
+    debug:
+        description: 'Build Focus/Klar for Android from source code'
+        run-on-tasks-for: [github-pull-request, github-push]
+        run:
+            gradle-build-type: debug
+            gradle-build-name: focusDebug
```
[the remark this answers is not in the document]
pocmo: If we can cut down the build time by running them in separate tasks (like we did in A-C) then I think I'd prefer that.

Same for all build types: If there are no other constraints (resource usage, cost, ..) then I would prefer to have all variants build for PRs (or at the very least when merged). It will make sure that we find beta/release-specific build errors immediately. They don't happen often, but if they happen right before a release then it's always super annoying.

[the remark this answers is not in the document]
sarah-clements: hrm, so if I run `assembleFocusDebug`, which is what those two lines of code above are doing 'under the hood', I see all variants are built, including klar. Is that expected? https://firefox-ci-tc.services.mozilla.com/tasks/MG6UAfEFQaefwhfGVIO58A/runs/0/logs/public/logs/live.log

taskcluster/docker/ui-tests/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+RUN mkdir -p ${TEST_TOOLS} && \
+    mkdir -p ${HOME}/.config/gcloud
+
+RUN curl https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${GOOGLE_SDK_VERSION}.0.0-linux-x86_64.tar.gz --output ${GOOGLE_SDK_DOWNLOAD} \
+    && tar -xvf ${GOOGLE_SDK_DOWNLOAD} -C ${TEST_TOOLS} \
+    && rm -f ${GOOGLE_SDK_DOWNLOAD} \
+    && ${TEST_TOOLS}/google-cloud-sdk/install.sh --quiet \
+    && ${TEST_TOOLS}/google-cloud-sdk/bin/gcloud --quiet components update
+
+# Flank v21.07.1
```
AaronMT, line 34: v21.08.1

taskcluster/docker/ui-tests/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+RUN URL_FLANK_BIN="$($CURL --silent 'https://api.github.com/repos/Flank/flank/releases/latest' | jq -r '.assets[] | select(.browser_download_url | test("flank.jar")) .browser_download_url')" \
```
AaronMT, line 36: Mind replacing `latest` with `48276753` (the previous release) – since this Docker will be recreated, this will download the latest version of Flank with `latest` and it has an issue I'm familiar with. I recently replaced this in Fenix. https://github.com/mozilla-mobile/fenix/commit/bca177541e8e16708fa3bd83c89a8dadc5064094

taskcluster/ci/ui-test/kind.yml
```diff
@@ -0,0 +1,61 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: focus_android_taskgraph.loader.multi_dep:loader
+
+transforms:
+    - focus_android_taskgraph.transforms.multi_dep:transforms
+    - taskgraph.transforms.job:transforms
+    - taskgraph.transforms.task:transforms
+
```
AaronMT, line 11: @sarah-clements The APKs are not being found by Flank. The job-defaults declared in the kind file map the APK file names. The kind file in the Fenix project provides an example of how they are used and setup.  

https://github.com/mozilla-mobile/fenix/blob/6a3adb5aeba976a7146eb96a2ef3794fae909304/taskcluster/ci/ui-test/kind.yml#L21-L31

taskcluster/ci/ui-test/kind.yml
```diff
@@ -0,0 +1,61 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: focus_android_taskgraph.loader.multi_dep:loader
```
ahal, line 5: I could be wrong, but I *think* these tasks should be using the `taskgraph.loader.transform:loader` instead of `multi_dep`. While working on the `signing` kind I kept running into issues with this kind not having the proper keys expected from `multi_dep`.

I don't think this needs to block landing, especially if everything is working as is. But we'll likely need to change it after the fact. Note if we do switch it, we'll need to set up the build dependency manually, e.g:
```
dependencies:
    build: <build-label>
```

ahal: Correspondingly, I think we also need to remove the `multi_dep` transforms from the transform list.

taskcluster/ci/build/kind.yml
```diff
@@ -0,0 +1,97 @@
...
+        run-on-tasks-for: [github-pull-request, github-push]
+        run:
+            gradle-build-type: debug
+            gradle-build-name: klarDebug
+        treeherder:
+            symbol: debug(Bkl)
+
+    focus-release:
+        description: 'Release Focus build for debugging'
+        run-on-tasks-for: [github-pull-request]
```
bhearsum, line 62: I'm pretty sure we'll want to remove this and the other release/beta/nightly builds from pull requests before this lands. (I'm guessing this was just for testing purposes - I just want to make sure we don't land it by accident.)

taskcluster/focus_android_taskgraph/target_tasks.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@_target_task("nightly")
+def target_tasks_nightly(full_task_graph, parameters, graph_config):
+    """Select the set of tasks required for a nightly build."""
+
+    def filter(task, parameters):
+        # We don't want to ship nightly while Google Play is still behind manual review.
+        # See bug 1628413 for more context.
+        return task.attributes.get("nightly", False) and task.kind != "push-apk"
```
bhearsum, line 41: We can drop the special case for `push-apk` now. The noted reason why we skipped is long since fixed. (We did this for Fenix shortly after you copied this code over.)

taskcluster/focus_android_taskgraph/target_tasks.py
```diff
@@ -0,0 +1,90 @@
...
+
+    return [l for l, t in full_task_graph.tasks.iteritems() if filter(t, parameters)]
+
+
+@_target_task("nightly-on-google-play")
```
bhearsum, line 46: ...which means we can drop this target entirely as well.

bhearsum (changes requested):
I know you're still doing a bit of work here, but I wanted to get ahead of the review a bit. This largely looks fine - just a couple of small things to tweak.

taskcluster/ci/ui-test/kind.yml
ahal: Hm, Ben mentioned he may have asked you to use `multi_dep` but can't remember. So I'd vote we leave it as is for this PR and maybe / maybe not change it later.

sarah-clements: Yes, he mentioned that would likely solve the issue of waiting the ui-test task not waiting on the focus-debug build to finish before running. I did that and switched to the job template and that solved that issue.

taskcluster/ci/build/kind.yml
sarah-clements: It was intentional. @pocmo said he wanted to run all builds so that they're not caught by a surprise if something breaks in one of these builds but not on debug. But I believe he said 'on merge' was also okay (please correct me if I'm wrong Sebastian).

bhearsum (approved):
I know you're still working through the last ui-test issue - but this is ready to go other than that as far as I'm concerned.

automation/taskcluster/androidTest/flank-x86.yml
```diff
@@ -0,0 +1,66 @@
...
+
+  auto-google-login: true
+  use-orchestrator: true
+  environment-variables:
+    clearPackageData: true
+  directories-to-pull:
+    - /sdcard/screenshots
+  performance-metrics: true
+
+  test-targets:
```
rpappalax, line 40: These are sufficient for now to get the UI tests running again.  @sv-ohorvath @AaronMT can you please review these (after this pr is merged) and make sure these are the packages we want to execute?

automation/taskcluster/androidTest/ui-test.sh
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env bash
```
rpappalax, line 1: 👍🏼

taskcluster/ci/ui-test/kind.yml
```diff
@@ -0,0 +1,67 @@
...
+        toolchain:
+            - android-sdk-linux
+            - android-gradle-dependencies
+    worker-type: b-android
+    worker:
+        docker-image: {in-tree: ui-tests}
+        max-run-time: 7200
+        env:
+            GOOGLE_APPLICATION_CREDENTIALS: '.firebase_token.json'
+            GOOGLE_PROJECT: 'moz-fx-mobile-firebase-testlab'
```
rpappalax, line 33: this is the old GCP project and won't work.  Should be:
GOOGLE_PROJECT: 'moz-focus-android'

automation/taskcluster/androidTest/flank-x86.yml
```diff
@@ -0,0 +1,66 @@
+# gcloud args match the official gcloud cli
+# https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run
+gcloud:
+  results-bucket: staging.moz-fx-mobile-firebase-testlab.appspot.com
```
rpappalax, line 4: This is the old storage space.  Needs to be changed to:
results-bucket: focus_android_test_artifacts

taskcluster/ci/ui-test/kind.yml
```diff
@@ -0,0 +1,67 @@
...
+kind-dependencies:
+    - toolchain
+    - build
+
+job-defaults:
+    attributes:
+        build-type: debug
+        code-review: true
+        retrigger: true
+    # dependencies:
```
rpappalax, line 20: do we still need these commented deps?

taskcluster/ci/ui-test/kind.yml
```diff
@@ -0,0 +1,67 @@
...
+        using: gradlew
+        dummy-secrets:
+            - content: "faketoken"
+              path: .adjust_token
+        secrets:
+            - name: project/mobile/focus-android/firebase
+              key: firebaseToken
+              path: .firebase_token.json
+              json: true
+        # TODO retrieving these from two separate tasks means they're signed differently; need to create a signing
```
rpappalax, line 54: I assume this TODO is still relevant or was this resolved?

rpappalax:
Note: In regards to UI tests, once 2 issues noted above are resolved, those are ready to land.  A few tests are failing but that's expected and shouldn't block this PR.  We'll follow up on fixing the failing test(s) soon.

rpappalax (approved):

taskcluster/ci/ui-test/kind.yml
sarah-clements: Its a TODO for once this gets merged and we merge ahal's signing kind. So we'll switch over to before, where we use the focusDebug build that ran in a separate task, and then run the androidTest in a separate task to speed things up, but the dep-signing will resolve the issue with the apks being signed differently.

taskcluster/ci/ui-test/kind.yml
sarah-clements: They'll be needed for the artifact retrieval but I can remove it now since its cleaner. The history will be preserved in this pr.

sarah-clements:
@pocmo This pr is ready to merge. I've addressed the issues rpapa is referring to here: https://github.com/mozilla-mobile/focus-android/pull/5145#issuecomment-920792515

ahal:
We're pretty sure (but not 100%) that the errors from the `firefoxci-taskcluster` bot above are happening because it is performing validation based on values from the `main` branch which this PR is invalidating. We pushed this branch up to [staging-focus-android](https://github.com/mozilla-releng/staging-focus-android/runs/3654872751) and the decision task builds correctly.

So I think we'll just have to land this PR and see how it goes. We can always backout if it causes problems.

jonalmeida:
I only have a question here that shouldn't block this landing: there are (rare) times when we need to make release builds from an older tag for security patches or regression testing. Once we enable taskgraph in `main`, will we still be able to create releases on pre-taskgraph branches?

sarah-clements:
> I only have a question here that shouldn't block this landing: there are (rare) times when we need to make release builds from an older tag for security patches or regression testing. Once we enable taskgraph in `main`, will we still be able to create releases on pre-taskgraph branches?

We've kept the release logic to how it was previously (per [here](https://github.com/mozilla-mobile/focus-android/blob/b79f977e7c3cbc296cc89cf843b1c7054c2f12ad/.taskcluster.yml#L100)), so that should remain unaffected. What this pr does is set up the scaffolding to gradually move all release logic (plus add support for Nightly and Beta) to taskgraph and it runs all tests, lints and most builds via taskgraph.

sarah-clements:
> > I only have a question here that shouldn't block this landing: there are (rare) times when we need to make release builds from an older tag for security patches or regression testing. Once we enable taskgraph in `main`, will we still be able to create releases on pre-taskgraph branches?
> 
> We've kept the release logic to how it was previously (per [here](https://github.com/mozilla-mobile/focus-android/blob/b79f977e7c3cbc296cc89cf843b1c7054c2f12ad/.taskcluster.yml#L100)), so that should remain unaffected. What this pr does is set up the scaffolding to gradually move all release logic (plus add support for Nightly and Beta) to taskgraph and it runs all tests, lints and most builds via taskgraph.

Regarding releases from older branches, I don't see why you wouldn't be able to since the release logic is unaffected. Maybe @escapewindow or @bhearsum can confirm.

pocmo:
Merging manually. Will look at updating the mergify rules to work with the new tasks separately once this is in.

Issue #5130, opened 2021-08-13
Intermittent UI test failure - <CustomTabTest.testCustomTabUI>


AndiAJ:
### Firebase Test Run: 
[Firebase link](https://console.firebase.google.com/u/0/project/moz-fx-mobile-firebase-testlab/testlab/histories/bh.2b4ca2c7fdb2f0f0/matrices/9142254862940283030/executions/bs.875192e085ebe0b2/testcases/1)
[Firebase link](https://console.firebase.google.com/u/0/project/moz-fx-mobile-firebase-testlab/testlab/histories/bh.2b4ca2c7fdb2f0f0/matrices/9142254862940283030/executions/bs.d97989a0c391b816/testcases/1)

### Stacktrace:
androidx.test.espresso.NoMatchingViewException: No views in hierarchy found matching: with text: is "Open in Firefox Focus"

View Hierarchy:
+>PopupDecorView{id=-1, visibility=VISIBLE, width=741, height=1026, has-focus=false, has-focusable=true, has-window-focus=true, is-clickable=false, is-enabled=true, is-focused=false, is-focusable=false, is-layout-requested=false, is-selected=false, layout-params={(981,77)(wrapxwrap) gr=TOP START CENTER DISPLAY_CLIP_VERTICAL sim={state=unchanged} ty=APPLICATION_PANEL fmt=TRANSPARENT wanim=0x7f130101 surfaceInsets=Rect(45, 45 - 45, 45) (manual)
fl=SPLIT_TOUCH HARDWARE_ACCELERATED FLAG_LAYOUT_ATTACHED_IN_DECOR
pfl=WILL_NOT_REPLACE_ON_RELAUNCH LAYOUT_CHILD_WINDOW_IN_PARENT_FRAME}, tag=null, root-is-layout-requested=false, has-input-connection=false, x=0.0, y=0.0, child-count=1}
|
+->PopupBackgroundView{id=-1, visibility=VISIBLE, width=741, height=1026, has-focus=false, has-focusable=true, has-window-focus=true, is-clickable=false, is-enabled=true, is-focused=false, is-focusable=false, is-layout-requested=false, is-selected=false, layout-params=android.widget.FrameLayout$LayoutParams@edd9905, tag=null, root-is-layout-requested=false, has-input-connection=false, x=0.0, y=0.0, child-count=1}
|
[...]

Issue #4935, opened 2021-06-10
Intermittent UI test failure - <SafeBrowsingTest.verifyPageSecurityIconAndInfo>


AndiAJ:
### Stacktrace:
java.lang.AssertionError
at org.junit.Assert.fail(Assert.java:86)
at org.junit.Assert.assertTrue(Assert.java:41)
at org.junit.Assert.assertTrue(Assert.java:52)
at org.mozilla.focus.activity.robots.BrowserRobot.verifySiteConnectionInfoIsSecure(BrowserRobot.kt:143)
at org.mozilla.focus.activity.SafeBrowsingTest$verifyPageSecurityIconAndInfo$2.invoke(SafeBrowsingTest.kt:114)
at org.mozilla.focus.activity.SafeBrowsingTest$verifyPageSecurityIconAndInfo$2.invoke(SafeBrowsingTest.kt:21)
at org.mozilla.focus.activity.robots.SearchRobot$Transition.loadPage(SearchRobot.kt:80)
at org.mozilla.focus.activity.SafeBrowsingTest.verifyPageSecurityIconAndInfo(SafeBrowsingTest.kt:111)
at java.lang.reflect.Method.invoke(Native Method)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
at androidx.test.internal.runner.junit4.statement.RunBefores.evaluate(RunBefores.java:80)
at androidx.test.internal.runner.junit4.statement.RunAfters.evaluate(RunAfters.java:61)
at androidx.test.rule.ActivityTestRule$ActivityStatement.evaluate(ActivityTestRule.java:531)
at org.junit.rules.RunRules.evaluate(RunRules.java:20)
at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
[...]

PR #5156, opened 2021-08-17
Fix UI tests #4935, #5130, #5077

sv-ohorvath:
Ran with the fixes on Firebase and all passed.


sv-ohorvath:
Some things got lost from the first commit, need to redo that one.

AaronMT (approved):

PR #5159, opened 2021-08-18
Adds support for window requests and app links

jonalmeida:
We needed to add support for `window=_blank` requests for opening new tabs and app links for sites that redirect to other apps.

I refrained from adding `CustomTabWindowFeature` because it seems like we need to add better support for redirecting intents back to us, similar to Fenix, or some other form of `customTabsUseCases.addTab(url, parentTabId)` - I think we should do that in the following release instead of this one.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
@@ -37,6 +37,7 @@ import mozilla.components.browser.state.state.content.DownloadState
 import mozilla.components.browser.state.state.createTab
 import mozilla.components.browser.toolbar.BrowserToolbar
 import mozilla.components.concept.engine.EngineView
+import mozilla.components.feature.app.links.AppLinksFeature
 import mozilla.components.feature.contextmenu.ContextMenuCandidate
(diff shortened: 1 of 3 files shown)


jonalmeida:
@Mergifyio backport releases_v92.0 main

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -238,6 +243,19 @@ class BrowserFragment :
             tabId = tab.id
         ), this, view)
 
+        appLinksFeature.set(
+            feature = AppLinksFeature(
+                requireContext(),
+                store = components.store,
+                sessionId = tabId,
+                fragmentManager = parentFragmentManager,
+                launchInApp = { true },
```
pocmo, line 252: This will always show a dialog before switching to a third-party app, right?

pocmo (approved):

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
jonalmeida: Yes, I verified that locally.

PR #5160, opened 2021-08-18
Temporarily remove whats new menu item

jonalmeida:
We don't have an accurate SUMO "what's new" link so we're hiding this menu item until we get one, similar to iOS.


jonalmeida:
@Mergifyio backport releases_v92.0

Backporting for 92 as well, and we can see what happens in 93.

mcarare:
> @Mergifyio backport releases_v92.0
> 
> Backporting for 92 as well, and we can see what happens in 93.

In 93 it's moved to settings anyway.

mcarare (approved):

pocmo (approved):

Issue #5057, opened 2021-08-03
FADPV2-2805: Capture Telemetry for Shield Icon - Tracking Protection


amedyne:
https://mozilla-hub.atlassian.net/browse/FADPV2-2805

**Description**

As a Product Manager, I want to see if adding the Shield Icon to the Search bar entices users to modify their tracking protection more frequently. 

**Acceptance Criteria**

Telemetry to Be Collected:

    How often are users clicking the Shield Icon (Tracking Protection) in the Search Bar / Awesome Bar?

    Are users managing Tracking Protection more from the Search Bar Shield Icon or from Settings?

    How many users (percentage of users) are modifying the Enhanced Tracking Protection?

    How many users (percentage of users) are modifying the Advertising, Analytics, Social or Content Trackers from the default on setting to off?

    Do users prefer to have a certain tracker (advertising, analytic, social, content) turned on?

mcarare:
@AndiAJ Please recheck this after https://github.com/mozilla-mobile/focus-android/pull/5322 lands.
Please note that 

> ❌ Enable/Disable TP options from Settings - Nothings is generated

should already worked since it uses the same method that the panel does and those metrics are registered correctly.

AndiAJ:
Hi @mcarare , just re-checked yesterday's inconsistencies and seem to be working properly.

✅  Event Ping 3c8dc805-ca32-4f8c-ad0f-8af1b5ae6259
• Disable Advertising from panel
• Disable Analytics from panel
• Disable Social from panel
• Enable Content from panel
```
"events": [
        {
          "category": "tracking_protection",
          "extra": {
            "is_enabled": "false",
            "source_of_change": "Panel",
            "tracker_changed": "Advertising"
          },
          "name": "tracker_setting_changed",
          "timestamp": 0
        },
        {
          "category": "tracking_protection",
          "extra": {
            "is_enabled": "false",
            "source_of_change": "Panel",
            "tracker_changed": "Analytics"
          },
          "name": "tracker_setting_changed",
          "timestamp": 463
        },
        {
          "category": "tracking_protection",
          "extra": {
            "is_enabled": "false",
            "source_of_change": "Panel",
            "tracker_changed": "Social"
          },
          "name": "tracker_setting_changed",
          "timestamp": 965
        },
        {
          "category": "tracking_protection",
          "extra": {
            "is_enabled": "true",
            "source_of_change": "Panel",
            "tracker_changed": "Content"
          },
[...]

Issue #5056, opened 2021-08-03
FADPV2-2804: Capture Telemetry for Focus "Shortcuts" / Quick Access Sites


amedyne:
https://mozilla-hub.atlassian.net/browse/FADPV2-2804

**Description**

As a Product Manager, I want to see how effective the new feature of Top Sites is on Focus so that I know if it is useful for our users. 

**Acceptance Criteria**

    Collect telemetry data for the following: 

        Do users add shortcuts to their home page? What’s the percentage of users that are adding shortcuts?

        How often are users clicking on the shortcuts?

        Are users adding one, two, three or four shortcuts to their home page? What’s the average?

        How often are users deleting a shortcut?

        How often are users adding a shortcut?

mcarare:
waiting on #5001, #5002, #5003.

AndiAJ:
Hi, I've just checked this matter on a nightly build created from main using a Google Pixel 3a (Android 11)

STR
• Added 4 Shortcuts
• Opened each of them 2x 
• Removed 2 Shortcuts 
• The other 2 Shortcuts remained displayed on the home screen

Metrics Ping 2e293c58-4cca-4f02-bcb3-77df0c392549
✅   `"shortcuts.shortcut_added_counter": 4`
✅   `"shortcuts.shortcut_opened_counter": 8`
✅   ```"shortcuts.shortcut_removed_counter": {
            "removed_from_home_screen": 2
          }```
❌   ```"quantity": {
          "shortcuts.shortcuts_on_home_number": 0
        }```

[Logcat](https://github.com/mozilla-mobile/focus-android/files/7104466/Shortcuts.txt)
[Glean dashboard](https://debug-ping-preview.firebaseapp.com/pings/aj-0903001)

Just to be 100% sure regarding the counting inconsistency for `"shortcuts.shortcuts_on_home_number"` I've made a separate run in which, added 1 shortcut, generated the metrics ping and the count remained set to 0.
Afterwards, added another shortcut, generated the metrics ping and again the count remained set to 0.

@mcarare , please review and share your thoughts ☺️ 
Thanks!

AndiAJ:
Hi, **verified as fixed** on a nightly build created from latest main using the same STR as mentioned above on a Google Pixel 3a (Android 11)

Metrics Ping b1d643b9-b90c-4431-a3c3-6b9edbd041a4
✅   `"shortcuts.shortcut_added_counter": 4`
[...]

PR #5163, opened 2021-08-18
For #5057: Add Telemetry for Tracking Protection.

mcarare:


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -840,6 +844,8 @@ class BrowserFragment :
             }
             sessionUseCases.reload(tab.id)
         }
+
+        TrackingProtection.hasEverChangedEtp.set(true)
```
mcarare, line 848: I have doubts here... I originally set a preference to follow this, but as it changes once for true it will always remain the same. Is this setting the value forever also for the metric, or should it be set along with the others when initializing Glean ?

mcarare:
# Request for data collection review form

**All questions are mandatory. You must receive review from a data steward peer on your responses to these questions before shipping new data collection.**

_1) What questions will you answer with this data?_
This data will provide usage statistics related to the new feature implemented, the tracking protection panel.

_2) Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements?_
Measure whether this feature is used by users and how often and compare it to usage statistics of using controls from the settings screen.

_3) What alternative methods did you consider to answer these questions? Why were they not sufficient?_
This is the only way.

_4) Can current instrumentation answer these questions?_
No. We need a new specific metric for a specific app feature.

_5) List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Firefox/Data_Collection) found on the Mozilla wiki._

_**Note that the data steward reviewing your request will characterize your data collection based on the highest (and most sensitive) category.**_

<table>
  <tr>
    <td>Measurement Description</td>
    <td>Data Collection Category</td>
    <td>Tracking Bug #</td>
  </tr>
  <tr>
    <td>open the panel from shield icon</td>
    <td>Category 2</td>
    <td>https://github.com/mozilla-mobile/focus-android/issues/5056</td>
  </tr>

 <tr>
    <td>tracking protection changed for a site</td>
    <td>Category 2</td>
    <td>https://github.com/mozilla-mobile/focus-android/issues/5056</td>
  </tr>

  <tr>
    <td>Open a shorcut</td>
    <td>Category 2</td>
    <td>https://github.com/mozilla-mobile/focus-android/issues/5056</td>
  </tr>

  <tr>
    <td>Enabling or disabling specific tracker from panel or settings </td>
    <td>Category 2</td>
    <td>https://github.com/mozilla-mobile/focus-android/issues/5056</td>
  </tr>

  <tr>
    <td>Individual tracker </td>
    <td>Category 2</td>
    <td>https://github.com/mozilla-mobile/focus-android/issues/5056</td>
  </tr>
</table>

_6) How long will this data be collected?  Choose one of the following:_
Until 2022-08-20.

_7) What populations will you measure?_
All release channels and locales.

_8) If this data collection is default on, what is the opt-out mechanism for users?_
 Users can opt of of data collection by disabling Usage and technical data from Settings -> Privacy and security -> Data choices.

_9) Please provide a general description of how you will analyze this data._
 Glean 

_10) Where do you intend to share the results of your analysis?_
Only on Glean, mobile teams and BD have internal access.

_12) Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection?_
No.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
travis79: If this is only ever set once, you can make this metric a `lifetime: user` in the metrics.yaml and Glean will maintain the state for you and you shouldn't have to set it again. Be aware, that there currently isn't a way to "un-set" a user lifetime metric.

app/metrics.yaml
```diff
@@ -130,3 +130,130 @@ settings_screen:
...
+      The user has changed the setting for enhanced tracking protection.
+    bugs:
+      - https://github.com/mozilla-mobile/focus-android/issues/5057
+    data_reviews:
+      - https://github.com/mozilla-mobile/focus-android/pull/5163
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - android-probes@mozilla.com
+    expires: "2022-08-17"
```
travis79, line 162: If you added an [event "extra"](https://mozilla.github.io/glean/book/reference/metrics/event.html#extra-metric-parameters) to this, you could track the new value that it was changed to in the "extra" info.

app/metrics.yaml
```diff
@@ -130,3 +130,130 @@ settings_screen:
...
+
+  has_ever_changed_etp:
```
travis79, line 164: Per my other comment, you can make this a [`lifetime: user`](https://mozilla.github.io/glean/book/user/metrics/adding-new-metrics.html#user) and you wouldn't have to set it with every execution, just once and Glean would remember the value for you.

app/metrics.yaml
```diff
@@ -130,3 +130,130 @@ settings_screen:
...
+    extra_keys:
+      source_of_change:
+        description: The source of interaction, "Panel" or "Settings"
+        type: string
+      tracker_changed:
+        description: |
+          The tracker changed, "Advertising", "Analytics", "Social", "Content"
+        type: string
+
+  has_social_blocked:
```
travis79, line 201: Since this will be sent on the metrics ping, this will show up in the data as the last value that was recorded before the metrics ping is sent (which is only once a day). If you wanted to know this for every "session", you would need to add a custom ping and include this metric in it, and then you can transmit the custom ping on going to background (or whenever this metric is recorded during the app execution)

app/metrics.yaml
```diff
@@ -130,3 +130,130 @@ settings_screen:
...
+      - https://github.com/mozilla-mobile/focus-android/issues/5057
+    data_reviews:
+      - https://github.com/mozilla-mobile/focus-android/pull/5163
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - android-probes@mozilla.com
+    expires: "2022-08-17"
+
+  has_advertising_blocked:
```
travis79, line 216: This is similar to the `has_social_blocked` above, see the comment there for tips on recording user settings like this.

app/metrics.yaml
```diff
@@ -130,3 +130,130 @@ settings_screen:
...
+      - https://github.com/mozilla-mobile/focus-android/issues/5057
+    data_reviews:
+      - https://github.com/mozilla-mobile/focus-android/pull/5163
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - android-probes@mozilla.com
+    expires: "2022-08-17"
+
+  has_analytics_blocked:
```
travis79, line 231: This is similar to the `has_social_blocked` above, see the comment there for tips on recording user settings like this.

app/metrics.yaml
```diff
@@ -130,3 +130,130 @@ settings_screen:
...
+      - https://github.com/mozilla-mobile/focus-android/issues/5057
+    data_reviews:
+      - https://github.com/mozilla-mobile/focus-android/pull/5163
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - android-probes@mozilla.com
+    expires: "2022-08-17"
+
+  has_content_blocked:
```
travis79, line 246: This is similar to the `has_social_blocked` above, see the comment there for tips on recording user settings like this.

app/metrics.yaml
```diff
@@ -130,3 +130,130 @@ settings_screen:
...
+      - android-probes@mozilla.com
+    expires: "2022-08-17"
+    extra_keys:
+      source_of_change:
+        description: The source of interaction, "Panel" or "Settings"
+        type: string
+      tracker_changed:
+        description: |
+          The tracker changed, "Advertising", "Analytics", "Social", "Content"
+        type: string
```
travis79, line 199: Curious why you don't also include the value that the setting was changed to here?

travis79 (commented):
This review is from my perspective as a telemetry peer and includes some suggestions to help you get the information you are looking for from the Glean metrics.

I'll await any questions or comments before I continue to review this from a data review perspective.

app/metrics.yaml
mcarare: I think we would want to have this as an average because we are trying to ascertain:  

> Do users prefer to have a certain tracker (advertising, analytic, social, content) turned on?

So we want to know if, on average there is a preference for having one of the trackers on or off, as opposed to the others.

app/metrics.yaml
mcarare: Responded above.

app/metrics.yaml
mcarare: Responded above.

app/metrics.yaml
mcarare: Responded above.

app/metrics.yaml
mcarare: I added  `isEnabled` extra

app/metrics.yaml
mcarare: I added  `isEnabled` extra.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
mcarare: Thanks! I added `lifetime: user` for this metric, we don't need to set it ever to `false`. 
One question, though... I did not see this in the documentation. Does it have a default value? Meaning if we do not set it, is it being sent with a default value?

Issue #5152, opened 2021-08-17
Replace Focus icons with AC icons


ionutbedregeanu:
### Why/User Benefit/User Problem
Remove all icons from Focus which are available in AC
Import AC icons in order to restore the current design

ionutbedregeanu:
Blocked by [10621](https://github.com/mozilla-mobile/android-components/issues/10621)

ionutbedregeanu:
Note for QA: To test this ticket you need to check if the browser menu have the same icons as in [Figma](https://www.figma.com/file/2V6wgf9fRFdYfLaf5c3oPj/Product-UI?node-id=2133%3A282824) for both scenarios(normal tab, custom tab)

lobontiumira:
Verified on the latest debug build from main from 8/25, with Lenovo Tab M10 (Android 10):

![AC icons](https://user-images.githubusercontent.com/40694158/130756981-0ebdc05e-19d4-439f-8f8e-75e92e206a3b.png)

**Note: I've filed #5212 for the extra space in custom tab's toolbar.

lobontiumira:
Verified as fixed also on RC 93.1.0 with Lenovo Tab M10 (Android 10).

PR #5175, opened 2021-08-19
Replace Focus icons with AC icons

ionutbedregeanu:
#### For #5152

Remove all menu icons which can be imported from AC
Some of the icons were also used in the old implementation of browser menu, implementation which is not used anymore, so I decided to delete it


jonalmeida (approved):
Thanks for upstreaming the icons!

Issue #5098, opened 2021-08-11
Implement updates and changes for Context Menu screens


ionutbedregeanu:
### Why/User Benefit/User Problem
New [Figma design](https://www.figma.com/file/2V6wgf9fRFdYfLaf5c3oPj/Product-UI?node-id=2133%3A282824) is available for context menus

### What / Requirements
Update long press image/link context menus according to figma design

ionutbedregeanu:
Please read before testing: https://github.com/mozilla-mobile/android-components/issues/10822#issuecomment-900819370

lobontiumira:
Verified on the latest debug build from main from 8/30 with Lenovo Tab M10 (Android 10), and here are my findings:

![context menus](https://user-images.githubusercontent.com/40694158/131344478-bc7a82be-1175-40a7-8ece-aa67535407e0.png)

Long-tapping on an image, the context menu displays the following options:
- Copy link
- Share link
- Share image
- Open image in new tab
- Save image
- Copy image location

Long-tapping on a link, the context menu displayes the following options:
- Open link in private tab
- Copy link
- Download link
- Share link

Long-tapping on a linked-image, the context menu displayes the following options:
- Open link in private tab
- Copy link
- Download link
- Share link
- Share image
- Open image in new tab
- Save image
- Copy image location

Long-tapping on a link from a page which has an external app, the context menu displayes the following options:
- Open link in private tab
- Copy link
- Download link
- Share link
- Share image
- Open image in new tab
- Save image
- Copy image location
- Open link in external app.

All the context menus display the Focus-specific design element “Saved and shared images **will not be** deleted when you erase Firefox Focus history”. 

The Figma designs are:
![Figma context menus](https://user-images.githubusercontent.com/40694158/131345527-5f15272e-7ab5-4b39-ad2a-263687ae95b9.png)

@ionutbedregeanu please review.
Thank you!

lobontiumira:
Verified also on RC 93.1.0 with Lenovo Tab M10 (Android 10), and noticed that the "Download link" option is missing from the context menus.
![93 1 0](https://user-images.githubusercontent.com/40694158/135239135-eb5b708d-dc2b-459d-b6ce-aa325dd8e822.png)

@ionutbedregeanu is this intended?
Thank you!

ionutbedregeanu:
@lobontiumira Yes, it's intended, in order display that option our selected url should have a specific schema

PR #5178, opened 2021-08-19
Add context menu additional note as string

ionutbedregeanu:
For #5098

Add only the string resource in order to merge it as soon as possible


mcarare (approved):

ionutbedregeanu:
@Delphine FYI

PR #5179, opened 2021-08-19
For 5172: Update whats-new SUMO link.

mcarare:
There is a question regarding different pages for Klar, but that can be added in a different PR if necessary.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/SupportUtils.kt
+++ b/app/src/main/java/org/mozilla/focus/utils/SupportUtils.kt
@@ -42,7 +42,7 @@ object SupportUtils {
         AUTOCOMPLETE("autofill-domain-android"),
         TRACKERS("trackers"),
         USAGE_DATA("usage-data"),
-        WHATS_NEW("whats-new-focus-android-8"),
+        WHATS_NEW("whats-new-firefox-focus-android-version-93"),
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/utils/SupportUtils.kt
```diff
@@ -42,7 +42,7 @@ object SupportUtils {
         AUTOCOMPLETE("autofill-domain-android"),
         TRACKERS("trackers"),
         USAGE_DATA("usage-data"),
-        WHATS_NEW("whats-new-focus-android-8"),
+        WHATS_NEW("whats-new-firefox-focus-android-version-93"),
```
mcarare, line 45: If the 93 is hardcoded, does it mean we would have to change it manually for every new version?

Amejia481: Yes, Is there any particular reason why we have the app hardcode? We have some code that adds the app version dynamically to link https://github.com/mozilla-mobile/focus-android/blob/baa92c122369515b8c1696c4d995c998a338a8b9/app/src/main/java/org/mozilla/focus/utils/SupportUtils.kt#L52

Amejia481 (approved):

PR #5186, opened 2021-08-19
For 5182: Add domain name as fallback to website title.

mcarare:


mcarare:
should we also add this to shorcuts?

jonalmeida:
If it's not too much work, let's.

jonalmeida (approved):

mcarare:
> If it's not too much work, let's.
> 

https://github.com/mozilla-mobile/focus-android/pull/5198

PR #5188, opened 2021-08-20
For 4931: Replace icons for tracking protection.

mcarare:


app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -63,6 +64,21 @@ class BrowserToolbarIntegration(
...
+                )!!,
+                trackingProtectionNothingBlocked = AppCompatResources.getDrawable(
+                    context,
+                    R.drawable.mozac_ic_shield
+                )!!,
+                trackingProtectionException = AppCompatResources.getDrawable(
+                    context,
+                    R.drawable.mozac_ic_shield_disabled
+                )!!
+            )
```
mcarare, line 81: Instead of this, we could update the default in AC, just wanted to make sure we have these changes here regardless.

Amejia481: That sounds good!

Amejia481 (approved):

Issue #5133, opened 2021-08-13
Add onboarding override for performance testing


mcomella:
The onboarding screen shows up on first run and afaict, the only ways to clear it are to:
- Manually clear it by tapping the screen
- (speculative) With a rooted device, put the shared preferences file in the app directory before launching

These aren't great solutions for automated performance testing, in particular, for our page load tests. We should implement an override. Perhaps like fenix:
```
--firefox.android.intentArgument=--ez --firefox.android.intentARgument=performancetest true
```

PR #5232, opened 2021-08-26
For #5133: Allow intent to bypass onboarding screen for perf

MarcLeclair:


app/src/main/java/org/mozilla/focus/perf/Performance.kt
```diff
@@ -0,0 +1,54 @@
...
+import android.os.BatteryManager
+import org.mozilla.focus.state.AppAction
+import org.mozilla.focus.state.AppStore
+import android.provider.Settings as AndroidSettings
+
+object Performance {
+
+    private const val EXTRA_IS_PERFORMANCE_TEST = "performancetest"
+
+    fun processIntentIfPerformanceTest(intent: Intent, context: Context, appStore: AppStore) {
```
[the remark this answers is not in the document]
MarcLeclair: I think that would be better in case it gets refactored / loss. Also, any change in one might be needed in the other. I think its just better for overall code health to have the duplicated code in one place

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -61,6 +62,11 @@ open class MainActivity : LocaleAwareAppCompatActivity() {
 
         val intent = SafeIntent(intent)
 
+        if (Settings.getInstance(this).shouldShowFirstrun() &&
+                !Performance.processIntentIfPerformanceTest(intent, this)){
+            components.appStore.dispatch(AppAction.ShowFirstRun)
```
mcomella, line 67: nitpick (blocking): we need to update the kdoc on `ShowFirstRun` to say it's not just for tests.

mcomella, line 67: thought (non-blocking): this is just a summary of my understanding of this pattern. This is a timeline of the events:
```
appStore.dispatch(ShowFirstRun) // dispatches to appStore dispatcher thread
lifecycle.addObserver(navigator)

navigator.start() -> subscribe() // subscribes to appStore in onStart

navigator.subscribe$collect // changes screens on state change
```

One thing I don't understand is why the navigator is guaranteed to receive the event. Since the dispatch occurs on a background thread, could the state update occur before we subscribe in `onStart`? I'm not familiar with the flow API: maybe it's propagating the initial state to the subscribers and that's why it's safe? Or perhaps it's because the framework (probably) doesn't yield the UI thread until after `onStart` so we have time to subscribe before we collect?

@jonalmeida Do you know if this is safe?

---

A second thought is if there is anything in `HomeActivity.onCreate/onStop/etc.` that shouldn't be called because we've navigated to the onboarding screen. I don't think I see anything. It may be added in the future but I'm not sure how to protect against that.

mcomella (commented):
Besides my open question about race conditions, this seems okay to me but I'd like to defer to jonalmeida's review.

btw, let's squash before we push.

mcomella (approved):
Apparently I requested changes so I have to approve to land this but I'd like to wait for jonalmeida's review before we land this.

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
jonalmeida: >  It may be added in the future but I'm not sure how to protect against that.

Does the intent that we send go directly to the `MainActivity` instead of the `IntentReceiverActivity` first? We could protect against this by processing the intent at the other activity and fire the action immediately before the `MainActivity` is even created.

Just a thought.

jonalmeida (approved):
I've left some comments for thought, but I don't think that should block us today if you need this to land for getting startup testing from intents to work.

app/src/main/java/org/mozilla/focus/perf/Performance.kt
```diff
@@ -0,0 +1,47 @@
...
+import android.content.Intent
+import android.content.IntentFilter
+import android.os.BatteryManager
+import mozilla.components.support.utils.SafeIntent
+import android.provider.Settings as AndroidSettings
+
+/**
+ * A collection of objects related to app performance.
+ */
+object Performance {
```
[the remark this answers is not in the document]
MarcLeclair: Agreed, I think @mcomella  pointed that above. It is a duplicated class

Issue #5185, opened 2021-08-19
Firefox Focus crashes when opening "About Firefox Focus" screen


cadeyrn:
### Steps to reproduce

1. compile Focus from the main branch
2. open Settings > Mozilla > About Firefox Klar / Firefox Focus

### Expected behavior

"About Firefox Klar" screen opens, no crash.

### Actual behavior

Firefox Klar / Firefox Focus crashes:

```
2021-08-19 15:40:53.141 6516-6516/? W/ScreenSize: getCutoutPathDataHeight() - 
    java.lang.reflect.InvocationTargetException
        at java.lang.reflect.Method.invoke(Native Method)
        at com.oneplus.base.ScreenSize.getCutoutPathDataHeight(ScreenSize.java:235)
        at com.oneplus.base.ScreenSize.<init>(ScreenSize.java:146)
        at com.oneplus.base.ScreenSize.<init>(ScreenSize.java:74)
        at com.oneplus.camera.CameraActivity.checkScreenAndWindowSize(CameraActivity.kt:1449)
        at com.oneplus.camera.CameraActivity.access$checkScreenAndWindowSize(CameraActivity.kt:92)
        at com.oneplus.camera.CameraActivity$displayListener$1.onDisplayChanged(CameraActivity.kt:584)
        at android.hardware.display.DisplayManagerGlobal$DisplayListenerDelegate.handleMessage(DisplayManagerGlobal.java:741)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:233)
        at android.app.ActivityThread.main(ActivityThread.java:8010)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:631)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:978)
     Caused by: java.lang.IllegalArgumentException: Path string cannot be empty.
        at android.util.PathParser.nParseStringForPath(Native Method)
        at android.util.PathParser.createPathFromPathData(PathParser.java:38)
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.oneplus.base.ScreenSize.getCutoutPathDataHeight(ScreenSize.java:235) 
        at com.oneplus.base.ScreenSize.<init>(ScreenSize.java:146) 
        at com.oneplus.base.ScreenSize.<init>(ScreenSize.java:74) 
[...]

PR #5242, opened 2021-08-27
For #5185: Refactor about page.

mcarare:
Comparison to old screen:

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/AboutFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/AboutFragment.kt
@@ -1,0 +1,182 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.fragment
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.ColumnScope
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Alignment.Companion.Start
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.res.colorResource
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.style.TextDecoration
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import kotlinx.coroutines.Job
+import mozilla.components.browser.state.state.SessionState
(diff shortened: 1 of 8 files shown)


brampitoyo:
@mcarare Good call, and good-looking design, too!

A couple of suggestions:

1. Is the logo the same size as the one found on the New Tab Page? Perhaps it’s better if the size is consistent everywhere.
2. Let’s make sure that the text style and colour matches one we use elsewhere in the app, and matches the Fenix about page. What’s the line-height value of the old design, by the way?
3. The second line of each bullet list item should be indented, so it sits together with the first line.

mcarare:
@brampitoyo I have updated the screenshot for the new version:

1. I have removed the small padding from the logo so it shows in the original size (same as in the splash screen).
2. I have added the line height specified in the HTML.
3. There isn't much I can do about that list, it is a string formatted for being displayed in a HTML file. I removed the tabulation before the bullet symbol and now the text wraps, but without the space before. IMO it looks better. For future iterations, when I can add new strings I can separate that part and format it better, but for now, it is riskier to split a localized string into so many parts.

cadeyrn:
> IMO it looks better.

IMO it was previously more pleasant to read. Now the list is a bit cramped due to the reduced spacing and line height…

mcarare:
@brampitoyo I can also replace &#x2022; with a bigger dot,  &#x25CF; if neccessary.

app/src/main/res/layout/fragment_about.xml
```diff
@@ -0,0 +1,36 @@
...
+        android:padding="@dimen/preference_padding_horizontal"
+        android:background="@color/settings_background">
+
+        <androidx.compose.ui.platform.ComposeView
+            android:id="@+id/about_page_content"
+            android:layout_width="match_parent"
+            android:layout_height="match_parent" />
+
+    </LinearLayout>
+</LinearLayout>
```
pocmo, line 36: nit: Would be nice to have no layout here at all. But makes sense with the toolbar for now, I guess. Maybe we can file a follow-up. :)

app/src/main/java/org/mozilla/focus/fragment/AboutFragment.kt
```diff
@@ -0,0 +1,182 @@
...
+        val content =
+            aboutContent
+                .replace("<li>", "\u2022 \u0009 ")
+                .replace("</li>", "\n")
+                .replace("<ul>", "\n \n")
+                .replace("</ul>", "")
+                .replace("<p>", "\n")
+                .replace("</p>", "")
+                .replaceAfter("<br/>", "")
+                .replace("<br/>", "")
```
pocmo, line 84: This feels a bit hacky, given that we can't really control the number of line breaks in the translations. Maybe we should consider splitting the strings. Especially if it also allows us to improve the layout. And it will confuse translators less...

Maybe we can ask @Delphine if that would be okay, I'll approve here anyways since we could try this for now (but we should look at some of the translations to see if that splitting approach works)

pocmo (approved):

app/src/main/res/layout/fragment_about.xml
mcarare: I guess eventually we should consider having a screen composable that automatically includes a toolbar and we can migrate more screens to compose.

app/src/main/java/org/mozilla/focus/fragment/AboutFragment.kt
mcarare: Yes, this was my initial proposal, but it has been postponed until after MVP when we would eventually even change the text on this screen.

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -672,6 +672,16 @@ class UrlInputFragment :
     }
 
     private fun openUrl(url: String, searchTerms: String?) {
+
```
gabrielluong, line 675: nit: no new line here

brampitoyo (approved):
Looks good to me! But I also agree with @cadeyrn that we should fix the alignment of the bullet points, when we’re able to, in the future.

Issue #5241, opened 2021-08-27
Exceptions menu: hide or change "Remove all websites" button when in multi-select mode


sv-ohorvath:
## Steps to reproduce
1. Visit a few websites, turn off ETP on each, to add them to the Exceptions list.
2. Go to Privacy settings> Exceptions.
3. Tap the 3 dot menu in the right-hand corner. Tap Remove.
4. Select some websites from the list, leaving at least one unselected.
5. Observe the "Remove all websites" button
6. Tap the "Remove all websites" button
### Expected behavior
I think the "Remove all websites" button should either be hidden or it should remove just the selected websites.
### Actual behavior
The "Remove all websites" button is misleading in this case, being the most visible action button. The user can barely notice the small trash icon in the right corner.
Tapping on "Remove all websites" has a weird behavior:
- first, it dismisses your selection and all the exceptions are still there as if they weren't deleted.
- pressing anything again: the 3 dot button or the "remove all" button, will exit the Exceptions list screen.
- the Exceptions list can't be accessed back, though it still looks enabled.
- all the websites that were in the exceptions list now have ETP back on. So pressing the "Remove all websites" button did remove all of them.
- the Exceptions list still appears an enabled after you restart the app
### Device information

* Android device: Pixel 2 (Android 9)
* Focus version: Pre-release 93.0.0 & Debug from latest main 8/27.

LaurentiuApahideanSV:
Verified as fixed on the latest debug build from main 8/31.

### Devices used:
- Huawei MediaPad M3 (Android 7)
- OnePlus 6T (Android 9)

delia-pop:
Verified as fixed as well on 93.1.0 RC build with Huawei P9 Lite (Android 8).

PR #5248, opened 2021-08-27
For #5241: Hide delete all websites button when in select mode.

mcarare:


app/src/main/java/org/mozilla/focus/exceptions/ExceptionsListFragment.kt
```diff
@@ -36,7 +37,7 @@ import org.mozilla.focus.state.AppAction
 import org.mozilla.focus.state.Screen
 import org.mozilla.focus.telemetry.TelemetryWrapper
 import org.mozilla.focus.utils.ViewUtils
-import java.util.Collections
+import java.util.*
```
Amejia481, line 40: nit: Let's each import individually.

Amejia481 (approved):

PR #5262, opened 2021-08-30
Issue #5216 - Display Shortcuts in the AwesomeBar

### LINKED ISSUE #5216: Display Shortcuts in the AwesomeBar

gabrielluong:
Fixes #5216

### DIFF
--- a/app/src/main/java/org/mozilla/focus/home/HomeScreen.kt
+++ b/app/src/main/java/org/mozilla/focus/home/HomeScreen.kt
@@ -11,32 +11,26 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.dp
-import mozilla.components.feature.top.sites.TopSite
-import org.mozilla.focus.topsites.TopSiteMenuItem
+import kotlinx.coroutines.DelicateCoroutinesApi
+import mozilla.components.lib.state.ext.observeAsComposableState
+import org.mozilla.focus.components
 import org.mozilla.focus.topsites.TopSites
 
 /**
  * The home screen.
- *
- * @param topSites List of [TopSite] to display.
- * @param topSitesMenuItems List of [TopSiteMenuItem] to display in a top site dropdown menu.
- * @param onTopSiteClick Invoked when the user clicks on a top site.
  */
+@OptIn(DelicateCoroutinesApi::class)
 @Composable
-fun HomeScreen(
-    topSites: List<TopSite>,
-    topSitesMenuItems: List<TopSiteMenuItem>,
-    onTopSiteClick: (TopSite) -> Unit = {}
-) {
+fun HomeScreen() {
+    val topSitesState = components.appStore.observeAsComposableState { state -> state.topSites }
+
     Column(horizontalAlignment = Alignment.CenterHorizontally) {
+        val topSites = topSitesState.value!!
+
         if (topSites.isNotEmpty()) {
             Spacer(modifier = Modifier.height(24.dp))
 
-            TopSites(
-                topSites = topSites,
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/focus/searchsuggestions/ui/SearchSuggestionsFragment.kt
```diff
@@ -116,7 +135,38 @@ class SearchSuggestionsFragment : Fragment(), CoroutineScope {
...
+                onTopSiteClick = { topSite ->
+                    Shortcuts.shortcutOpenedCounter.add()
+
+                    requireComponents.tabsUseCases.addTab(
+                        url = topSite.url,
+                        source = SessionState.Source.Internal.HomeScreen,
+                        selectTab = true,
+                        private = true
+                    )
+                }
```
gabrielluong, line 167: This is more or less copied over from UrlInputFragment https://searchfox.org/mozilla-mobile/source/focus-android/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt#303-324.

I am wondering if there is a recommended way to extract this out so that it can be reusable in both fragments.

app/src/main/java/org/mozilla/focus/searchsuggestions/ui/SearchSuggestionsFragment.kt
```diff
@@ -116,7 +135,38 @@ class SearchSuggestionsFragment : Fragment(), CoroutineScope {
         }
 
         val searchSuggestionsView = view.findViewById<ComposeView>(R.id.search_suggestions_view)
-        searchSuggestionsView.setContent { SearchOverlay(searchSuggestionsViewModel) }
+
+        searchSuggestionsView.setContent {
```
pocmo, line 139: nit: Let's move all this composable code into a function so that `onViewCreated()` does not contain all that compose code?

app/src/main/java/org/mozilla/focus/searchsuggestions/ui/SearchSuggestionsFragment.kt
```diff
@@ -126,14 +176,36 @@ class SearchSuggestionsFragment : Fragment(), CoroutineScope {
...
 @Composable
 private fun SearchOverlay(
-    viewModel: SearchSuggestionsViewModel
+    viewModel: SearchSuggestionsViewModel,
+    topSites: List<TopSite>,
+    topSitesMenuItems: List<TopSiteMenuItem>,
+    onTopSiteClick: (TopSite) -> Unit = {}
 ) {
     val state = viewModel.state.observeAsState()
+    val query = viewModel.searchQuery.observeAsState()
```
pocmo, line 185: Now that this whole tree will get recomposed for every text change, maybe we should pass down the text to `SearchSuggestions()` to not have to subscribe there again? 🤔

app/src/main/java/org/mozilla/focus/home/HomeScreen.kt
```diff
@@ -11,32 +11,25 @@ import androidx.compose.runtime.Composable
...
+@OptIn(DelicateCoroutinesApi::class)
 @Composable
-fun HomeScreen(
-    topSites: List<TopSite>,
-    topSitesMenuItems: List<TopSiteMenuItem>,
-    onTopSiteClick: (TopSite) -> Unit = {}
-) {
+fun HomeScreen() {
+    val topSitesState = components.appStore.observeAsComposableState { state -> state.topSites }
+    val topSites = topSitesState.value!!
```
pocmo, line 26: If you don't need this value here then I would avoid calling `.value` because afaik this will mean this scope is subscribed to it and will get recomposed whenever this value changes. But if you only need it down the tree then read the value only down the tree: Then only this subtree will need to get recomposed.

pocmo (approved):

Issue #5257, opened 2021-08-30
[bug] Can't add shortcuts on (Device)homescreen with Huawei MediaPad M2 (Android 5.1.1)


sflorean:
## Steps to reproduce
1.Open an webpage.
2. Select "Add to Home screen" from the menu.

### Expected behavior
Add to Home screen dialog pops up and the page is added to the homescreen. 

### Actual behavior
The dialog might not be displayed (1/3) at all.
The dialog is displayed, and after the "Add" button is taped, the shortcut is not available on the device's screen. 

### Device information
* Android device: Huawei MediaPad M2 (Android 5.1.1)
* Focus version: 93.0.0, 91.3.0.
** Not reproducible on: Samsung Galaxy Tab A6 (Android 5.1.1).

mcarare:
@sflorean Can you add any other shortcuts (from any other app, maybe try Fenix) on that device?
We might need to hide that menu item on devices that have a launcher that does not support pinning.

mcarare:
For QA: The option should not show up in the menu for devices that do not allow shortcuts to be added.

sflorean:
> @sflorean Can you add any other shortcuts (from any other app, maybe try Fenix) on that device?
> We might need to hide that menu item on devices that have a launcher that does not support pinning.

@mcarare sorry for late reply here.
Tested on Focus 93.0.2 and the issue is still reproducible:
- the menu option is still displayed
- on Fenix, the shortcuts are added to the device home screen without issues

sflorean:
Tested again on the main and here is a log: [focusHomescsreenOption.txt](https://github.com/mozilla-mobile/focus-android/files/7178490/focusHomescsreenOption.txt).

mcarare:
This part is relevant:
` I/HwLauncher: InstallShortcutReceiver Do not install shortcut ,because has the ComponentName activity with launcher caterory. component=ComponentInfo{org.mozilla.focus/org.mozilla.focus.activity.MainActivity}`

mcarare:
This can be retested on the main branch.

sflorean:
Verified as fixed with Huawei MediaPad M2 (Android 5.1.1), shortcuts can be added to the homescreen. 

But when tapping on the shortcut, the following message is displayed "App is not installed". 
@mcarare ^

PR #5265, opened 2021-08-30
For #5257: Do not show add to home menu item if pinning is not supported

mcarare:


pocmo (commented):
I wonder why this is needed? The code used to use the legacy way of adding shortcuts if the new feature wasn't available, I thought (that was through a broadcast, I think?)

mcarare:
`installShortCutViaManager` is used. And it does nothing when  `ShortcutManagerCompat.isRequestPinShortcutSupported(context)` is false. So instead of showing the dialog and do nothing, we do not show the menu item at all.

pocmo:
Looks like the code I remembered was replaced with `ShortcutManagerCompat` here:
https://github.com/mozilla-mobile/focus-android/commit/342613b2be15ca788240219c9a0936bc80cd5d90#diff-55fc6ebddb489ebdaf2a980a5de4d2c80d1258354f693ba10eaaf927da8e1537

[The docs of `ShortcutManagerCompat`](https://developer.android.com/reference/androidx/core/content/pm/ShortcutManagerCompat?hl=en#requestPinShortcut(android.content.Context,%20androidx.core.content.pm.ShortcutInfoCompat,%20android.content.IntentSender)) say: _"On API <= 25 it creates a legacy shortcut with the provided icon, label and intent."_ So I wonder why this is not working. Or is just the check not working properly?

pocmo:
Looks like this change was done in #1291 - but for no other reason than using the AppCompat implementation .... that doesn't work? 😅

mcarare:
It seems it does not on some devices.

mcarare:
Either way, `ShortcutManagerCompat` 's`requestPinShortcut` also checks for `isRequestPinShortcutSupported` and returns false.

pocmo:
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:core/core/src/main/java/androidx/core/content/pm/ShortcutManagerCompat.java;l=148-169;drc=0d673a8ebdd62c3610c26c389dce9a41af280ebe

Hm, still curious which one we hit here, the permission?

mcarare:
My guess is `queryBroadcastReceivers`, for some launchers.

ionutbedregeanu (approved):

mcarare:
Talked with @pocmo offline, I opened a separate issue to further investigate failure to add shortcut: https://github.com/mozilla-mobile/focus-android/issues/5354.

mcarare:
> What went wrong:
> Execution failed for task ':service-telemetry:extractReleaseAnnotations'.
> Could not resolve all files for configuration ':service-telemetry:lintClassPath'.
   > Could not download lint-api-30.0.0.jar (com.android.tools.lint:lint-api:30.0.0)
      > Could not get resource 'https://dl.google.com/dl/android/maven2/com/android/tools/lint/lint-api/30.0.0/lint-api-30.0.0.jar'.
         > Premature end of Content-Length delimited message body (expected: 931,074; received: 524,288)`

Issue #5240, opened 2021-08-27
All trackers are "unblocked" with Content TP off


sv-ohorvath:
## Steps to reproduce
1. With default TP settings, go to https://senglehardt.com/test/trackingprotection/test_pages/tracking_protection.html
2. Check ads, social and analytics trackers
3. Open the ETP panel.
4. Disable the Content option.
5. Refresh the TP test page.
6. Check ads, social and analytics tracking results.
7. Re-enable the Content option.
8. Refresh the page.
9. Check ads, social and analytics tracking results.
### Expected behavior
Step 6: As also reported in https://github.com/mozilla-mobile/focus-android/issues/4864, I think the other trackers should be blocked if Content TP is off.
Step 9: everything is blocked.
### Actual behavior
Step 6: Same as https://github.com/mozilla-mobile/focus-android/issues/4864, the other trackers are not blocked if Content TP is off.
Step 9: The test page still shows everything as "not blocked".
With default options: all 4 TP options on, everything is "blocked".
### Device information

* Android device: Pixel 2 (Android 9)
* Focus version: Pre-release build of vers. 93.0.0

mcarare:
@Amejia481 Can you also take a look at this since you have experience with ETP? From what I have seen we send the correct tracker list in `createTrackingProtectionPolicy`, not sure why the first ones are disabled.

Also, I think here: https://github.com/mozilla-mobile/focus-android/blob/54842a92c30a971b41ea831da82e397c90800ada/app/src/main/java/org/mozilla/focus/utils/Settings.kt#L57

we should have blocked `CONTENT`, not `SCRIPTS_AND_SUB_RESOURCES`, although the  `shouldBlockOtherTrackers` name is misleading, this should refer to content trackers. What do you think?

lobontiumira:
Verified as fixed on the latest debug build from 9/1 from main, with Samsung Galaxy Note 8 (Android 9).

Amejia481:
Thanks @lobontiumira!

lobontiumira:
Verified also on the latest RC 92.1.0, with Samsung Galaxy Tab A6 (Android 5.1.1).

PR #5266, opened 2021-08-30
For #5240: Update default setting for trackers in panel.

mcarare:
This ensures the default values are the ones actually used in the policy.


app/src/main/java/org/mozilla/focus/settings/privacy/TrackingProtectionPanel.kt
```diff
@@ -10,6 +10,7 @@ import androidx.appcompat.content.res.AppCompatResources
 import androidx.core.view.isVisible
 import com.google.android.material.bottomsheet.BottomSheetBehavior
 import com.google.android.material.bottomsheet.BottomSheetDialog
+import kotlinx.android.synthetic.main.dialog_tracking_protection_sheet.*
```
Amejia481, line 13: nit: do we need this import?

mcarare: It added itself while I was typing the view id. I think it knows I plan to remove synthetics. 🤖 😄

Issue #4992, opened 2021-07-15
Implement Beta Channel Dark Mode


amedyne:
Implement Beta channel Landing screen design for Light and Dark Mode

- [ ] Logo 
- [ ] Landing page

amedyne:
latest designs: https://www.figma.com/file/jgsM2hWimZbPmtUJn7IGIB/Landing-Page?node-id=1941%3A61459

lobontiumira:
Verified on the latest beta build 93.0 (GV 93.0a1-20210831093704) with Google Pixel (Android 10):

![Beta Dark](https://user-images.githubusercontent.com/40694158/131858760-56e0310e-99cc-4197-b45a-314d32b24393.png)

PR #5270, opened 2021-08-30
Implement beta dark mode

ionutbedregeanu:
For #4992

[Fixed] Focus production use the same background as nightly
[Added] Background and logo for Focus Beta


ionutbedregeanu:
@brampitoyo I noticed that in Figma we are using two different colors for menu #342F6D and #271948 and I'm not sure which is the correct one. Can you help me with that? Also the color used for menu should be used for dialogs and settings screen, am I right?

brampitoyo:
@ionutbedregeanu The colour value that @Kate-Galetski has specified in the `Landing Page` document is the correct one: `Ink 10` `#342F6D`.

I’ve now updated all menus in the other two documents – `Product UI` and `Shortcuts` – to use the same value. Sorry for this oversight!

Amejia481 (approved):

jonalmeida:
Ah, this change updated the debug icon and not the release ones. 🙂

ionutbedregeanu:
> Ah, this change updated the debug icons as well and not the release ones. 🙂

[This pr](https://github.com/mozilla-mobile/focus-android/pull/5360) should fix this

Issue #5203, opened 2021-08-24
[debug build] "Add to shortcut"/"Remove from Shortcut" options are still available after adding/removing the shortcut


lobontiumira:
## Steps to reproduce
1. Open a page and add it as a shortcut.
2. Re-open the three-dot menu and observe the "Add to shortcut".
3. Clear the browsing history.
4. From the app's homescreen, tap on one of the added shortcuts.
5. Open the three-dot menu and tap on "Remove from Shortcuts".
6. Open again the three-dot menu, and observe.

### Expected behavior
After adding or removing a page from shortcuts, the options should interchange.
If a page is added already, there should only be "Remove from Shortcuts" option available, and viceversa.

### Actual behavior
The user can add a page multiple times (I added a page 10 times, the option is still available, but there are only 4 shortcuts on the app's homescreen).
The user has the option "Remove from Shortcuts" even after all the shortcuts have been removed, even after refreshing the page.

### Device information
* Android device: Lenovo Tab M10 (Android 10)
* Focus version: debug build from 8/24 main

lobontiumira:
Verified as fixed on the 93.0.1 pre-release build, with Lenovo Tab M10 (Android 10).

delia-pop:
Verified as fixed as well on 93.1.0 RC build with Huawei P9 Lite (Android 8).

PR #5279, opened 2021-08-31
For #5203: Update app state when shortcuts are added or removed.

mcarare:


mcarare:
Needs https://github.com/mozilla-mobile/android-components/pull/10906 to be merged and AC version bump.

app/src/main/java/org/mozilla/focus/state/AppAction.kt
```diff
@@ -80,4 +80,14 @@ sealed class AppAction : Action {
...
+
+    /**
+     * An item has been added to the list of [TopSite]s.
+     */
+    data class TopSiteAdded(val topSiteUrl: String, val topSiteName: String) : AppAction()
+
+    /**
+     * An item has been removed from the list of [TopSite]s.
+     */
+    data class TopSiteRemoved(val topSiteUrl: String) : AppAction()
```
gabrielluong, line 92: I am very curious if we need these 2 actions since `TopSitesChange` should update the state through its view when topSitesUseCases is called. I will test this out when we have the AC part landed.

mcarare: We need them because the app state should be updated even if we do not have the home screen displayed. The way it is implemented now when you are on the browser screen the app state does not update.

mcarare: Also, making the composition of a composable responsible for updating a state does not seem like a good pattern to me.

mcarare:
Closed in favor of https://github.com/mozilla-mobile/focus-android/pull/5294.

Issue #5272, opened 2021-08-30
link is missing for first tip in carousel


cadeyrn:
### Steps to reproduce

1. open Focus

### Expected behavior

I don't know if there is a link planned for the first tip in the carousel on the start screen because all your documents are access restricted but from reading the first tip ("read more") I would expect that there is a link:

![Screenshot_20210830-212716](https://user-images.githubusercontent.com/1100614/131396781-386de361-ae4a-406f-87a2-1cf251da08b3.jpg)

### Actual behavior

There is no link.

### Device information

* Android device: OnePlus 7T Pro McLaren, Oxygen OS 11 (Android 11)
* Focus version: main branch revision 56ab7b8a872f19f4397553b516837b80b750c6f1

LaurentiuApahideanSV:
Verified as fixed on the latest Focus build from main 09/14.
### Devices used:
- Oppo find X3 (Android 11)
- OnePlus 6T (Android 9)

PR #5280, opened 2021-08-31
Add whats new url as a deeplink for the first tip

ionutbedregeanu:
For #5272

When the first tip from ProTips carousel is tapped we should be redirected to "What's New" page

### DIFF
--- a/app/src/main/java/org/mozilla/focus/tips/TipManager.kt
+++ b/app/src/main/java/org/mozilla/focus/tips/TipManager.kt
@@ -10,8 +10,10 @@ import mozilla.components.browser.state.state.SessionState
 import org.mozilla.focus.R
 import org.mozilla.focus.ext.components
 import org.mozilla.focus.telemetry.TelemetryWrapper
+import org.mozilla.focus.utils.AppConstants
 import org.mozilla.focus.utils.Settings
 import org.mozilla.focus.utils.SupportUtils
+import org.mozilla.focus.utils.SupportUtils.getSumoURLForTopic
 
 class Tip(
     @StringRes
@@ -29,7 +31,7 @@ class Tip(
             val deepLink = {
                 context.components.tabsUseCases.addTab(
                     url,
-                    source = SessionState.Source.Internal.Menu,
+                    source = SessionState.Source.Internal.HomeScreen,
                     selectTab = true,
                     private = true
                 )
@@ -55,11 +57,27 @@ class Tip(
             val id = R.string.tip_fresh_look
             val name = context.getString(R.string.app_name)
 
+            val sumoTopic = if (AppConstants.isKlarBuild) {
+                SupportUtils.SumoTopic.WHATS_NEW_KLAR
+            } else {
+                SupportUtils.SumoTopic.WHATS_NEW_FOCUS
+            }
+
+            val whatsNewUrl = getSumoURLForTopic(context, sumoTopic)
+            val deepLink = {
+                context.components.tabsUseCases.addTab(
+                    url = whatsNewUrl,
+                    source = SessionState.Source.Internal.HomeScreen,
+                    selectTab = true,
+                    private = true
+                )
+                TelemetryWrapper.pressTipEvent(id)
+            }
             val shouldDisplayFreshLookTip = {
                 Settings.getInstance(context).getAppLaunchCount() == 0
             }
 
-            return Tip(id, name, shouldDisplayFreshLookTip)
+            return Tip(id, name, shouldDisplayFreshLookTip, deepLink)
         }
 
         fun createShortcutsTip(context: Context): Tip {
@@ -96,7 +114,7 @@ class Tip(
             val deepLinkRequestDesktop = {
                 context.components.tabsUseCases.addTab(
                     requestDesktopURL,
-                    source = SessionState.Source.Internal.Menu,
+                    source = SessionState.Source.Internal.HomeScreen,
                     selectTab = true,
                     private = true
                 )


app/src/main/java/org/mozilla/focus/tips/TipManager.kt
```diff
@@ -49,11 +51,26 @@ class Tip(val id: Int, val text: String, val shouldDisplay: () -> Boolean, val d
...
+                    private = true
+                )
+                TelemetryWrapper.pressTipEvent(id)
+            }
             val shouldDisplayFreshLookTip = {
                 Settings.getInstance(context).getAppLaunchCount() == 0
             }
 
-            return Tip(id, name, shouldDisplayFreshLookTip)
+            return Tip(id, name, shouldDisplayFreshLookTip,deepLink)
```
eliserichards, line 73: This should fix the ktlint error 😄 

```suggestion
            return Tip(id, name, shouldDisplayFreshLookTip, deepLink)
```

app/src/main/java/org/mozilla/focus/tips/TipManager.kt
```diff
@@ -49,11 +51,26 @@ class Tip(val id: Int, val text: String, val shouldDisplay: () -> Boolean, val d
             val id = R.string.tip_fresh_look
             val name = context.resources.getString(id, context.getString(R.string.app_name))
 
+            val sumoTopic = if (AppConstants.isKlarBuild)
+                SupportUtils.SumoTopic.WHATS_NEW_KLAR
+            else
+                SupportUtils.SumoTopic.WHATS_NEW_FOCUS
```
eliserichards, line 57: Nit: 
```suggestion
            val sumoTopic = if (AppConstants.isKlarBuild) {
                SupportUtils.SumoTopic.WHATS_NEW_KLAR
            } else {
                SupportUtils.SumoTopic.WHATS_NEW_FOCUS
            }
```

Amejia481:
The link works, but as the whole text is purple there is not a clear indication that the text is a link. 
image

eliserichards (dismissed):
Looks good! Just a nit & a fix for lint and this will be perfect 🚀

Amejia481:
Maybe having just the **"Read more"** as link could will be clear.

eliserichards:
I agree with @Amejia481. We should also have a line break before "Read more"

app/src/main/res/values/strings.xml
```diff
@@ -703,7 +703,11 @@ The new line here must be kept as the second half of the string is clickable for
 
     <!-- Tip displayed on the home view about the changes in the new release.
     %1$s is a placeholder for the app name -->
-    <string name="tip_fresh_look">Why yes, we do have a fresh new look!\n Read more about this and other updates to %1$s.</string>
+    <string name="tip_fresh_look">Why yes, we do have a fresh new look! Read more about this and other updates to %1$s.</string>
+
+    <!-- Tip displayed on the home view about the changes in the new release.
+   %1$s is a placeholder for the app name -->
+    <string name="tip_fresh_look2">Why yes, we do have a fresh new look!\n Read more about this and other updates to %1$s.</string>
```
ionutbedregeanu, line 710: @Delphine To highlight just the second line of this tip I added a new string(tip_fresh_look2) with "\n". Can you tell me if you are ok with this change or I should revert it(I know that we are in a strings freeze)?

Delphine: So we updated that string on our side last week (more or less) but with no change in string ID. See example in Pontoon with French here!: https://pontoon.mozilla.org/fr/focus-for-android/mozilla-mobile/focus-android/app/src/main/res/values/strings.xml/?search=fresh&string=230880
I flagged it to localizers, but not all have updated their strings with that yet.
As I understand it, is the issue here that the link won't show up at all without the `/n`?
If that's the case, I think we should change the string ID as you suggest. I will make sure localizers are aware, if we do this

Delphine: Update: actually I see the link works from past conversations, but is not highlighted. Let me know if you want to update the string itself, or only add the `\n`

ionutbedregeanu: Oh, but that's great, I didn't know that you already updated that string with "\n". Then I should revert my change(remove tip_fresh_look2) and let tip_fresh_look as it was

Delphine: Ok awesome! I have a list of locales that haven't updated that yet, so I will manually update those in our translation tool. It's not too much of a pain and that way we are sure it looks good across the board (and we don't break string freeze, yay!)
I guess the only thing you'd have to update on your side is adding it to en-US, unless that's good there.
Thanks for CCing me!

Delphine: All locales have been updated.
One thing I noticed though on the latest build I got is I could not even click that link - even for locales that already had the `\n` fix in such as French (cf [comment above](https://github.com/mozilla-mobile/focus-android/pull/5280#issuecomment-910695022))

ionutbedregeanu: Yep, this pr should fix that

ionutbedregeanu:
> The link works, but as the whole text is purple there is not a clear indication that the text is a link.

@Amejia481 @eliserichards That string was already updated with a '\n' before 'Read more', but this happens because not all translations were updated. I already discussed with Delphine about that and made the updates in Pontoon.

app/src/main/java/org/mozilla/focus/tips/TipManager.kt
```diff
@@ -55,11 +57,27 @@ class Tip(
...
+                SupportUtils.SumoTopic.WHATS_NEW_KLAR
+            } else {
+                SupportUtils.SumoTopic.WHATS_NEW_FOCUS
+            }
+
+            val whatsNewUrl = getSumoURLForTopic(context, sumoTopic)
+            val deepLink = {
+                context.components.tabsUseCases.addTab(
+                    url = whatsNewUrl,
+                    source = SessionState.Source.Internal.Menu,
```
jonalmeida, line 70: The source is not the menu, but `Internal.HomeScreen`. It's wrong for the other tips too, so Ieave it up to you if you want to fix them all here or in a separate PR. 🙂

jonalmeida (approved):

jonalmeida:
(It's not clear to me if you want this landed now or not, so I'll leave that up to you. 🙂 )

ionutbedregeanu:
> (It's not clear to me if you want this landed now or not, so I'll leave that up to you. 🙂 )

For now can't be merged because I need also an approve from @eliserichards

PR #5294, opened 2021-09-01
For #5203: Update state with shortcuts list when in browser fragment.

mcarare:
Needs https://github.com/mozilla-mobile/android-components/pull/10906 to be merged and AC version bump.


pocmo (approved):
👍 As a workaround I think this is okay for now.

But I would really like to move the top sites handling out of features that are bound to UI and instead move it to a middleware.

Issue #5301, opened 2021-09-01
 Home screen menu button ripple animation is shown behind the toolbar. 


mcarare:
On the home screen, the menu button ripple animation is shown behind the toolbar.

mcarare:
Fix included in https://github.com/mozilla-mobile/focus-android/pull/5220/files.

PR #5302, opened 2021-09-01
For #5301: Allow menu ripple animation to be shown on parent view.

mcarare:


mcarare:
Fix included in https://github.com/mozilla-mobile/focus-android/pull/5220/files.

Issue #5305, opened 2021-09-01
Most menu items truncate in priority languages


Delphine:
## Steps to reproduce
Go to a website
Open the menu

### Expected behavior
Strings are not truncated

### Actual behavior
Strings are truncated and incomprehensible

### Device information
Pixel 2

* Android device: v 11
* Focus version: Nightly (Focus crashes when I try to see the Nightly version)

Note: this has initially been reproduced by @flodolo who has another device, and I confirm I see all the same issues. This is really concerning because most menu items are incomprehensible at this point
Also reproducible in Italian, French, German, and other "priority" locales. See screenshot in Italian below
CC @amedyne for visibility.
![menu](https://user-images.githubusercontent.com/399020/131731984-df8894a9-a4a5-4728-9252-7af312407254.jpg)

amedyne:
Tagging @brampitoyo for input from UX.

mcarare:
The way we handle this in Fenix is to allow the menu to stretch to a maximum width (set in Fenix ) while also allowing a minimum of 48dp (constraint in AC) space to allow the menu to be dismissed on taping outside of it. 
@Bram, I implemented this with the same sizes we use in Fenix here: https://github.com/mozilla-mobile/focus-android/pull/5315.

brampitoyo:
@mcarare Excellent suggestion. I was going to say “Let’s follow Fenix’s implementation”, so I agree completely.

lobontiumira:
Verified on the 93.0 (build 11 GV 93.0a1-20210906031657), and on 93.0.2 pre-release with OnePlus 5T (Android 9).
The page's menu is wider, but some options in some languages are still truncated:

![languages](https://user-images.githubusercontent.com/40694158/132862734-4433d812-93da-4193-890f-91cfb470897b.png)

@Delphine please review and advise.
Thank you!

Delphine:
@brampitoyo @mcarare: Is it technically impossible to have the text wrap to a second line (we do that in other areas)? That would be preferable given this is the main menu and affects top tier locales

brampitoyo:
@Delphine @mcarare I am happy to have the text wrap to the second line.

Follow this specs for two-line list items:
https://material.io/components/lists#specs

Except in our case, the second line does not contain secondary text. It’s simply the item title that spills over.

mcarare:
[...]

PR #5315, opened 2021-09-02
For #5305: Allow menu to stretch up to 314dp width.

mcarare:
Before and after the fix, for the language presented in #5305

### DIFF
--- a/app/src/main/res/values/dimens.xml
+++ b/app/src/main/res/values/dimens.xml
@@ -71,4 +71,8 @@
 
     <dimen name="search_suggestions_padding_start">56dp</dimen>
     <dimen name="search_suggestions_padding_with_icon">16dp</dimen>
+
+    <!-- Override the default menu width values from AC. -->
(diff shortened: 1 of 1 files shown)


mcarare:
@Delphine This would fix the issue with important locales, and it would be the same behaviour Fenix has.

mcarare:
cc @brampitoyo

Delphine:
Thanks for the help here. Please note though that the last string on the screenshot being truncated, and isn't comprehensible that way.

brampitoyo:
@mcarare @Delphine Let me know if any UI help is needed here. This decision should also apply to Fenix, if we do make another change.

mcarare:
UX approved this approach: https://github.com/mozilla-mobile/focus-android/issues/5305#issuecomment-913896455

Delphine:
> Thanks for the help here. Please note though that the last string on the screenshot being truncated, and isn't comprehensible that way.

Thanks for the help! I misstated that the last string on the screenshot was truncated as well, because it's not :) This LGTM now, thank you!

ionutbedregeanu (approved):

PR #5322, opened 2021-09-02
Adjust tracking protection metrics.

mcarare:


st3fan:
@mcarare What was the reason for this PR? I'm working on the same for iOS so I am curious why this addition happened.

Amejia481 (approved):
This address the issue, I'm not sure how pressing is this issue, I'm approving in case we can't find a better alternative. 

The real source is we are using a [PreferenceSwitch](https://github.com/mozilla-mobile/focus-android/blob/d7d1bbc43cd718cc19e7e1e643805f5627ab733d/app/src/main/res/layout/dialog_tracking_protection_sheet.xml#L81) for the toggles instead of normal switches, the  tracking protection panel, when the toggles change  we get the event on [EngineSharedPreferencesListener](https://github.com/mozilla-mobile/focus-android/pull/5322/files#diff-e0a2fe13e74c921afb44d25e6361db5c7df442ab0f5aa0fbea2ea1c3dcb80fa0R21) then we get another event on the [click listener](https://github.com/mozilla-mobile/focus-android/pull/5322/files#diff-c5ba7cc8d21d0905f90e730cbab45b612b9ee331c83555339cd02bd60727dcd5R136) that updates the preference again.

mcarare:
> 
> The real source is we are using [PreferenceSwitch](https://github.com/mozilla-mobile/focus-android/blob/d7d1bbc43cd718cc19e7e1e643805f5627ab733d/app/src/main/res/layout/dialog_tracking_protection_sheet.xml#L81)es for the toggles instead of normal switches, the tracking protection panel, when the toggles change we get the event on [EngineSharedPreferencesListener](https://github.com/mozilla-mobile/focus-android/pull/5322/files#diff-e0a2fe13e74c921afb44d25e6361db5c7df442ab0f5aa0fbea2ea1c3dcb80fa0R21) then we get another event on the [click listener](https://github.com/mozilla-mobile/focus-android/pull/5322/files#diff-c5ba7cc8d21d0905f90e730cbab45b612b9ee331c83555339cd02bd60727dcd5R136) that updates the preference again.

Yes, that is true, but I still prefer the preference toggles to be handled at the call site by `onPreferenceTreeClick` instead of a different listener in onSharedPreferenceChanged, because that can be triggered when you set the value to the existing value and might trigger a metric event from code, not from user interaction.

PR #5329, opened 2021-09-03
Use separate application ID for Nightly builds

pocmo:
To install/ship Nightly as a separate app, we need to use a separate application ID.


jonalmeida (approved):
✨

Issue #5333, opened 2021-09-03
 The tracking protection data panel is cutoff in landscape mode and cannot be scrolled 


delia-pop:
## Steps to reproduce
1. Using the landscape mode, open a page.
2. Tap on the ETP icon displayed in front of the URL.
3. Notice that the tracking protection data panel is cutoff under the "Analytics" text and toggle and the user cannot scroll on the panel.

### Expected behavior
The tracking protection data panel is correctly displayed, all the trackers names/ toggles and the number of trackers blocked on the page are visible when scrolling on the panel.

### Actual behavior
The tracking protection data panel is cutoff, the items under the "Analytics" text and toggle are not visible and the user cannot scroll on the panel.

### Device information
* Android device: Huawei P9 Lite (Android 8), Huawei P20 (Android 10), Google Pixel (Android 10).
* Focus version: 93.0 (Build 352452037 ; 93.0a1-20210831093704).
Please note that this issue is not encountered when using Tablets.

https://user-images.githubusercontent.com/89388888/132010891-328daebb-05ed-428d-bd0f-69ca4badf888.mp4

lobontiumira:
Verified as fixed on the latest debug from main from 9/9, using a Google Pixel (Android 10), and OnePlus 5T (Android 9).

PR #5336, opened 2021-09-03
For #5333: Make ETP panel scrollable.

mcarare:


mcarare:
The diff looks strange, but the only change is adding NestedScrollView as a parent for the layout and change the `ConstraintLayout`'s height to `wrap_content`

ionutbedregeanu (approved):

Issue #5212, opened 2021-08-25
[custom tabs] Too much space in the toolbar before the back arrow


lobontiumira:
## Steps to reproduce
1. Make Firefox Focus your default browser.
2. Open from Gmail a link in custom tab.
3. Observe the toolbar.

### Expected behavior
There should be no space in front of the back arrow in the toolbar.

### Actual behavior
There is a large space before the back arrow.

### Device information
* Android device: Lenovo Tab M10 (Android 10)
* Focus version: RC 91.3.0, and debug build from main from 8/25

![custom tab](https://user-images.githubusercontent.com/40694158/130757267-e7d21926-b529-4af6-b47c-1e63f336b4ea.png)

lobontiumira:
I'll re-open this because the space is again visible on the latest debug build from main from 9/3:

![aaa](https://user-images.githubusercontent.com/40694158/131986783-4a999ceb-2e35-419d-9a30-4eebaf9a15af.jpg)

lobontiumira:
Reproducible on 93.0.1 build with Lenovo Tab M10 (Android 10).

lobontiumira:
Verified as fixed on the 93.0 build 11-GV 93.0a1-20210906031657 with Lenovo Tab M10 (Android 10).

LaurentiuApahideanSV:
Verified as fixed on Focus 93.1.0.
### Device used:
- Huawei MediaPad M3 (Android 9)

lobontiumira:
I'll re-open the issue, because I was able to reproduce it again on RC 97.1.1 on the Lenovo Tablet M10 (Android 10):

![Screenshot_3](https://user-images.githubusercontent.com/40694158/154040497-398245c5-4f6b-4719-97d8-e00d7ffaf4e7.jpg)

** Not reproducible on RC 97.1.0 build with the same device:
![Screenshot_4](https://user-images.githubusercontent.com/40694158/154041209-4396256f-f970-4d4f-956f-6f45dead8a80.jpg)

ionutbedregeanu:
I didn't succeed to reproduce it. I will move it back to backlog to keep an eye on it

ionutbedregeanu:
I succeed to reproduce it with device dark theme. Thanks, @lobontiumira

lobontiumira:
Verified as fixed on the latest Focus Nightly 100.0a1 (build 360870510 with GV 100.0a1-20220327093354) with Lenovo Tab M10 (Android 10).

PR #5340, opened 2021-09-06
Change the color of navigation buttons and url background

ionutbedregeanu:
For #5212 
Update the color of navigation buttons for that scenario when a custom tab is open and the app theme is light

This pr also fix https://github.com/mozilla-mobile/focus-android/issues/5344


app/src/main/java/org/mozilla/focus/browser/integration/NavigationButtonsIntegration.kt
```diff
@@ -29,11 +30,16 @@ class NavigationButtonsIntegration(
...
-    private val disabledColorRes = context.theme.resolveAttribute(R.attr.disabled)
+    private var enabledColorRes = context.theme.resolveAttribute(R.attr.primaryText)
+    private var disabledColorRes = context.theme.resolveAttribute(R.attr.disabled)
 
     init {
 
+        if (store.state.findCustomTabOrSelectedTab(customTabId)?.ifCustomTab() != null) {
+            enabledColorRes = R.color.enabled_toolbar_button_color
+            disabledColorRes = R.color.disabled_toolbar_button_color
+        }
```
pocmo, line 41: Custom tabs can have different toolbar colors, so we need to pick the tint color dynamically. For the text we do:
https://github.com/mozilla-mobile/android-components/blob/main/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt#L114

jonalmeida (approved):
Thanks for fixing!

PR #5350, opened 2021-09-07
Add light theme and clean up color resources

mcarare:
This is the first step in trying to implement a light theme, so it would be easier to change the colors for the light theme. 

For the moment both colors.xml files have the same content. This step would involve browsing through the app to see if there are any inconsistencies with the design.


app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -43,10 +42,6 @@ open class MainActivity : LocaleAwareAppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        // Because some MVP characteristics are set from xml files we need to choose the mode of the theme
-        // according to MVP flag and not by system theme mode
-        AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_YES)
```
jonalmeida, line 48: This is temporary right? 🙂

jonalmeida (commented):
external-content duckduckgo com

app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
mcarare: Yes,
temporary-8ld3i6

mcarare:
This will also need https://github.com/mozilla-mobile/android-components/pull/10954 to land and AC bump.

jonalmeida:
Putting 'do not land' so we don't mistakenly merge this for 93.

mcarare:
This still needs assets for the home background in light theme and I have to adjust the value for some light theme colors. Otherwise, I think it can be code reviewed.

For the color usage review we have 2 options:
 - Bram reviews them based on a build we provide before this gets merged
 - We merge this after the release cut and Bram and QA can review them. QA can also test for contrast in places where we do not have light theme designs in Figma and I had to try to match them based on Fenix night/day color correspondence, for example.

There might be colors I switched when restructuring the colors.xml files so we definitely need a color review.

@brampitoyo @jonalmeida Thoughts?

jonalmeida:
Sounds like a great plan! 

> I would also prefer to handle bugs or color adjustments separately on individual issues rather than having another large PR.

This is a good idea and goes well with your second recommendation:

> We merge this after the release cut and Bram and QA can review them. QA can also test for contrast in places where we do not have light theme designs in Figma and I had to try to match them based on Fenix night/day color correspondence, for example.

app/src/main/java/org/mozilla/focus/FocusApplication.kt
```diff
@@ -71,6 +76,44 @@ open class FocusApplication : LocaleAwareApplication(), CoroutineScope {
...
 
+    private fun setTheme(context: Context) {
+        val settings = Settings.getInstance(context)
+        when {
+            settings.lightThemeSelected -> {
+                AppCompatDelegate.setDefaultNightMode(
+                    AppCompatDelegate.MODE_NIGHT_NO
+                )
+            }
+            settings.darkThemeSelected -> {
```
gabrielluong, line 87: Should we have a new line above?

mcarare: Done.

app/src/main/res/drawable/home_background.xml
```diff
@@ -0,0 +1,27 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
gabrielluong, line 1: Sorry this is a drive by. Should this also have a license

mcarare: This will be replaced when I have the assets from UX. But I can add them now...

app/src/main/res/values-night/colors.xml
```diff
@@ -3,70 +3,147 @@
...
+
+    <!-- Snackbar-->
+    <color name="snackbarBackground">#FF414146</color>
+    <color name="snackbarTextColor">#ffffffff</color>
+    <color name="snackbarActionText">@color/photonBlue50</color>
+
+    <!-- Search Suggestions -->
+    <color name="searchSuggestionsBackground">@color/photonInk80</color>
+    <color name="searchSuggestionPromptBackground">@color/photonInk80</color>
+    <color name="searchSuggestionPromptTitleColor">@color/photonLightGrey05</color>
```
gabrielluong, line 72: I only came around because I was talking to Mugurell about design systems and colors since I am providing some engineering context for Crystal who is working on the design system. 

Can we match all these instances of `photonLightGrey05` to `colorPrimary`? It seems to me overall, we want to map to these "Primary", "Secondary" variables in the long run instead of the photon color directly. Since UX can easily make the decision to change "Primary" from say "photonLightGrey05" to "photonLightGrey20", that way we will just propagate the changes all around.

gabrielluong: This is not intended to block the PR or anything. Just more design system thinking from my side and trying to rationalize how we should go about this.

gabrielluong: Feel free to do this as a followup.

mcarare: I do not have designs for the light theme of this, so I cannot know if the corresponding light theme color would still be `colorPrimary`. But yes, as a rule, I tried merging colors to refer to a small number of colors.

Issue #5318, opened 2021-09-02
Adding the same search string twice will result in a server not found error message


LaurentiuApahideanSV:
## Steps to reproduce
1. Go to the three-dot menu -> Settings -> Search -> Search engine;
2. Tap on "+ Add another search engine";
3. Enter a name for the engine and the string to use and tap on save.
4. Tap on "+ Add another search engine" again;
5. Enter a different name for the search engine but the same search string used at step 3 and tap on save.
### Expected behavior
The search engine is added a second time or an error message stating that the search engine is already added should be displayed.
### Actual behavior
The server not found error message is displayed.

### Additional information
If a user added search engine is deleted it can not be added again, the server not found error message is displayed when trying to do so.
Might be related to https://github.com/mozilla-mobile/focus-android/issues/5317.
### Device information

* Android device: Oppo Find X3 Neo (Android 11), OnePlus 6T (Android 9)
* Focus version: 92.1.0

mcarare:
This issue is for tracking the name error message. For the same search string the issue is: https://github.com/mozilla-mobile/focus-android/issues/5318

mcarare:
I can fix the server not found error message (we get a `SocketTimeoutException` ), but the question remains if we want to save the engine with the same search string (different name tough).

IMO there are use cases (even privacy reasons) where a user might want to save the same search engine with different names. Adding to this, we do not restrict this in Fenix either. 

@brampitoyo, what do you think?

brampitoyo:
@mcarare I think this will be a rare corner case, where the same search engine is added twice under different names.

I would be OK for Focus not to restrict this addition if the user desires it. Let’s follow Fenix’s behaviour.

delia-pop:
Verified as fixed on the latest Focus build from main 09/14 with Huawei P20 (Android 10).

PR #5356, opened 2021-09-08
For #5318: Close response stream to allow another request.

mcarare:
This also needs UX confirmation in order to land.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.kt
@@ -206,7 +206,11 @@ class ManualAddSearchEngineSettingsFragment : BaseSettingsFragment() {
             )
 
             return try {
-                client.fetch(request).status < VALID_RESPONSE_CODE_UPPER_BOUND
+                val response = client.fetch(request)
+                // Close the response stream to ensure the body is closed correctly. See https://bugzilla.mozilla.org/show_bug.cgi?id=1603114.
+                response.body.close()
+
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/settings/ManualAddSearchEngineSettingsFragment.kt
```diff
@@ -206,7 +206,11 @@ class ManualAddSearchEngineSettingsFragment : BaseSettingsFragment() {
             )
 
             return try {
-                client.fetch(request).status < VALID_RESPONSE_CODE_UPPER_BOUND
+                val response = client.fetch(request)
+                // Close the response stream to ensure the body is closed correctly. See https://bugzilla.mozilla.org/show_bug.cgi?id=1603114.
+                response.body.close()
```
jonalmeida, line 211: Why does this not happen in Fenix as well? 🤔

mcarare: It did. In Fenix, we use `response.body.string()` to read the stream (that also automatically closes it. ) At the moment of the fix the `close()` method had a bug that has been fixed now.  We should also use close there too. I will put up a PR for that.

mcarare: https://github.com/mozilla-mobile/fenix/pull/21203

mcarare:
UX also gave the go-ahead: https://github.com/mozilla-mobile/focus-android/issues/5318#issuecomment-915724307

jonalmeida (approved):
Brilliant, thanks for following up so quickly!

PR #5358, opened 2021-09-08
Stand up build signing in taskgraph

ahal:
This depends on @sarah-clements 's PR #5145 . I'm submitting now for testing purposes (as `staging-focus-android` doesn't have necessary scopes yet).

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -116,10 +116,6 @@ tasks:
                         requires: all-completed   # Must be explicit because of Chain of Trust
                         priority: highest
                         retries: 5
-                        scopes:
-                          - assume:repo:github.com/mozilla-mobile/focus-android:release
-                        routes:
-                          - statuses  # Automatically added by taskcluster-github. It must be explicit because of Chain of Trust
                         payload:
                           maxRunTime: 600   # Decision should remain fast enough to schedule a handful of tasks
                           image: mozillamobile/focus-android:1.7
--- a/taskcluster/ci/signing/kind.yml
+++ b/taskcluster/ci/signing/kind.yml
@@ -1,0 +1,40 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: focus_android_taskgraph.loader.multi_dep:loader
+
+transforms:
+    - focus_android_taskgraph.transforms.multi_dep:transforms
+    - focus_android_taskgraph.transforms.signing:transforms
+    - taskgraph.transforms.task:transforms
+
+kind-dependencies:
+    - build
+
+primary-dependency: build
+
+group-by: build-type
+
+job-template:
+    description: Sign Focus for Android
+    worker: {}
+    signing-format: autograph_focus
+    index:
+        by-tasks-for:
+            (action|cron|github-release):
+                by-build-type:
+                    nightly:
+                        type: signing
+                    default: {}
+            default: {}
+    run-on-tasks-for: [github-push, github-release]
+    treeherder:
+        job-symbol:
+            by-build-type:
+                focus-.*: Bfs
+                klar-.*: Bkls
+                default: Bs
+        kind: build
+        platform: android-all/opt
+        tier: 1
--- a/taskcluster/focus_android_taskgraph/transforms/signing.py
+++ b/taskcluster/focus_android_taskgraph/transforms/signing.py
@@ -1,0 +1,99 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+"""
+Apply some defaults and minor modifications to the jobs defined in the build
+kind.
+"""
+
+from taskgraph.transforms.base import TransformSequence
+from taskgraph.util.schema import resolve_keyed_by
+
+
+transforms = TransformSequence()
+
+PRODUCTION_SIGNING_BUILD_TYPES = [
+    "nightly",
+    "beta",
+    "focus-release",
+    "klar-release",
+]
+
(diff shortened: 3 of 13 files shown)


taskcluster/ci/ui-test/kind.yml
```diff
@@ -2,23 +2,14 @@
...
+loader: taskgraph.loader.transform:loader
 
 transforms:
-    - focus_android_taskgraph.transforms.multi_dep:transforms
     - taskgraph.transforms.job:transforms
     - taskgraph.transforms.task:transforms
 
 kind-dependencies:
     - build
-    - toolchain
```
sarah-clements: Once you add the signing kind, I think you'll see why the the multi_dep transform is needed, unless there's some other way to have one task pending until another task completes. Also, if you look at fenix, the dependency should change once you add the signing kind; this task will no longer be primarily dependent on the build to finish (that will be the signing kind), and this in turn will depend on the signing kind to finish. The reason why I didn't set this kind up is because for the first pr, the `focusDebug` is signed by default with a debug key, and that's all that's needed for the ui-test kind.

.taskcluster.yml
```diff
@@ -2,267 +2,484 @@ version: 1
...
+                        schedulerId: ${scheduler_id}   # Must be explicit because of Chain of Trust
+                        provisionerId: mobile-3
+                        workerType: decision
+                        requires: all-completed   # Must be explicit because of Chain of Trust
+                        priority: highest
+                        retries: 5
+                        scopes:
+                          - assume:repo:github.com/mozilla-mobile/focus-android:release
+                        routes:
+                          - statuses  # Automatically added by taskcluster-github. It must be explicit because of Chain of Trust
```
escapewindow, line 121: Removing the two lines above may fix the CoT issues. I think we use `checks` instead of `statuses` now.

escapewindow: Ben adds `adding https://github.com/mozilla-mobile/fenix/blob/main/.taskcluster.yml#L3 might be good too, regardless`

`verify_cot` was useful in debugging here; it may be a good tool to start using.

taskcluster/focus_android_taskgraph/transforms/signing.py
```diff
@@ -0,0 +1,91 @@
...
+def set_signing_type(config, tasks):
+    for task in tasks:
+        signing_type = "dep-signing"
+        if (
+            str(config.params["level"]) == "3"
+            and config.params["tasks_for"] in ("cron", "github-release", "action")
+        ):
+            if task["attributes"]["build-type"] in ("beta", "release"):
+                signing_type = "fennec-production-signing"
+            elif task["attributes"]["build-type"] in ("nightly", "android-test-nightly", "beta-mozillaonline", "release-mozillaonline"):
```
bhearsum, line 58: I'm a little bit surprised we're signing test builds with real certs, but Fenix is already doing it, so there's probably a good reason....

taskcluster/focus_android_taskgraph/transforms/signing.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@transforms.add
+def set_index(config, tasks):
+    for task in tasks:
+        index = {}
+        if (
+            config.params["tasks_for"] in ("cron", "github-release", "action")
+            and task["attributes"]["build-type"]
+            in ("nightly", "debut", "nightly-simulation", "beta", "release")
```
bhearsum, line 71: Typo here: s/debut/debug/

bhearsum (approved):
This seems like it should be good, modulo the one typo below!

taskcluster/focus_android_taskgraph/transforms/signing.py
ahal: I don't know.. this was just copy / pasted over from Fenix. Might be better to start without it and then add it later if it turns our it's necessary? Also I guess we can remove these `mozillaonline` values?

taskcluster/focus_android_taskgraph/transforms/signing.py
ahal: Good catch, this typo is actually in `fenix` too...

taskcluster/focus_android_taskgraph/transforms/signing.py
ahal: On closer look, none of these (except `nightly`) are even valid build types in `focus`.. I'll remove them.

bhearsum: Yes, actually - the mozillaonline ones can defiintely be removed...

ahal:
I dug into why some of the signed arfitacts have `unsigned` in their name. Turns out it's just because that's what the build apk artifacts have as well. The build artifact names in turn come from the command `./gradlew --no-daemon --quiet printVariants`. As far as I can tell there's no mechanism for `signingscript` to change the name of the artifact being signed.. so that means if folks want the `unsigned` gone, it will need to be adjusted in the gradle configuration. That is not something I know how to do, so I'll leave it up to the focus team to make whatever tweaks they would like there as a follow-up to this PR :).

This PR is ready to go and just waiting on #5145 to merge.

escapewindow:
+1, I recomment `target.apk` as a name since we've standardized on it elsewhere.

ahal:
I've made a fair amount of changes since last time, so re-requesting review.

taskcluster/ci/signing/kind.yml
```diff
@@ -0,0 +1,40 @@
...
+    signing-format: autograph_focus
+    index:
+        by-tasks-for:
+            (action|cron|github-release):
+                by-build-type:
+                    nightly:
+                        type: signing
+                    default: {}
+            default: {}
+    run-on-tasks-for: [github-push, github-pull-request]
```
bhearsum, line 31: Eventually we'll want `github-release` here too, I think? It can probably be added now without causing any issuse, if you want. Definitely not blocking though.

bhearsum (approved):

ahal:
I think I saw that the ui-tests were removed from mergify's check, so hopefully this lands automatically once other tasks are green.

Issue #937, opened 2017-07-14
Add support for file uploads (input type=file)


feikname:
I tried to upload a file to a website which has a `<input type="file">` element, and noticed that Focus doesn't bring any UI for file selection currently.

**Edit**: Support for the `multiple` attribute would be nice too, as it seems no major mobile browser really supports it currently.

**Edit 2:** It seems now Chrome supports it. I haven't tested other mobile browsers.

Eskuero:
Yeah I'm really looking forward upload support, would like to be able to use Firefox Send from the phone without having to download a different browser.

bbinto:
https://play.google.com/apps/publish?account=7083182635971239206#ReviewDetailsPlace:p=org.mozilla.focus&reviewid=gp:AOqpTOGW9BN46OARE0sbLYhXvR6ejrwijywQ-gWG7l0hTe7Ty9vuScwoP_Wf_i4ZOopSrN-5HxFZxENpYgT7JsI

pocmo:
This should work in GeckoView (this bug is only for WebView) - but it doesn't seem like it does. I can select files but they do not get uploaded.

pocmo:
* WebView -> Feature
 * GeckoView -> Bug?

bbinto:
Let's keep it simple and only fix the bug for GeckoView (and forget about the WebView aspect)

cpeterso:
I filed a GeckoView bug about fixing <input type="file">:

https://bugzilla.mozilla.org/show_bug.cgi?id=1476810

pocmo:
From the GV bug:

> So the Focus Activity (or Crow Activity) has to customize `GeckoViewPrompt.filePickerRequestCode` > [1] for the file picker to work properly. geckoview_example does that at [2] and [3].
> 
> [1] https://github.com/mozilla-mobile/focus-android/blob/6507f302900fa7f087768dd36a2160d2734ca0a9/app/src/gecko/java/org/mozilla/focus/web/GeckoViewPrompt.java#L57
> [2] https://searchfox.org/mozilla-central/rev/88199de427d3c5762b7f3c2a4860c10734abd867/mobile/android/geckoview_example/src/main/java/org/mozilla/geckoview_example/GeckoViewActivity.java#165
[...]

PR #5359, opened 2021-09-08
For #937: fix file uploads

Amejia481:


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -213,7 +213,7 @@ class BrowserFragment :
             PromptFeature(
                 fragment = this,
                 store = components.store,
-                customTabId = tab.id,
+                customTabId = tryGetCustomTabId(),
```
Amejia481, line 216: We were passing the tab `id` instead of the `customTabId`, causing `PromptFeature` to get confused [here](https://github.com/mozilla-mobile/android-components/blob/c41a8e569b83ac1c6b48a673aadfca522cab3e13/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FilePicker.kt#L137)

jonalmeida (approved):
Nice find!

Issue #5355, opened 2021-09-08
Update landing page background(nightly/production) and launch icons(production)


ionutbedregeanu:
## Steps to reproduce
Open a release build
### Expected behavior
The app should have different backgrounds for nightly and production
Production launch icons should be updated
 [Figma design](https://www.figma.com/file/jgsM2hWimZbPmtUJn7IGIB/Landing-Page?node-id=1941%3A61457)
### Actual behavior
Wrong [nightly background](https://www.figma.com/file/jgsM2hWimZbPmtUJn7IGIB/Landing-Page?node-id=1941%3A61458) is displayed for nightly and production
### Device information

* Android device: any
* Focus version: 93.0.1

lobontiumira:
Verified on the latest Release, and Nightly builds from main from 9/15, with OnePlus 5T (Android 10). The first row of screenshots is from Release, the second is from Nightly:

![release vs nightly](https://user-images.githubusercontent.com/40694158/133408863-3b2217ec-1476-40fc-bc79-665d29d0ec4e.png)

![icons](https://user-images.githubusercontent.com/40694158/133408925-b9b7e859-cc66-43a8-b318-022dec6aadcc.jpg)

Issue #5180, opened 2021-08-19
Firefox Klar logo has wrong colors


cadeyrn:
### Steps to reproduce

1. compile Focus from the main branch with Klar branding

### Expected behavior

The logo has the same colors as with the Focus branding.

### Actual behavior

Wrong colored Firefox logo. A purple foreground on purple background does, obviously, not work very well.

The Klar logo is also noticeable larger than the Focus logo which seems wrong to me.

![klar-logo](https://user-images.githubusercontent.com/1100614/130062746-feee4c1d-c951-4d53-992f-76064b46b04a.png)

### Device information

* Android device: OnePlus 7T Pro McLaren, Oxygen OS 11 (Android 11)
* Focus version: main branch revision baa92c122369515b8c1696c4d995c998a338a8b9

ionutbedregeanu:
Because both logos suffered some changes I will update Focus logo as well in this ticket

Focus logo resource: https://drive.google.com/drive/folders/1ZUwG06Bq4YrTGHQal18sdsJrlYCbUz-r
Klar logo resource: https://drive.google.com/drive/folders/1-Kb3ejWKjw1GTmPEiDXGr_Mfuys1CDu4

cadeyrn:
> QAReady

The logo is still purple on a purple background and the logo of Firefox Klar is still noticeable larger than the logo of Firefox Focus: The fox logo is larger as well as the font size of the word mark:

<img width="1433" alt="Bildschirmfoto 2021-08-25 um 11 39 55" src="https://user-images.githubusercontent.com/1100614/130767658-8755c208-e53a-4453-87a0-230e88d9a3a9.png">

ionutbedregeanu:
@cadeyrn Thank you! You are right, is a minor difference in size

@brampitoyo From what I see Focus logo has 384 width and 80 height while the Klar logo has 338 width and 80 height
Can you help me with another resource in order to have the same dimensions for both? Not sure which are the intended ones?

mcarare:
@cadeyrn These are the color from the design team, not sure what ales are to do about it, except for them reconsidering.

lobontiumira:
Verified as fixed on the latest debug build from main from 8/30, with Lenovo Tab M10 (Android 10)

![Untitled](https://user-images.githubusercontent.com/40694158/131352881-697a6d56-7727-4754-88d9-b3454928f49f.png)

PR #5360, opened 2021-09-09
Update production main background and launch icons

ionutbedregeanu:
For #5355 

Move all launch icons from debug to production
Move main page background from debug to production
Update nightly background


jonalmeida (approved):
Thanks for fixing!

The icon seems bigger than the recommended icon size (this can affect Samsung's Android launcher, etc if we draw outside the margins), so maybe that's something @brampitoyo might know how we can resolve too.

Let's land this for now so at least we can correct the icons and background.

Screen Shot 2021-09-13 at 2 32 50 PM

brampitoyo:
@jonalmeida Odd. I’m opening our Firefox and Focus icon assets side by side, and found that they both have identical margins from the side:

Screen Shot 2021-09-14 at 2 44 33 PM

Here are the two folders that I’m using for comparison. They both contain ready to use assets:
- Firefox `ic_launcher`: https://drive.google.com/drive/folders/12u1kRCAU62jLeaNdL9KHEhwdkd6i6ZMq
- Focus `ic_launcher`: https://drive.google.com/drive/folders/1Y8-0OaEALfYANdRlK-aF7TST5ldxnWyZ

Just to be doubly sure, I’ve also re-exported the Focus assets and re-checked them so they match Firefox’s.

cadeyrn:
@brampitoyo Could you please clarify a few questions regarding the logo?

1. Do I see it correctly that the Focus logo was changed in v91 and will be changed again in v93 (with other colors)? Or is this a pre-release-only thing?
2. Is it really expected to have such a low contrast between logo and background? For me it's almost impossible to distinguish the bottom part of the logo from the background because it's almost the same color. See also #5180 for the same issue with the start screen background. Now the launcher icon has the same issue. Reducing / Removing contrasts seems to be Mozilla's new design philosophy (see also the Proton UI for desktop where some user interface elements like the address bar are almost invisible) but is it really a good idea for the brand logo?

jonalmeida:
>     1. Do I see it correctly that the Focus logo was changed in v91 and will be changed again in v93 (with other colors)? Or is this a pre-release-only thing?

Yes, this is correct. That logo change was a previous iteration that we didn't intend to be the final version but it was released as part of 91.

brampitoyo:
Just to confirm that @Kate-Galetski will be working on this logo contrast issue, although I might end up making the actual change in the production file and re-export them!

brampitoyo:
@jonalmeida I’ve just darkened the logo background, from an old `Ink` value of `#202340`, to our darkest shade of `Ink 90` `#1d1133`.

The result should be a slightly improved contrast between the icon and its background.

You can find all the logo assets here: https://drive.google.com/drive/folders/1Y8-0OaEALfYANdRlK-aF7TST5ldxnWyZ

ionutbedregeanu:
> @jonalmeida I’ve just darkened the logo background, from an old `Ink` value of `#202340`, to our darkest shade of `Ink 90` `#1d1133`.
> 
> The result should be a slightly improved contrast between the icon and its background.
> 
> You can find all the logo assets here: https://drive.google.com/drive/folders/1Y8-0OaEALfYANdRlK-aF7TST5ldxnWyZ

@brampitoyo In order to have a single source of truth, can you please update also in Figma?

Issue #5372, opened 2021-09-13
Several Carousel tips are displayed truncated after changing the language in portrait mode


delia-pop:
## Steps to reproduce
1. Go to Settings -> General -> Language, and change to a different language (e.g. Deutsch, Netherlands).
2. Exit settings and observe the carousel tips.

### Expected behavior
The carousel tips are fully displayed.

### Actual behavior
Some carousel tips appear truncated:
2nd tip - incomplete string displayed in portrait mode.
3rd tip - the link is not displayed in portrait mode.
4th tip - only Russian language - incomplete string displayed in portrait mode.

### Device information
* Android device: Huawei P9 Lite (Android 8), Huawei P20 (Android 10).
* Focus version: 93.0.2 (Build #352522329 93.0a1-20210906031657).

![Carousel tips are displayed truncated after changing the language_2_DE_Portrait](https://user-images.githubusercontent.com/89388888/133055253-92c177b2-a6d3-4ec6-adbf-04aa225191b6.jpg)
![Carousel tips are displayed truncated after changing the language_2_DE_Landscape](https://user-images.githubusercontent.com/89388888/133055264-1c251bf1-7af3-4028-88d1-b88c47f3f157.jpg)
![Carousel tips are displayed truncated after changing the language_3_DE_Portrait](https://user-images.githubusercontent.com/89388888/133055297-ef0380e7-622d-4899-a812-0f7b489b0918.jpg)
![Carousel tips are displayed truncated after changing the language_3_DE_Landscape](https://user-images.githubusercontent.com/89388888/133055304-6b9ee767-69e3-46b8-b4e9-16205d8aa7ce.jpg)
![Carousel tips are displayed truncated after changing the language_2_Netherlands_Portrait](https://user-images.githubusercontent.com/89388888/133055348-7cc961f3-28fc-48f6-bd78-7dd32cd4be33.jpg)
![Carousel tips are displayed truncated after changing the language_2_Netherlands_Landscape](https://user-images.githubusercontent.com/89388888/133055357-a095a01c-298e-4a8c-a1f0-ee46f4deaf96.jpg)
![Carousel tip displayed truncated after changing the language_4_RU_Portrait](https://user-images.githubusercontent.com/89388888/133057529-cd756a00-7c90-4063-815e-5cddb02bac4f.jpg)
[...]

PR #5386, opened 2021-09-14
Increase the number of lines for a home tip

ionutbedregeanu:
For #5372


app/src/main/res/layout/tip_carousel_item.xml
```diff
@@ -5,11 +5,11 @@
 <com.google.android.material.textview.MaterialTextView xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:gravity="center"
+    android:layout_marginBottom="16dp"
+    android:gravity="center|bottom"
     android:lineSpacingExtra="7sp"
+    android:lines="5"
```
mcarare, line 11: Do we want to have **exactly** this many lines, or to just **allow** this many lines?

ionutbedregeanu: To allow. Probably you will suggest to use maxLines instead of lines but I tried that and it didn't work. Probably because we are using layout_weight in fragment_urlinput2

mcarare: No, it's because the first item displayed will give the height of the next ones in the current adapter used. But I guess it would be more work to make it adapt to each item's height so the workaround of adding bottom gravity should be enough for this layout in its present state.

mcarare (approved):

Issue #5369, opened 2021-09-10
Show the broken lock icon when visiting a HTTP site


Amejia481:

lobontiumira:
@Amejia481, I've encountered a situation on the OnePlus 5T (Android 10) device: trying to access http://permission.site loads an unusable page, trying to load http://muntzomani.ro loads a blank insecure page. Not reproducible on 93.0.2 build.

![OnePlus](https://user-images.githubusercontent.com/40694158/133745295-48640cf6-7a08-486a-8c25-fc98ed26c669.png)

Not reproducible on Samsung Galaxy Note 8 (Android 9), nor on Lenovo Tab M10 (Android 10).
I'll remove the QAReady label for now.

Amejia481:
> @Amejia481, I've encountered a situation on the OnePlus 5T (Android 10) device: trying to access http://permission.site loads an unusable page, trying to load http://muntzomani.ro loads a blank insecure page. Not reproducible on 93.0.2 build.
> 
> ![OnePlus](https://user-images.githubusercontent.com/40694158/133745295-48640cf6-7a08-486a-8c25-fc98ed26c669.png)
> [logcat.txt](https://github.com/mozilla-mobile/focus-android/files/7183734/logcat.txt)
> 
> Not reproducible on Samsung Galaxy Note 8 (Android 9), nor on Lenovo Tab M10 (Android 10).
> I'll remove the QAReady label for now.

Is this also reproduceable in Fenix?

Nickoriginal:
@lobontiumira do you have JavaScript disabled? Broken error page due to disabled JavaScript is a known issue.
<img src = https://user-images.githubusercontent.com/85299944/133878978-85f289ee-a391-41f3-baeb-fa1bd309a286.jpg width=35%>

lobontiumira:
@Amejia481 this does not reproduce on Fenix Nightly 94.a01 from 9/20 with the same device.
But I am unable to reproduce it again on the debug build (352581202 with GV 93.0a1-2021090631657).
Not reproducible on 93.0.3 build either.
I'm closing the issue.

PR #5402, opened 2021-09-14
For #5369 show the broken lock icon when visiting a HTTP site

Amejia481:


app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -150,12 +159,35 @@ class BrowserToolbarIntegration(
...
+                                indicators = indicators + Indicators.SECURITY
+                            }
+                        }
+                    } else if (Indicators.SECURITY in toolbar.display.indicators && it.content.securityInfo.secure) {
+                        toolbar.display.apply {
+                            indicators = indicators - Indicators.SECURITY
+                        }
+                    }
+                }
+        }
```
pocmo, line 183: nit: There is an opportunity to write this more compact:
- Since we only need the boolean flag, we can map to that flag and then use `ifChanged()`
- On the indicator list we can use `+=` and `-=" which avoids using those multi-line `apply()` statements
- Currently one of the if cases is nested and the other is not

So we end up with something like this:

```Kotlin
        securityIndicatorScope = store.flowScoped { flow ->
            flow
                .mapNotNull { state -> state.findCustomTabOrSelectedTab(customTabId)?.content?.securityInfo?.secure }
                .ifChanged()
                .collect { secure ->
                    if (secure && Indicators.SECURITY in toolbar.display.indicators) {
                        toolbar.display.indicators -= Indicators.SECURITY
                    } else if (!secure && Indicators.SECURITY !in toolbar.display.indicators) {
                        toolbar.display.indicators += Indicators.SECURITY
                    }
                }
        }
```

app/src/main/res/drawable/mozac_ic_broken_lock.xml
```diff
@@ -0,0 +1,19 @@
...
+    <path
+        android:fillColor="?attr/primaryText"
+        android:pathData="M7 7.562V11H6.5A2.5 2.5 0 0 0 4 13.5v6A2.5 2.5 0 0 0 6.5 22h11c1.066 0 1.97-0.67 2.329-1.61L18.5 19.061V19.7l-0.8 0.8H6.3l-0.8-0.8v-6.4l0.8-0.8h5.638l-1.5-1.5H8.5V9.062L7 7.562z" />
+    <path
+        android:fillColor="#FF4F5E"
+        android:pathData="M4.78 3.22l17.5 17.5A0.748 0.748 0 0 1 21.749 22a0.748 0.748 0 0 1-0.53-0.22L3.719 4.281A0.75 0.75 0 1 1 4.78 3.22z" />
+    <path
+        android:fillColor="?attr/primaryText"
+        android:pathData="M12 3.5c1.93 0 3.5 1.57 3.5 3.5v4h-0.809l1.5 1.5H17.7l0.8 0.8v1.509l1.5 1.5V13.5a2.5 2.5 0 0 0-2.5-2.5H17V7a5 5 0 0 0-5-5 4.99 4.99 0 0 0-4.127 2.182L8.971 5.28C9.574 4.222 10.698 3.5 12 3.5z" />
+</vector>
```
pocmo, line 19: nit: I think now there's a case for having this icon in A-C and even use them as default in the toolbar. :)

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -150,12 +159,35 @@ class BrowserToolbarIntegration(
 
         customTabsFeature?.start()
         navigationButtonsIntegration?.start()
+        observerSecurityIndicatorChanges()
+    }
+
+    private fun observerSecurityIndicatorChanges() {
```
pocmo, line 165: For now this workaround in Focus may be okay. But we should consider moving this behavior to A-C (into the `Presenter`) to avoid having potentially different behavior implementations in our apps (of course this needs to be configurable, since Fenix doesn't need it for now). In A-C, in the `Presenter`, we can also cover it with tests.. :)

pocmo (approved):

mcarare:
I think this might be a better solution, but it seems to be different than the UX proposed by @brampitoyo on Slack with the shield icon being replaced by the lock icon.

Amejia481:
> I think this might be a better solution, but it seems to be different than the UX proposed by @brampitoyo on Slack with the shield icon being replaced by the lock icon.

Thanks letting know about it :)

app/src/main/res/drawable/mozac_ic_broken_lock.xml
Amejia481: Actually, this is the icon that we are using as a default in [AC](https://github.com/mozilla-mobile/android-components/blob/main/components/ui/icons/src/main/res/drawable/mozac_ic_broken_lock.xml), we are re-adding it here to customize the `fillColor` as we don't have API for it on AC 😅 .

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
Amejia481: ✨

Issue #5382, opened 2021-09-14
Dismissing the keyboard clears the address bar


Vinnl:
**Describe the bug**

If I type in the address bar, then dismiss the keyboard (e.g. to be able to see more search suggestions), everything I typed into the address bar is removed.

**To Reproduce**

1. Type something in the address bar.
2. Press the Android down button.
3. (Optional) Press outside the search suggestions.

**Expected behavior**

I can press in the address bar again to complete the URL I was typing.

**Actual behavior**

I have to retype the full URL and make sure not to dismiss the keyboard.

**Smartphone (please complete the following information):**

- Device: Fairphone 2
 - OS: Android 9
 - Focus Version: the one built for the Sep 13- Sep 17, 2021 Foxfooding

mcarare:
@brampitoyo What should be the behavior here? In Fenix, I think we also hide the search suggestions when we dismiss the keyboard.
Note that more search suggestions can be viewed by scrolling the suggestions list, not by dismissing the keyboard.

Vinnl:
@mcarare 

>  Note that more search suggestions can be viewed by scrolling the suggestions list, not by dismissing the keyboard.

Maybe that's where I went wrong: if I scroll through the suggestions in Fenix, the keyboard is dismissed so that I can see more suggestions at the same time (without clearing the address bar). In Focus, the keyboard remains open, obscuring the view of a couple of search suggestions.

mcarare:
@Vinnl You are right, that behavior was implemented in Fenix, but I know there were some UX discussions around it, so I think @brampitoyo can add more insight to what we should try to do in order to provide the best experience for the users.

brampitoyo:
@mcarare @Vinnl I think the right behaviour here should be:

- If you scroll through search suggestions, the keyboard will be dismissed but your search query remains
- If, on the other hand, you dismiss the keyboard by pressing the down arrow or swiping, we count it as a “Back” action. In this case, the keyboard will be dismissed _and_ the address bar will be un-selected. Since the address bar is no longer active, your search query will reset as well

Vinnl:
@brampitoyo For that second bullet, one thing that also confused me was that search suggestions remained visible, even though the query they were based on was removed. Possibly they should be cleared as well?

mcarare:
Yes, in Fenix the suggestions are cleared when the keyboard is closed and the search field cleared. We should do the same here.

brampitoyo:
[...]

PR #5412, opened 2021-09-16
For #5382: Improve user interaction with suggestions list.

mcarare:
Editing cancellation is being triggered by pressing the back button.


pocmo (approved):
Works well and code looks good too :) 👍

Issue #5368, opened 2021-09-10
Intermittent UI test failure - < CustomTabTest. testCustomTabUI >


AndiAJ:
### Firebase Test Run: 
[Firebase link](https://console.firebase.google.com/u/0/project/moz-fx-mobile-firebase-testlab/testlab/histories/bh.2b4ca2c7fdb2f0f0/matrices/5532166430602730345/executions/bs.ce8710894895e76c/testcases/1/logs)
[Firebase link](https://console.firebase.google.com/u/0/project/moz-fx-mobile-firebase-testlab/testlab/histories/bh.2b4ca2c7fdb2f0f0/matrices/5532166430602730345/executions/bs.41b982a67e186372/testcases/1/logs)

### Stacktrace:

--------- beginning of crash
09-09 16:57:29.097: E/AndroidRuntime(8791): FATAL EXCEPTION: main
09-09 16:57:29.097: E/AndroidRuntime(8791): Process: org.mozilla.focus.debug, PID: 8791
09-09 16:57:29.097: E/AndroidRuntime(8791): android.content.res.Resources$NotFoundException: Drawable org.mozilla.focus.debug:drawable/toolbar_url_dark_background with resource ID #0x7f080193
09-09 16:57:29.097: E/AndroidRuntime(8791): Caused by: android.content.res.Resources$NotFoundException: File res/drawable/toolbar_url_dark_background.xml from drawable resource ID #0x7f080193
09-09 16:57:29.097: E/AndroidRuntime(8791): 	at android.content.res.ResourcesImpl.loadDrawableForCookie(ResourcesImpl.java:847)
09-09 16:57:29.097: E/AndroidRuntime(8791): 	at android.content.res.ResourcesImpl.loadDrawable(ResourcesImpl.java:631)
09-09 16:57:29.097: E/AndroidRuntime(8791): 	at android.content.res.Resources.getDrawableForDensity(Resources.java:888)
09-09 16:57:29.097: E/AndroidRuntime(8791): 	at android.content.res.Resources.getDrawable(Resources.java:827)
09-09 16:57:29.097: E/AndroidRuntime(8791): 	at androidx.core.content.res.ResourcesCompat.getDrawable(ResourcesCompat.java:95)
09-09 16:57:29.097: E/AndroidRuntime(8791): 	at org.mozilla.focus.browser.integration.BrowserToolbarIntegration.setUrlBackground(BrowserToolbarIntegration.kt:139)
[...]

PR #5415, opened 2021-09-16
Fix CustomTabTest.testCustomTabUI

ionutbedregeanu:
For #5368 
This test is trying to get toolbar_url_dark_background from light theme resources,
were it doesn't exist because it was declared only for dark theme


AndiAJ:
Hi @ionutbedregeanu , due to the taskgraph changes currently being done we can't run the UI tests on Firebase for the moment.
Nevertheless, I've just checked your fix locally and the `testCustomTabUI` crash seems to be fixed.
Thanks for your help ! ☺️

mcarare (approved):

PR #5420, opened 2021-09-17
For #5257: Use activity alias for launcher activity.

mcarare:
This allows for flexibility in handling shortcuts.

Fenix also uses <activity-alias> for the launcher activity and there are no pitfalls mentioned in the official documentation: https://developer.android.com/guide/topics/manifest/activity-alias-element.

Tested on a device with shortcuts added before the change, no regressions found.


jonalmeida (approved):
> Fenix also uses for the launcher activity and there are no pitfalls mentioned in the official documentation: https://developer.android.com/guide/topics/manifest/activity-alias-element.

I think the only pitfall would be that shortcuts/widgets that depend on the alias in the future require us to always keep the alias around.

Could we keep a small comment above the alias about this? [Fenix mentions](https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/AndroidManifest.xml#L37-L43) why we inherited one from Fennec as well.

Issue #5424, opened 2021-09-20
Browser toolbar remains invisible on pages where scroll is not possible.


mcarare:
## Steps to reproduce

1. Open the mozilla-mobile.github.io/testapp/.
2. Scroll down to hide the browser toolbar.
3. Click on the senglehardt.com link

Note: This STR can be used with any page that has a link to a page that is not scrollable.

### Expected behavior

Browser toolbar should be visible or there should be a mechanism to render it visible.

### Actual behavior

Browser toolbar remains hidden.

### Device information

* Android device: Any.
* Focus version: main branch / 93.0.3

LaurentiuApahideanSV:
Verified as fixed on the latest debug build from main 09/21.
### Devices used:
- Oppo Find X3 (Android 11)
- OnePlus 6T (Android 9)

PR #5425, opened 2021-09-20
For #5424: Show browser toolbar when navigating to another page.

mcarare:


app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -164,6 +169,17 @@ class BrowserToolbarIntegration(
         customTabsFeature?.start()
         navigationButtonsIntegration?.start()
         observerSecurityIndicatorChanges()
+        observeCurrentUrlChanges()
+    }
+
+    private fun observeCurrentUrlChanges() {
+        fragment.consumeFlow(store) { flow ->
+            flow.map { state -> state.selectedTab?.content?.url }
+                .ifChanged()
```
jonalmeida, line 178: Should this be something like:

```suggestion
            flow.map { state -> state.selectedTab?.content }
                .ifAnyChanged { arrayOf(loading, url) }
```

Or maybe just loading? We can get URL updates without navigation changes (like anchor links).

jonalmeida (approved):
Left a comment for consideration, but this is still fine to land for now since it's an improvement nonetheless.

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
jonalmeida: Ah, looks like AC's [toolbar controller](https://searchfox.org/mozilla-mobile/source/android-components/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarBehaviorController.kt#41) does just that - observes `content.loading`.

Related: I wonder why we aren't using the `TabsFeature` here and getting this for free..

Might be another one of those toolbar things to investigate after 93. 😄

mcarare: The AC handles its own toolbar, we still have in Focus a separate DisplayToolbar that currently handles the hiding on scroll...

mcarare: We can rely on `content.showToolbarAsExpanded `when we remove the Focus `DisplayToolbar` and set the behavior for the AC `BrowserToolbar`.

PR #5428, opened 2021-09-21
Update Mergify rules to verify status of taskhgraph tasks.

pocmo:
The taskgraph changes just landed on `main`. Since taskgraph creates tasks with different names, we need to update `mergify.yml` so that Mergify knows when a PR is ready to be merged.


pocmo:
With this change Mergify will not look at `ui-test-x86-debug` yet, until it is stable and passing.

Issue #5429, opened 2021-09-21
[debug] Focus freezes after tapping on the trashcan


lobontiumira:
## Steps to reproduce
1. Install the latest debug build from main from 9/21.
2. Open a page, then tap on the trashcan.

### Expected behavior
The page closes, the homescreen is displayed with the "Browsing history cleared" snackbar.

### Actual behavior
Focus freezes. The page is unusable.
The URL search field is empty, but the three-dot menu can be accessed.
Re-tapping on the trashcan, displays a "Erase browsing history" snackbar".

### Device information
* Android device: Samsung Galaxy Note 8 (Android 9), OnePlus 5T (Android 10)
* Focus version: 93.0 (build 11 GV 93.0a1-20210906031657)

[logcat.txt](https://github.com/mozilla-mobile/focus-android/files/7203375/logcat.txt)

https://user-images.githubusercontent.com/40694158/134178368-abd00e7f-5a38-4d27-9f7b-1042ab21465f.mp4

lobontiumira:
Verified as fixed on the latest debug build from main from 9/22, with Samsung Galaxy Tab A6 (Android 5.1.1), Samsung Galaxy Note 8 (Android 9), and Google Pixel (Android 10).

PR #5434, opened 2021-09-21
For #5429: Remove initial workaround for opening search from another app

mcarare:
This prevented navigation to home screen on using the erase button.


jonalmeida (approved):
That makes sense. ✅

PR #5437, opened 2021-09-21
Fix .taskcluster.yml rendering to avoid error when no tasks are produced

bhearsum:
This is horrible to read, but it's almost exclusively indentation changes. I've dropped the highest level `mergeDeep` in favour of nested `if/then/else` for each type of task we want to generate. This avoids a `tasks` variable that ends up with a single `{}` in it, and instead gives us an empty list if nothing matches.


escapewindow (approved):
Looking good. Stamping; thank you!

Issue #5450, opened 2021-09-22
Pro tip touch target area is too small


brampitoyo:
We have the same issue on https://github.com/mozilla-mobile/focus-ios/issues/2376, which had since been fixed.

STR
- open focus
- tap on a pro tip on the home screen

The touch target size for the link to learn more (e.g. "Read more about this and other updates to Focus") doesn't seem to be meeting our min touch target size. It seems like the touch target might be just limited to the actual purple text?

See the Focus iOS issue on how to make the entire area tappable, instead of only the text.

delia-pop:
Verified as fixed on the latest debug build from main 09/30 (Build 11 GV 94.0a1-20210927093458) with Huawei P9 Lite (Android 8).
Tapping around the area of the pro tip on homescreen will open the link.

PR #5458, opened 2021-09-22
Pro tip touch area increase

ionutbedregeanu:
For #5450


jonalmeida (approved):
Nice, took me a minute to understand what you did there. 😄

PR #5461, opened 2021-09-22
Fix chain of trust error in build signing tasks

ahal:
The comment in 'Dockerfile' was modified to force docker-image rebuilds.
Otherwise taskgraph would replace it with an older image task and we
would continue to hit this CoT issue until the tasks were rebuilt.


ahal:
Waiting for some testing to complete on `staging-focus-android` before requesting review.

Issue #5254, opened 2021-08-30
[bug] Closing one tab at a time (by tapping back) displays [0] tabs briefly


lobontiumira:
## Steps to reproduce
1. Open a page with multiple links.
2. Long-tap a link and open it in a new tab, multiple times in order to have at least 3 tabs opened.
3. Tap the back button, and observe the tabs counter.

### Expected behavior
The tabs counter displays the remaining opened tabs, without glitches.

### Actual behavior
The tabs counter displays 0 (zero) remaining tabs briefly - please see the attached slow motion short video.

### Device information
* Android device: Lenovo Tab M10 (Android 10), Samsung Galaxy Note 8 (Android 9)
* Focus version: 93.0.0, the debug build from 8/27 main

https://user-images.githubusercontent.com/40694158/131309143-bac38421-e401-407b-8026-ce90c547ad82.mp4

lobontiumira:
Reproducible on RC 91.4.0 with Samsung Galaxy Note 8 (Android 9).

lobontiumira:
Reproducible also on RC 92.1.1 with Lenovo Tab M10 (Android 10).

sflorean:
This is still reproducible on 93.0.2 with Huawei MediaPad M2 (Android 5.1.1), and instead of [0] the trashcan is displayed.

LaurentiuApahideanSV:
The issue still occurs on the latest debug build from main 09/21.
### Device used:
- Huawei MediaPad M3 (Android 7)

ionutbedregeanu:
> The issue still occurs on the latest debug build from main 09/21.
> 
> ### Device used:
> * Huawei MediaPad M3 (Android 7)

@LaurentiuApahideanSV tabs counter is still displays 0?

LaurentiuApahideanSV:
I've retested the issue on the latest debug build from main 09/22. The tab counter does no longer display 0, however it does display the trash can icon when using the device back button. I've attached a slow down video showing the issue.

https://user-images.githubusercontent.com/50983519/134310213-64ffe107-8c1a-4480-9cd4-e17445ad124e.mp4

lobontiumira:
@ionutbedregeanu, here is the current behavior on the latest debug build from main from 9/28:

https://user-images.githubusercontent.com/40694158/135050672-a09b152a-fafb-46b2-aa10-8d8c2b78e023.mp4

[...]

PR #5472, opened 2021-09-23
Update keepHidden default value

ionutbedregeanu:
For #5254
Erase button state is not persisted between tabs and it is restored at every switch. 
keepHidden default value was false which makes the button to pops up even when the number of tabs was bigger than 1
By updating keepHidden default value to true we avoid those scenarios when the visibility of button is set to VISIBLE and imediately after that is set to GONE


jonalmeida (approved):
I think this is fine. I'm not sure what the side-effects (if any) might be though.

Issue #5452, opened 2021-09-22
Implement AC dynamic toolbar behavior. 


mcarare:
As a follow-up on issues like https://github.com/mozilla-mobile/focus-android/issues/5381, we should remove the Focus DisplayToolbar and rely only on the AC toolbar behavior to hide and show the toolbar and the info "exchange with the engine view.

Tagging @Mugurell for adding more context and action points if needed.

PR #5476, opened 2021-09-24
For #5452 - Replace AppBarLayout.Behavior with BrowserToolbarBehavior

Mugurell:
This behavior in correlation with EngineViewBrowserToolbarBehavior would ensure
the dynamic toolbar considers what happens in the browser for whether it should
be translated or not and the browser is in sync with toolbar's translation to
continuously update the display height so that the webpage is always displayed
fully even with the toolbar collapsed or expanded.

For the moment this will regress the toolbar fading when it is being collapsed.

https://user-images.githubusercontent.com/11428869/134632613-a2bc2e95-e6af-4221-b675-191889ade216.mov


Mugurell:
This patch has everything working the same or better (with only one exception) but it's too late for it to make get in the next release so will keep it as a draft for now.

The one exception - something that is regressed - is the fade effect for when the toolbar is collapsed.
That is a bit hard to implement now because all the views that are animated together - toolbar, engineView, fabs have to be siblings in the same CoordinatorLayout parent.
But the toolbar needs a new View as parent to serve as the static background that isn't fading.
Tried to loosen the AC behavior which [currently works with only a BrowserToolbar](https://github.com/mozilla-mobile/android-components/blob/71e124a5dbafffb800168ea615fdbc54d7398f50/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/behavior/BrowserToolbarBehavior.kt#L131), make it work with any View but then saw that even the ToolbarController is of course [tied to a BrowserToolbar](https://github.com/mozilla-mobile/android-components/blob/17b5545a70ce4fb83bc72f50edf580452b7b02c4/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarBehaviorController.kt#L26) and [calls specific BrowserToolbar methods](https://github.com/mozilla-mobile/android-components/blob/17b5545a70ce4fb83bc72f50edf580452b7b02c4/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarBehaviorController.kt#L67) so these changes would probably be too much.

One approch would be to have another BrowserToolbar implementation, maybe something like `DynamicFadingBrowserToolbar` and then maybe also expand `BrowserToolbarbehavior` for this new toolbar.
Another aproach would be to look into improving the AC behaviors, bring them more in line with how `AppBarLayout.Behavior`s work and get everything working here instead of needing a new toolbar implementation. (We'd probably need new behaviors though)

Since this is not needed in the next immediate release we can take the time and think about what the best approach would be.

Mugurell:
@pocmo @jonalmeida Now that 94 started I'd say migration to the AC behavior can happen.
But the issue from above with the fading effect remains.
Thinking about how big the changes would be to support this now I'd propose to migrate without the fading effect and leave this to be reimplemented later in Compose where it should be pretty straighforward.
Though there we'd also need something in the likes of what was proposed in https://github.com/mozilla-mobile/android-components/issues/10585 -  a new simpler API from GV to allow syncing gestures.
Thoughts?

pocmo:
> Thinking about how big the changes would be to support this now I'd propose to migrate without the fading effect and leave this to be reimplemented later in Compose where it should be pretty straighforward.

Agree. The fading effect was nice, but it's not essential.

Issue #5473, opened 2021-09-23
Clicking the secure/ETP icon in a Custom Tab crashes


jonalmeida:
## Steps to reproduce
1. Set Focus as the default browser.
2. Clear app data for Focus.
1. Open a site in a Custom Tab.
2. Click on the secure lock or the ETP icon (which ever shows up first)

### Expected behavior
- You see the site security/ETP panel.

### Actual behavior
- 💥

### Device information

* Android device: Pixel 3
* Focus version: `main` from 2 days ago (sorry, filling on mobile, will update later if needed)

mcarare:
I can't seem to reproduce this with on the latest master, with opening a custom tab from youtube. 
@jonalmeida Is there a specific page you have opened?

jonalmeida:
There wasn't a specific page. I'm still able to reproduce this:

```
2021-09-27 18:12:21.723 27701-27701/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.mozilla.focus.debug, PID: 27701
    android.view.InflateException: Binary XML file line #39 in org.mozilla.focus.debug:layout/dialog_tracking_protection_sheet: Binary XML file line #39 in org.mozilla.focus.debug:layout/dialog_tracking_protection_sheet: Error inflating class com.google.android.material.textview.MaterialTextView
    Caused by: android.view.InflateException: Binary XML file line #39 in org.mozilla.focus.debug:layout/dialog_tracking_protection_sheet: Error inflating class com.google.android.material.textview.MaterialTextView
    Caused by: java.lang.reflect.InvocationTargetException
        at java.lang.reflect.Constructor.newInstance0(Native Method)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
        at android.view.LayoutInflater.createView(LayoutInflater.java:852)
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:1004)
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:959)
        at android.view.LayoutInflater.rInflate(LayoutInflater.java:1121)
        at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:1082)
        at android.view.LayoutInflater.rInflate(LayoutInflater.java:1124)
        at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:1082)
        at android.view.LayoutInflater.inflate(LayoutInflater.java:680)
[...]

PR #5494, opened 2021-09-28
For #5473: Temporary add night resources in light theme colors.

mcarare:
These will be changed in https://github.com/mozilla-mobile/focus-android/pull/5350 but will fix the crashes when color resources cannot be resolved.


mcarare:
Thanks to @jonalmeida for testing the fix.

PR #5496, opened 2021-09-28
Add nightly support

sarah-clements:
Tested on staging-focus-android repo, and triggered the cron task hook - all nightly tasks pass: https://firefox-ci-tc.services.mozilla.com/tasks/groups/UIprIWGVQhu2LQWL1G5z8g

I'll work on the nightly ui test in a separate pr.


sarah-clements:
@pocmo Did you still want to have beta and nightly to build on every pr?

.taskcluster.yml
```diff
@@ -167,87 +167,10 @@ tasks:
...
-                          cron: {$json: {$eval: 'cron'}}
-                          tasks_for: ${tasks_for}
-                        metadata:
-                          name: (Focus for Android) Focus/Klar Nightly Builds (Public)
-                          description: Decision task scheduled by cron task [${cron.task_id}](https://tools.taskcluster.net/tasks/${cron.task_id})
-                          owner: ${event.sender.login}@users.noreply.github.com
-                          source: ${repository}/raw/${event.release.tag_name}/.taskcluster.yml
-                    else:
-                      # cron commented out until we port it to taskgraph
-                      #tasks_for in ["action", "cron"]
```
bhearsum, line 246: You may need to dedent the block beneath this, depending how fussy yaml is about consisent indentation. If it worked on staging it's probably fine...but if things stop firing after this is merged, it could be something to investigate.

taskcluster/focus_android_taskgraph/target_tasks.py
```diff
@@ -31,38 +31,11 @@ def filter(task, parameters):
...
-def target_tasks_bump_android_components(full_task_graph, parameters, graph_config):
-    """Select the set of tasks required to update android components."""
-
-    def filter(task, parameters):
-        return task.attributes.get("bump-type", "") == "android-components"
-
-    return [l for l, t in full_task_graph.tasks.items() if filter(t, parameters)]
 
+    def filter(task):
+        return 'nightly' in task.label.split('-')
```
bhearsum, line 39: I think it would be better if you did what Fenix does, and look at attributes for this (https://github.com/mozilla-mobile/fenix/blob/main/taskcluster/fenix_taskgraph/target_tasks.py#L30). This is a bit more brittle and implicit.

bhearsum (changes requested):
r=me with the target task change.

pocmo:
> @pocmo Did you still want to have beta and nightly to build on every pr?

If that is easy to keep/maintain then yeah, I would like to keep it, so that we notice build-specific issues in the PR already.

> Also, I can see that [Fenix's taskgraph](https://github.com/mozilla-mobile/fenix/blob/main/taskcluster/fenix_taskgraph/transforms/build.py#L111) includes `Pofficial` flag when running the gradle command to build nightly; is that needed for Focus?

It's not needed at the moment. But we could add it here too - then we can make use of it in the future to enable certain features only for CI builds (e.g. telemetry - CC @jonalmeida).

taskcluster/focus_android_taskgraph/target_tasks.py
sarah-clements: The problem with what Fenix did is that `attributes: nightly: true` is only explicitly used in the builds kind and its not at all clear how signing and push-apk nightly tasks are run by the cron task. In my mind, this approach is more obvious. Nightly tasks are all clearly labeled, so are release and beta. If you were to insist that we should go by attribute instead of label, then I'd rename it to something less obscure like `attributes: nightly-task: true` but then I'd do that for everything, so every single kind would have an extra attribute (for release-task, for beta-task, etc).

sarah-clements:
> > @pocmo Did you still want to have beta and nightly to build on every pr?
> 
> If that is easy to keep/maintain then yeah, I would like to keep it, so that we notice build-specific issues in the PR already.

Sure, not a problem.

> > Also, I can see that [Fenix's taskgraph](https://github.com/mozilla-mobile/fenix/blob/main/taskcluster/fenix_taskgraph/transforms/build.py#L111) includes `Pofficial` flag when running the gradle command to build nightly; is that needed for Focus?
> 
> It's not needed at the moment. But we could add it here too - then we can make use of it in the future to enable certain features only for CI builds (e.g. telemetry - CC @jonalmeida).

The build.py transform has already been added so you can use this feature whenever you're ready.

pocmo (approved):

taskcluster/focus_android_taskgraph/target_tasks.py
sarah-clements: I added `nightly-task` attribute per our discussion in the relounge channel and tested it on the staging repo.

sarah-clements:
@pocmo This is ready to be merged whenever you're ready.

jonalmeida:
> > @pocmo Did you still want to have beta and nightly to build on every pr?
> 
> If that is easy to keep/maintain then yeah, I would like to keep it, so that we notice build-specific issues in the PR already.

If we have a PR against `main` or a particular branch, we would only be building against a specific GV version. So if we're building prod/beta/nightly there would always be 2/3 variant builds that are using the wrong GV version for it, right? Is the intention to make sure we're always building and testing against variant specific code? 

The latter works too, but I wanted to clarify for my own understanding. 🙂

pocmo:
> If we have a PR against `main` or a particular branch, we would only be building against a specific GV version. So if we're building prod/beta/nightly there would always be 2/3 variant builds that are using the wrong GV version for it, right?

What do you mean with "wrong" GV version in this case? We are building the nightly/beta Gradle build type - so with a potentially different build configuration. It's not a release or beta version of GeckoView, but it's the version this branch will always be build against, no matter where its in the train.

If we have feature / build flags then this prevents us from accidentally breaking the code paths that are not part of the "debug" build.

jonalmeida:
> It's not a release or beta version of GeckoView, but it's the version this branch will always be build against, no matter where its in the train.

I thought there was an expectation that the GV version was also expected to match the gradle build types. Thanks for clarifying. 🙂

PR #5499, opened 2021-09-29
possibly fix taskgraph release

escapewindow:


jonalmeida (approved):
✅ 

These are the same changes we landed and tested in `releases_93.0`.

PR #5503, opened 2021-09-29
Update dev launcher icon

jonalmeida:
For our dev builds, we were using the wrong icon, making it hard to distinguish against release builds. Instead of just fixing the icon reference, I made a refreshed version. 🙃 

| **Before** | **After** |
|------------|----------|
| Screen Shot 2021-09-29 at 1 49 58 AM | Screen Shot 2021-09-29 at 2 46 27 AM |

(there is a smiley face on the fox but it's too light and doesn't come through. 😅 )


pocmo (approved):

pocmo:
Does the Fox have an eye? 🙃

jonalmeida:
It does, but you have to squint! 😄

Issue #5487, opened 2021-09-28
Exiting a youtube video fullscreen is not possible because of invisible overlapping tab switcher


jonalmeida:
## Steps to reproduce
1. Ensure you **DO NOT** have multiple tabs.
2. Open a tab to youtube.com; navigate to a video page.
3. Ensure the trash icon is visible and then rotate to landscape mode (this seems to matter) 
4. Start the video in fullscreen mode.
3. Attempt to exit the video by clicking on exit mode button in the lower right hand corner.

### Expected behaviour
- You can successfully exit full screen mode and continue browsing.

### Actual behaviour
- You are unable to click the button because of an overlapping task switch FAB (see video).
- The task switcher should not be enabled because there is only one tab.
- Rotating back to portrait persists the bug until you erase the session.

https://user-images.githubusercontent.com/1370580/135010885-fa67bf68-395d-4e1e-b143-f6e9d3839425.mp4

### Device information

* Android device: All devices
* Focus version: main @ e595eda

jonalmeida:
Looks like the FAB is only `View.INVISIBLE` and not `View.GONE` so our clicks are not passing to the `EngineView` below.

<img width="421" alt="Screen Shot 2021-09-27 at 10 03 10 PM" src="https://user-images.githubusercontent.com/1370580/135010976-6b57913d-c31c-4aaa-a524-437d89a9e694.png">

PR #5507, opened 2021-09-29
Update floating button auto hide state in real time

ionutbedregeanu:
For #5487

By setting auto hide state to false we are able to control manually the visibility of a floating button, in this manner we avoid those scenarios when we are trying to show the button but the default behavior class hides it
"enable" variable was removed because put the button in an inconsistent state, not visible but enabled


jonalmeida (approved):
Excellent find! ✨

PR #5525, opened 2021-10-01
Add Release push apk support

sarah-clements:
Did a trial run on staging-focus-android repo, with fake pushes to google play store: https://firefox-ci-tc.services.mozilla.com/tasks/groups/NeJ984HaRVSAzByfCy45qw

@rvandermeulen and @pocmo @jonalmeida With this pr, one change you'll notice is that the push-apk tasks are broken into two - one each for klar and focus (pushapk scriptworker has been updated), so its consistent with how we've broken them into two different task for builds and signing.

### DIFF
--- a/taskcluster/focus_android_taskgraph/target_tasks.py
+++ b/taskcluster/focus_android_taskgraph/target_tasks.py
@@ -11,8 +11,16 @@ def target_tasks_default(full_task_graph, parameters, graph_config):
     """Target the tasks which have indicated they should be run on this project
     via the `run_on_projects` attributes."""
 
-    filter = filter_for_tasks_for
-    return [l for l, t in full_task_graph.tasks.items() if filter_for_tasks_for(t, parameters)]
+    # Temporary until we have official beta support
+    def releases(task, parameters):
+        return task.attributes.get("release-type", "") == 'release'
+
+    _filter = filter_for_tasks_for
+
+    if parameters["tasks_for"] == "github-release":
+        _filter = releases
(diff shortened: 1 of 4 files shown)


taskcluster/focus_android_taskgraph/target_tasks.py
```diff
@@ -11,8 +11,16 @@ def target_tasks_default(full_task_graph, parameters, graph_config):
...
+
+    _filter = filter_for_tasks_for
+
+    if parameters["tasks_for"] == "github-release":
+        _filter = releases
+
+    return [l for l, t in full_task_graph.tasks.items() if _filter(t, parameters)]
 
 
 @_target_task('release')
```
sarah-clements, line 26: I didn't use the target_task logic yet because in Fenix the `target-tasks-for` to hit this custom release target task is set in the custom parameters.py since it doesn't look like it'll be needed once we switch to shipit releases as part of the next segment of beta/release work.

sarah-clements:
Please merge this as soon as its approved so it'll be ready for relman next week.

taskcluster/focus_android_taskgraph/target_tasks.py
bhearsum: Yeah, I don't think we'll need to pay attention `github-release` at all after that work is done (because Ship It should always be used to kick off releases).

bhearsum (approved):

pocmo (approved):

PR #5537, opened 2021-10-04
Update version to 94.0

jonalmeida:


jonalmeida:
Note: We're 95 on nightly (main), but we're need to cut the beta branch with 94.

Issue #5536, opened 2021-10-04
Focus nightly/beta app name doesn't include nightly/beta in the app_name


jonalmeida:
## Steps to reproduce
 - Install a build of Nightly/Beta.
 - Look at the app name in the launcher or app info.

### Expected behaviour
- The app name should be 'Firefox Focus Nightly' or 'Firefox Focus Beta'.

### Actual behaviour
- All the variants have the same app name.

### Device information

* Android device: All devices
* Focus version: `main` with a [sample nightly build](https://firefox-ci-tc.services.mozilla.com/tasks/YNEc6wQZQTywXaQIZEEi9Q#artifacts).

ionutbedregeanu:
Should we update also the app name used inside of the app(tips, settings, etc.)?

jonalmeida:
> @jonalmeida Should we update also the app name used inside of the app(tips, settings, etc.)?

If we override the `app_name` reference correctly (see review comments), we should get those for free.

PR #5539, opened 2021-10-05
Update app name for beta and nightly

ionutbedregeanu:
For #5536


app/src/klar/res/values/app.xml
```diff
@@ -4,4 +4,6 @@
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <resources>
     <string name="app_name" translatable="false">Firefox Klar</string>
+    <string name="beta_app_name" translatable="false">Firefox Klar Beta</string>
```
jonalmeida, line 7: We don't ship beta/nightly for Klar so we don't need this. 🙂

app/src/main/res/values/app.xml
```diff
@@ -4,6 +4,8 @@
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <resources>
     <string name="app_name" translatable="false">Firefox Focus</string>
+    <string name="beta_app_name" translatable="false">Firefox Focus Beta</string>
```
jonalmeida, line 7: We have build variants for strings so we could put these in the correct directories and the Android OS would know how to pick the right `app_name`:

| Directory | `app_name` in the strings.xml | Notes |
|----------|--------------------------------|-------|
| `/main` | `Firefox Focus` | Default string for all variants (including debug) |
| `/focusBeta` | `Firefox Focus Beta` | |
| `/focusNightly` | `Firefox Focus Nightly` | |
| `/klar` | `Firefox Klar` | Default string for all Klar variants (although we don't ship nightly/beta for it) |

jonalmeida: Ah, in Focus it looks like we put the app name in the `app.xml` instead of the `strings.xml` probably so that we don't mistakenly try to translate it.

app/build.gradle
```diff
@@ -63,18 +63,22 @@ android {
                 println ("All builds will be debuggable")
                 debuggable true
             }
+            manifestPlaceholders = [appLabel:"@string/app_name"]
```
jonalmeida, line 66: With the changes in the other comments, I don't think we will need this.

PR #5543, opened 2021-10-06
For #5057: Fix user_lifetime_expiration warning.

mcarare:
Metrics with 'user' lifetime cannot have an expiration date.


mcarare:
This was the complete warning:

>   WARNING: USER_LIFETIME_EXPIRATION: tracking_protection.has_ever_changed_etp: Metrics with 'user' lifetime cannot have an expiration date. They live as long as the user profile does. Set expires to 'never'.

@travis79 Is there anything that needs to be changed so these warnings always appear? I stumbled upon this when I had other errors. Otherwise, it does not show up, even when you have syntax errors.

travis79:
@mcarare I'm not certain about what you can do to ensure this warning always shows up, I'm pinging @mdboom as he may know off the top of his head if there is a solution since he has more experience working on the glean_parser, where this error is coming from.

mdboom:
I'm not sure why those errors aren't always showing up.  At the glean_parser level there's no flag to hide them or anything.  I suspect it's *something something* Gradle.  I'll probably need to build Focus myself and play around with it.  Are you seeing this behavior in AndroidStudio or the commandline (because occasionally that makes a difference).

Issue #5316, opened 2021-09-02
Regression: Focus does not return to the "Add search engine" entry page


delia-pop:
## Steps to reproduce
1. Go to the three-dot menu -> Settings -> Search -> Search engine.
2. Tap on "+ Add another search engine". 
Enter a name for the engine.
3. Tap on the (i) icon in the upper right corner, next to "SAVE".
4. Tap the back arrow from the "Add search engine" SUMO page.

### Expected behavior
The user is returned to the "Add search engine" entry page and the progress is saved.

### Actual behavior
Focus returns incorrectly to the home screen and the browsing history is erased, losing the progress. 

Regression after issue https://github.com/mozilla-mobile/focus-android/issues/3297

### Device information
* Android device: Huawei P9 Lite (Android 8)
* Focus version: RC 92.1.0 (Build 352442046 ; 92.0-20210901142158)

https://user-images.githubusercontent.com/89388888/131824543-09b9e525-dcfd-4893-b6af-0296c6c5c587.mp4

delia-pop:
This issue was reproduced again on 93.0.2 build (352522329 GV 93.0a1-20210906031657) using Huawei P9 Lite (Android 8).

ionutbedregeanu:
At this moment we don't have a navigation back stack, to keep a history. Focus navigation is based on AppState

To solve this we are having multiple options:
- to add a tag in AppStore which will be used to identify the previous screen and decide the next destination
- implement our own back stack navigation
- implement a navigation graph 

@pocmo What do you think? Can you help with a suggestion?

pocmo:
tbh. I think we should update the test case and not implement anything here. The help page used to be displayed in a separate activity at some point (like some other things from settings), but nowadays they open regular tabs and are treated like normal tabs. I don't think it's worth adding a bunch of complex code to handle this edge case.

pocmo:
The only (low effort) solution I can think of, is maybe trying to launch a custom tab instead of opening a regular tab. 🤔 That'd somewhat keep the navigation hierarchy.

PR #5550, opened 2021-10-07
Open add search engine info as a custom tab

ionutbedregeanu:
For #5316

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -318,7 +318,7 @@ android.applicationVariants.all { variant ->
     println("Build type:        " + variant.buildType.name)
     println("Flavor:            " + variant.flavorName)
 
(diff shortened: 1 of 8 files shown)


ionutbedregeanu:
> Neat solution! Only have a nit that we should make this into a more re-usable method. It would help with making the about pages (Your rights/Privacy notice) into Custom Tabs like this when 🚫 📑 lands.

Please take a look again and confirm if it's better now

mcarare:
> Neat solution! Only have a nit that we should make this into a more re-usable method. It would help with making the about pages (Your rights/Privacy notice) into Custom Tabs like this when 🚫 📑 lands.

There might be some updates on those pages, maybe we can open them on a separate screen like the About screen. We could then create them in Compose.

Issue #5549, opened 2021-10-07
Add to home screen dialog is not completely readable on light theme


mcarare:
@brampitoyo Do we already have specs for the add-to-home screen dialog (for that purpose, should we have a style for all dialogs? ) on dark/ light theme.

This also presents the opportunity to change the background of the icons we add to the home screen. The current one is #4C0074.

brampitoyo:
@mcarare We do.

See this Figma: https://www.figma.com/file/2V6wgf9fRFdYfLaf5c3oPj/Product-UI?node-id=2133%3A282824

Under `Specs` → `Android` → Image 5 & 6 from the left

<img width="360" alt="Android - Browser - Add to home screen@2x-1" src="https://user-images.githubusercontent.com/51007/136497724-bc52cb26-fe9f-4214-8cd2-0273fa700147.png"> <img width="360" alt="Android - Browser - Add to home screen@2x" src="https://user-images.githubusercontent.com/51007/136497729-ffcbba44-0561-40e6-ada2-288031e3d478.png">

Let’s make sure that the background colour of this “Add to home screen” dialog uses the same colour value as the background colour we use in other menus.

As for the background of the icon, we should use the same background colour of the Shortcut Tile on Dark Theme: `Ink 60``#271948`

I’m guessing that we can’t have different backgrounds based on theme, so I’ve picked the Dark Theme variant.

PR #5560, opened 2021-10-08
For #5549: Update add to home screen dialog UI

mcarare:


mcarare:
@brampitoyo That color is not something we set from Focus, but, just like Fenix, that is a color chosen in AC depending on the website snippet to match as best it can the website predominant color. 
 The colors we choose from are:

    <array name="mozac_browser_icons_photon_palette">
        <item>#9A4C00</item>
        <item>#AB008D</item>
        <item>#4C009C</item>
        <item>#002E9C</item>
        <item>#009EC2</item>
        <item>#009D02</item>
        <item>#51AB00</item>
        <item>#36385A</item>
    </array>

IMO this feature is important to be kept as such, as it helps differentiate between different launcher shortcuts that we add for multiple sites, when the site icon is not accessible. We can change the colors in that list as they do not seem to actually be photon colors (maybe they are old values).

PR #5606, opened 2021-10-13
Use incrementing version numbers for beta builds

jonalmeida:


jonalmeida:
Skipping CI to test with a release.

Issue #5601, opened 2021-10-13
Implement a Nimbus Settings Screen in Focus Android


channingcarter:
The minimal requirement for settings to ship Nimbus (and start doing experiments) is to allow the user to globally opt-out of all experiments.

(follow designs based on Fenix https://www.figma.com/file/bZyKtDQECrnkiW6uhGxlwG/Nimbus?node-id=104%3A17984 ) 

This issue asks for a Nimbus SUMO page at the location:

https://support.mozilla.org/1/mobile/\(AppInfo.appVersion)/iOS/\(languageIdentifier)/studies

Notes for QA:

- Test if Studies state is updated properly 
- Tapping on Learn more text a SUMO page is open
- At this moment the list of studies is not loaded

ionutbedregeanu:
> @brampitoyo Where should we place "Data collection" option?

https://github.com/mozilla-mobile/focus-android/pull/5661#issuecomment-946647470

lobontiumira:
Verified on the latest debug build from main from 11/5 with OnePlus 5T (Android 10):
* the Studies option is displayed in Data Choices section, below "Send usage data",
* the Studies are set to OFF by default,
* tapping on the toggle triggers the "Studies The application will quit to apply changes" prompt, with the CANCEL, and OK options,
* enabling/disabling the toggle, closes the app,
* tapping on the "Learn more" link, opens the https://support.mozilla.org/1/mobile/95.0/Android/en-US/studies in Portuguese?! - please see the last screenshot:

![studies](https://user-images.githubusercontent.com/40694158/140520790-ebafe45b-0d91-4650-bfee-a7410fe0de45.png)

@ionutbedregeanu please review and advise regarding the SUMO opened page.
Thanks!

ionutbedregeanu:
@amedyne @channingcarter Seems that https://support.mozilla.org/1/mobile/\(AppInfo.appVersion)/Android/\(languageIdentifier)/studies doesn't exist. 
We are using the correct url? 
I noticed that in Fenix  is used https://support.mozilla.org/en-US/kb/how-opt-out-studies-firefox-android which is working.

Amejia481:
@ionutbedregeanu  we recently got a patch on Fenix for fixing the link issue, you can take a look at https://github.com/mozilla-mobile/fenix/commit/1809bea49d8d1de61107de8f2273378b9ff779d1

lobontiumira:
[...]

PR #5661, opened 2021-10-19
Implement Nimbus settings screen

ionutbedregeanu:
For #5601
Add a new screen for studies where we can:
- Open Learn more page
- Enable/Disable studies, just UI implementation

Marked as Draft until I receive an approve from UX about Studies option position


ionutbedregeanu:
@brampitoyo At this moment Studies option is placed under Data Choices. Let me know if it's ok as it is or it should placed somewhere else

Issue #5671, opened 2021-10-20
The newly added search engine's icon (magnifying glass) is barely visible


lobontiumira:
## Steps to reproduce
1. Go to Settings - Search - Search engine - + Add another search engine.
2. Add a valid search engine and save it.
3. Observe the icon of the newly added search engine.

### Expected behavior
The magnifying glass icon is visible in both light and dark theme.

### Actual behavior
The magnifying glass icon is barely visible.

### Device information
* Android device: all
* Focus version: latest debug build from main from 10/20, Beta 94.0.0-beta.2

![magnifying glass](https://user-images.githubusercontent.com/40694158/138111363-cc3dd2af-33f8-44ef-a074-fd6ddda5cc0e.png)

PR #5672, opened 2021-10-20
For #5671: Replace engine static icon with engine specific icon.

mcarare:
The new icon is generated using the search string of the engine.

cc @brampitoyo: Instead of using a default icon, I thought we could do the same as we do in Fenix and try to obtain the real engine icon.


brampitoyo:
@mcarare Yes. I agree with your suggestion. Let’s use the same approach as Fenix’s!

Issue #5714, opened 2021-10-27
Menu options have wrong style.


mcarare:
The menus from the toolbar have the wrong background and text color.

Example from the autocomplete menu:
<img src = https://user-images.githubusercontent.com/48995920/139056041-24b4939c-4adc-4840-8d36-dc4a32b4ddff.png  width = 45%>

LaurentiuApahideanSV:
I tested the issue on the latest debug build from main 2021-11-05 and checked the menu options from the following sections: Exceptions, Chose search engine and Manage sites.
![Menu options](https://user-images.githubusercontent.com/50983519/140539481-f3da86b0-afdd-423a-bcfe-3b77a244812a.png)
### Devices used:
- OnePlus 6T (Android 9)
- Oppo Find X3 (Android 11)

lobontiumira:
Verified as fixed also on Beta 96.0.0-beta.1 with Sony Xperia Z5 Premium (Android 7.1.1), and Lenovo Tab M10 (Android 10).

PR #5715, opened 2021-10-27
For #5714: Update style for toolbar menu items.

mcarare:
Before / After


brampitoyo (approved):
@mcarare r+ for visual design.

PR #5754, opened 2021-11-02
No issue: Remove Firebase shippable secrets

jonalmeida:
In Fenix, we use Firebase Cloud Messaging for WebPush and Firebase Testlab for UI testing. In Focus, we only use Firebase for testing, so we do not need to include the push keys as a shippable secret.


jonalmeida:
Slack tells me sclements is on PTO, so added aki too.

escapewindow (approved):

sarah-clements (commented):
lgtm, just keep in mind you'll need to get using the [`secrets`](https://github.com/mozilla-mobile/focus-android/blob/main/taskcluster/ci/ui-test/kind.yml#L47) block in the ui-test kind.

PR #5759, opened 2021-11-02
Update to first Android Components 96.0 Nightly build

pocmo:


Amejia481 (approved):
I don't have access to see the UI test failure, but it's very likely to be related to the AC changes to the tracking protection exception

pocmo:
That patch is not in it yet. Triggered a new Nightly with it. Will update this PR here.

pocmo:
Updated to newest 96.0 Nightly.

jonalmeida:
> I don't have access to see the UI test failure, but it's very likely to be related to the AC changes to the tracking protection exception

Those failures are being fixed in https://github.com/mozilla-mobile/focus-android/pull/5756.

Issue #5261, opened 2021-08-30
Intermittent UI test failure - <CustomTabTest.openCustomTabInFocusTest>


AndiAJ:
### Stacktrace:
androidx.test.espresso.PerformException: Error performing 'single click - At Coordinates: 709, 735 and precision: 16, 16' on view 'with text: is "Open in Firefox Focus"'.
at androidx.test.espresso.PerformException$Builder.build(PerformException.java:86)
at androidx.test.espresso.base.DefaultFailureHandler.getUserFriendlyError(DefaultFailureHandler.java:87)
at androidx.test.espresso.base.DefaultFailureHandler.handle(DefaultFailureHandler.java:59)
at androidx.test.espresso.ViewInteraction.waitForAndHandleInteractionResults(ViewInteraction.java:322)
at androidx.test.espresso.ViewInteraction.desugaredPerform(ViewInteraction.java:178)
at androidx.test.espresso.ViewInteraction.perform(ViewInteraction.java:119)
at org.mozilla.focus.activity.robots.CustomTabRobot.clickOpenInFocusButton(CustomTabRobot.kt:57)
at org.mozilla.focus.activity.CustomTabTest$openCustomTabInFocusTest$4.invoke(CustomTabTest.kt:106)
at org.mozilla.focus.activity.CustomTabTest$openCustomTabInFocusTest$4.invoke(CustomTabTest.kt:102)
at org.mozilla.focus.activity.robots.CustomTabRobotKt.customTab(CustomTabRobot.kt:82)
at org.mozilla.focus.activity.CustomTabTest.openCustomTabInFocusTest(CustomTabTest.kt:102)
at java.lang.reflect.Method.invoke(Native Method)
at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
at androidx.test.internal.runner.junit4.statement.RunBefores.evaluate(RunBefores.java:80)
at androidx.test.internal.runner.junit4.statement.RunAfters.evaluate(RunAfters.java:61)
at androidx.test.rule.ActivityTestRule$ActivityStatement.evaluate(ActivityTestRule.java:531)
[...]

PR #5771, opened 2021-11-03
For #5261 - Avoid FLAG_ACTIVITY_NEW_TASK for the custom tab intent

Mugurell:
Debugging the normal behavior shows that custom tabs launched from other apps
will have no flags set for the Intent while the previous
ActivityTestRule#launchActivity will set FLAG_ACTIVITY_NEW_TASK flag on the
intent which after the recent changes will break the custom tabs functionality.

Replacing the now deprecated ActivityTestRule with ActivityScenario for
starting a new custom tabs intent will have no flags set on the intent and so
ensure the intended behavior without having to hardcode removing the NEW_TASK
flag in app code.


mcarare (approved):

ionutbedregeanu (approved):

sv-ohorvath (approved):
LGTM. Thank you!

jonalmeida:
Thank you all for figuring this out!

Issue #5795, opened 2021-11-05
Replace "multiple tabs" with Nimbus experiment


pocmo:
Feature Flag: https://github.com/mozilla-mobile/focus-android/issues/5572

This feature should be controlled by a Nimbus experiment instead of a static feature flag.

pocmo:
The code is in. The actual experiment will run at a later point.

PR #5800, opened 2021-11-05
Issue #5795: Replace "multiple tabs" feature flag with Nimbus experiment

pocmo:
@travis79 @jhugman Can you take a look if the experiment integration is correct?


app/src/main/java/org/mozilla/focus/experiments/ExperimentalFeatures.kt
```diff
@@ -0,0 +1,33 @@
...
+private const val BRANCH_TREATMENT = "treatment"
+
+/**
+ * Feature flags driven by Nimbus experiments.
+ */
+class ExperimentalFeatures(nimbus: NimbusApi) {
+    /**
+     * Whether or not this browser should have multiple tabs or run in single-tab mode.
+     */
+    val hasMultipleTabs: Boolean by lazy {
```
pocmo, line 19: I am curious about the impact of using `by lazy` here: I want to check only once per app launch for this experiment. Because it would be confusing for the user if they have multiple tabs and then we switch to single-tab mode. So while the app is running this should not change.
But I guess the experiment configuration in Nimbus could change while the app is running. Would this cause us to report the wrong data (e.g. "User is in treatment" but actually this feature is not enabled because they were not when we checked?)

travis79: The way Nimbus is meant to be used, the experiment treatment should _never_ change during the execution of the app. The only API function that can cause this to happen is `Nimbus.applyPendingExperiments` and only if there are experiments that need applied. This function should be set up to only be called when the app is launched so that Nimbus will only return the same value for a feature configuration during that execution. Nimbus even caches the value in memory so that retrieving it shouldn't trigger a read from storage.

pocmo: Great, thank you. That sounds like exactly the behavior I need. And since we only call `applyPendingExperiments` when constructing `NimbusApi` this seems to be safe: https://github.com/mozilla-mobile/focus-android/blob/5a990f5d90110be0d24710e216e82139a87430a2/app/src/main/java/org/mozilla/focus/experiments/NimbusSetup.kt#L81

app/src/main/java/org/mozilla/focus/experiments/ExperimentalFeatures.kt
```diff
@@ -0,0 +1,33 @@
...
+        // We are wrapping this in "by lazy" because we do not want to switch between a
+        // multi-tab/single-tab browser at runtime while the user is using the browser. Instead we
+        // decide once on startup.
+
+        @Suppress("TooGenericExceptionCaught")
+        val branch = try {
+            nimbus.getExperimentBranch(EXPERIMENT_SINGLE_TAB_BROWSER)
```
travis79, line 26: I would strongly suggest that you don't use this since it ties the application to experiment internals. We have deprectated the use of the `getExperimentBranch` function in favor of using the Feature API: https://experimenter.info/feature-variables-and-me

travis79: Here's an example of the Feature API instrumentation in Fenix for a reference: https://github.com/mozilla-mobile/fenix/blob/3f9d277566cceed8449c63b19939afba4c8ea104/app/src/main/java/org/mozilla/fenix/experiments/NimbusFeatures.kt

travis79 (changes requested):
In order to keep a disconnect between how the experiment id and branch ids are generated from forcing changes to the experiment instrumentation to ride the trains if an experiment needs changed in Experimenter, we should rely on the Nimbus Feature API.  cc @jhugman in case he has anything to add.

pocmo:
@travis79 Thank you for the review! I updated the PR to use the feature API instead. Can you take another look to confirm this follows best practices now?

app/src/main/java/org/mozilla/focus/experiments/ExperimentalTabsFeatures.kt
```diff
@@ -0,0 +1,30 @@
...
+ * Experimental features regarding the usage of "browser tabs".
+ */
+class ExperimentalTabsFeatures(
+    variables: Variables
+) {
+    /**
+     * Do we support multiple tabs or is this a single tab browser?
+     */
+    val isMultiTab: Boolean by lazy {
+        when (variables.getString(VARIABLE_TAB_MODE)) {
```
travis79, line 24: The feature API supports boolean types, if you wanted to you could simplify this a bit with something like `variables.getBool(IS_MULTI_TAB)`

travis79 (approved):
Just one suggestion for a possible way to simplify this a bit, but other than that this is looking good!

app/src/main/java/org/mozilla/focus/experiments/ExperimentalTabsFeatures.kt
pocmo: Thanks! I updated the patch to use a boolean instead.

jhugman (approved):
This looks nice and clean.

Amejia481:
@jhugman @travis79 should `recordExposureEvent` be called at some point? I remembered we used it on the [Fenix home screen experiments](https://github.com/mozilla-mobile/fenix/pull/21411/)

pocmo:
> @jhugman @travis79 should `recordExposureEvent` be called at some point? I remembered we used it on the [Fenix home screen experiments](https://github.com/mozilla-mobile/fenix/pull/21411/files#diff-41731714e97083a1b73584da2837ab11c22796c25b892b1e86ce4190ebba2633R374)

I think that's not needed here. `getVariables()` has a `recordExposureEvent` that is `true` by default. In Fenix we explicitly pass in `false`, which we do not here.

travis79:
If you rely on the "automatic" exposure events (the default `true` state), be aware that every time you request the variables it will trigger an exposure event. If this is when the user is exposed and it only happens once per exposure, then this is fine. If, on the other hand, you request the config from Nimbus multiple times but the user is only "exposed" to the values once, then you should pass in `false` and trigger the `recordExposureEvent` when the user is presented with the feature.

jonalmeida (approved):
This looks good to me and should work now with our Nimbus integration. The discussion above was also insightful too.

✅

pocmo:
> [task 2021-11-22T14:24:10.489Z] org.mozilla.focus.searchsuggestions.SearchSuggestionsViewModelTest > alwaysSearchSelected FAILED
[task 2021-11-22T14:24:10.489Z]     java.lang.IllegalStateException at SearchSuggestionsViewModelTest.kt:35

That looks unrelated. I'll restart.

PR #5815, opened 2021-11-08
Update sumo url builder

ionutbedregeanu:
For #5601

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/SupportUtils.kt
+++ b/app/src/main/java/org/mozilla/focus/utils/SupportUtils.kt
@@ -50,7 +50,13 @@ object SupportUtils {
         WHATS_NEW_KLAR("whats-new-firefox-klar-android"),
         SEARCH_SUGGESTIONS("search-suggestions-focus-android"),
         ALLOWLIST("focus-android-allowlist"),
-        STUDIES("studies")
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/utils/SupportUtils.kt
```diff
@@ -50,15 +50,13 @@ object SupportUtils {
...
     }
 
-    fun getSumoURLForTopic(context: Context, topic: SumoTopic): String {
+    fun getSumoURLForTopic(topic: SumoTopic): String {
         val escapedTopic = getEncodedTopicUTF8(topic.topicStr)
-        val appVersion = getAppVersion(context)
-        val osTarget = "Android"
         val langTag = Locales.getLanguageTag(Locale.getDefault())
-        return "https://support.mozilla.org/1/mobile/$appVersion/$osTarget/$langTag/$escapedTopic"
+        return "https://support.mozilla.org/$langTag/kb/$escapedTopic"
```
jonalmeida, line 59: For the "What's new" link, we need to know the version, so I don't think we should remove it here.

Maybe create a separate function for the Studies link? I think Fenix did that too.

jonalmeida (approved):

jonalmeida:
Thanks!

Issue #5738, opened 2021-10-29
The Exceptions button does not grey out after clearing all exceptions


LaurentiuApahideanSV:
## Steps to reproduce
1. Open a few websites and disable ETP in order to add them to the exceptions list;
2. Got to Settings -> Privacy & Security -> Exceptions;
3. Tap the Remove all websites button.

### Expected behavior
After removing all websites from the exceptions list the Exceptions button should become grey.
### Actual behavior
After removing all websites from the exceptions list the Exceptions button can be tapped once more resulting in a flicker before becoming grey.
### Device information

* Android devices: Oppo Find X3 (Android 11), OnePlus 6T (Android 9), Huawei P9 Lite (Android 8)
* Focus version: RC 94.1.1

https://user-images.githubusercontent.com/50983519/139405890-ca9eeb5e-3280-45cf-a930-d601b5a7e5a7.mp4

Amejia481:
We are planing it, to land it after 96 cycle starts

Amejia481:
This landed on https://github.com/mozilla-mobile/focus-android/commit/96812489c3687e97a0bfcf1016f9555b7cf4f252 we can re-test now.

lobontiumira:
Reproducible on Beta 95.0.0-beta.3, and on the latest debug build from main, with Samsung Galaxy Tab A6 (Android 5.1.1).

ionutbedregeanu:
@jeffreygee As Arturo suggested [here](https://github.com/mozilla-mobile/focus-android/pull/5827#pullrequestreview-802748007) I added a callback to know when the removing process ends and the navigation should be done to  Privacy & Security screen
But I'm not sure if a spinner should appears while the exceptions are removed because everything is almost instant and the spinner it will be like a flicker effect.

Should we give up on this ideea or should we add a minimum duration for that spinner?
What do you think?

Also if we decide to add a spinner we need design for that, at this moment we don't have a loader spinner in the app.

jeffreygee:
> @jeffreygee As Arturo suggested [here](https://github.com/mozilla-mobile/focus-android/pull/5827#pullrequestreview-802748007) I added a callback to know when the removing process ends and the navigation should be done to Privacy & Security screen But I'm not sure if a spinner should appears while the exceptions are removed because everything is almost instant and the spinner it will be like a flicker effect.
> 
> Should we give up on this ideea or should we add a minimum duration for that spinner? What do you think?
> 
[...]

PR #5827, opened 2021-11-09
Add callback to finish exceptions removal before navigation

ionutbedregeanu:
For #5738 
Navigate to the previous screen only after all the exceptions were removed

### DIFF
--- a/app/src/main/java/org/mozilla/focus/exceptions/ExceptionsListFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/exceptions/ExceptionsListFragment.kt
@@ -40,6 +40,7 @@ import org.mozilla.focus.utils.ViewUtils
 import java.util.Collections
 import kotlin.coroutines.CoroutineContext
 
+private const val REMOVE_EXCEPTIONS_DISABLED_ALPHA = 0.5f
 typealias DomainFormatter = (String) -> String
 
 /**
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/mozilla/focus/exceptions/ExceptionsListFragment.kt
```diff
@@ -121,16 +123,18 @@ open class ExceptionsListFragment : BaseSettingsLikeFragment(), CoroutineScope {
...
 
-            val exceptions = (exceptionList.adapter as DomainListAdapter).selection()
-            TelemetryWrapper.removeAllExceptionDomains(exceptions.count())
-
-            requireComponents.appStore.dispatch(
-                AppAction.NavigateUp(
-                    requireComponents.store.state.selectedTabId
+                val exceptions = (exceptionList.adapter as DomainListAdapter).selection()
+                TelemetryWrapper.removeAllExceptionDomains(exceptions.count())
+                delay(REMOVE_EXCEPTIONS_DELAY)
```
ionutbedregeanu, line 131: I have mixed feelings about this but I didn't find a better solution without doing changes in AC(maybe that's the only way). I'm open to suggestions if you have 🙏

Amejia481: No problem, I'm taking a look to what alternatives we have. I'll report back :)

Amejia481 (changes requested):
I think this approach is bit fragile as we are not completely sure how much time it will take to remove all exceptions in different devices, in some it could take `100 MS` in others it could take more, as a result we will see the same issue on some devices. 

I believe the real problem, it's that the app doesn't have a way to tell when the exceptions have been completely removed. We can expose a callback on AC for the app to tell, similar to [what we are doing for other user cases of tracking protection](https://github.com/mozilla-mobile/android-components/blob/b37f3e46871c201252fdc7107418cf063f1cc20b/components/feature/session/src/main/java/mozilla/components/feature/session/TrackingProtectionUseCases.kt#L145), then we can add loading spinner after the remove all button is clicked and we can show it until the callback tells us the all exceptions have been removed .

ionutbedregeanu:
> I think this approach is bit fragile as we are not completely sure how much time it will take to remove all exceptions in different devices, in some it could take `100 MS` in others it could take more, as a result we will see the same issue on some devices.
> 
> I believe the real problem, it's that the app doesn't have a way to tell when the exceptions have been completely removed. We can expose a callback on AC for the app to tell, similar to [what we are doing for other user cases of tracking protection](https://github.com/mozilla-mobile/android-components/blob/b37f3e46871c201252fdc7107418cf063f1cc20b/components/feature/session/src/main/java/mozilla/components/feature/session/TrackingProtectionUseCases.kt#L145), then we can add loading spinner after the remove all button is clicked and we can show it until the callback tells us the all exceptions have been removed .

I completely agree with you. Thanks, @Amejia481

app/src/main/res/layout/fragment_exceptions_domains.xml
```diff
@@ -38,8 +39,9 @@
         <androidx.recyclerview.widget.RecyclerView
             android:id="@+id/exceptionList"
             android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            android:orientation="vertical" />
+            android:layout_height="0dp"
+            android:orientation="vertical"
+            android:layout_weight="0.9"/>
```
Amejia481, line 44: Is there any specific reason why we are changing the `RecyclerView` layout configurations ? :)

ionutbedregeanu: If we are having many exceptions "remove exceptions" buttons is pushed out of the screen

Amejia481: Thanks for the clarification 👍🏽

Amejia481 (approved):

Issue #5829, opened 2021-11-09
Implement new design for Trash Icon and Tabs


amedyne:
This ticket is to implement the design changes for the Trash Icon location and the number of tabs indicator.

Figma designs here: https://www.figma.com/file/8Xyk89M3lk8n2c5aP79AHh/Delete-Action?node-id=2522%3A7249

This should help address and answer the following: 
https://github.com/mozilla-mobile/focus-android/issues/5774
https://github.com/mozilla-mobile/focus-android/issues/5757
https://github.com/mozilla-mobile/focus-android/issues/5707
https://github.com/mozilla-mobile/focus-android/issues/5706
https://github.com/mozilla-mobile/focus-android/issues/5704
https://github.com/mozilla-mobile/focus-android/issues/5466

lobontiumira:
Verified as fixed on the latest Focus Nightly build 95.0 (353271124 with GV 96.0a1-20211122093912), withthe following devices:
* Samsung Galaxy Tab A6 (Android 5.1.1),
* Samsung Galaxy Note 8 (Android 9),
* Google Pixel (Android 10), and
* Sony Xperia Z5 Premium (Android 7.1.1).

The trashcan, and the tabs counter are separated, both moved to the left/right of the search field:
![5829](https://user-images.githubusercontent.com/40694158/143006329-3bceb603-0d61-4b75-8502-d5d93df8e477.jpg)


I'll start verifying the above mentioned issues as well.

lobontiumira:
Verified as implemented successfully on Beta 96.0.0-beta.1 as well.
Tested with:
* Sony Xperia Z5 Premium (Android 7.1.1),
* Google Pixel (Android 10),
* Samsung Galaxy Tab a6 (Android 5.1.1).

PR #5852, opened 2021-11-11
Relocate erase and tabs counter to toolbar

ionutbedregeanu:
For #5829
Remove erase and tabs floating buttons and their custom behavior
Add their equivalent buttons in toolbar using the action buttons from AC
Fix unit tests in order to pass after the changes mentioned above
Remove erase and tabs counter buttons from UI tests and mark with Ignore those tests which are failing because of that

#### Normal tabs

https://user-images.githubusercontent.com/11731652/141286946-1aaa1056-bc93-4349-b033-7184b9126b3a.mp4


ionutbedregeanu:
@sv-ohorvath Erase and tabs counter floating buttons were removed, as a result many UI tests are failing and require some refactoring.
I marked those methods and classes with Ignore, but please take a look on this pr when you have time and confirm if it's ok. Maybe some more changes needs to be done

sv-ohorvath:
Thanks, @ionutbedregeanu, I logged an issue for me to take care of these https://github.com/mozilla-mobile/focus-android/issues/5854.

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -153,9 +162,44 @@ class BrowserToolbarIntegration(
...
+                iconTintColorResource = R.color.primaryText,
+                listener = {
+                    val openedTabs = store.state.tabs.size
+                    TabCount.eraseButtonTapped.record(TabCount.EraseButtonTappedExtra(openedTabs))
+                    eraseActionListener.invoke()
+                }
+            )
+            toolbar.addNavigationAction(eraseAction)
+
+            store.flowScoped { flow ->
```
pocmo, line 182: This method returns a `CoroutineScope`. We should make sure to keep a reference to it and cancel it (at the latest from `stop()`).

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -153,9 +162,44 @@ class BrowserToolbarIntegration(
...
+                flow.ifChanged { state -> state.tabs.size }
+                    .collect {
+                        if (store.state.tabs.size > 1) {
```
pocmo, line 185: nit: If we are only interested in this check then we could make it part of the "changed" check above. Something like:

```Kotlin
flow.map { state -> state.tabs.size > 1 }
    .ifChanged()
    .collect { hasTabs ->
     // ...
     }

File: app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt [hunk 3]
```diff
@@ -105,6 +112,8 @@ class BrowserToolbarIntegration(
             fragment.showTrackingProtectionPanel()
         }
 
+        setActionButtons(context)
```
pocmo, line 115: nit: We could move the "is custom tab" check out of the method to here. `setUrlBackground()` needs it too. So we could check once here and pass it down.

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -153,9 +162,44 @@ class BrowserToolbarIntegration(
...
+                        if (store.state.tabs.size > 1) {
+                            val tabsAction = TabCounterToolbarButton(
+                                lifecycleOwner = fragment,
+                                showTabs = {
+                                    toolbar.hideKeyboard()
+                                    tabCounterListener.invoke()
+                                },
+                                store = store
+                            )
+                            toolbar.addBrowserAction(tabsAction)
```
pocmo, line 194: We need to prevent it from adding this button multiple times :D

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -153,9 +162,44 @@ class BrowserToolbarIntegration(
...
+                        }
```
pocmo, line 195: I also wonder if we need to remove the button too. In the past when navigating back in a tab ... if we couldn't navigate back anymore then we closed that tab. But that doesn't seem to be the case anymore, now the whole session gets removed. That may be a bug? 🤔

sv-ohorvath:
@ionutbedregeanu I just gave this PR a better look and saw that the trash and tabs icons aren't really removed, just have different selectors. Could you please revert what was deleted (trashButtonTest and some steps involving the erase button)? Just tag all of them as Ignored and I'll take care of updating their selectors. Thanks!

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
ionutbedregeanu: yep, seems a regression added by me in this pr. Thanks

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -153,9 +159,44 @@ class BrowserToolbarIntegration(
...
+                        },
+                        store = store
+                    )
+                    if (state.tabs.size > 1) {
+                        toolbar.addBrowserAction(tabsAction)
+                    } else {
+                        toolbar.removeBrowserAction(tabsAction)
+                    }
+                }
+        }
```
pocmo, line 195: There's a mismatch now. We subscribe in `init()`, but we unsubscribe in `stop()`. The start and stop methods are getting called with the lifecycle. So if you open a tab, minimize the app, resume the app and then open another tab, then we will have unsubscribed and will not add the button.

We could observe/cancel from start/stop, like we do for the security indicator changes. But then we also need to make sure we do not add the button multiple times. :)

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -193,6 +234,7 @@ class BrowserToolbarIntegration(
     @VisibleForTesting
     internal fun stopObserverSecurityIndicatorChanges() {
         securityIndicatorScope?.cancel()
+        tabsCounterScope?.cancel()
```
pocmo, line 237: nit: This should probably happen in `stop()` directly and not in `stopObserverSecurityIndicatorChanges()` which is unrelated.

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
ionutbedregeanu: I tried to reproduce this scenario but I didn't succeed. Do you think that you can help me with a video?

pocmo: https://user-images.githubusercontent.com/89638/141994422-310af25b-595c-43d1-8365-a7c21644405e.mp4

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -153,13 +172,32 @@ class BrowserToolbarIntegration(
         toolbar.display.setUrlBackground(urlBackground)
     }
 
+    private fun setActionButtons() {
+        toolbar.addNavigationAction(eraseAction)
```
pocmo, line 176: Since this action will not get added/removed dynamically, I think we could keep this one in `init()`? 🤔 (Which also avoids us having to remove it in stop() again).

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -188,6 +226,9 @@ class BrowserToolbarIntegration(
         customTabsFeature?.stop()
         navigationButtonsIntegration?.stop()
         stopObserverSecurityIndicatorChanges()
+        toolbar.removeBrowserAction(tabsAction)
```
pocmo, line 229: I wonder why we have to remove the tab counter here? Shouldn't the observer handle all cases already? 🤔

ionutbedregeanu: Is still added every time that's why I chose to remove it at stop()

pocmo (approved):

PR #5890, opened 2021-11-16
Close #5879: Add proguard rules for native components

### LINKED ISSUE #5879: Nimbus: viaduct::backend::ffi: Missing HTTP status
After https://github.com/mozilla-mobile/focus-android/issues/5781 landed, we're new viaduct error that ends up giving up the same stacktrace:

```
2021-11-15 15:31:59.708 26250-26284/? E/viaduct::backend::ffi: Viaduct Ffi Error: 
doFetch failed: Field method_ for mozilla.appservices.httpconfig.MsgTypes$Request not found. 

Known fields are [
  public com.google.protobuf.ByteString mozilla.appservices.httpconfig.MsgTypes$Request.body_, 
  public com.google.protobuf.MapFieldLite mozilla.appservices.httpconfig.MsgTypes$Request.headers_, 
  public byte mozilla.appservices.httpconfig.MsgTypes$Request.memoizedIsInitialized, 
  public java.lang.String mozilla.appservices.httpconfig.MsgTypes$Request.url_, 
  public static final mozilla.appservices.httpconfig.MsgTypes$Request mozilla.appservices.httpconfig.MsgTypes$Request.DEFAULT_INSTANCE, 
  public static volatile com.google.protobuf.Parser mozilla.appservices.httpconfig.MsgTypes$Request.PARSER
]
```

jonalmeida:
In Fenix, [we were doing this already](https://github.com/mozilla-mobile/fenix/blob/17865bdf66ccca9fc89e34db532a5e2b75315e85/app/proguard-rules.pro#L60-L64), probably because it was necessary from other components. For Focus, we're seeing the need for it when integrating Nimbus.

Verified this is a bug by building the beta variant and a debug one from `main`. On debug, we were successfully receiving the experiments. Disabling minification on the beta build also enabled experiments to be fetched, which further proved that it was related to proguarding.

Re-enabling minification and adding the proguard rules from Fenix also allowed the experiments to be fetched. 🙂


pocmo (approved):
👍 👍

jonalmeida:
@Mergifyio backport releases_v95.0

app/proguard-rules.pro
```diff
@@ -104,6 +104,12 @@
     <init>(android.app.Application);
 }
 
+####################################################################################################
+# Mozilla Application Services
+####################################################################################################
+
+-keep class mozilla.appservices.** { *; }
+
```
jhugman, line 112: I wonder if 

```
-keepclassmembers class * extends com.sun.jna.** {
    <fields>;
    <methods>;
}
```

would do the maximal thing you want— though **only** for APK size.

jhugman (approved):
I've added another suggestion, but I'm unsure if you should take it :)

app/proguard-rules.pro
jonalmeida: I think we would get this for free via https://github.com/mozilla/application-services/blob/main/proguard-rules-consumer-jna.pro .

Is the idea to add this for all fields/methods and not just the public ones? I'm not sure if that would keep more than we use. 🤔

jonalmeida: Thinking about it more, this rule is the same one used in Fenix and since this change is blocking the A/A experiment, I'd like to keep the rule identical to what we have there.

Happy to re-visit this again. Maybe as part of https://github.com/mozilla-mobile/focus-android/issues/4922?

Issue #5908, opened 2021-11-17
WebCompatReporterFeature is using the wrong productName


denschub:
When adding the WeCompatReporterFeature, the productName `focus` was chosen:

https://github.com/mozilla-mobile/focus-android/blob/7f8879347df74c3b56a0d18185ec582ec8b7d697/app/src/main/java/org/mozilla/focus/Components.kt#L105

Unfortunately, the webcompat.com backend does not allow that value. As the [component's documentation says](https://github.com/mozilla-mobile/android-components/tree/main/components/feature/webcompat-reporter),

> Note that simply inventing new values here does not work, as each product name has to be safelisted by the WebCompat team on webcompat.com, so [please get in touch](https://wiki.mozilla.org/Compatibility#Core_Team) when you need to add a new product name.

However, instead of adding that new productName, I'll change that to `focus-geckoview`. We have many thousand reports already filed with that productName, and we don't want to re-write all those reports just to have it say `focus`. I'll submit a PR to have that changed. :)

denschub:
Fixed by #5909

PR #5909, opened 2021-11-17
Issue #5908 - WebCompatReporterFeature - Use `focus-geckoview` as the productName.

denschub:
See #5908 - this change allows webcompat.com to add a matching `browser` label, and makes sure the reports end up in the same pile as all previous Focus reports we have in our system.


jonalmeida (approved):
Thanks for reporting and fixing the issue!

Issue #5594, opened 2021-10-13
java.lang.IllegalStateException: SET SESSION: Current activity: CustomTabActivity hashcode 5303933 Other activity: MainActivity hashcode 16940721


AndiAJ:
## Prerequisites
`openCustomTabInFocusTest` crashes with this signature on Firebase
Tracking the intermittent failures in #5283

## Steps to reproduce
1. Launch Focus
2. Skip the onboarding tour
3. Load a website
4. Launch the previously loaded website in a custom tab
5. Tap the ⋮ button and open the main menu
6. Tap the "Open in Firefox Focus" button

### Expected behavior
The tab is properly opened in Focus and can interact with it

### Actual behavior
After tapping the "Open in Firefox Focus" button Focus crashes

### Device information

* Android device: GCE x86 phone API 28
* Focus version: latest Main

### Notes
❗ AC issue [#11021](https://github.com/mozilla-mobile/android-components/issues/11021)
❗ Similar issue encountered on Fenix [#21030](https://github.com/mozilla-mobile/fenix/issues/21030)
❗ Similar issue encountered on RB [#1675](https://github.com/mozilla-mobile/reference-browser/issues/1675#issuecomment-942083574)

► Video

https://user-images.githubusercontent.com/51314259/137105160-4bef6a40-5b9c-462b-91b1-a51ff82a3b80.mp4

[Firebase link](https://console.firebase.google.com/u/0/project/moz-focus-android/testlab/histories/bh.2189b040bbce6d5a/matrices/8985550466989399501/executions/bs.296c3af575d9ea25/testcases/1/logs)

AaronMT:
@jonalmeida Checking in with this one since it pops up daily I our UI test suite. It looks like @AndiAJ has STR above, is that what we were looking for on this one?

mcarare:
Let's keep an eye on this due to the fact the new fix has landed.

PR #5911, opened 2021-11-18
For #5594: Also Stop GeckoSession when releasing acquired GeckoDisplay.

mcarare:


pocmo (approved):
That's an awesome find! 👍 

I verified that this is working locally and preventing the crash. You may end up with a not completely loaded page in Focus (which was also not loaded in the custom tab yet) if you are really fast. But that may still be better than crashing.

Since that seems to be a GeckoView bug I filed this here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1741899

Let's add a comment referencing the bug. :)

mcarare:
Added a comment referencing the Bugzilla bug.

pocmo:
Great. Let's land this to unblock other PRs.

pocmo:
@Mergifyio backport backport releases_v95.0

Issue #5884, opened 2021-11-16
Web content that is sticky to the bottom is hidden until refresh


jonalmeida:
## Steps to reproduce

1. Navigate to reddit.com
2. Scroll the toolbar away before the page loaded.
3. Scroll the toolbar away after the page is loaded.

(see examples below)

### Expected behavior

- At step 2, the toolbar is made visible and persists in that position with the full viewport visible **after the page is done loading**.
- At step 3, the the toolbar is made visible and persists in that position with the full viewport visible **after a touch event is recorded**.

Expected results are seen in Fenix:

https://user-images.githubusercontent.com/1370580/142020187-1b6753d3-d269-46be-a32e-2de50eac0434.mp4

https://user-images.githubusercontent.com/1370580/142020200-5112484d-c241-4db3-91e5-378fc8fcd88f.mp4

### Actual behavior

- At step 2 and step 3, the toolbar is still hidden, the viewport is visible, but scrolling then locks the toolbar visible and cuts out the bottom viewport.

https://user-images.githubusercontent.com/1370580/142019418-b0657e20-80d0-4a26-aeb2-f4def13b90c6.mp4 

### Device information

* Android device: All devices
* Focus version: Firefox Nightly 96

lobontiumira:
Verified as fixed on the latest Focus Nightly from 11/24 (build 353281028 GV 96.0a1-20211123094249) with Google Pixel (Android 10):


https://user-images.githubusercontent.com/40694158/143263179-c098e852-2bf1-4c09-8351-0b308f639751.mp4

lobontiumira:
Verified as fixed on the RC 95.2.0 with Lenovo tablet M10 (Android 10).

PR #5932, opened 2021-11-23
Use the correct value of toolbar height at dynamic behavior

ionutbedregeanu:
For #5884


Mugurell (approved):
A bit interesting that the toolbar height was otherwise `0`.
This seems like a good fix that could probably solve more issues.
I see below that for `showAsFixed` the dimension and not height is already used so we're good there also.

Amejia481:
@Mergifyio backport releases_v95.0

Issue #5741, opened 2021-10-29
Add option to close tabs selectively.


mcarare:
### Why/User Benefit/User Problem

Users do not have the option to close just one or more, not all opened tabs.

### What / Requirements

Add a close button to items in the session list.

### Acceptance Criteria (how do I know when I’m done?)

Users have the option to close tabs selectively.

P.S.: This request was also encountered in play store reviews.

OysterP:
Already raised under #5445

kelimuttu:
We've seen a request to have this feature from the [Google Play Store](https://play.google.com/store/apps/details?id=org.mozilla.focus&reviewId=gp:AOqpTOF8Wl5M-rLQ1uCttb8G0E79btdR0BvyDbWbYv72dJJDarqJNMWOSyrSl6Lm3HLHYVagclUnr_t3qNROtFU).

lobontiumira:
Verified as fixed on the latest Focus Nightly build from 12/8, and on Beta 96.0.0-beta.1 with multiple devices:
* Sony Xperia Z5 Premium (Android 7.1.1),
* Samsung Galaxy Tab A6 (Android 5.1.1),
* Xiaomi mi4i (Android 5.0.2),
* Google Pixel (Android 10),
* Lenovo tablet M10 (Android 10).

The tabs can be closed in any order. If the user selects to close the active tab, the next teb under it is loading and displayed.
Switching from portrait to landscape and viceversa while the tabs list is displayed, keeps the tabs list displayed, does not dismiss it.

![5741 landscape](https://user-images.githubusercontent.com/40694158/145197339-6bcbb8ba-e87b-457b-abdf-5f2e621edecd.png)

![5741 portrait](https://user-images.githubusercontent.com/40694158/145197384-60673545-b206-443a-812e-169268d7daec.png)

PR #5934, opened 2021-11-23
For #5741: Add option to close sessions selectively.

mcarare:
Refactor TabsAdapter to implement ListAdapter.
Refactor TabViewHolder.
Remove state observing and pass initial tab list:
The tab list and the selected tab do not change while fragment is displayed.

### DIFF
--- a/app/src/main/java/org/mozilla/focus/session/ui/TabsAdapter.kt
+++ b/app/src/main/java/org/mozilla/focus/session/ui/TabsAdapter.kt
@@ -6,47 +6,35 @@ package org.mozilla.focus.session.ui
 
 import android.view.LayoutInflater
 import android.view.ViewGroup
-import android.widget.TextView
 import androidx.recyclerview.widget.RecyclerView
-import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.collect
-import mozilla.components.browser.state.selector.privateTabs
-import mozilla.components.browser.state.state.BrowserState
 import mozilla.components.browser.state.state.TabSessionState
-import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifAnyChanged
+import org.mozilla.focus.databinding.ItemSessionBinding
 
 /**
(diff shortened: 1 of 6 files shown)


mcarare:
This is one of the most requested features in Playstore reviews.

Added @jeffreygee to review the UI implementation (it is based on the current design with the addition of the close button, the standard one from AC also used in Fenix)

Adding @channingcarter for the go-ahead on adding this feature.

Below are screenshots in both themes and a video demo of the feature.

 

https://user-images.githubusercontent.com/48995920/143579024-aa254140-3584-4cee-a4d6-4c9b6a29b9db.mp4

app/src/main/java/org/mozilla/focus/session/ui/TabsAdapter.kt
```diff
@@ -6,47 +6,35 @@ package org.mozilla.focus.session.ui
...
+            isCurrentSession = isCurrentSession.invoke(currentItem),
+            selectSession = selectSession,
+            closeSession = closeSession,
+        )
     }
 
-    private fun onUpdate(tabs: List<TabSessionState>) {
-        this.tabs = tabs
-        notifyDataSetChanged()
-    }
```
pocmo, line 51: There are a lot of changes, so it's not immediately clear to me: The old code seems to have updated tabs dynamically, while this new code uses a fixed list, why is that?

Ah, now I think I understand, the UI gets closed and therefore we do not need to update it, right?

pocmo (approved):
LGTM. Tested on my device too. 👍

app/src/main/java/org/mozilla/focus/session/ui/TabsAdapter.kt
mcarare: Yes, the updating never happens while the list is displayed, so there is no need to observe changes.

mcarare:
PR update: Updated UI test to reflect new view hierarchy.

Issue #5881, opened 2021-11-16
Current app shows up in share via list app 


mcarare:
Breakout from https://github.com/mozilla-mobile/focus-android/issues/5824.

## Steps to reproduce
1. Install the latest debug build from main
2. Open a page, and from the three-dot menu select the Share option

3. Observe the installed apps displayed in the dialogue.

### Expected behavior
The current app should not be in the list,

### Actual behavior
The current app is in the list.

lobontiumira:
Reproducible on Beta 96.0.0-beta.1 with Samsung Galaxy A6 (Android 5.1.1).

delia-pop:
Reproducible on RC 96.1.0 with Google Pixel 6 (Android 12).

delia-pop:
Verified as fixed on Firefox Focus Nightly 98.0 (Build 360330514 🦎 98.0a1-20220201093942) with Google Pixel 6 (Android 12).
The current app is no longer displayed in the share list. 
![Screenshot_20220202-154211](https://user-images.githubusercontent.com/89388888/152165112-181966fb-68f0-4049-b917-2341db9b4e24.png)

PR #5948, opened 2021-11-26
Remove current app from share list

ionutbedregeanu:
For #5881


app/src/main/java/org/mozilla/focus/utils/IntentUtils.kt
```diff
@@ -3,142 +3,68 @@
...
-            0 -> handleUnsupportedLink(context, intent)
-            1 -> {
-                val info = matchingActivities[0]
-                val externalAppTitle = info?.loadLabel(packageManager) ?: "(null)"
+    fun getIntentChooser(
+        context: Context,
+        intent: Intent,
+        chooserTitle: CharSequence? = null
+    ): Intent? {
+        val chooserIntent: Intent?
```
Mugurell, line 20: Am I correct in saying that this will never be null?
So we can declare this as non-nullable, maybe as a `lateinit var` and then the code would be a bit more straightforward in clients also - no need to check the Intent for null before `startActivity()`
Otherwise we might have to think about how to handle cases in which this is null <=> users tap to share but nothing happens.

app/src/main/java/org/mozilla/focus/utils/IntentUtils.kt
```diff
@@ -3,142 +3,68 @@
...
-        // see what domains we're visiting, so there's no loss of privacy here:
-        val packageManager = context.packageManager
-        val matchingActivities = packageManager.queryIntentActivities(intent, 0)
-
-        when (matchingActivities.size) {
-            0 -> handleUnsupportedLink(context, intent)
-            1 -> {
-                val info = matchingActivities[0]
-                val externalAppTitle = info?.loadLabel(packageManager) ?: "(null)"
+    fun getIntentChooser(
```
Mugurell, line 15: This needs to be documented.

app/src/main/java/org/mozilla/focus/utils/IntentUtils.kt
```diff
@@ -3,142 +3,68 @@
...
+        val resolveInfos = context.packageManager.queryIntentActivities(intent, 0).toHashSet()
+
+        val excludedComponentNames = HashSet<ComponentName>()
+        resolveInfos.forEach {
+            val activityInfo = it.activityInfo
+            if (activityInfo.packageName == context.packageName) {
+                val componentName = ComponentName(activityInfo.packageName, activityInfo.name)
+                excludedComponentNames.add(componentName)
+            }
+        }
```
Mugurell, line 30: nit: This can be replaced with just `resolveInfos.filter { .. }`

app/src/main/java/org/mozilla/focus/utils/IntentUtils.kt
```diff
@@ -3,142 +3,68 @@
...
-                    externalAppTitle
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            chooserIntent = Intent.createChooser(intent, chooserTitle)
+                .putExtra(Intent.EXTRA_EXCLUDE_COMPONENTS, excludedComponentNames.toTypedArray())
+        } else {
+            val targetIntents: MutableList<Intent> = ArrayList()
+            for (resolveInfo in resolveInfos) {
+                val activityInfo = resolveInfo.activityInfo
+                if (excludedComponentNames.contains(ComponentName(activityInfo.packageName, activityInfo.name)))
+                    continue
```
Mugurell, line 40: This suggests we can also work with the result of `resolveInfos.filterNot { .. }`.

app/src/main/java/org/mozilla/focus/utils/IntentUtils.kt
```diff
@@ -3,142 +3,68 @@
...
+                val targetIntent = Intent(intent)
+                targetIntent.setPackage(activityInfo.packageName)
```
Mugurell, line 42: nit: If we're setting a `ComponentName` below there's no need to also set the package separately.
With `ComponentName` we specify a particular component of the app, with package allows targeting an exact application but with all of it's components.

app/src/main/java/org/mozilla/focus/utils/IntentUtils.kt
```diff
@@ -3,142 +3,68 @@
...
-                val chooserIntent = Intent.createChooser(intent, chooserTitle)
-                context.startActivity(chooserIntent)
+            // deal with M list seperate problem
+            chooserIntent = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                // create chooser with empty intent in M could fix the empty cells problem
+                Intent.createChooser(Intent(), chooserTitle)
+            } else {
+                // create chooser with one target intent below M
+                Intent.createChooser(targetIntents.removeAt(0), chooserTitle)
             }
```
Mugurell, line 62: Offering more context about these issues and the need for separate approaches would be nice.

Mugurell (changes requested):
I think there are some small improvements to be made but overall this looks good 👍.
Would've been nice to have a common approach between Fenix and Focus shared from AC, maybe later.

Mugurell (approved):
🚢 it!

Issue #5581, opened 2021-10-11
[debug] Tapping on a shortcut displays the "tapping" action briefly before opening the page


lobontiumira:
## Steps to reproduce
1. Build and install the latest debug build from 10/11 (build 11 GV 95.0a1-20211005094529).
2. Add a shortcut to the app's homescreen.
3. Go to the homescreen, and tap on the created shortcut, and observe.

### Expected behavior
Tapping on the shortcut, opens the page in a new tab, without glitches.

### Actual behavior
Tapping on the shortcut displays briefly the "tapping" action, by highlighting the thumbnail and the title of the shortcut - please see the slow-motion short video.

### Device information
* Android device: Google Pixel (Android 10)
* Focus version: the latest debug build from 10/11 (build 11 GV 95.0a1-20211005094529)
* Reproducible on both light, and dark theme, but it is more obvious in light theme.
* It was not reproducible on the debug build from 10/8

https://user-images.githubusercontent.com/40694158/136755157-bd2e57ba-184b-4457-ab06-d2aa9dffcceb.mp4

lobontiumira:
Still reproducible on Google Pixel (Android 10) on the latest Focus Nightly 95.0 (build 353061431 GV 95.0a1-20211101084651).
Not reproducible on Sony Xperia Z5 Premium (Android 7.1.1).

lobontiumira:
Verified as fixed on the 12/3 Focus Nightly with Google Pixel (Android 10).

PR #5951, opened 2021-11-26
For #5581: Do not show an indication on top site item pressed.

mcarare:


pocmo (approved):

mcarare:
`Exception in thread "main" java.net.SocketException: Broken pipe (Write failed)`

Issue #5740, opened 2021-10-29
Add option to provide Shortcuts name


mcarare:
### Why/User Benefit/User Problem

Shortcuts on the home screen cannot be renamed and the name is chosen by default.

### What / Requirements

Add a dialog before saving a shortcut, similar to adding shortcuts to the device launcher screen.

### Acceptance Criteria (how do I know when I’m done?)

Users have the option to name their shortcuts, at least at creation time.

P.S.: This request was also encountered in play store reviews.

brampitoyo:
@mcarare For maximum simplicity and least interruption, I’d prefer for shortcut renaming to happen only after the user long-presses the shortcut tile and select “Rename”.

It’s detailed in this Figma: https://www.figma.com/file/KggJlo8pg2T2Q8nn0Ec5c3/Shortcuts?node-id=1875%3A63694

Otherwise, Fenix already has a rename UI, and we should reuse it on Focus.

mcarare:
@brampitoyo I want to confirm shortcuts behavior when they show up in search fragment:

Currently, on long-press, you still have the option to remove or rename the top sites. 

Should this behavior be kept, or should the shortcuts on the search fragment only be clickable to open the URL?

lobontiumira:
Verified on the 97.0 Focus Nightly (build 353560954 GV 97.0a1-20211220094500) with Google Pixel (Android 10).
When the user deletes all the characters, the "Ok" button becomes inactive.
However, I filed #6110 for the UI.

![rename shortcut ui](https://user-images.githubusercontent.com/40694158/147073766-f655040f-899c-4348-b9d0-977c883f806a.png)

PR #5975, opened 2021-12-02
Add option to provide Shortcuts name

ionutbedregeanu:
For #5740
Create separate composable components for Focus default menu
Remove the duplication of TopSitesOverlay
Send callbacks as parameters to TopSites
Extract SearchOverlay in another file
Integrate top site rename dialog
Create a dedicated typography style for the input field of the dialog

https://user-images.githubusercontent.com/11731652/145040368-94ef455b-85bd-467f-b202-c11b75dfa84b.mp4


app/src/main/java/org/mozilla/focus/ui/dialog/FocusDialog.kt
```diff
@@ -139,21 +149,37 @@ fun DialogTextButton(
...
+            modifier = modifier
+                .wrapContentHeight()
+                .fillMaxWidth(),
+            value = text,
+            onValueChange = onValueChange,
+            textStyle = focusTypography.dialogInput,
+            decorationBox = { innerTextField ->
+                if (text.isEmpty()) {
+                    Text(
+                        text = placeholder,
```
mcarare, line 166: Instead of using this workaround, we can investigate using a `TextField` composable that has a `placeholder` parameter that also comes with a customisable default style.

app/src/main/java/org/mozilla/focus/ui/menu/CustomDropdownMenu.kt
```diff
@@ -0,0 +1,46 @@
...
+@Composable
+fun CustomDropdownMenu(
+    menuItems: List<MenuItem>,
+    isExpanded: Boolean,
+    onDismissClicked: () -> Unit
+) {
+    DropdownMenu(
+        expanded = isExpanded,
+        onDismissRequest = onDismissClicked,
+        modifier = Modifier.background(color = focusColors.topSiteMenuBackground)
```
mcarare, line 30: nit: As we now made this composable reusable and generic, we should also rename this color to reflect this.

app/src/main/java/org/mozilla/focus/ui/menu/CustomDropdownMenu.kt
```diff
@@ -0,0 +1,46 @@
...
+        for (item in menuItems) {
+            DropdownMenuItem(
+                onClick = {
+                    item.onClick()
+                    onDismissClicked.invoke()
+                }
+            ) {
+                Text(
+                    text = item.title,
+                    color = focusColors.topSiteMenuText
```
mcarare, line 41: nit: same as `topSiteMenuBackground`

app/src/main/java/org/mozilla/focus/searchsuggestions/ui/SearchOverlay.kt
```diff
@@ -0,0 +1,120 @@
...
+    val nestedScrollConnection = remember {
+        object : NestedScrollConnection {
+            override fun onPreScroll(available: Offset, source: NestedScrollSource): Offset {
+                onListScrolled.invoke()
+                return Offset.Zero
+            }
+        }
+    }
+
+    FocusTheme {
```
mcarare, line 105: This should not be needed here, as this should be in the fragment when we call `setContent` on a `ComposeView` or in a topmost composable in a complete Compose screen.

app/src/main/java/org/mozilla/focus/ui/dialog/FocusDialog.kt
mcarare: It seems that as `TextField` is a Material design implementation of a text field, they have chosen to hardcode some style-related values, like padding for example: `internal val TextFieldPadding = 16.dp.
This means that using `TextField` will give us no option to adjust the start padding for the input field.

I still feel we should go with the Material design implementation that includes the padding because it comes with out-of-the-box benefits like placeholder handling and optional icons. (And as the name suggests, it follows Material design guidelines)

As we do not have a design for the rename dialog, I think @jeffreygee will need to help with some input regarding the padding and general design of the dialog.
`

mcarare: This is the dialog with the mentioned padding, as a reference:
with text / with the placeholder

mcarare (approved):
LGTM! Great work! 👍

PR #5983, opened 2021-12-03
5293 The "Language" page has extra elements 

iorgamgabriel:
I added a new screen for Languages . I deleted the old dialog and I refactored some of the old code .


pocmo:
Something seems to be wrong with this branch. There are a lot of unrelated commits in this PR.

Issue #5293, opened 2021-09-01
The "Language" page has extra elements


lobontiumira:
## Steps to reproduce
1. Install the latest debug build from main from 9/1.
2. Go to Settings - General - Language, and observe the page.

### Expected behavior
The design is like in Figma https://www.figma.com/file/2V6wgf9fRFdYfLaf5c3oPj/Product-UI?node-id=2140%3A71629.
The page has a back arrow, and no "Cancel" button.

### Actual behavior
The "Language" page has a "Cancel" button in the bottom-right corner, no back arrow, and scrolling through it, displays some borders - please see the attached screenshot.

### Device information
* Android device: Samsung Galaxy Note 8 (Android 9).
* Focus version: latest debug build from main from 9/1

![language - debug](https://user-images.githubusercontent.com/40694158/131624288-ae03cb11-74a6-4c6e-a3b6-47ce0961a586.jpg)

mcarare:
The refactoring of language and block cookies preferences has been postponed after 93.

lobontiumira:
Tested on Focus Nightly 95.0 (build 353060511 GV 95.0a1-20211101084651) with Samsung Galaxy Note 8 (Android 9), and the design in Figma is different:

![language](https://user-images.githubusercontent.com/40694158/139812968-7d9ed80c-c632-44d7-bb3a-b9f82f4cb6ba.png)

iorgamgabriel:
I'm working on this

lobontiumira:
Verified on 1/6 Focus Nightly (build 360061030 GV 97.0a1-20220105093353) with several devices:
* Oppo Find X3 Lite (Android 11),
* Samsung Galaxy Tab A6 (Android 5.1.1),
* Lenovo tab M10 (Android 10), and
* Xiaomi mi4i (Android 5.0.2).

Found some regressions, and filed #6170, and #6171.
I'll remove the QAReady label, and leave this ticket opened until the two new tickets are fixed.

Thank you!

delia-pop:
Verified as fixed also on Beta 97.0.0-beta.1 with Google Pixel 6 (Android 12). 
We'll close this ticket, and the above mentioned issues will be treated separately.

delia-pop:
Reproducible on RC 96.3.0 with Google Pixel 6 (Android 12).

PR #5985, opened 2021-12-03
For #5293 Implemented new Languages screen 

iorgamgabriel:
I implemented a new language screen using compose and the architecture from fenix . I deleted the old dialog for languages . I refactored the class LocaleListPreference.java using kotlin . I deleted CharacterValidator.class because I think it's useless for the current android versions.


Mugurell:
This presents an interesting approach, having all functionality in a Jetpack ViewModel and drive the composable from that 🤔

app/src/main/res/layout/fragment_locale.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?><!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
Mugurell, line 5: This seems like a simple enough layout that could be implemented directly in Compose without the need for the xml wrapper.

app/src/main/java/org/mozilla/focus/locale/screen/LocaleFragment.kt
```diff
@@ -0,0 +1,149 @@
...
+                    saveCurrentLocaleInSharePref(language.tag)
+                    setCurrentLanguage(language.tag)
+                }
+            )
+            languageListItems.add(languageListItem)
+        }
+        return languageListItems
+    }
+
+    private fun getCurrentLocale(): String {
```
Mugurell, line 109: If we are using a ViewModel to handle the business logic then the responsibility of deciding where/how to save the current locale should be extracted from the Fragment.

Mugurell (commented):
Other than some style cleanup I see two important changes needed:
- use only a composable, without the xml wrapper
- refactor all functionality from the Fragment to the ViewModel.
Though I'm curious on whether going with this ViewModel / MVVM approach suits the app going forward.
@pocmo

Mugurell:
Talked with Sebastian about the ViewModel approach which would be something new to the app.
He suggested trying out simpler classes for each particular functionality:
 - `*ScreenDataSource` for gathering data from various sources and mapping it to an observable state to be displayed
 - `*ScreenInteractor` for responding to all user events, handling all business logic
 All data updates should follow the flow used on Fenix where each screen has a Store which is updated in a Reducer following various Actions.

app/src/main/java/org/mozilla/focus/components/StoreProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+ */
+class StoreProviderFactory<T : Store<*, *>>(
+    private val createStore: () -> T
+) : ViewModelProvider.Factory {
+
+    @Suppress("UNCHECKED_CAST")
+    override fun <VM : ViewModel?> create(modelClass: Class<VM>): VM {
+        return StoreProvider(createStore()) as VM
+    }
+}
```
pocmo, line 44: Attaching the `Store` to a `ViewModel` is an interesting idea. But I wonder if it is needed. If we want to keep track of a `Store` instance across recompositions then we could/should use `remember()`.

To create a `Store` that saves/restores it state through configuration changes, and including process restores, we have a `composableStore()` helper [in A-C](https://github.com/mozilla-mobile/android-components/blob/main/components/lib/state/src/main/java/mozilla/components/lib/state/ext/ComposeExtensions.kt#L84).

Since all data the locale store holds is already persisted, it seems like we could away with just using `remember()` for recompositions and otherwise, on configuration changes, can live with the store (and its state) getting recreated?

app/src/main/java/org/mozilla/focus/locale/screen/LanguageStorage.kt
```diff
@@ -0,0 +1,97 @@
...
+     * This method filters down the list before generating the descriptor array.
+     */
+    private fun getUsableLocales(): Array<LocaleDescriptor?> {
+        val shippingLocales = LocaleManager.getPackagedLocaleTags()
+        val initialCount: Int = shippingLocales.size
+        val locales: MutableSet<LocaleDescriptor> = HashSet(initialCount)
+        for (tag in shippingLocales) {
+            val descriptor = LocaleDescriptor(tag)
+            if (!descriptor.isUsable()) {
+                Log.w(LOG_TAG, "Skipping locale $tag on this device.")
```
pocmo, line 83: We do not use `Log` directly [at least in new code]. Instead we usually create a `Logger()` (from `support-base`). The logger allows us to have more control about what gets logged when (e.g. in Fenix for example we limit what we log in release builds).

app/src/main/java/org/mozilla/focus/locale/screen/LocaleDescriptor.kt
```diff
@@ -0,0 +1,152 @@
...
+        languageCodeAndNameMap["ay"] = "Aimara"
+        languageCodeAndNameMap["quc"] = "K'iche'"
+        languageCodeAndNameMap["tsz"] = "P'urhepecha"
+        languageCodeAndNameMap["jv"] = "Basa Jawa"
+        languageCodeAndNameMap["ppl"] = "Náhuat Pipil"
+        languageCodeAndNameMap["su"] = "Basa Sunda"
+        languageCodeAndNameMap["hus"] = "Tének"
+        languageCodeAndNameMap["co"] = "Corsu"
+        languageCodeAndNameMap["sn"] = "ChiShona"
+        languageCodeAndNameMap["zh-CN"] = "中文 (中国大陆)"
```
pocmo, line 54: (Off-topic): Eventually we should have this map somewhere in A-C, to avoid having this in Focus and Fenix.

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -61,7 +62,7 @@ object Dependencies {
     const val androidx_lifecycle_extensions = "androidx.lifecycle:lifecycle-extensions:${Versions.AndroidX.lifecycle}"
 
     const val google_material = "com.google.android.material:material:${Versions.Google.material}"
-
+    const val google_accompanist_insets_ui = "com.google.accompanist:accompanist-insets-ui:${Versions.Google.accompanist}"
```
pocmo, line 65: Is this strictly needed? Just curious whether we require `accompanist` and whether we want to "officially" make it part of our compose tool set (CC-ing some folks for opinions, @jonalmeida @Mugurell @gabrielluong @mcarare)

pocmo (commented):
Sorry, this took me a while to get to it. Overall I think this is really great! I added some questions/comments, but I think this could land soon. :)

app/src/main/java/org/mozilla/focus/navigation/MainActivityNavigation.kt
```diff
@@ -1,7 +1,6 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
```
gabrielluong, line 4: nit: looks like an unwanted change

buildSrc/src/main/java/Dependencies.kt
Mugurell: This accompanist lib was needed to ensure the `Scaffold` with a toolbar will be placed below the status bar.
This was needed because the`MainActivity` which shows this fragment [is set as fullscreen](https://github.com/mozilla-mobile/focus-android/blob/67129f722a3717e4b67e542e7fcb82c1de70f877/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt#L68) .

app/src/main/java/org/mozilla/focus/locale/screen/LocaleFragmentCompose.kt
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Displays a single language Text
+ *
+ * @param language of the item to be display in the textView
+ * @param onClick Callback when the user taps on Language text , the same like on the radioButton
+ */
+@Composable
+private fun LanguageDisplayName(language: Language, onClick: (String) -> Unit) {
+    if (language.displayName != null) {
```
gabrielluong, line 162: We already check for `language.displayName?` in Line 125 before calling `LanguageDisplayName`. We could simply just pass the `displayName` so we can avoid this check.

buildSrc/src/main/java/Dependencies.kt
gabrielluong: I do have `accompanist` as part of my toolset for my Top Sites rewrite on Fenix because I needed the ViewPager equivalent. I don't know much about what about Android insets to provide much opinion on this particular dependency.

pocmo: Okay, if it will be used for Fenix too, then I guess this is already decided. :) 👍

app/src/main/java/org/mozilla/focus/locale/screen/LocaleDescriptor.kt
gabrielluong: We actually do this quite differently in Fenix. https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleManagerExtension.kt#16

pocmo: We still seem to have a map though. For some reason containing a lot of strings that should be in Android already though.
https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleViewHolders.kt#67

app/src/main/java/org/mozilla/focus/components/StoreProvider.kt
iorgamgabriel: org.mozilla.fenix.components.StoreProvider is the same class like in Fenix  . It is used in all the fragment there . I think is a good idea to move the class in A-C , to eliminate  duplicated code.

app/src/main/java/org/mozilla/focus/navigation/MainActivityNavigation.kt
iorgamgabriel: I made the changes.

app/src/main/java/org/mozilla/focus/locale/screen/LocaleFragmentCompose.kt
iorgamgabriel: I made the changes.

app/src/main/java/org/mozilla/focus/locale/screen/LanguageStorage.kt
iorgamgabriel: I made the changes.

app/src/main/java/org/mozilla/focus/components/StoreProvider.kt
pocmo: I think that emerged early in Fenix when we wanted to keep the state of a `Store` through configuration changes. For a `View`-based app using a `ViewModel` seems like the most convenient way. Now, with Jetpack Compose, we may have other patterns/mechanisms that are not as heavy as a `ViewModel`. Here the `ViewModel` only acts as a container that holds the `Store`. There are more low-level APIs of Compose that we could use (some examples above). But I think that we do not need to solve this now - something similar will have to happen in Fenix at some point too. (CC @csadilek, just to be aware of this).

In this particular case I wonder if we could not do anything fancy at all and just create a Store and attach it to `LanguageFragment`. It seems like there's no state we need to keep around since it can be recreated from the storage anyways?

app/src/main/java/org/mozilla/focus/locale/screen/DefaultLanguageScreenInteractor.kt
```diff
@@ -0,0 +1,51 @@
...
+    private val languageScreenStore: LanguageScreenStore,
+    private val localeUseCase: LocaleUseCases
+) : LanguageScreenInteractor {
+    override fun handleLanguageSelected(language: Language) {
+        if (languageScreenStore.state.selectedLanguage == language) {
+            return
+        }
+        languageScreenStore.dispatch(LanguageScreenAction.Select(selectedLanguage = language))
+        LanguageStorage(context = activity).saveCurrentLanguageInSharePref(language.tag)
+        setCurrentLanguage(language.tag)
```
pocmo, line 30: A pattern, that has been really successful, is to not expose any storage to the UI at all and instead handle this in a Middleware. The UI only displays state from the store, and only dispatches actions on the store. Then a Middleware takes care of calling actions on the store (or the system APIs) whenever it sees specific actions; or initializes the state when it sees the `InitAction`. This completely avoids that you have to have such "synchronizing" code where you need to let multiple things know about the new language; which doesn't scale so well.

The ContainerMiddleware is a nice example. It's the only piece of code interacting with the storage. Other code only reads and updates state:
https://github.com/mozilla-mobile/android-components/blob/main/components/feature/containers/src/main/java/mozilla/components/feature/containers/ContainerMiddleware.kt

app/src/main/java/org/mozilla/focus/locale/screen/LanguageScreenStore.kt
```diff
@@ -0,0 +1,62 @@
...
+ * The initial LanguageScreenState when the user enters first in the screen
+ */
+fun createInitialLanguagesScreenState(context: Context): LanguageScreenState {
+    val storage = LanguageStorage(context = context)
+    val languageList = storage.getLanguages()
+    return LanguageScreenState(
+        languageList = languageList,
+        selectedLanguage = storage.getSelectedLanguageTag()
+    )
+}
```
pocmo, line 38: See other comment. This could get initialized from a middleware.

app/src/main/java/org/mozilla/focus/locale/screen/LanguageFragment.kt
```diff
@@ -0,0 +1,127 @@
...
+import org.mozilla.focus.ext.requireComponents
+import org.mozilla.focus.settings.BaseSettingsComposeFragment
+import org.mozilla.focus.state.AppAction
+import org.mozilla.focus.state.Screen
+
+class LanguageFragment : BaseSettingsComposeFragment() {
+    private lateinit var browserStore: BrowserStore
+    private lateinit var localeUseCase: LocaleUseCases
+    private lateinit var languageScreenStore: LanguageScreenStore
+    private lateinit var screenInteractorDefault: DefaultLanguageScreenInteractor
```
pocmo, line 35: Instead of `lateinit`, could some of them be initialized from a `lazy { .. } ` block? `lateinit` has bitten us a couple of times. Code gets refactored/moved and suddenly there are edge cases where non initialized properties get accessed. I try to avoid it whenever possible.

iorgamgabriel:
thank you all. updated the patch based on your suggestions

app/src/main/java/org/mozilla/focus/locale/screen/LanguageFragment.kt
```diff
@@ -0,0 +1,119 @@
...
+            languageScreenStore = languageScreenStore
+        )
+    }
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View {
+        screenInteractorDefault.initLanguages()
```
pocmo, line 53: nit: You could consider moving the dispatching of the init action into the store class if it should happen for every new store. that's how we do it in `BrowserStore`:
https://github.com/mozilla-mobile/android-components/blob/main/components/browser/state/src/main/java/mozilla/components/browser/state/store/BrowserStore.kt#L47

app/src/main/java/org/mozilla/focus/locale/screen/LanguageMiddleware.kt
```diff
@@ -0,0 +1,65 @@
...
+                setCurrentLanguage(action.selectedLanguage.tag)
+            }
+            is LanguageScreenAction.UpdateLanguages -> {
+                next(action)
+            }
+            is LanguageScreenAction.InitLanguages -> {
+                /**
+                 * The initial LanguageScreenState when the user enters first in the screen
+                 */
+                val storage = LanguageStorage(context = activity)
```
pocmo, line 37: nit: Does it make sense to keep just one storage instance in the middleware instead of creating a new one for whenever we process an action?

app/src/main/java/org/mozilla/focus/locale/screen/LanguageMiddleware.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    override fun invoke(
+        context: MiddlewareContext<LanguageScreenState, LanguageScreenAction>,
+        next: (LanguageScreenAction) -> Unit,
+        action: LanguageScreenAction
+    ) {
+        when (action) {
+            is LanguageScreenAction.Select -> {
+                LanguageStorage(context = activity).saveCurrentLanguageInSharePref(action.selectedLanguage.tag)
+                setCurrentLanguage(action.selectedLanguage.tag)
```
pocmo, line 28: We are not calling `next()` for select? Then this will never end up in the reducer.

app/src/main/java/org/mozilla/focus/locale/screen/LanguageMiddleware.kt
```diff
@@ -0,0 +1,65 @@
...
+                 * The initial LanguageScreenState when the user enters first in the screen
+                 */
+                val storage = LanguageStorage(context = activity)
+                context.dispatch(
+                    LanguageScreenAction.UpdateLanguages(
+                        storage.getLanguages(),
+                        storage.getSelectedLanguageTag()
+                    )
+                )
+            }
```
pocmo, line 44: In the `else` case we may want to invoke `next()`, otherwise this middleware will intercept all other (future) actions.

pocmo (approved):

app/src/main/java/org/mozilla/focus/locale/screen/LanguageFragment.kt
iorgamgabriel: I made the changes

app/src/main/java/org/mozilla/focus/locale/screen/LanguageMiddleware.kt
iorgamgabriel: I made the changes

app/src/main/java/org/mozilla/focus/locale/screen/LanguageMiddleware.kt
iorgamgabriel: I made the changes

Issue #5935, opened 2021-11-23
Add a setting for indicating the HTTPS-Only Mode configuration


Amejia481:
Users should be able to select the HTTPS-Only mode configuration.

<img width="703" alt="image" src="https://user-images.githubusercontent.com/773158/143082220-5617496d-5146-489d-9c9d-3ca9bf35fb8c.png">

Amejia481:
Related AC ticket https://github.com/mozilla-mobile/android-components/issues/11305

Amejia481:
As Focus by policy we don't persist any information, I think we may not need to add a exception page

pocmo:
Since Focus is always in private mode, I assume we only need an ON/OFF option here.

Amejia481:
Agree!

pocmo:
Opened a draft PR that introduces the setting and uses the desktop strings:

![https_only_setting](https://user-images.githubusercontent.com/89638/144649627-bdef4a71-5ca9-4ff9-bfaf-425a3d9e6793.png)

lobontiumira:
Verified on the Focus Nightly 97.0 (build 353441046 GV 97.0a1-20211208095339) with Samsung Galaxy Note 8 (Android 9) that the new option HTTPS-Only Mode is displayed in Settings - Privacy & Security - Security area.
The toggle can be enabled/disabled without glitches.
The "Learn more" link when tapped redirects the user to https://support.mozilla.org/en-US/kb/https-only-prefs SUMO page.

![Screenshot_4](https://user-images.githubusercontent.com/40694158/145550830-75abd2c7-b8d6-4970-bd09-ae5fffd4b846.jpg)

PR #5988, opened 2021-12-03
Issue #5935: Introduce setting for HTTPS-Only mode

pocmo:
This patch introduces HTTPS-Only mode. _It's a draft_.

* For the preference: Should we have a "learn more" URL in settings like desktop? ([Desktop page](https://support.mozilla.org/en-US/kb/https-only-prefs))
* Preference placement: Like desktop, I created a "Security" section, that contains SafeBrowsing and this new setting. Is that okay?
* This is only a minimal version of the error page compared to desktop (See https://github.com/mozilla-mobile/android-components/issues/11306). The reason for that is that we do not support such rich error pages in A-C yet. But the error page is fully functional.
* All strings I used are from desktop. Do they work for mobile?

CC @jeffreygee 

https_only_setting

https_only_errorpage

### DIFF
--- a/app/src/main/java/org/mozilla/focus/Components.kt
+++ b/app/src/main/java/org/mozilla/focus/Components.kt
@@ -95,7 +95,8 @@ class Components(
             trackingProtectionPolicy = settings.createTrackingProtectionPolicy(),
             javascriptEnabled = !settings.shouldBlockJavaScript(),
             remoteDebuggingEnabled = settings.shouldEnableRemoteDebugging(),
-            webFontsEnabled = !settings.shouldBlockWebFonts()
+            webFontsEnabled = !settings.shouldBlockWebFonts(),
+            httpsOnlyMode = settings.getHttpsOnlyMode()
         )
     }
 
--- a/app/src/main/java/org/mozilla/focus/settings/HttpsOnlyModePreference.kt
+++ b/app/src/main/java/org/mozilla/focus/settings/HttpsOnlyModePreference.kt
@@ -1,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.settings
+
+import android.content.Context
+import android.util.AttributeSet
+import mozilla.components.concept.engine.Engine
+import org.mozilla.focus.R
+import org.mozilla.focus.ext.components
+import org.mozilla.focus.utils.Features
+
+/**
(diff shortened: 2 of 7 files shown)


pocmo:
Needs A-C build with https://github.com/mozilla-mobile/android-components/pull/11352

jeffreygee:
@pocmo 

For the preference: Should we have a "learn more" URL in settings like desktop? (Desktop page) /   
All strings I used are from desktop. Do they work for mobile? **Yes, I think it's good to keep consistent with desktop. I will double check with our content designers to make sure**

Preference placement: Like desktop, I created a "Security" section, that contains SafeBrowsing and this new setting. Is that okay? **Looks good to me!**

This is only a minimal version of the error page compared to desktop - **Have we ever thought of having a secondary CTA/link below like 'Manage your settings' where we can viewport them back to their security settings?**
---

What are your thoughts on setting the default state as toggled on? Something we can probably discuss with the group in the weekly.

pocmo:
> This is only a minimal version of the error page compared to desktop - **Have we ever thought of having a secondary CTA/link below like 'Manage your settings' where we can viewport them back to their security settings?**

That's an interesting idea. We would need to find a way to  communicate this from our "error pages" component to the app. Not impossible. But something we need to figure out.

> **_What are your thoughts on setting the default state as toggled on to enhance the security experience for our users? Something we can probably discuss with the group in the weekly._**

That sounds great. Yeah, let's discuss this. Maybe this requires us to change the error page text though. Otherwise the user may not really understand it, if they haven't enabled this themselves? 🤔

pocmo:
Updated the PR.

* HTTPS-Only mode is now behind a feature flag that is only enabled in Nightly (and debug builds)
* The setting is now enabled by default

mcarare (approved):
👌

pocmo:
Filed follow-up issues:
* https://github.com/mozilla-mobile/focus-android/issues/6008
* https://github.com/mozilla-mobile/focus-android/issues/6009
* https://github.com/mozilla-mobile/focus-android/issues/6010

jeffreygee:
@pocmo I worked with one of our content designers for strings updates:

**[Settings screen]**
Header: HTTPS-only mode
Subheader: Automatically attempts to connect to sites using HTTPS for increased security. HTTPS connections provide a secure, encrypted connection to websites you visit. 
CTA: Learn more

**[Error Page]**
Title: The site you requested does not support HTTPS
Body: By default, Firefox Focus attempts to connect using HTTPS for increased security. To change this setting or to learn more, go to **Settings > Privacy & Security > Security**. 
CTA: Continue to HTTP Site

pocmo:
@jeffreygee Awesome, thank you! I will work on getting this updated in #6009.

Issue #5767, opened 2021-11-03
Migrate the deprecated Kotlin Android Extensions compiler plugin


mcarare:
Since Kotlin 1.4.20 `android-kotlin-extensions` is deprecated and it is to be removed ( [timeline](https://android-developers.googleblog.com/2020/11/the-future-of-kotlin-android-extensions.html) is after September 2021 Kotlin release).

Migration guide: https://developer.android.com/topic/libraries/view-binding/migration.

Note for QA:  A smoke test through the entire app is required to test this ticket, to check if all the screen are loaded properly, if all the buttons are working and the behaviour wasn't altered

lobontiumira:
I have encountered this [crash](https://github.com/mozilla-mobile/focus-android/issues/6094) while accessing the Cookies options.
I'll remove the QAReady label for now.

ionutbedregeanu:
> I have encountered this [crash](https://github.com/mozilla-mobile/focus-android/issues/6094) while accessing the Cookies options. I'll remove the QAReady label for now.

Fixed

lobontiumira:
I'll close this ticket as verified as fixed on the 97.0 Focus Nightly (build 353551145 GV 97.0a1-20211220094500) with Xiaomi Mi8 Lite (Android 10), and Samsung Galaxy Note 8 (Android 9).

PR #5997, opened 2021-12-06
Migrate to Jetpack view binding

ionutbedregeanu:
For #5767
Migrate from Kotlin synthetics to Jetpack view binding
Remove some unused or unnecessary imports

### DIFF
--- a/app/src/main/java/org/mozilla/focus/exceptions/ExceptionsListFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/exceptions/ExceptionsListFragment.kt
@@ -22,7 +22,6 @@ import androidx.recyclerview.widget.ItemTouchHelper
 import androidx.recyclerview.widget.ItemTouchHelper.SimpleCallback
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
-import kotlinx.android.synthetic.main.fragment_exceptions_domains.*
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
@@ -31,6 +30,7 @@ import mozilla.components.concept.engine.content.blocking.TrackingProtectionExce
 import org.mozilla.focus.GleanMetrics.TrackingProtectionExceptions
 import org.mozilla.focus.R
 import org.mozilla.focus.autocomplete.AutocompleteDomainFormatter
+import org.mozilla.focus.databinding.FragmentExceptionsDomainsBinding
 import org.mozilla.focus.ext.components
 import org.mozilla.focus.ext.requireComponents
 import org.mozilla.focus.settings.BaseSettingsLikeFragment
@@ -50,6 +50,9 @@ open class ExceptionsListFragment : BaseSettingsLikeFragment(), CoroutineScope {
     private var job = Job()
     override val coroutineContext: CoroutineContext
         get() = job + Dispatchers.Main
+    private var _binding: FragmentExceptionsDomainsBinding? = null
+    protected val binding get() = _binding!!
+
     /**
      * ItemTouchHelper for reordering items in the domain list.
      */
@@ -105,29 +108,35 @@ open class ExceptionsListFragment : BaseSettingsLikeFragment(), CoroutineScope {
         inflater: LayoutInflater,
         container: ViewGroup?,
         savedInstanceState: Bundle?
-    ): View = inflater.inflate(R.layout.fragment_exceptions_domains, container, false)
+    ): View {
+        _binding = FragmentExceptionsDomainsBinding.inflate(inflater, container, false)
+        return binding.root
+    }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        exceptionList.layoutManager =
-            LinearLayoutManager(activity, RecyclerView.VERTICAL, false)
-        exceptionList.adapter = DomainListAdapter()
-        exceptionList.setHasFixedSize(true)
-
-        removeAllExceptions.isVisible = !isSelectionMode()
-
-        if (!isSelectionMode()) {
-            itemTouchHelper.attachToRecyclerView(exceptionList)
+        binding.exceptionList.apply {
+            layoutManager =
+                LinearLayoutManager(activity, RecyclerView.VERTICAL, false)
+            adapter = DomainListAdapter()
+            setHasFixedSize(true)
         }
 
-        removeAllExceptions.setOnClickListener { removeButton ->
+        binding.removeAllExceptions.isVisible = !isSelectionMode()
+
+        if (!isSelectionMode()) {
+            itemTouchHelper.attachToRecyclerView(binding.exceptionList)
+        }
+
+        binding.removeAllExceptions.setOnClickListener { removeButton ->
             removeButton.apply {
                 isEnabled = false
                 alpha = REMOVE_EXCEPTIONS_DISABLED_ALPHA
             }
             requireComponents.trackingProtectionUseCases.removeAllExceptions {
-                val exceptionsListSize = (exceptionList.adapter as DomainListAdapter).itemCount
+                val exceptionsListSize =
+                    (binding.exceptionList.adapter as DomainListAdapter).itemCount
                 TrackingProtectionExceptions.allowListCleared.record(
                     TrackingProtectionExceptions.AllowListClearedExtra(exceptionsListSize)
                 )
@@ -147,8 +156,8 @@ open class ExceptionsListFragment : BaseSettingsLikeFragment(), CoroutineScope {
 
         updateTitle(R.string.preference_exceptions)
 
-        (exceptionList.adapter as DomainListAdapter).refresh(requireActivity()) {
-            if ((exceptionList.adapter as DomainListAdapter).itemCount == 0) {
+        (binding.exceptionList.adapter as DomainListAdapter).refresh(requireActivity()) {
+            if ((binding.exceptionList.adapter as DomainListAdapter).itemCount == 0) {
                 requireComponents.appStore.dispatch(
                     AppAction.NavigateUp(requireComponents.store.state.selectedTabId)
                 )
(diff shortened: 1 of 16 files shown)


app/src/main/java/org/mozilla/focus/exceptions/ExceptionsListFragment.kt
```diff
@@ -105,29 +107,35 @@ open class ExceptionsListFragment : BaseSettingsLikeFragment(), CoroutineScope {
         inflater: LayoutInflater,
         container: ViewGroup?,
         savedInstanceState: Bundle?
-    ): View = inflater.inflate(R.layout.fragment_exceptions_domains, container, false)
+    ): View {
+        FragmentExceptionsDomainsBinding.inflate(inflater, container, false)
+        return binding.root
```
Mugurell, line 112: Seems like the binding property is not initialized?
It will also need to be set to null in `onDestroyView`.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -135,30 +128,23 @@ class BrowserFragment :
...
 
     @Suppress("LongMethod", "ComplexMethod")
-    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
-        val view = inflater.inflate(R.layout.fragment_browser, container, false)
-
-        statusBar = view.findViewById(R.id.status_bar_background)
-
-        popupTint = view.findViewById(R.id.popup_tint)
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+        binding = FragmentBrowserBinding.inflate(inflater, container, false)
```
Mugurell, line 132: The binding instance needs to be freed in `onDestroyView`.

app/src/main/java/org/mozilla/focus/fragment/CrashReporterFragment.kt
```diff
@@ -10,19 +10,23 @@ import android.view.View
...
     var onCloseTabPressed: ((sendCrashReport: Boolean) -> Unit)? = null
+    private lateinit var binding: FragmentCrashReporterBinding
 
     override fun onCreateView(
         inflater: LayoutInflater,
         container: ViewGroup?,
         savedInstanceState: Bundle?
-    ): View? = inflater.inflate(R.layout.fragment_crash_reporter, container, false)
+    ): View {
+        binding = FragmentCrashReporterBinding.inflate(inflater, container, false)
```
Mugurell, line 27: The binding instance needs to be freed in `onDestroyView`.
But better yet, if we only need access to views inside `onViewCreated` then we don't need to keep around the binding instance, and the ceremony for freeing it.

We can have
```
class CrashReporterFragment : Fragment(R.layout.fragment_crash_reporter) {
    // no need for the `onCreateView` override since the ctor receiving a layout will itself call `onCreateView`
    ...
    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
        super.onViewCreated(view, savedInstanceState)
        val binding = FragmentCrashReporterBinding.bind(view)
        ...
    }
}
```

app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.kt
```diff
@@ -39,47 +38,21 @@ class FirstrunFragment : Fragment(), View.OnClickListener {
...
-        viewPager!!.addOnPageChangeListener(object : ViewPager.OnPageChangeListener {
-            override fun onPageSelected(position: Int) {
-                Onboarding.pageDisplayed.record(Onboarding.PageDisplayedExtra(0))
-
-                viewPager!!.contentDescription = adapter.getPageAccessibilityDescription(position)
-            }
-
-            override fun onPageScrolled(position: Int, positionOffset: Float, positionOffsetPixels: Int) {}
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+        binding = FragmentFirstrunBinding.inflate(inflater, container, false)
```
Mugurell, line 42: The binding instance needs to be freed in `onDestroyView`.

app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.kt
```diff
@@ -97,6 +70,34 @@ class FirstrunFragment : Fragment(), View.OnClickListener {
         }
     }
 
+    @Suppress("MagicNumber")
+    private fun setPager() {
```
Mugurell, line 74: This refactoring while not needed for the ticket is a nice touchup 👍. 
Could the method be named `setupPager`? Just `set` to me says this is a setter.

app/src/main/java/org/mozilla/focus/searchsuggestions/ui/SearchSuggestionsFragment.kt
```diff
@@ -118,29 +119,32 @@ class SearchSuggestionsFragment : Fragment(), CoroutineScope {
         inflater: LayoutInflater,
         container: ViewGroup?,
         savedInstanceState: Bundle?
-    ): View? = inflater.inflate(R.layout.fragment_search_suggestions, container, false)
+    ): View {
+        binding = FragmentSearchSuggestionsBinding.inflate(inflater, container, false)
```
Mugurell, line 123: The binding instance needs to be freed in `onDestroyView`.

app/src/main/java/org/mozilla/focus/session/ui/TabSheetFragment.kt
```diff
@@ -18,50 +18,44 @@ import androidx.recyclerview.widget.RecyclerView
...
-
     private var scope: CoroutineScope? = null
+    private lateinit var binding: FragmentSessionssheetBinding
 
-    private lateinit var store: BrowserStore
-
-    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
-        val view = inflater.inflate(R.layout.fragment_sessionssheet, container, false)
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+        binding = FragmentSessionssheetBinding.inflate(inflater, container, false)
```
Mugurell, line 38: The binding instance needs to be freed in `onDestroyView`.

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -242,28 +242,27 @@ class UrlInputFragment :
         inflater: LayoutInflater,
         container: ViewGroup?,
         savedInstanceState: Bundle?
-    ): View? {
-        val view = inflater.inflate(R.layout.fragment_urlinput, container, false)
+    ): View {
+        binding = FragmentUrlinputBinding.inflate(inflater, container, false)
```
Mugurell, line 246: The binding instance needs to be freed in `onDestroyView`.

app/src/main/java/org/mozilla/focus/autocomplete/AutocompletePreference.kt
```diff
@@ -1,36 +0,0 @@
...
-import android.widget.TextView
-import androidx.preference.Preference
-import androidx.preference.PreferenceViewHolder
-import org.mozilla.focus.R
-import org.mozilla.focus.ext.settings
-
-/**
- * Autocomplete preference that will show a sub screen to configure the autocomplete behavior.
- */
-class AutocompletePreference(context: Context?, attrs: AttributeSet?) : Preference(context, attrs) {
```
Mugurell, line 18: This seems to have been unused for some time.
Nice to have cleaned it though it would've been nice to be in a separate commit.
If there are more such situations in code maybe a ticket just for cleaning unused classes and resources would be useful.

Mugurell (changes requested):
Looks good but could use a bit of cleaning especially around following the recommended pattern of freeing to binding instance in `onDestroyView`.
Separating now extraneous changes not needed for the ticket is not really needed but I'd say is a good practice to keep in mind for the future.

ionutbedregeanu:
I create another ticket https://github.com/mozilla-mobile/focus-android/issues/6049 to fix those code warnings which doesn't imply a major refactoring

ionutbedregeanu:
@sv-ohorvath I marked some UI tests with Ignore since HomeScreenRobot implies an update after this migration

app/src/test/java/org/mozilla/focus/searchsuggestions/SearchSuggestionsViewModelTest.kt
```diff
@@ -32,7 +31,6 @@ class SearchSuggestionsViewModelTest {
     @Before
     fun setup() {
         lifecycle = LifecycleRegistry(mock(LifecycleOwner::class.java))
-        lifecycle.handleLifecycleEvent(Lifecycle.Event.ON_RESUME)
```
Mugurell, line 35: Good find that this isn't really needed 👍

app/src/main/java/org/mozilla/focus/fragment/CrashReporterFragment.kt
```diff
@@ -5,35 +5,27 @@
...
-        inflater: LayoutInflater,
-        container: ViewGroup?,
-        savedInstanceState: Bundle?
-    ): View? = inflater.inflate(R.layout.fragment_crash_reporter, container, false)
-
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-
-        view.findViewById<View>(R.id.background).background =
+        val binding = FragmentCrashReporterBinding.bind(view)
```
Mugurell, line 21: We are trying to bind here a View which wasn't previously created - inflated.
As such this method - `onViewCreated` won't even be called.

Nice that you've removed the almost empty `onCreateView` but if not inflating the layout from there we should use the Fragment constructor taking the layout resource as a parameter like in 
```
class CrashReporterFragment : Fragment(R.layout.fragment_crash_reporter) { .. }
```

This fragment's functionality can then be tested using `about:crashtab`.

Mugurell (commented):
Small change needed in `CrashReporterFragment`, LGTM otherwise.

Mugurell (approved):

Issue #5996, opened 2021-12-06
[L10N] The default cookies option isn't selected after switching to a different language


lobontiumira:
## Steps to reproduce
1. Go to Settings - General - Language, and select a different language then the system's.
2. Go to Settings - Privacy & Security - Cookies and Site Data, and tap on the "Block cookies".

### Expected behavior
The default cookies option is selected.

### Actual behavior
The default cookies option isn't selected. None of the options is selected.
Switching to the system's language fixes the issue.

### Device information
* Android device: Sony Xperia Z5 Premium (Android 7.1.1), Samsung Galaxy Note 8 (Android 9)
* Focus version: Focus Nightly 12/6, RC 95.1.0

![cookies](https://user-images.githubusercontent.com/40694158/144847498-ff82e820-e809-42fb-aa93-f2b78e86a717.png)

lobontiumira:
Verified as fixed on the 97.0 Focus Nightly build 353441054 (GV 97.0a1-20211208095339) with Samsung Galaxy Note 8 (Android 9), and a Sony Xperia Z5 Premium (Android 7.1.1).

lobontiumira:
Verified also on Beta 96.0.0-beta.2 with Google Pixel (Android 10).

PR #6002, opened 2021-12-06
For #5996: Improve cookie option preference handling for all locales.

mcarare:


app/src/main/java/org/mozilla/focus/widget/CookiesPreference.kt
```diff
@@ -31,11 +31,16 @@ class CookiesPreference(context: Context?, attrs: AttributeSet?) : ListPreferenc
 
     fun updateSummary() {
         val settings = context.settings
-        super.setSummary(settings.shouldBlockCookiesValue())
+
+        val cookieOptionIndex =
+            context.resources.getStringArray(R.array.cookies_options_entry_values)
+                .asList().indexOf(settings.shouldBlockCookiesValue())
+        this.summary =
+            context.resources.getStringArray(R.array.cookies_options_entries)[cookieOptionIndex]
```
ionutbedregeanu, line 39: I'm receiving ArrayIndexOutOfBoundsException: length=5; index=-1 at this line

mcarare: Thanks for testing. I think you already had a setting saved with another locale. We have to find a solution for that, but I think it will involve resetting user changes to the default value for other locales.

PR #6011, opened 2021-12-07
Set version to 96.0.0.

mcarare:


jonalmeida (approved):

mcarare:
This seems to be stuck. Reopening.

Issue #6049, opened 2021-12-13
Address code warnings


ionutbedregeanu:
Remove unused imports, classes, methods etc
Address those code warnings which doesn't imply a major refactoring

Note for QA: Theoretically this ticket shouldn't add regressions because just remove unused code, but maybe a smoke testing should be done just to be sure

lobontiumira:
I finished a smoke & sanity testing run on the latest Focus Nightly from 1/20 with Oppo Find X3 Lite (Android 11), and there seems to be no regressions from removing unused code.

PR #6102, opened 2021-12-21
Address those warnings which doesn't imply code refactoring

ionutbedregeanu:
For #6049
Remove unused imports, methods, classes
Remove redundant imports
Make methods private(where it was an warning)
Replace if with when(where it was an warning)
Extract listener as lambda(where it was an warning)

I know it's a pretty big pr but we tend to gather too much unused code and many minor warnings


Mugurell (approved):
chris-pratt-wow

ionutbedregeanu:
error: RPC failed; curl 56 GnuTLS recv error (-54): Error in the pull function.

PR #6111, opened 2021-12-22
Document all the steps required to create a new release branch

ionutbedregeanu:


ionutbedregeanu:
@amedyne @rvandermeulen I documented all the steps required to create a new release branch on Focus, but in project file(ReleaseBranch.md).
Should I add those steps in Wiki and delete ReleaseBranch.md or should be kept in the both places?

jonalmeida (approved):
The documentation looks great!

Depending on where you decide to place the documentation, please remove the [old release docs](https://github.com/mozilla-mobile/focus-android/wiki/Release-Process) here so we have one source of truth. 🙂

docs/ReleaseBranch.md
```diff
@@ -0,0 +1,38 @@
+### Creating a new Release Branch
+
+1. Create a branch name with the format `releases_v[version]` (for example: `releases_v87.0.0`).
```
ionutbedregeanu, line 3: Maybe for consistency we should use the same versioning format for the name of the branch as in the versionName  from gradle(for example: `releases_v87.0` and not `releases_v87.0.0`)
@rvandermeulen What do you think?

jonalmeida: Ideally, we did want to have it as v87.0 as we do for AC, but we stuck to consistency with Fenix had started the v87.0.0 and _I think_ some automation is built around that format for now - so we need to check that nothing breaks there during release.

We should definitely switch them both to dropping the patch version though.

ionutbedregeanu:
I will close this pr since we decide to update the old wiki documentation and be the single source of truth
https://github.com/mozilla-mobile/focus-android/wiki/Release-Process

Issue #6110, opened 2021-12-22
"Rename" shortcut's prompt should have the same UI as the other prompts


lobontiumira:
## Steps to reproduce
1. Add a page as a shortcut.
2. Long-tap on the shortcut.
3. Select the "Rename" option.
4. Observe the UI.

### Expected behavior
The UI is similar to other prompts from Focus, please see the attached screenshots.

### Actual behavior
The prompt has grey colored "Cancel", "Ok" options, and the edit line is purple, not pink.

### Device information
* Android device: Google Pixel (Android 10)
* Focus version: 97.0 Focus Nightly

![rename shortcut ui](https://user-images.githubusercontent.com/40694158/147073257-ce849e22-5c3e-4449-9103-62e719237bdd.png)

![Screenshot_2](https://user-images.githubusercontent.com/40694158/147073498-20312e42-6c79-4df3-b2fb-16e114cd34df.jpg)

lobontiumira:
Verified as fixed on the Focus Nightly 97.0 (build 353610924 GV 97.0a1-20211226092042) from 12/27 with Google Pixel (Android 10):

![Untitled](https://user-images.githubusercontent.com/40694158/147449359-e95d551a-cf19-480a-8389-41984a4e40ae.png)

PR #6112, opened 2021-12-22
Update rename shortcut dialog colors

ionutbedregeanu:
For #6110


app/src/main/java/org/mozilla/focus/ui/theme/FocusTheme.kt
```diff
@@ -48,6 +48,7 @@ val focusColors: FocusColors
 
 private fun darkColorPalette(): FocusColors = FocusColors(
     material = darkColorsMaterial(),
+    accent = PhotonColors.Pink40,
```
mcarare, line 51: nit: I think we should choose an explicit name and move away from the old XML naming that was a bit confusing. The name should explicitly reflect it is being used for active controls.

mcarare (approved):

Issue #5245, opened 2021-08-27
Adding a page on the home screen with ETP disabled will not open with it off


LaurentiuApahideanSV:
## Steps to reproduce
1. Open any website;
2. Disable Enhanced Tracking Protection;
3. From the 3 dot menu select Add to Home screen;
4. Place the shortcut on the Home screen;
5. Close Focus;
6. Tap on the home screen shortcut from step 4.
### Expected behavior
The webpage should open in Focus with ETP disabled.
### Actual behavior
The webpage opens in Focus with ETP enabled.
### Device information
* Android device: Oppo Find X3 (Android 11), OnePlus 6T (Android 9)
* Focus version: 93.0.0

brampitoyo:
@mcarare Yes. It makes sense to me that website shortcuts should open according to the ETP setting that has been stored in the Exceptions list.

mcarare:
@brampitoyo, then should we remove the warning from the add to homes screen dialog?

brampitoyo:
@mcarare I think the warning in the “Add to home screen” dialog should be kept, just so people don’t accidentally open sites without ETP.

mcarare:
@brampitoyo Thank you for your feedback. Following the comments on this, I am going to close this issue, as there is nothing to fix.

lobontiumira:
I will re-open this issue, since it is not clear to me what is the expected behaviour when adding a page with ETP off to the homescreen.
After discussing with Bram, this is the behaviour that we wanted:
* If you visited a page, disable ETP, and Add to Homescreen – that page should open with ETP off.
* If you visit the same page and re-enable ETP, then erase browsing history – Focus “remembers” the state of the ETP between sessions. When you open it from the homescreen, ETP should be on.

In this case, on 93.0.2, with Lenovo Tab M10 (Android 10), when adding a page with ETP disabled on the homescreen, that page is opened with ETP enabled when accessed through homescreen.
Not reproducible on the latest debug build from main from 9/14.

mcarare:
@lobontiumira The state the page is opened in has nothing to do with the state of the ETP when the page was saved. It only follows the rule from the exception list.

lobontiumira:
I've noticed a strange behavior now on the latest Focus Nightly from 11/22 (build 353260949 GV 96.0a1-20211120215911) with Samsung Galaxy Note 8 (Android 9):
1. Opened a page.
2. Disabled the ETP.
3. Added the page to the device's homescreen.
[...]

PR #6172, opened 2022-01-06
Update the initialisation order of toolbar indicators

ionutbedregeanu:
For #5245
Even if the tracking protection indicator were added at initialisation the security indicator observer were attached after that, fact which leads to tracking protection indicator overwriting
In setTrackingProtectionState method from DisplayToolbar.kt AC component we are checking if our current list of indicators contains a tracking protection indicator, condition which was false since it was replaced by a security one


ionutbedregeanu:
I will close this draft since the main problem it will be fixed in AC

Issue #6171, opened 2022-01-06
The "System default" text is missing from General - Language settings page


lobontiumira:
## Steps to reproduce
1. Install the latest Focus Nightly from 1/6.
2. Go to Settings - General, and observe the Languages option.

### Expected behavior
The "System default" text is missing from under Languages option.
Selecting a different language than the system's, displays the name of the selected languages under "Languages".

### Actual behavior
The "System default" text is displayed under the Languages option.

### Device information
* Android device: Lenovo tab M10 (Android 10), Xiaomi mi4i (Android 5.0.2)
* Not reproducible on Oppo Find X3 Lite (Android 11), nor on Samsung Galaxy Tab A6 (Android 5.1.1)
* Focus version: Focus Nightly 97.0 from 1/6 (build 360061030 GV 97.0a1-20220105093353)
* Regression from #5293 

![missing text jpeg](https://user-images.githubusercontent.com/40694158/148356145-f2d73633-7aca-4ab5-8dd5-c7e0ede69e86.jpg)

iorgamgabriel:
@lobontiumira  can you test please with an older version to see if this is happening . I don't have Lenovo tab M10 (Android 10), Xiaomi mi4i (Android 5.0.2)

lobontiumira:
I did not reproduce this on older versions of Nightly.
Also, cleaning the profile fixes the issue.

iorgamgabriel:
This bug happens only at update from an older version with languages list in a pop-up window to this version. I checked SharedPreferences value for locale if is empty and I set the value to "System default".

lobontiumira:
Verified as fixed on the Focus Nightly 97.0 (build 360070831 GV 97.0a1-20220106155649) from 1/7 with Lenovo Tab M10 (Android 10), and Xiaomi mi4i (Android 5.0.2).

lobontiumira:
Verified as fixed also on Beta 97.0.0-beta.1 with Xiaomi mi4i (Android 5.0.2), and Lenovo Tab M10 (Android 10).

PR #6174, opened 2022-01-06
For #6171 - The "System default" text is missing from General - Language settings page

iorgamgabriel:
This bug happens only at update from an older version with languages list in a pop-up window to this version. I checked SharedPreferences value for locale if  is empty and I set the value to "System default".


Mugurell (approved):
This was tested and multiple devices.
All good as fresh install or as update.

PR #6180, opened 2022-01-07
Switch from manual releases to shipit releases

sarah-clements:
This was all copied over from the Fenix repository, with a few minor changes. What these changes will do:

1) Disables the triggering of a decision task based on a github-release action (via the current, manual way of creating a github release in the UI).
2) In the shipit UI, once Relman creates a release and hits the 'ship' action, an action callback (in taskcluster/focus_android_taskgraph/release_promotion.py below) will be triggered. Similar to Fenix, you'll see a "Ship Focus/Klar" task and that creates the graph of tasks (based on whether its for beta or release).
3) Same as Fenix, these ship release tasks include a version-bump to update version.txt and publishes the apks in the "releases" section of the repository (note: I just copied what is done for Fenix, if you don't want all apks published or you want them renamed differently, let me know).

I've been able to successfully test a [release](https://github.com/mozilla-releng/staging-focus-android/releases) on the staging-focus-android repo by creating a release and triggering the ship action on shipit staging (note: even though the apks have been published to the staging github repo it didn't actually push apks to the google play store).

The release process will be the same as for Fenix, so someone will want to copy over release instructions on the Fenix wiki to here (assuming the android-components part is the same).

Happy to go over this in more detail if people want a deeper understanding of the changes.


sarah-clements:
Note to self: There needs to be a scriptworker prod deploy before these changes go live.

pocmo (approved):
As far as I understand everything ... 👍  from me :)

version.txt
```diff
@@ -0,0 +1 @@
+97.0.0-beta.1
```
jcristau, line 1: We probably shouldn't add this here.

jcristau (approved):
Looks good to me!

version.txt
sarah-clements: Can you clarify what you mean? The version number is needed here for both shipit and the version-bump kind (basically same as what Fenix has).

jcristau: Right but neither of those things happen off the `main` branch, so I thought the file could be created at the branch point.  It's probably fine to have it on main like on fenix, in which case maybe bump to 98 though?

sarah-clements: Probably better to be consistent and mirror Fenix, but yeah, I'll update it to 98 now.

Issue #5743, opened 2021-10-29
Add autoplay settings


mcarare:
Some reviews on the play store are mentioning the lack of setting to disable video/ audio autoplay.

Did we use to have this on the previous versions?

We should consider adding a setting to allow/block autoplay. Maybe not as complex as the one in Fenix, or maybe just disable autoplay by default?

brampitoyo:
@mcarare Focus is a browser where it makes sense to disable autoplay by default. In other words, our default policy should be “Block audio and video”.

ionutbedregeanu:
I took a look in Fenix to see how it was implemented the Autoplay feature and I noticed that we are using SitePermissionsFeature component.
In order to use SitePermissionsFeature in Focus we need to add some additional functionality which is not necessarily helpful because we just want to block everything related to autoplay, without user input

Also from this discussion https://bugzilla.mozilla.org/show_bug.cgi?id=1577596#c18 seems that it doesn't exists a shortcut to set directly in GV the state of Autoplay

@Amejia481 Maybe you have more context about this. 
Should I add a new component in AC in order to be able to block Autoplay directly or should I use SitePermissionsFeature as in Fenix?

Amejia481:
👋🏽 @ionutbedregeanu 

In case, you want to give a global value for a permission you could use `SitePermissionsFeature.sitePermissionsRules`,
with it you can control how a permission should behave globally by default behavior on each page:

For example:
 
If you want to **block** `autoplayAudible` and **allow** `autoplayInaudible` (This is the recommended default value as blocking both could cause some pages to not work properly for more details see https://github.com/mozilla-mobile/fenix/issues/16557#issuecomment-729820241).
 
```kotlin
val sitePermissionsRules = SitePermissionsRules(
            notification = Action.ASK_TO_ALLOW,
            microphone = Action.ASK_TO_ALLOW,
            location = Action.ASK_TO_ALLOW,
            camera = Action.ASK_TO_ALLOW,
            autoplayAudible = AutoplayAction.BLOCKED,
            autoplayInaudible = AutoplayAction.ALLOWED,
            persistentStorage = Action.ASK_TO_ALLOW,
            mediaKeySystemAccess = Action.ASK_TO_ALLOW
)

SitePermissionsFeature(
  ..., // Other properties
  sitePermissionsRules = sitePermissionsRules
  , ... // Other properties
)
```
[...]

PR #6191, opened 2022-01-10
Reload the session with the new autoplay rules

ionutbedregeanu:
For #5743
Until now the the session was redundant updated with the old values and this commit add a new flag "reloadSession" to update the session when the autoplay option is changed

https://user-images.githubusercontent.com/11731652/148803226-3308f565-4a7e-485e-ac9d-f009f200cc47.mp4

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
@@ -332,6 +332,10 @@ class BrowserFragment :
             owner = this,
             view = rootView
         )
+        if (requireComponents.appStore.state.autoplayRulesChanged) {
+            requireComponents.sessionUseCases.reload(tabId)
+            requireComponents.appStore.dispatch(AppAction.AutoplayChange(false))
+        }
     }
 
     override fun onAccessibilityStateChanged(enabled: Boolean) = when (enabled) {
--- a/app/src/main/java/org/mozilla/focus/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/focus/utils/Settings.kt
@@ -14,7 +14,6 @@ import mozilla.components.concept.engine.Engine
 import mozilla.components.concept.engine.EngineSession
 import mozilla.components.feature.sitepermissions.SitePermissionsRules
 import org.mozilla.focus.R
-import org.mozilla.focus.ext.components
 import org.mozilla.focus.fragment.FirstrunFragment
 import org.mozilla.focus.searchsuggestions.SearchSuggestionsPreferences
 import org.mozilla.focus.settings.permissions.AutoplayOption
@@ -195,7 +194,6 @@ class Settings(
         preferences.edit()
             .putString(getPreferenceKey(R.string.pref_key_autoplay), prefKey)
             .apply()
-        context.components.sessionUseCases.reload.invoke(context.components.store.state.selectedTabId)
         currentAutoplayOption = getValueByPrefKey(autoplayPrefKey = prefKey, context = context)
     }
 
--- a/app/src/main/java/org/mozilla/focus/settings/permissions/AutoplayFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/settings/permissions/AutoplayFragment.kt
@@ -11,6 +11,7 @@ import org.mozilla.focus.ext.requirePreference
 import org.mozilla.focus.settings.BaseSettingsFragment
 import org.mozilla.focus.settings.RadioButtonPreference
 import org.mozilla.focus.settings.addToRadioGroup
+import org.mozilla.focus.state.AppAction
 
 class AutoplayFragment : BaseSettingsFragment() {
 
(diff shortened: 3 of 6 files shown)


app/src/main/java/org/mozilla/focus/utils/Settings.kt
```diff
@@ -195,11 +195,18 @@ class Settings(
         preferences.edit()
             .putString(getPreferenceKey(R.string.pref_key_autoplay), prefKey)
             .apply()
-        context.components.sessionUseCases.reload.invoke(context.components.store.state.selectedTabId)
-        currentAutoplayOption = getValueByPrefKey(autoplayPrefKey = prefKey, context = context)
+        autoplayState = autoplayState.copy(
+            option = getValueByPrefKey(autoplayPrefKey = prefKey, context = context),
+            reloadSession = true
+        )
```
Mugurell, line 201: This is an interesting approach but seems like a long departure from our arch direction of updating the UI and passing state through the MVI stores.
Wondering if we could rather have this state best uplifted in BrowserState.

app/src/main/java/org/mozilla/focus/utils/Settings.kt
```diff
@@ -195,11 +195,18 @@ class Settings(
         preferences.edit()
             .putString(getPreferenceKey(R.string.pref_key_autoplay), prefKey)
             .apply()
-        context.components.sessionUseCases.reload.invoke(context.components.store.state.selectedTabId)
-        currentAutoplayOption = getValueByPrefKey(autoplayPrefKey = prefKey, context = context)
+        autoplayState = autoplayState.copy(
+            option = getValueByPrefKey(autoplayPrefKey = prefKey, context = context),
+            reloadSession = true
```
Mugurell, line 200: This seems like misplaced responsibility.
If this is just "state" it should not dictate how the app will function - "reload session".
I understand it serve the purpose of just informing when the value has been "updated" so maybe it can be named as such?

app/src/main/java/org/mozilla/focus/settings/permissions/AutoplayState.kt
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package org.mozilla.focus.settings.permissions
```
Mugurell, line 4: nit: newline

app/src/main/java/org/mozilla/focus/settings/permissions/AutoplayState.kt
```diff
@@ -0,0 +1,9 @@
...
+
+data class AutoplayState(
```
Mugurell, line 6: This needs documentation.
Though using it like it is now seems like a different approach than of our MVI direction.
I would be curious of other approaches to this.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -332,6 +332,9 @@ class BrowserFragment :
             owner = this,
             view = rootView
         )
+        if (requireComponents.settings.autoplayState.reloadSession) {
+            requireComponents.sessionUseCases.reload(tabId)
+        }
```
Mugurell, line 337: Don't see where this `reloadSession` is reset to `false`.
Would it mean that once set to `true` everytime the user navigates back to this Fragment the session would be reloaded?

Mugurell (changes requested):
The new `AutoplayState` and the way it is used seems like a big departure from the current arch direction.
Seems like we need an in-memory state containing the value of the autoplay permission.
Until having a proper `BrowserScreenStore` I propose having just a `autoplayStatus` property inside `BrowserFragment`, checked, updated and acted upon in `onCreateView`. This seems like a smaller change, easier to reason and refactor later.
Or maybe use `AppState` which seems like a good fit for the autoplay status since that is read/updated from multiple places in the app.

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -332,6 +332,10 @@ class BrowserFragment :
             owner = this,
             view = rootView
         )
+        if (requireComponents.appStore.state.autoplayRulesChanged) {
```
Mugurell, line 335: Would've been nice to observe the store and do a direct mapping of the state instead of treating the AppStore as just a singleton containing a value but we can refactor later to this flow and with Compose it will just work.
Tested it and solves the issue. Thanks!

Mugurell (approved):

ionutbedregeanu:
Restart this pr since we are receiving this error Could not GET 'https://jcenter.bintray.com/org/gradle/kotlin/gradle-kotlin-dsl-plugins/2.1.4/gradle-kotlin-dsl-plugins-2.1.4.module'.

PR #6196, opened 2022-01-11
Set version to 97.0.0

ionutbedregeanu:


rvandermeulen:
@ionutbedregeanu Why is versionName being bumped to 98.0? That should happen on main AFAIK.

rvandermeulen (approved):

ionutbedregeanu:
> @ionutbedregeanu Why is versionName being bumped to 98.0? That should happen on main AFAIK.

My mistake. Solved. Thanks

Issue #6170, opened 2022-01-06
Tapping on "Languages" displays a glitching Language page


lobontiumira:
## Steps to reproduce
1. Install the latest Focus Nightly from 1/6.
2. Go to Settings - General, and tap on Languages.
3. Observe the transition from General page to Languages page.

### Expected behavior
The Languages page is displayed without visual glitches.

### Actual behavior
The Languages page is displayed with a visual glitch - please see the video below.

### Device information
* Android device: Samsung Galaxy Tab A6 (Android 5.1.1), Lenovo tab M10 (Android 10), Xiaomi mi4i (Android 5.0.2)
* Focus version: 1/6 Focus Nightly
* Regression from #5293 

https://user-images.githubusercontent.com/40694158/148355367-e927acd0-0264-48bf-ad48-1776e9044726.mp4

iorgamgabriel:
I can't do nothing about this glitch . It takes a little to show all the languages . I think this is expected .

lobontiumira:
@iorgamgabriel isn't it possible to make the transition as smooth as in Fenix - please see the short video below:


https://user-images.githubusercontent.com/40694158/148507130-f28ad2b6-9139-47e3-908e-e136595e2794.mp4

Mugurell:
Drive-by comment: Fenix has the same general behavior and shows the same issue - for a split second the screen will be empty

<img width="428" alt="image" src="https://user-images.githubusercontent.com/11428869/148508728-25a585e8-fb7f-480a-82f9-4ecd9191c209.png">

though indeed at a quick glance it seems that on Fenix the delay to display languages is smaller.

The difference may be because of the way the locale list is initialized:
- on Fenix Fenix being more direct - [[onCreateView -> createInitialLocaleSettingsState](https://github.com/mozilla-mobile/fenix/blob/46b5dd27065a50643dd96145ae39131014d45a45/app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleSettingsFragment.kt#L52)] -> [[onViewCreated -> "adapter".update(...)](https://github.com/mozilla-mobile/fenix/blob/46b5dd27065a50643dd96145ae39131014d45a45/app/src/main/java/org/mozilla/fenix/settings/advanced/LocaleSettingsFragment.kt#L100)]
[...]

PR #6214, opened 2022-01-13
For #6170 - Tapping on "Languages" displays a glitching Language page

iorgamgabriel:
Fix by adding additionalTop to contentPadding TopAppBar .


app/src/main/java/org/mozilla/focus/settings/BaseSettingsComposeFragment.kt
```diff
@@ -12,65 +16,78 @@ import androidx.compose.material.Text
...
+import org.mozilla.focus.utils.StatusBarUtils
 
-open class BaseSettingsComposeFragment : Fragment() {
+/**
+ * BaseSettingsComposeFragment displays a toolbar in compose for the screen
+ */
+abstract class BaseSettingsComposeFragment : Fragment() {
+    /**
+     * title of the screen.
+     */
```
Mugurell, line 39:
```suggestion
    /**
     * Screen title shown in toolbar. 
     */
```

app/src/main/java/org/mozilla/focus/settings/BaseSettingsComposeFragment.kt
```diff
@@ -12,65 +16,78 @@ import androidx.compose.material.Text
...
+    abstract fun title(): String
+
+    /**
+     * onClickUpButton callback for the back arrow of the toolbar
+     */
```
Mugurell, line 44:
```suggestion
    /**
     * Callback for the up navigation button shown in toolbar.
     */
```

app/src/main/java/org/mozilla/focus/settings/BaseSettingsComposeFragment.kt
```diff
@@ -12,65 +16,78 @@ import androidx.compose.material.Text
...
-                                ),
-                                navigationIcon = {
-                                    IconButton(
-                                        onClick = onClickUpButton
-                                    ) {
-                                        Icon(
-                                            Icons.Filled.ArrowBack,
-                                            stringResource(R.string.go_back),
-                                            tint = focusColors.toolbarColor
+    abstract fun ContentScreen()
```
Mugurell, line 51: nit: Maybe just `Content`?

Mugurell (commented):
Tested this and it works beautifully!
Thanks to @mcarare who suggested this approach also.
I suggested some documentation improvements that I think will help the abstract class for the future but otherwise LGTM!

Curios if Mihai has any other input.

app/src/main/java/org/mozilla/focus/settings/BaseSettingsComposeFragment.kt
iorgamgabriel: Doesn't work with Content() because Content method from Compose Class will get called and the app will crash

app/src/main/java/org/mozilla/focus/locale/screen/LanguageFragment.kt
```diff
@@ -5,71 +5,68 @@
...
 import org.mozilla.focus.state.AppAction
 import org.mozilla.focus.state.Screen
 
-class LanguageFragment : BaseSettingsComposeFragment() {
-    private val browserStore by lazy { requireContext().components.store }
-    private val localeUseCase by lazy { LocaleUseCases(browserStore) }
-    private val languageScreenStore by lazy {
-        LanguageScreenStore(
+class LanguageFragment : BaseComposeFragment() {
+
```
mcarare, line 27: nit: I do not think the space is necessary and we d not have it in other fragments.

app/src/main/java/org/mozilla/focus/locale/screen/LanguageFragment.kt
```diff
@@ -5,71 +5,68 @@
...
+    private lateinit var browserStore: BrowserStore
+    private lateinit var localeUseCase: LocaleUseCases
+    private lateinit var languageScreenStore: LanguageScreenStore
+    private lateinit var screenInteractorDefault: DefaultLanguageScreenInteractor
```
mcarare, line 31: nit: I think a name that follows the class name and thus is directly identifiable with the type is preferable.

app/src/main/java/org/mozilla/focus/settings/BaseSettingsComposeFragment.kt
mcarare: I think the pattern in the API is to have the content as a composable function with a return type. 
So we could name it `content()` with a  `() -> Unit` return type.

app/src/main/java/org/mozilla/focus/locale/screen/LanguageFragment.kt
```diff
@@ -5,71 +5,68 @@
...
-                        .observeAsComposableState { state -> state.languageList }.value
-                    val languageSelected = languageScreenStore.observeAsComposableState { state ->
-                        state.selectedLanguage
-                    }.value
-                    if (languageSelected != null && languagesList != null) {
-                        Languages(languageSelected = languageSelected, languages = languagesList)
-                    }
-                }
-            }
+    override val title: String
```
mcarare, line 50: nit: This could be a `stringRes` and it would make the getter obsolete, but I am fine with either.

Mugurell: Passing just the Int for `titleRes` sounds great!

app/src/main/java/org/mozilla/focus/settings/BaseSettingsComposeFragment.kt
mcarare: I had a comment that went missing 😮 .
You can either: 
1. keep the `Content` naming and use `this@BaseComposeFragment.Content()` to avoid calling the `ComposeView` method (that is happening because the calling place is in the `ComposeView` scope)
2. Use the API pattern to have it as a `val` with a `@Composable ()-> Type` that you can simply name `content`. (This will involve changing the way you overrode the content.

mcarare (approved):
LGTM. 
As possible future improvements (not in the scope of this PR) we might consider improving animation when a locale is selected.

Issue #6223, opened 2022-01-13
Target SDK 31 in Focus


jonalmeida:
We should align with Fenix and target SDK 31 in general, but also because there is a desire to upgrade to a newer WorkManager version to fix a bug in Fenix.

See: https://github.com/mozilla-mobile/android-components/pull/11541

PR #6225, opened 2022-01-14
Update the target sdk to 31

ionutbedregeanu:
For #6223
Update the target sdk to 31
Starting with Android 12 if the app component includes the LAUNCHER category android:exported should
be set to true
Also IntentReceiverActivity is marked as exported, in this way we can continue using the action.View

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -37,7 +37,9 @@
         android:name=".FocusApplication"
         android:usesCleartextTraffic="true">
 
-        <activity android:name=".activity.IntentReceiverActivity">
+        <activity
+            android:name=".activity.IntentReceiverActivity"
+            android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
                 <category android:name="android.intent.category.DEFAULT" />
@@ -71,6 +73,7 @@
 
         <activity android:name=".activity.MainActivity"
             android:launchMode="singleTask"
+            android:exported="true"
             android:windowSoftInputMode="adjustResize|stateAlwaysHidden"
             android:configChanges="orientation|screenSize|screenLayout"
             android:label="@string/app_name">
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -11,7 +11,7 @@ import com.android.build.OutputFile
 import groovy.json.JsonOutput
 
 android {
-    compileSdkVersion 30
+    compileSdkVersion 31
 
     if (project.hasProperty("testBuildType")) {
         // Allowing to configure the test build type via command line flag (./gradlew -PtestBuildType=beta ..)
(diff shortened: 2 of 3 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -37,7 +37,9 @@
         android:name=".FocusApplication"
         android:usesCleartextTraffic="true">
 
-        <activity android:name=".activity.IntentReceiverActivity">
+        <activity
+            android:name=".activity.IntentReceiverActivity"
+            android:exported="true">
```
Mugurell, line 42: Why is this exported?
Seems like an important change that should be better documented - in the ticket and in commit message.

app/src/main/AndroidManifest.xml
```diff
@@ -71,6 +73,7 @@
 
         <activity android:name=".activity.MainActivity"
             android:launchMode="singleTask"
+            android:exported="true"
```
Mugurell, line 76: Why is this exported?
Seems like an important change that should be better documented - in the ticket and in commit message.

ionutbedregeanu: If the app component includes the LAUNCHER category, set android:exported to true.
https://developer.android.com/about/versions/12/behavior-changes-12

app/src/main/AndroidManifest.xml
ionutbedregeanu: To keep the action  <action android:name="android.intent.action.VIEW" />

Mugurell (approved):

app/src/main/AndroidManifest.xml
jonalmeida: I don't think `MainActivity` should be exported.

In Fenix, the equivalent is `HomeActivity` and it doesn't seem to need it: https://github.com/mozilla-mobile/fenix/blob/main/app/src/main/AndroidManifest.xml#L44-L56

If you look directly at the activity, instead of the alias too, it is not an exported activity. 

From the docs:
> If the app component includes the LAUNCHER category, set android:exported to true. **In most other cases, set android:exported to false.**

The idea is that we only have one entry point into the app that we can control the flow of intents from i.e. the IntentReceiverActivity, which will then tell the appropriate activity to start up.

jonalmeida: @ionutbedregeanu ^

jonalmeida:
Thank you for doing this work so quickly! ❤️

app/src/main/AndroidManifest.xml
ionutbedregeanu: If we are not setting a value for `android:exported` for `MainActivity` the app doesn't compile and if `android:exported` is set to false we can't launch the app
I'm not sure why in Fenix is working, an investigation should be done

Issue #6226, opened 2022-01-14
Address AC update conflicts


ionutbedregeanu:
I tried to update the Android Components version to 98.0.20220114002255 but it seems that OnLifecycleEvent is [deprecated now](https://developer.android.com/jetpack/androidx/releases/lifecycle#2.4.0-beta01) and we can't push the update
To fix this we need to use LifecycleEventObserver or DefaultLifecycleObserver where we are using OnLifecycleEvent
<img src="https://user-images.githubusercontent.com/11731652/149527692-b4f5d0e8-acc7-4d13-a947-72b460a58271.png" width="100%"/>

ionutbedregeanu:
Note for QA: Please test if the fingerprint unlock works as expected

lobontiumira:
Verified on Google Pixel (Android 10), Oppo Find X3 (Android 11), and Google Pixel 6 (Android 12) on Focus nightly from 1/19, and 1/20, and here are our findings:
- the fingerprint unlock is requested when the "Use fingerprint to unlock app" is enabled,'
- the fingerprint is requested only if a tab is opened in Focus,
- if there are no tabs openes, the fingerprint is not requested (even if a shortcut is added on the Focus homescreen),
- the fingerprint is requested when a page is added to the device's homescreen,
- the autocomplete works as expected.

PR #6228, opened 2022-01-14
Address AC update conflicts

ionutbedregeanu:
For #6226
Update to Android-Components 98.0.20220114002255
OnLifecycleEvent is deprecated - to fix this we can use LifecycleEventObserver or DefaultLifecycleObserver
Now instead of using '@OnLifecycleEvent(Lifecycle.Event.ON_PAUSE)' LockObserver.kt extends DefaultLifecycleObserver and override onPause method


Amejia481 (approved):

Mugurell (approved):

Mugurell (approved):
Thanks for the quick fix!

jonalmeida:
Thank you!

PR #6260, opened 2022-01-20
Restore old telemetry.

mcarare:
Notes:

1. This restores the old telemetry in the same state that it was when the migration to Glean process began. Meaning, if the previous implementation was not proper, this just restores it. Some of the obvious errors were fixed while restoring.
2. The tips telemetry has not been restored, the previous one even had a wrong trigger.
3. Some of the context menu/browser telemetry was not recorded after the migration to AC. I reintroduced the ones that are still valid.
4. Some telemetry was obsolete even before the migration, for example, switch to gecko view from the web view. This obsolete telemetry was not reinstated. 

### Pull Request checklist

3. click the `Artifacts` row,

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
@@ -80,6 +80,7 @@ import org.mozilla.focus.session.ui.TabsPopup
 import org.mozilla.focus.settings.privacy.ConnectionDetailsPanel
 import org.mozilla.focus.settings.privacy.TrackingProtectionPanel
 import org.mozilla.focus.state.AppAction
+import org.mozilla.focus.telemetry.TelemetryWrapper
 import org.mozilla.focus.topsites.DefaultTopSitesStorage.Companion.TOP_SITES_MAX_LIMIT
 import org.mozilla.focus.topsites.DefaultTopSitesView
 import org.mozilla.focus.utils.FocusSnackbar
@@ -506,6 +507,9 @@ class BrowserFragment :
 
             DownloadState.Status.COMPLETED -> {
                 Downloads.downloadCompleted.record(NoExtras())
+
+                TelemetryWrapper.downloadDialogDownloadEvent(sentToDownload = true)
+
                 showDownloadCompletedSnackbar(state, extension)
             }
 
@@ -610,6 +614,7 @@ class BrowserFragment :
                 Browser.backButtonPressed.record(
                     Browser.BackButtonPressedExtra("erase_to_external_app")
                 )
+                TelemetryWrapper.eraseBackToAppEvent()
 
                 // This session has been started from a VIEW intent. Go back to the previous app
                 // immediately and erase the current browsing session.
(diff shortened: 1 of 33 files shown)


ionutbedregeanu (approved):

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -506,6 +507,9 @@ class BrowserFragment :
 
             DownloadState.Status.COMPLETED -> {
                 Downloads.downloadCompleted.record(NoExtras())
+
+                TelemetryWrapper.downloadDialogDownloadEvent(sentToDownload = true)
```
jonalmeida, line 511: self note: we always send true only here?

app/src/main/java/org/mozilla/focus/fragment/CrashReporterFragment.kt
```diff
@@ -24,10 +25,14 @@ class CrashReporterFragment : Fragment(R.layout.fragment_crash_reporter) {
 
         CrashReporter.displayed.record(NoExtras())
 
+        TelemetryWrapper.crashReporterOpened()
+
         binding.closeTabButton.setOnClickListener {
             val wantsSubmitCrashReport = binding.sendCrashCheckbox.isChecked
             CrashReporter.closeReport.record(CrashReporter.CloseReportExtra(wantsSubmitCrashReport))
 
+            TelemetryWrapper.closeTabButtonTapped(wantsSubmitCrashReport)
```
jonalmeida, line 34: self note: correct event for _crash reporter_ close?

app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.kt
```diff
@@ -62,6 +64,8 @@ class FirstrunFragment : Fragment(), View.OnClickListener {
             R.id.skip -> {
                 finishFirstrun()
                 Onboarding.skipButtonTapped.record(Onboarding.SkipButtonTappedExtra(currentItem))
+                TelemetryWrapper.skipFirstRunEvent()
+                TelemetryWrapper.finishFirstRunEvent()
```
jonalmeida, line 68: Shouldn't finish be below for `R.id.finish`?

jonalmeida (approved):
I left two self-notes for myself to check some events out that I can't recall and one question.

If this is blocking a release or experiment, feel free to land it and I'll get back to my notes later as a follow-up.

app/src/main/java/org/mozilla/focus/fragment/FirstrunFragment.kt
mcarare: Good catch! Thank you!

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
mcarare: Yes, we send `false` when the download is canceled: https://github.com/mozilla-mobile/focus-android/blob/b34f1ebf56eb6dbb9f1ef214a994c7fca87fc72d/app/src/main/java/org/mozilla/focus/telemetry/TelemetryMiddleware.kt#L61-L66

Issue #6332, opened 2022-02-02
The CFR for Shield Icon is incorrectly placed


delia-pop:
## Steps to reproduce
1. Install Firefox Focus Nightly.
2. Open a webpage and observe the CFR prompt.
3. Clear data.
4. Go to Settings - Mozilla  - Your Rights and observe the CFR prompt.
### Expected behavior
After step 2, the CFR prompt is displayed under the shield icon.
After step 4, the CFR prompt is displayed under the shield icon.

### Actual behavior
After step 2, the CFR prompt is firstly incorrectly placed in the left upper corner above the trash icon, and then jumps under the shield icon.
After step 4, the CFR prompt is incorrectly placed and remains displayed in the left upper corner above the trash icon.
### Device information

* Android device: Google Pixel 6 (Android 12), Motorola Moto G30 (Android 11), Samsung Galaxy Tab S3 (Android 9), Huawei MediaPad M2 (Android 5.1.1).

* Focus version: Focus Nightly 98.0 (Build 360330514 🦎 98.0a1-20220201093942).

https://user-images.githubusercontent.com/89388888/152144873-7fc6dd06-35e9-4fec-990f-06d8a85a7fc8.mp4

https://user-images.githubusercontent.com/89388888/152145100-f64144e9-2832-4848-aaff-999c9a87ebdf.mp4

https://user-images.githubusercontent.com/89388888/152145386-58186e25-b20b-4f8d-904d-b92b919feddf.mp4

lobontiumira:
This issue is not fixed yet.
The CFR is positioned incorrectly on Samsung Galaxy Tab A6 (Android 5.1.1), and on Xiaomi mi4i (Android 5.0.2).
Tested with the latest Focus Nightly build 98.0 (360390513 with GV 98.0a1-20220207065816) from 2/8.

![Screenshot_1](https://user-images.githubusercontent.com/40694158/152959893-34067e5c-47b1-4307-91e2-deb8241f66d0.jpg)

I'll remove the QAReady label for now.

Mugurell:
@iorgamgabriel @ionutbedregeanu On Android <= 23 we might have to use `showAtAnchorLocation` instead of `showAsDropdown` when showing the PopupWindow. The solution needs to be validated for when rotating the screen.

Mugurell:
@ QA: This should be addressed by https://github.com/mozilla-mobile/focus-android/pull/6411
Please recheck.

lobontiumira:
[...]

PR #6344, opened 2022-02-03
For #6332 - Properly anchor the Tracking Protection CFR

iorgamgabriel:
Add cfrForShieldToolbarIconListener in BrowserToolbarIntegration and show cfr only after Shield Icon is visible.
 Add toolbarShieldIconCfrPopupWindow?.setOnDismissListener(null) in onDestroyView . 
 Rename some files .

### Pull Request checklist

3. click the `Artifacts` row,

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
@@ -47,6 +47,7 @@ class BrowserToolbarIntegration(
     private val onUrlLongClicked: () -> Boolean,
     private val eraseActionListener: () -> Unit,
     private val tabCounterListener: () -> Unit,
+    private val onTrackingProtectionShown: () -> Unit,
     private val customTabId: String? = null,
     inTesting: Boolean = false
 ) : LifecycleAwareFeature {
@@ -216,6 +217,7 @@ class BrowserToolbarIntegration(
                     val url = it.content.url
                     if (secure && Indicators.SECURITY in toolbar.display.indicators) {
                         addTrackingProtectionIndicator()
+                        onTrackingProtectionShown()
                     } else if (!secure && Indicators.SECURITY !in toolbar.display.indicators &&
                         !url.trim().startsWith("about:")
                     ) {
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -47,6 +47,7 @@ class BrowserToolbarIntegration(
     private val onUrlLongClicked: () -> Boolean,
     private val eraseActionListener: () -> Unit,
     private val tabCounterListener: () -> Unit,
+    private val cfrForShieldToolbarIconListener: () -> Unit,
```
Mugurell, line 50: Nice separations of concerns and control flow here with using a callback instead of calling methods directly on the fragment which contains this integration.
But the naming is still leaking the expected behavior which should not be known to this class.
Maybe just `onTrackingProtectionShown` and then let the layer above - the fragment decide if this is a good indicator to do some random action like maybe showing a popup?

app/src/main/java/org/mozilla/focus/utils/CfrUtils.kt
```diff
@@ -26,14 +26,15 @@ class CfrUtils {
...
+         * Shows CFR for Shield Toolbar Icon  if content is secure.
+         * This should be shown only when Shield Icon is visible.
          *
          * @property rootView of Browser Toolbar.
          * @property context where CFR should be shown.
          * @property isContentSecure true if the tab is currently pointed to a URL
          * with a valid SSL certificate, otherwise false.
          */
-        fun shouldShowCFRForShieldToolbarIcon(
+        fun showCFRForShieldToolbarIconIfNeeded(
```
Mugurell, line 37: 👍
Now this does express an action instead of a boolean.

app/src/main/java/org/mozilla/focus/utils/Settings.kt
```diff
@@ -176,7 +176,7 @@ class Settings(
                 .commit()
         }
 
-    var isCfrForForShieldToolbarIconVisible: Boolean
+    var shouldShowCfrForShieldToolbarIcon: Boolean
```
Mugurell, line 179: 👍
This seems to express more clearly what the value represents and where it can be used.

Mugurell (approved):
LGTM.
Like the cleanup!
One more suggestion about the callback naming in `BrowserToolbarIntegration` and one nit to maybe update the title of the commit/PR to better express the result of your changes - right now it seems that after this patch the CFR is incorrectly placed.

PR #6348, opened 2022-02-03
For #6347 - Upgrade Espresso libraries for API 30

AaronMT:
- Upgrades all Espresso libraries to support API 30
- Adjusts some tests to run on API 30
- Updates Flank configuration to use an API 30 device

Note: Currently `ActivityTestRule` is deprecated. https://developer.android.com/reference/android/support/test/rule/ActivityTestRule.html

Note: Switching to `ActivityScenario` or `ActivityScenarioRule` has some caveats. This initial work will `@supress` `ActivityTestRule`.

Opening as draft to see how Firebase runs look like.


AaronMT:
@isabelrios @sv-ohorvath I think this is ready for review. We can treat any other flaky tests as any other flaky afterwards. Similar patches to expected on Fenix next.

app/src/androidTest/java/org/mozilla/focus/activity/DownloadFileTest.kt
```diff
@@ -104,7 +108,9 @@ class DownloadFileTest {
             clickDownloadButton()
             verifySnackBarText("finished")
             mDevice.openNotification()
-            verifyDownloadNotification()
+            notificationTray {
+                expandEraseBrowsingNotification()
```
sv-ohorvath, line 112: I think this should be verifySystemNotificationExists, as the expandEraseBrowsingNotification does something else.

app/src/androidTest/java/org/mozilla/focus/activity/robots/NotificationRobot.kt
```diff
@@ -31,39 +33,20 @@ class NotificationRobot {
...
+        val notificationInTray = mDevice.wait(
+            Until.hasObject(
+                By.res(NOTIFICATION_SHADE).hasDescendant(
+                    By.text(notificationMessage)
+                )
+            ),
+            waitingTime
+        )
+
+        assertTrue(notificationInTray)
```
sv-ohorvath, line 49: I'm still a little concerned that the notification won't always be visible and will need a bit of scrolling to reach it because that's how it started being very flaky in the past, that's why we had this complicated hack. But as we saw, on this new API they behave differently, so let's hope it won't be the case, or we'll fix them as they come.

sv-ohorvath (changes requested):
Just one change needed, otherwise lgtm.

AaronMT:
> Just one change needed, otherwise lgtm.

Disabling this unreliable download notification test. Scrolling to the silent notifications on Firebase is unreliable.

sv-ohorvath:
@AaronMT mind trying out one of the hack methods I wrote for that download notification test? https://github.com/mozilla-mobile/focus-android/commit/695b8b30fc7187dd28aba48471f0301831026e52#diff-11ab976f957c6a5203d0d6488bbe2628cd36cb8e6f5655f7a4d784069c5f8d8cL60 
It seemed to work fine on my branch.

sv-ohorvath (approved):
Thanks for fixing the tests

isabelrios (approved):
LGTM, great work with this change! 👏

Issue #6211, opened 2022-01-12
Tool Tip for: Privacy and Security Settings page


amedyne:
Display Inline tool tip on the Privacy and Security Settings page.
<img width="318" alt="Screen Shot 2022-01-12 at 1 35 39 PM" src="https://user-images.githubusercontent.com/68711330/149201389-e6b5d804-af41-4464-a465-7620e1de24cb.png">

Figma design [here](https://www.figma.com/file/ahqwL3sByHprH2Rru9R58b/Onboarding-2022?node-id=2522%3A7249).
Content/Strings is still being finalized.

@jeffreygee please add comment or update once behavior on the following is finalized:

- [ ] condition/trigger when the CFR/tool tip appears
- [ ] behavior to dismiss the CFR/tool tip (click on X or click anywhere on the screen)

jeffreygee:
**Trigger logic**: This will be an inline CFR that is triggered by a new updated version. The user will see it within the Privacy & Security window above the configurable settings. 
**Behavior to dismiss**: CFR can be dismissed by tapping a ‘close’ icon in the opposite corner of the pointer, or anywhere outside of the CFR. Once dismissed, will not be shown again.

PR #6351, opened 2022-02-04
For #6211 - Tool Tip for: Privacy and Security Settings page

iorgamgabriel:
This pull request contains the Tool Tip for Privacy and Security Settings page and  Feature Flags for:
 1.Tool Tip for: Privacy and Security Settings page
 2.Home page tips
 3.CFR for Shield toolbar icon

### Pull Request checklist

3. click the `Artifacts` row,

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
@@ -353,7 +353,11 @@ class UrlInputFragment :
 
         binding.browserToolbar.editMode()
         setHomeMenu()
-        updateTipsLabel()
+        if (Features.SHOULD_SHOW_HOME_PAGE_PRO_TIPS) {
+            updateTipsLabel()
+        } else {
+            binding.homeTips.visibility = View.GONE
+        }
     }
 
     private fun setHomeMenu() {
--- a/app/src/main/java/org/mozilla/focus/settings/privacy/PreferenceToolTipCompose.kt
+++ b/app/src/main/java/org/mozilla/focus/settings/privacy/PreferenceToolTipCompose.kt
@@ -1,0 +1,137 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.focus.settings.privacy
+
+import android.content.Context
+import android.util.AttributeSet
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.wrapContentHeight
+import androidx.compose.foundation.layout.wrapContentSize
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.Icon
+import androidx.compose.material.IconButton
+import androidx.compose.material.Text
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Close
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.res.colorResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import androidx.preference.Preference
+import androidx.preference.PreferenceViewHolder
+import org.mozilla.focus.R
+import org.mozilla.focus.databinding.FocusPreferenceComposeLayoutBinding
+import org.mozilla.focus.ext.settings
+import org.mozilla.focus.ui.theme.FocusTheme
+import org.mozilla.focus.ui.theme.focusColors
+
+class PreferenceToolTipCompose(context: Context, attrs: AttributeSet?) :
+    Preference(context, attrs) {
+
+    override fun onBindViewHolder(holder: PreferenceViewHolder) {
+        super.onBindViewHolder(holder)
+        val binding = FocusPreferenceComposeLayoutBinding.bind(holder.itemView)
+        binding.toolTipContent.setContent {
+            ToolTipContent(
+                onDismissButton = {
+                    context.settings.shouldShowPrivacySecuritySettingsToolTip = false
+                    preferenceManager.preferenceScreen.removePreference(this)
+                }
+            )
+        }
+    }
+}
+
+@Composable
+@Preview
+private fun ToolTipContentPreview() {
+    ToolTipContent {
+    }
+}
+
+/**
+ * Displays a tool tip for Privacy Security Settings screen.
+ *
+ * @param onDismissButton Callback when the "x" Button from tool tip is pressed.
+ */
+@Composable
+fun ToolTipContent(onDismissButton: () -> Unit) {
+    FocusTheme {
+        Column(
+            modifier = Modifier
+                .fillMaxWidth()
+                .wrapContentSize(Alignment.Center)
+        ) {
+            Box(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .wrapContentHeight()
+                    .padding(16.dp)
+                    .clip(RoundedCornerShape(10.dp))
+                    .background(
+                        shape = RoundedCornerShape(10.dp),
+                        brush = Brush.linearGradient(
+                            colors = listOf(
+                                colorResource(R.color.cfr_pop_up_shape_end_color),
+                                colorResource(R.color.cfr_pop_up_shape_start_color)
+                            ),
+                            end = Offset(0f, Float.POSITIVE_INFINITY),
+                            start = Offset(Float.POSITIVE_INFINITY, 0f)
+                        )
+                    ),
+                contentAlignment = Alignment.CenterStart
+            ) {
+                Column(
+                    modifier = Modifier
+                        .wrapContentHeight()
+                        .padding(16.dp),
+                    verticalArrangement = Arrangement.spacedBy(
+                        10.dp
+                    )
+                ) {
+                    Text(
+                        text = stringResource(R.string.tool_tip_title),
+                        fontSize = 16.sp,
+                        letterSpacing = 0.5.sp,
+                        color = focusColors.privacySecuritySettingsToolTip,
+                        fontWeight = FontWeight.Bold,
+                        lineHeight = 20.sp
+                    )
+                    Text(
+                        text = stringResource(R.string.tool_tip_message),
+                        fontSize = 14.sp,
+                        color = focusColors.privacySecuritySettingsToolTip,
+                        lineHeight = 20.sp
+                    )
+                }
+                IconButton(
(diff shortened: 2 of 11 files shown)


iorgamgabriel:
Screenshot_20220204-124505

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -727,6 +731,8 @@ class UrlInputFragment :
     }
 
     override fun onSharedPreferenceChanged(sharedPreferences: SharedPreferences, key: String) {
-        if (key == activity?.getString(R.string.pref_key_homescreen_tips)) { updateTipsLabel() }
+        if (key == activity?.getString(R.string.pref_key_homescreen_tips)) {
```
mcarare, line 734: It would be preferable that the changes not related to this PR, especially indentation or code rearrangements are not included here.

app/src/main/java/org/mozilla/focus/settings/privacy/PrivacySecuritySettingsFragment.kt
```diff
@@ -193,11 +201,10 @@ class PrivacySecuritySettingsFragment :
     private fun updateStealthToggleAvailability() {
         val switch =
             preferenceScreen.findPreference(resources.getString(R.string.pref_key_secure)) as? SwitchPreferenceCompat
-        if (preferenceManager.sharedPreferences
-            .getBoolean(
-                    resources.getString(R.string.pref_key_biometric),
-                    false
-                )
+        if (preferenceManager.sharedPreferences.getBoolean(
```
mcarare, line 204: It would be preferable that the changes not related to this PR, especially indentation or code rearrangements are not included here.

app/src/main/java/org/mozilla/focus/settings/privacy/PrivacySecuritySettingsFragment.kt
```diff
@@ -41,6 +43,12 @@ class PrivacySecuritySettingsFragment :
         ) {
             preferenceScreen.removePreference(biometricPreference)
         }
+        if (!Features.SHOULD_SHOW_TOOLTIP_FOR_PRIVACY_SECURITY_SETTINGS_SCREEN ||
+            !requireContext().settings.shouldShowPrivacySecuritySettingsToolTip ||
+            !WhatsNew.shouldHighlightWhatsNew(requireContext())
```
mcarare, line 48: I think we do not need this check. 
The display conditions described [here](https://github.com/mozilla-mobile/focus-android/issues/6211#issuecomment-1016067914) are covered without it, as the app needs an update to get the new value of the feature flag. 
In turn, if the user does not click the close button, based on UX specs, the CFR should continue to be displayed, even after some time has passed and when the `shouldHighlightWhatsNew` would return false.

app/src/main/res/xml/privacy_security_settings.xml
```diff
@@ -7,6 +7,11 @@
         android:layout="@layout/focus_preference"
         android:summary="@string/preference_privacy_and_security_description"/> -->
 
+    <org.mozilla.focus.settings.privacy.PreferenceToolTipCompose
+        android:key="@string/pref_key_tool_tip"
+        android:layout="@layout/focus_preference_compose_layout"
+        android:selectable="false"/>
+
```
mcarare, line 14: I do not see a separator in the Figma file, so we should use ` app:allowDividerBelow="false" ` here

buildSrc/src/main/java/Dependencies.kt
```diff
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 object Versions {
-    const val compose_version = "1.0.4"
+    const val compose_version = "1.0.5"
```
mcarare, line 6: nit: maybe this should be done in a separate commit, if at all in this PR.

app/src/main/java/org/mozilla/focus/settings/privacy/PrivacySecuritySettingsToolTip.kt
```diff
@@ -0,0 +1,114 @@
...
+@Preview
+private fun ToolTipContentPreview() {
+    ToolTipContent {
+    }
+}
+
+/**
+ * Displays a tool tip for Privacy Security Settings screen.
+ *
+ * @param onDismissToolTip Callback when the "x" Button from tool tip is pressed.
```
mcarare, line 47: nit: I think we can call it the "dismiss button."

app/src/main/java/org/mozilla/focus/settings/privacy/PrivacySecuritySettingsToolTip.kt
```diff
@@ -0,0 +1,114 @@
...
+        ) {
+            Box(
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .wrapContentHeight()
+                    .padding(16.dp)
+                    .clip(RoundedCornerShape(10.dp))
+                    .background(
+                        shape = RoundedCornerShape(10.dp),
+                        brush = Brush.verticalGradient(
```
mcarare, line 65: The Figma designs have a gradient with an angle. This might be possible with a linear gradient with computed offsets or using an XML as `painterResource` for the background.

app/src/main/java/org/mozilla/focus/settings/privacy/PrivacySecuritySettingsToolTip.kt
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * Displays a tool tip for Privacy Security Settings screen.
+ *
+ * @param onDismissToolTip Callback when the "x" Button from tool tip is pressed.
+ */
+@Composable
+fun ToolTipContent(onDismissToolTip: () -> Unit) {
+    FocusTheme {
```
mcarare, line 51: This should be moved to the topmost composable, in our case the ComposeView. This pattern is followed by the other Compose implementations.

app/src/main/java/org/mozilla/focus/settings/privacy/PrivacySecuritySettingsToolTip.kt
```diff
@@ -0,0 +1,114 @@
...
+                        fontSize = 16.sp,
+                        letterSpacing = 0.5.sp,
+                        color = focusColors.privacySecuritySettingsToolTip,
+                        fontFamily = FontFamily(Font(R.font.metropolis_bold)),
+                        lineHeight = 20.sp
+                    )
+                    Text(
+                        text = stringResource(R.string.tool_tip_message),
+                        fontSize = 14.sp,
+                        letterSpacing = 0.5.sp,
```
mcarare, line 93: I think the letter spacing is only for the title.
@jeffreygee for confirmation.
Also @jeffreygee , can you confirm the message body text style? It seems it is not the same for all CFR messages. Thank you!

app/src/main/java/org/mozilla/focus/utils/CfrUtils.kt
```diff
@@ -39,41 +39,43 @@ class CfrUtils {
...
-            )
-            if (shieldToolbarIcon != null &&
-                context.settings.shouldShowCfrForShieldToolbarIcon &&
-                isContentSecure
-            ) {
+            if (Features.SHOULD_SHOW_CFR_FOR_SHIELD_TOOLBAR_ICON) {
+                val shieldToolbarIcon = rootView.findViewById<View>(
+                    R.id.mozac_browser_toolbar_tracking_protection_indicator
+                )
+                if (shieldToolbarIcon != null &&
```
mcarare, line 46: could we move all conditions to a single if expression?

app/src/main/java/org/mozilla/focus/utils/Features.kt
```diff
@@ -9,6 +9,9 @@ package org.mozilla.focus.utils
  */
 object Features {
     const val SEARCH_TERMS_OR_URL: Boolean = true
+    const val SHOULD_SHOW_CFR_FOR_SHIELD_TOOLBAR_ICON: Boolean = false
+    const val SHOULD_SHOW_TOOLTIP_FOR_PRIVACY_SECURITY_SETTINGS_SCREEN: Boolean = false
```
mcarare, line 13: I wonder if we should consider enabling these for debug versions so they can also be tested. @ionutbedregeanu what do you think?

app/src/main/java/org/mozilla/focus/settings/privacy/PreferenceToolTipCompose.kt
```diff
@@ -0,0 +1,27 @@
...
+
+class PreferenceToolTipCompose(context: Context, attrs: AttributeSet?) :
+    Preference(context, attrs) {
+
+    override fun onBindViewHolder(holder: PreferenceViewHolder) {
+        super.onBindViewHolder(holder)
+        val binding = FocusPreferenceComposeLayoutBinding.bind(holder.itemView)
+        binding.toolTipContent.setContent {
+            ToolTipContent {
+                context.settings.shouldShowPrivacySecuritySettingsToolTip = false
```
mcarare, line 22: nit: For readability, I think we should keep the lambda in parentheses and maybe also use a named param to avoid any confusion caused by also having a composable here.

app/src/main/java/org/mozilla/focus/settings/privacy/PrivacySecuritySettingsFragment.kt
iorgamgabriel: Done

app/src/main/java/org/mozilla/focus/settings/privacy/PreferenceToolTipCompose.kt
iorgamgabriel: Done

app/src/main/java/org/mozilla/focus/settings/privacy/PrivacySecuritySettingsFragment.kt
iorgamgabriel: CI will not work if I revert the code

app/src/main/res/xml/privacy_security_settings.xml
iorgamgabriel: Done

app/src/main/java/org/mozilla/focus/settings/privacy/PrivacySecuritySettingsToolTip.kt
iorgamgabriel: Done

app/src/main/java/org/mozilla/focus/utils/Features.kt
ionutbedregeanu: Agree, QA team will thank us

iorgamgabriel: Done

app/src/main/java/org/mozilla/focus/settings/privacy/PrivacySecuritySettingsToolTip.kt
mcarare: Also, looking at the designs, the gradient has an angle, it's not horizontal or vertical. It might be more difficult obtaining that in Compose (would need a custom element) or you can try having it in an XML and setting it as painter resource.

In either case, @jeffreygee should confirm the result.

iorgamgabriel:
Screenshot_20220207-155213

mcarare (approved):
Code LGTM. We should wait for the UX feedback on the background gradient.

app/src/main/java/org/mozilla/focus/settings/privacy/PrivacySecuritySettingsToolTip.kt
jeffreygee: @mcarare -  since this was a reused component that exists in Fenix right now - is there a way to crosscheck our CFR with theirs?

I was also looking at screenshot - the header looks correct (until we go get rid of Metropolis and completely to native fonts), but the message body should be the the Android native font (Roboto)

Issue #6209, opened 2022-01-12
CFR for Delete Icon


amedyne:
This is part of the Focus onboarding updates.
Display CFR(Contextual Feature Recommender)/Tool tip for the Delete Icon.

Figma design [here](https://www.figma.com/file/ahqwL3sByHprH2Rru9R58b/Onboarding-2022?node-id=2647%3A11565).

@jeffreygee please add comment or update once behavior on when the CFR appears is finalized.

jeffreygee:
**Trigger logic**: CFR will pop up after user visits at least 3 sites. we should have it so that it's triggered by sessions on sites, and not opening the app 3 times.  
**Behavior to dismiss**: CFR can be dismissed by tapping a ‘close’ icon in the opposite corner of the pointer, or anywhere outside of the CFR. Once dismissed, will not be shown again.

ionutbedregeanu:
@jeffreygee I exported  the cfr shape as SVG and I'm trying to import it in Android Studio but I'm receiving some errors
<img width="1064" alt="Screenshot 2022-02-02 at 16 44 25" src="https://user-images.githubusercontent.com/11731652/152176160-f52aeee8-33df-4ddd-a1ac-8b54d49e43b1.png">
Can you please take a look?

ionutbedregeanu:
> **Trigger logic**: CFR will pop up after user visits at least 3 sites. we should have it so that it's triggered by sessions on sites, and not opening the app 3 times. **Behavior to dismiss**: CFR can be dismissed by tapping a ‘close’ icon in the opposite corner of the pointer, or anywhere outside of the CFR. Once dismissed, will not be shown again.

@jeffreygee Maybe we should consider to show this CFR when the user open 3 tabs and not 3 sites.
Can be scenarios when the user open a single site but 3 redirections are made, so 3 urls are were loaded and we are showing the CFR after one navigation.
What do you think?

lobontiumira:
Verified on the 2/28 Focus Nightly build with several devices:
* Oppo Find X3 Lite (Android 11),
* Xiaomi Mi8 Lite (Android 10),
* Sony Xperia Z5 Premium (Android 7.1.1), and 
* Samsung Galaxy Tab A6 (Android 5.1.1).
The CFR for delete icon is displayed on the third opened tab.

I've encountered several bugs related to the CFR for delete icon:
* #6524 
* #6525 
* #6526

I'll close this ticket as resolved.

PR #6352, opened 2022-02-04
Implement CFR for erase tabs

ionutbedregeanu:
For #6209

https://user-images.githubusercontent.com/11731652/155296896-a85f802b-3fd0-47df-8597-978b11488235.mp4

### Pull Request checklist

3. click the `Artifacts` row,

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/activity/MozillaSupportPagesTest.kt
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/MozillaSupportPagesTest.kt
@@ -28,6 +28,7 @@ class MozillaSupportPagesTest {
     @Before
     fun setUp() {
         featureSettingsHelper.setShieldIconCFREnabled(false)
+        featureSettingsHelper.setNumberOfTabsOpened(4)
     }
 
     @After
--- a/app/src/main/java/org/mozilla/focus/Components.kt
+++ b/app/src/main/java/org/mozilla/focus/Components.kt
@@ -50,6 +50,7 @@ import mozilla.components.service.nimbus.NimbusApi
 import mozilla.components.support.locale.LocaleManager
 import org.mozilla.focus.activity.MainActivity
 import org.mozilla.focus.browser.BlockedTrackersMiddleware
+import org.mozilla.focus.cfr.CfrMiddleware
 import org.mozilla.focus.components.EngineProvider
 import org.mozilla.focus.downloads.DownloadService
 import org.mozilla.focus.engine.AppContentInterceptor
@@ -70,6 +71,7 @@ import org.mozilla.focus.tabs.MergeTabsMiddleware
 import org.mozilla.focus.telemetry.GleanMetricsService
 import org.mozilla.focus.telemetry.TelemetryMiddleware
 import org.mozilla.focus.topsites.DefaultTopSitesStorage
+import org.mozilla.focus.utils.Features
 import org.mozilla.focus.utils.IntentUtils
 import org.mozilla.focus.utils.Settings
 import java.util.Locale
@@ -130,6 +132,11 @@ class Components(
     }
 
     val store by lazy {
+        val cfrMiddleware = if (Features.SHOW_ERASE_CFR) {
+            listOf(CfrMiddleware(context.components))
+        } else {
+            listOf()
+        }
         BrowserStore(
             middleware = listOf(
                 PrivateNotificationMiddleware(context),
@@ -146,8 +153,8 @@ class Components(
                 AdsTelemetryMiddleware(adsTelemetry),
                 BlockedTrackersMiddleware(context),
                 MergeTabsMiddleware(context),
-                RecordingDevicesMiddleware(context)
-            ) + EngineMiddleware.create(engine)
+                RecordingDevicesMiddleware(context),
+            ) + EngineMiddleware.create(engine) + cfrMiddleware
         ).apply {
             MediaSessionFeature(context, MediaSessionService::class.java, this).start()
         }
--- a/app/src/main/java/org/mozilla/focus/cfr/CfrMiddleware.kt
+++ b/app/src/main/java/org/mozilla/focus/cfr/CfrMiddleware.kt
@@ -1,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package org.mozilla.focus.cfr
+
+import mozilla.components.browser.state.action.BrowserAction
+import mozilla.components.browser.state.action.TabListAction
+import mozilla.components.browser.state.state.BrowserState
+import mozilla.components.lib.state.Middleware
+import mozilla.components.lib.state.MiddlewareContext
+import org.mozilla.focus.Components
+import org.mozilla.focus.state.AppAction
+import org.mozilla.focus.utils.ERASE_CFR_LIMIT
+
+/**
+ * Middleware used to intercept browser store actions in order to decide when should we display a specific CFR
+ */
+class CfrMiddleware(private val components: Components) : Middleware<BrowserState, BrowserAction> {
+    override fun invoke(
+        context: MiddlewareContext<BrowserState, BrowserAction>,
+        next: (BrowserAction) -> Unit,
+        action: BrowserAction
+    ) {
+        next(action)
+        if (action is TabListAction.AddTabAction) {
+            components.settings.numberOfTabsOpened++
+            if (components.settings.numberOfTabsOpened == ERASE_CFR_LIMIT) {
+                components.appStore.dispatch(AppAction.ShowEraseTabsCfrChange(true))
+            }
+        }
+    }
+}
--- a/app/src/main/java/org/mozilla/focus/state/AppAction.kt
+++ b/app/src/main/java/org/mozilla/focus/state/AppAction.kt
@@ -82,7 +82,12 @@ sealed class AppAction : Action {
     data class AutoplayChange(val value: Boolean) : AppAction()
 
     /**
-     * State of secret settings state has changed.
+     * State of secret settings has changed.
      */
     data class SecretSettingsStateChange(val enabled: Boolean) : AppAction()
+
+    /**
+     * State of erase tabs CFR has changed
+     */
+    data class ShowEraseTabsCfrChange(val value: Boolean) : AppAction()
 }
--- a/app/src/main/java/org/mozilla/focus/state/AppState.kt
+++ b/app/src/main/java/org/mozilla/focus/state/AppState.kt
@@ -16,12 +16,14 @@ import java.util.UUID
  * @property autoplayRulesChanged A flag which reflects the state of autoplay rules,
  * whether they have been updated or not
  * @property secretSettingsEnabled A flag which reflects the state of debug secret settings
+ * @property showEraseTabsCfr A flag which reflects the state erase tabs CFR
  */
 data class AppState(
     val screen: Screen,
     val topSites: List<TopSite> = emptyList(),
     val autoplayRulesChanged: Boolean = false,
-    val secretSettingsEnabled: Boolean = false
+    val secretSettingsEnabled: Boolean = false,
+    val showEraseTabsCfr: Boolean = false
 ) : State
 
 /**
--- a/app/src/main/java/org/mozilla/focus/utils/Settings.kt
+++ b/app/src/main/java/org/mozilla/focus/utils/Settings.kt
@@ -19,6 +19,8 @@ import org.mozilla.focus.searchsuggestions.SearchSuggestionsPreferences
 import org.mozilla.focus.settings.permissions.AutoplayOption
 import org.mozilla.focus.settings.permissions.getValueByPrefKey
 
+const val ERASE_CFR_LIMIT = 3
+
 /**
  * A simple wrapper for SharedPreferences that makes reading preference a little bit easier.
  * This class is designed to have a lot of (simple) functions
@@ -369,6 +371,21 @@ class Settings(
         getPreferenceKey(R.string.pref_key_default_theme),
         false
     )
+
+    // Store how many tabs were opened until now, but the max value which can be stored is 4
+    // since this values is used to decide if we should display the erase cfr
+    var numberOfTabsOpened: Int
+        get() = preferences.getInt(
+            context.getString(R.string.number_of_opened_tabs),
+            0
+        )
+        set(value) {
+            if (value <= ERASE_CFR_LIMIT + 1) {
+                preferences.edit()
+                    .putInt(context.getString(R.string.number_of_opened_tabs), value)
+                    .apply()
+            }
+        }
 
     var shouldUseNimbusPreview: Boolean
         get() = preferences.getBoolean(getPreferenceKey(R.string.pref_key_use_nimbus_preview), true)
--- a/app/src/main/res/values/preference_keys.xml
+++ b/app/src/main/res/values/preference_keys.xml
@@ -86,4 +86,7 @@
     <string name="pref_key_light_theme" translatable="false">pref_key_light_theme</string>
     <string name="pref_key_dark_theme" translatable="false">pref_key_dark_theme</string>
     <string name="pref_key_default_theme" translatable="false">pref_key_default_theme</string>
+
+    <!-- Opened tabs counter used to trigger erase tabs CFR -->
+    <string name="number_of_opened_tabs" translatable="false">number_of_opened_tabs</string>
 </resources>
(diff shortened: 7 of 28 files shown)


ionutbedregeanu:
@jeffreygee I scanned the cfr with accessibility scanner and we are having two warnings
We need to add text for the close image button as content description which will be played for those users which are using an accessibility app





Also the shape of popup is not imported as expected in Android Studio, see my comment from the ticket https://github.com/mozilla-mobile/focus-android/issues/6209#issuecomment-1028010905

app/src/test/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegrationTest.kt
```diff
@@ -107,6 +108,24 @@ class BrowserToolbarIntegrationTest {
         verify(browserToolbarIntegration).observerSecurityIndicatorChanges()
     }
 
+    @Test
+    fun `WHEN start method is called THEN observe erase tabs CFR changes`() {
+
```
Mugurell, line 113: nit: Extraneous empty line.
Separating the 3 main sections: arrange, act, assert is nice 👍

app/src/test/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegrationTest.kt
```diff
@@ -107,6 +108,24 @@ class BrowserToolbarIntegrationTest {
...
+        doNothing().`when`(browserToolbarIntegration).observeEraseCfr()
+
+        browserToolbarIntegration.start()
+
+        verify(browserToolbarIntegration).observeEraseCfr()
+    }
+
+    @Test
+    fun `WHEN stop method is called THEN erase tabs scope is not active anymore`() {
+
```
Mugurell, line 123: nit: Extraneous empty line.
Separating the 3 main sections: arrange, act, assert would be nice though.

app/src/main/java/org/mozilla/focus/state/AppState.kt
```diff
@@ -16,12 +16,14 @@ import java.util.UUID
...
  * @property secretSettingsEnabled A flag which reflects the state of debug secret settings
+ * @property showEraseTabsCfr A flag which reflects the state of erase tabs CFR
  */
 data class AppState(
     val screen: Screen,
     val topSites: List<TopSite> = emptyList(),
     val autoplayRulesChanged: Boolean = false,
-    val secretSettingsEnabled: Boolean = false
+    val secretSettingsEnabled: Boolean = false,
+    val showEraseTabsCfr: Boolean = false
```
Mugurell, line 26: nit: Food for thought. `AppState` seems to become a sink for all things. Maybe a `BrowserScreenState` would be nice to have in the future and it could contain this kind of properties (cfr that only appears in the browser fragment).

app/src/main/java/org/mozilla/focus/cfr/CfrPopup.kt
```diff
@@ -0,0 +1,43 @@
...
+    private lateinit var binding: PopupCfrBinding
+
+    init {
+        initializeLayout()
+    }
+
+    private fun initializeLayout() {
+        val layoutInflater =
+            parentView.context.getSystemService(Context.LAYOUT_INFLATER_SERVICE) as LayoutInflater
+        binding = PopupCfrBinding.inflate(layoutInflater, parentView, false)
```
Mugurell, line 28:
```suggestion
binding = PopupCfrBinding.inflate(LayoutInflater.from(parentView.context), parentView, false)
```
Just for consistency. Seems like `LayoutInflater.from(..)` is preferred and used in many other places.

app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -693,6 +694,12 @@ class UrlInputFragment :
                 requireComponents.store.dispatch(ContentAction.UpdateSearchTermsAction(tabId, searchTerms))
             }
         }
+        if (Features.SHOW_ERASE_CFR) {
+            requireComponents.settings.numberOfTabsOpened++
+            if (requireComponents.settings.numberOfTabsOpened == ERASE_CFR_LIMIT) {
+                requireComponents.appStore.dispatch(AppAction.EraseTabsCfrChange(true))
+            }
```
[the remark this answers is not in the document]
Mugurell: Not sure this is the best place to observe the number of open tabs and dispatch the action to show the CFR.
I long clicked on links -> "Open link in private tab" and though I now have 3 tabs I didn't get the CFR. Not sure if I was should have.

app/src/main/java/org/mozilla/focus/cfr/CfrPopup.kt
```diff
@@ -0,0 +1,43 @@
...
+package org.mozilla.focus.cfr
+
+import android.content.Context
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import android.widget.PopupWindow
+import androidx.core.content.ContextCompat
+import org.mozilla.focus.databinding.PopupCfrBinding
+
+class CfrPopup(
```
Mugurell, line 14: Understood this is meant to be used for all CFRs? 👍
nit: Needs documentation.

app/src/main/res/drawable/cfr_shape_with_top_arrow.xml
```diff
@@ -0,0 +1,22 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
Mugurell, line 1: Use `android:autoMirrored="true"` to have this automatically flipped in RTL and so prevent
image

Mugurell (commented):
Generally this looks good though there are some things that need to be checked and maybe updated.

app/src/main/java/org/mozilla/focus/utils/Settings.kt
```diff
@@ -362,6 +364,21 @@ class Settings(
         false
     )
 
+    // Store how many tabs were opened until now, but the max value wich can be stored is 3
+    // since this values is used to decide if we should display the erase cfr
+    var numberOfTabsOpened: Int
```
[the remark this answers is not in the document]
ionutbedregeanu: We need too keep a counter even if the app is killed

app/src/main/java/org/mozilla/focus/state/AppState.kt
ionutbedregeanu: `BrowserFragment` is already a complex class, to add now a `BrowserScreenState` implies too many changes which will be out of the scope. I will investigate more to find a better approach

app/src/main/java/org/mozilla/focus/Components.kt
```diff
@@ -126,6 +128,10 @@ class Components(
         }
     }
 
+    val browserScreenStore by lazy {
```
Mugurell, line 131: This shows an issue with our first proposal of installing a Middleware to BrowserStore - which is a singleton per app so if the Middleware is to dispatch updates to another store like a `..ScreenStore` that also has to be a singleton and this diverges from our flow of having each store's lifecycle tied to it's component's - most likely a fragment - lifecycle.
Thinking about how this will evolve over time, we'd have multiple `..ScreenStore` singletons, if not set as private here accessible from any other component. This could bring the advantage of easier updates from everywhere in the app but at the same time could lead to strange bugs if improperly used - state updated from everywhere and leaking everywhere.
I'd still think `Store`s should be scoped to smaller app components and a `..ScreenStore` should only live while that Screen lives and only be available there.

The only other `Store` that is already a singleton is `AppStore` which maybe more suitable for the usecase of syncing the state of various components based on which a CFR should be shown.
Alternatively we could use an `AbstractBinding` in which to observe `BrowserStore` updates.

app/src/main/java/org/mozilla/focus/Components.kt
```diff
@@ -146,7 +147,8 @@ class Components(
                 AdsTelemetryMiddleware(adsTelemetry),
                 BlockedTrackersMiddleware(context),
                 MergeTabsMiddleware(context),
-                RecordingDevicesMiddleware(context)
+                RecordingDevicesMiddleware(context),
+                CfrMiddleware(context.components),
```
jonalmeida, line 151: Do we want existing users to see these CFRs as well? If yes, we should put this middleware behind a feature flag too. We we release this code without the flag enabled, the settings code in the middleware would continue to increment up. So when the flag is flipped, this CFR would potentially not show up because existing users would have passed that criteria.

app/src/main/java/org/mozilla/focus/cfr/CfrMiddleware.kt
```diff
@@ -0,0 +1,32 @@
...
+class CfrMiddleware(private val components: Components) : Middleware<BrowserState, BrowserAction> {
+    override fun invoke(
+        context: MiddlewareContext<BrowserState, BrowserAction>,
+        next: (BrowserAction) -> Unit,
+        action: BrowserAction
+    ) {
+        next(action)
+        if (action is TabListAction.AddTabAction) {
+            components.settings.numberOfTabsOpened++
+            if (components.settings.numberOfTabsOpened == ERASE_CFR_LIMIT) {
```
jonalmeida, line 27: Do we need this setting incrementing forever? Could we stop once we reached our threshold?

app/src/main/res/values/static_strings.xml
```diff
@@ -15,4 +15,5 @@
     <string name="about_debug_menu_toast_progress" translatable="false">Debug menu: %1$d click(s) left to enable</string>
     <string name="about_debug_menu_toast_done" translatable="false">Debug menu enabled</string>
 
+    <string name="number_of_opened_tabs" translatable="false">Number of opened tabs</string>
```
jonalmeida, line 18: Shouldn't this be in [`preference_keys.xml`](https://github.com/mozilla-mobile/focus-android/blob/main/app/src/main/res/values/preference_keys.xml)?

app/src/main/java/org/mozilla/focus/cfr/CfrMiddleware.kt
ionutbedregeanu: We already do a check in Settings in numberOfTabsOpened set method

app/src/main/java/org/mozilla/focus/state/AppAction.kt
```diff
@@ -82,7 +82,12 @@ sealed class AppAction : Action {
     data class AutoplayChange(val value: Boolean) : AppAction()
 
     /**
-     * State of secret settings state has changed.
+     * State of secret settings has changed.
      */
     data class SecretSettingsStateChange(val enabled: Boolean) : AppAction()
+
+    /**
+     * State of secret erase tabs CFR has changed
```
Mugurell, line 90: Why is the erase tabs CFR secret? 🕵️‍♂️

app/src/main/java/org/mozilla/focus/utils/Settings.kt
```diff
@@ -370,6 +372,21 @@ class Settings(
         false
     )
 
+    // Store how many tabs were opened until now, but the max value wich can be stored is 3
```
Mugurell, line 375:
```suggestion
    // Store how many tabs were opened until now, but the max value which can be stored is 3
```

app/src/main/java/org/mozilla/focus/utils/Settings.kt
```diff
@@ -370,6 +372,21 @@ class Settings(
...
+    // since this values is used to decide if we should display the erase cfr
+    var numberOfTabsOpened: Int
+        get() = preferences.getInt(
+            context.getString(R.string.number_of_opened_tabs),
+            0
+        )
+        set(value) {
+            if (value <= ERASE_CFR_LIMIT + 1) {
```
Mugurell, line 383: This contradicts the property description - "the max value which can be stored is `3`".
Is it okay to also store `4`?

ionutbedregeanu:
WHEN start method is called THEN observe erase tabs CFR changes FAILED

app/src/main/res/values/preference_keys.xml
```diff
@@ -86,4 +86,7 @@
     <string name="pref_key_light_theme" translatable="false">pref_key_light_theme</string>
     <string name="pref_key_dark_theme" translatable="false">pref_key_dark_theme</string>
     <string name="pref_key_default_theme" translatable="false">pref_key_default_theme</string>
+
+    <!--Opened tabs counter used to trigger erase tabs CFR-->
```
Mugurell, line 90: nit: For consistency with the others this comment should be padded with a space character (between it and the comment tags)

app/src/main/java/org/mozilla/focus/Components.kt
```diff
@@ -130,8 +132,13 @@ class Components(
     }
 
     val store by lazy {
+        val browserStoreMiddlewares = if (Features.SHOW_ERASE_CFR) {
```
Mugurell, line 135: nit: the actual browser store middlewares are a bit below and there is a list of them.
Can this property be named just "cfrMiddleware" and then be added after engineMiddleware - just to not do the addition both at the beginning and at the end of where the middlewares are set?

Mugurell (commented):
LGTM.
Curious if Jonathan has other feedback.

app/src/main/res/values/preference_keys.xml
```diff
@@ -86,4 +86,7 @@
     <string name="pref_key_light_theme" translatable="false">pref_key_light_theme</string>
     <string name="pref_key_dark_theme" translatable="false">pref_key_dark_theme</string>
     <string name="pref_key_default_theme" translatable="false">pref_key_default_theme</string>
+
+    <!-- Opened tabs counter used to trigger erase tabs CFR -->
+    <string name="number_of_opened_tabs" translatable="false">Number of opened tabs</string>
```
jonalmeida, line 91: key value should use snake casing.
```suggestion
    <string name="number_of_opened_tabs" translatable="false">number_of_opened_tabs</string>
```

jonalmeida (approved):
> Curious if Jonathan has other feedback.

Thanks for checking! This looks good to me. 🙂

app/src/test/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegrationTest.kt
```diff
@@ -107,6 +108,24 @@ class BrowserToolbarIntegrationTest {
         verify(browserToolbarIntegration).observerSecurityIndicatorChanges()
     }
 
+    @Test
+    fun `WHEN start method is called THEN observe erase tabs CFR changes`() {
+        doNothing().`when`(browserToolbarIntegration).observeEraseCfr()
+
+        browserToolbarIntegration.observeEraseCfr()
+
+        verify(browserToolbarIntegration).observeEraseCfr()
```
Mugurell, line 117: The `verify` here seems redundant.

To actually test that when  `state.showEraseTabsCfr` changes we'll try showing the intended CFR (can't test that we actually show one because the CFR instance is internally constructed) the code might be refactored such that:
- when the state changes we call a method to build the CFR  - this can then be tested
- the method to build the CFR can be tested to ensure all properties are the intended ones - can test container/text/anchor/dismiss callback.
- there is another method which receives a CFR instance and just calls `show()` on it. This can also be tested.

app/src/test/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegrationTest.kt
```diff
@@ -107,6 +108,24 @@ class BrowserToolbarIntegrationTest {
...
+    }
+
+    @Test
+    fun `WHEN stop method is called THEN erase tabs scope is not active anymore`() {
+        browserToolbarIntegration.observeEraseCfr()
+
+        browserToolbarIntegration.stop()
+
+        assertEquals(browserToolbarIntegration.eraseTabsCfrScope?.isActive, false)
```
Mugurell, line 126: This way of testing a loose indication that we listen for "erase tabs cfrs updates" could also be used for `browserToolbarIntegration.start()` where we could also test the feature flag and the different behavior depending on if the current tab is a custom one or not.

app/src/androidTest/java/org/mozilla/focus/activity/MozillaSupportPagesTest.kt
```diff
@@ -104,6 +105,7 @@ class MozillaSupportPagesTest {
 
     @SmokeTest
     @Test
+    @Ignore("Refactoring is required because of erase popup https://github.com/mozilla-mobile/focus-android/pull/6352")
```
Mugurell, line 108: When disabling UI tests a new ticket should be creating to inform their owners about this and allow them to prioritize refactoring these tests.

Mugurell (approved):
Looks good to me also but the tests can be improved.

app/src/androidTest/java/org/mozilla/focus/activity/MozillaSupportPagesTest.kt
ionutbedregeanu: https://github.com/mozilla-mobile/focus-android/issues/6509

app/src/test/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegrationTest.kt
ionutbedregeanu: After coroutines version update this test is failing but I'm not sure why.
In order to not block this merge will create a separate ticket to increase the coverage of this class

PR #6411, opened 2022-02-11
New CFR popup composable

Mugurell:
This is intended to fix all issues related to layout and anchoring and serve as
the default way for building and showing a CFR popup. (Anchored to top).

Because of the async nature of the values for insets and screen rotation immediately after an orientation change the popup will automatically get dismissed to prevent any anchoring issues.
While not ideal the effect of this is better than accepting layout issues after orientation changes and is the same approach used for other of our popups.

Chose to use compose because it can ease anchoring and also ties in with the compose refactoring.
Going by this route allows us easier control over the popup, with it being built at runtime and not depend on hardcoded drawables.
For example it would help with showing anchorage by having the arrow indicator at the right position while the hardcoded drawable would have issues with anchors immediately at the start of the screen since the arrow indicator has a fixed padding that may be too big to properly show the anchorage. Plus the colors now match perfectly!

https://user-images.githubusercontent.com/11428869/153623802-6f09556f-86f5-4aa3-bdec-1f349f8c7493.mp4

&nbsp;
-------------------
&nbsp;

How will these help the new CFRs:

| Security CFR in LTR  | Security CFR in LTR |
| --- | --- |
| <video src="https://user-images.githubusercontent.com/11428869/154315551-16a1e1a4-4f60-46b5-a9f0-69da8aeb5a05.mp4" /> | <video src="https://user-images.githubusercontent.com/11428869/154315908-1187eee9-d23a-4c4e-bff5-378317c45d52.mp4" /> |

| Erase tabs CFR in LTR  | Erase tabs CFR in LTR |
| --- | --- |
| <video src="https://user-images.githubusercontent.com/11428869/154316120-23ede547-2ef5-4055-9c71-f60346d6d129.mp4" /> | <video src="https://user-images.githubusercontent.com/11428869/154316188-c0b08fbb-9793-4522-a081-4720a82fda0b.mp4" /> |

### Pull Request checklist

3. click the `Artifacts` row,


Mugurell:
Also tagging @jeffreygee for reviewing the end result 
- there is a 10dp horizontal padding around the 256dp popup to allow the close button more drawing space so that is can be 48 dp while not distorting the original design. 
(the padding should not have any negative impact for the original design and the intended use of this popups)
- there is 10dp elevation (configurable for each popup) which gives a bit of shadow.

jeffreygee:
> Also tagging @jeffreygee for reviewing the end result
> 
>     * there is a 10dp horizontal padding around the 256dp popup to allow the close button more drawing space so that is can be 48 dp while not distorting the original design.
>       (the padding should not have any negative impact for the original design and the intended use of this popups)
> 
>     * there is 10dp elevation (configurable for each popup) which gives a bit of shadow.

@Mugurell  This looks good to me! thank you

ionutbedregeanu (approved):

iorgamgabriel (approved):

app/src/main/java/org/mozilla/focus/compose/CFRPopupShape.kt
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * A [Shape] describing a popup with an indicator triangle shown at top.
+ *
+ * @param indicatorArrowStartOffset Distance between the popup start and the indicator arrow start
+ * @param indicatorArrowHeight Height of the indicator triangle. This influences the base length.
+ * @param cornerRadius The radius of the popup's corners.
+ * If [indicatorArrowStartOffset] is `0` then the top-start corner will not be rounded.
+ */
+class CFRPopupShape(
```
gabrielluong, line 43: I wonder if we should just call this by its full name instead of acronym - ideally we shouldn't expect people to have to know our acronyms. `ContextualFeatureRecommendationShape`,  `FeatureRecommendationShape`, ???

app/src/main/java/org/mozilla/focus/compose/CFRPopupShape.kt
```diff
@@ -0,0 +1,179 @@
...
+            contentAlignment = Alignment.Center
+        ) {
+            Text(text = "This is just a test")
+        }
+    }
+}
+
+@Preview
+@Composable
+fun CFRPopupShapeRTLPreview() {
```
gabrielluong, line 155: Maybe we should make Preview functions private?

app/src/main/java/org/mozilla/focus/compose/CFRPopupShape.kt
gabrielluong: Just a drive by thought, don't need another review from me.

Mugurell: I do prefer explicative naming but in this case `ContextualFeatureRecommendation` seems just a tad long to be used all throughout the app. I see in Fenix `cfr` is used in many places and in tickets title / descriptions also so the entire feature might be named "CFR".

Mugurell:
`ui-test-x86-debug` failed because of 
image

Restarting CI tasks.

Mugurell:
Ignored the unrelated and failing `notificationOpenButtonTest` ui test. CI should be green now.

Mugurell:
Again `inconclusive` results for `ui-test-x86-debug`. Restarting CI.

Mugurell:
Same inconclusive ui test results.

PR #6444, opened 2022-02-15
Close #6443: Apply experiments after fetching during startup

### LINKED ISSUE #6443: Fetch experiments on startup first before applying
When we want to run experiments on startup _for new users_. We need to make sure we apply the experiments only after fetching them.

jonalmeida:
We need to fetch the experiments first, observe when it's completed, before finally applying it.

If this patch looks good, then:
 1. We need to uplift this to release/beta where needed.
 2. Do the same thing for Fenix. (Related: why was the code not doing this before (which is taken from Fenix)? Why can't these setup things also go into AC?)

3. click the `Artifacts` row,


jonalmeida:
> If you implement it this way then _all_ experiments will be subject to the delay for the network roundtrip to remote-settings to fetch the experiment recipes.

Ah sorry, maybe I misunderstood your recommendation. Was there another direction I should have gone to update only this experiment?

> That may be an acceptable delay for the app, but my concern is that if the first thing the user sees is controlled by an experiment then it may require delaying until the fetch is complete before showing that to the user which may cause the app to appear slow to start. I just want to be clear what the implications are for awaiting the fetch.

I'm not sure how concerned we should be here if we're optimizing for this particular experiment right now, because:
1. The tab experiment is for showing the tab counter & tab switcher, which are not visible by default _until_ you have tabs.
2. The experiment is targeted at new users. So they would be seeing our onboarding cards first, and then have to tap through or skip them.
2. To try and create multiple tabs you need to navigate to a website or searching, and then long press a link to open in a new tab or expect the link to open a new window event.

travis79:
> 

This approach is definitely acceptable, I just wanted to make sure that it was understood that the delay from the first fetch to applying those experiments may be a factor in how the UX works. As long as you are aware of that, and we are optimizing for this particular experiment (and there aren't any other experiments running that we would need to worry about that might be affected), then I'm onboard with doing this. I think James' main concerns in going with the "apply first, then fetch" approach was that some feature of the app could request experiment data from Nimbus locally, then an apply of updates happens, and then that app feature might get different values if it requested the experiment data again after the update.

As long as this is a controlled environment, I think this should work just fine for the tabs experiment, especially if the user needs to open multiple tabs before the UX for this experiment is obvious.

I'll give my r+ on this, thanks for clarifying!  :)

travis79 (approved):

Amejia481 (approved):

jonalmeida:
Thanks @travis79 and @Amejia481 !

jonalmeida:
@Mergifyio backport releases_v97.0 releases_v98.0

Issue #6428, opened 2022-02-14
Very frequent UI test failure - different shards with inconclusive results


Archaeopteryx:
### Firebase Test Run:
https://console.firebase.google.com/project/moz-focus-android/testlab/histories/bh.2189b040bbce6d5a/matrices/6479591213728734212
### Build:
https://github.com/mozilla-mobile/focus-android/commit/92bf0b66d24669272c3edfe28a9dc43d7fcab3dc

Archaeopteryx:
Started by #6348

AaronMT:
I've opened a ticket with Google for insight, not getting any replies in Firebase Community.

AaronMT:
No replies yet in ticket. Testing if a Pixel 2 is more stable in draft PR. From Slack reply

> For [matrix-2iu96w9y1otlk](https://console.firebase.google.com/u/0/project/moz-focus-android/testlab/histories/bh.2189b040bbce6d5a/matrices/8978338696274776020/details?stepId=), the issue is on the infrastructure. A related question, does your app have a lot log?

Do we have excessive logging enabled?

AaronMT:
I have a promising workaround while they continue to investigate based on discussion about reducing our sharding.

> If we assume the infrastructure failure rate is the same, more sharding will increase the chance for the matrix to fail.  Further, these test executions all have overhead in terms of device scheduling, preparation, post test run result processing, these time combined might be much longer than the time running the test. So in practice, more sharding might not help you at some point.

AaronMT:
> HI Aaron,
> 
> Thank you for your patience, After speaking with Daming this is a known issue within the Testing Infrastructure team for which there is an active ticket and ongoing investigation. The test matrix ID's you have provided have been noted and the product team has been made aware of your specific case.
> 
> At this time, there is no mitigation available, and I will continue to provide you with updates as they become available on my end. Please let me know if you have any questions in the meantime.
> 
> Best,
> 
> Desmon

AaronMT:
> Hi Aaron, 
> 
> Thank you for your patience. I have received correspondence from the product team regarding the results you have provided: 
> 
> this is an issue with our current API level 30 virtual devices (Pixel2, Pixel3 and NexusLowRes) and our engineering team is actively working on it. We have alternate devices you can test in the meantime that are currently in private alpha. We have whitelisted you at this time for these devices to try at your discretion. 
> 
> - Pixel2Emulator 
> - NexusLowResEmulator. 
> 
[...]

PR #6463, opened 2022-02-17
No issue: Reduce number of FTL shards by pairing tests

AaronMT:
For https://github.com/mozilla-mobile/focus-android/issues/6428, while the Firebase team investigates the `Inconclusive Result` infrastructure problem it was recommended that we try reducing our shard count by pairing UI tests (number of UI tests / 2 so 36 (72 tests at this time of writing)). 

From Firebase Community Slack

> If we assume the infrastructure failure rate is the same, more sharding will increase the chance for the matrix to fail.  Further, these test executions all have overhead in terms of device scheduling, preparation, post test run result processing, these time combined might be much longer than the time running the test. So in practice, more sharding might not help you at some point.

Based on runs in https://github.com/mozilla-mobile/focus-android/pull/6457 this looks promising. This PR reduces the shard count for debug where we run all tests (`-1` previously indicates create a shard for every test).


isabelrios (approved):
Let's give this a try! 🚀

Issue #6531, opened 2022-02-28
[UI smoke tests] Improve smoke tests coverage


sv-ohorvath:
Create new UI tests or improve coverage of existing ones base on the smoke tests suite in TestRail: https://testrail.stage.mozaws.net/index.php?/suites/view/1028&group_by=cases:section_id&group_order=asc&group_id=51537

sv-ohorvath:
Lets keep it

PR #6532, opened 2022-02-28
[UI smoke test] For #6531: ads UI test for close tab button

sv-ohorvath:
### Pull Request checklist

3. click the `Artifacts` row,


app/src/androidTest/java/org/mozilla/focus/helpers/TestHelper.kt
```diff
@@ -39,6 +39,7 @@ object TestHelper {
     @JvmField
     var mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
     const val waitingTime = DateUtils.SECOND_IN_MILLIS * 15
+    const val waitingTimeShort = DateUtils.SECOND_IN_MILLIS * 5
```
AaronMT, line 42: I'm OK with adding this back if we think it won't cause any Firebase issues with short timeouts. I figure if we wait longer on Firebase there's less chance of timing problems, that was my thinking when I removed it in the refactor - I was wondering the benefit. I'll leave it up to you.

AaronMT (approved):
Wait on landing for feedback on from my question.

app/src/androidTest/java/org/mozilla/focus/helpers/TestHelper.kt
sv-ohorvath: I thought we can use this when we don't expect a long wait as for page loading, if some view doesn't show in 5 sec, maybe it's not worth waiting longer thus increasing the total time of the test run. But I don't mind following your advice, as you think is better.

PR #6568, opened 2022-03-03
No issue: Revert group sharding to fix reporting

AaronMT:
I think by splitting the tests into shards it broke my daily reporting for flaky and failures. I didn't see too much of a difference in shard splitting times so let's revert it to fix my reporting.


isabelrios (approved):
👍

AaronMT:
As it's possible that more sharding might lessen the infra error, I'll leave this up for now until we get some sort of closure on the Google infra issue

Issue #6522, opened 2022-02-28
[crash] Focus Nightly crashes when accessing Your Rights page


lobontiumira:
## Steps to reproduce
1. Open Focus Nightly.
2. Go to Settings - Mozilla - Your Rights.
3. Tap on it.

### Expected behavior
The "Your Rights" page openes.

### Actual behavior
Nightly crashes.

### Device information
* Android device: Oppo Find X3 Lite (Android 11), Xiaomi Mi8 Lite (Android 9)
* Focus version: Focus Nightly from 2/28

1181f2ec-af46-47c2-9d1c-c7ccada4a570
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Boolean
* Sentry: <No URL>
* Socorro: https://crash-stats.mozilla.org/report/index/bp-9c051603-8185-4cc5-b7f1-4968b0220228
----
```
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Boolean
at org.mozilla.focus.settings.MozillaSettingsFragment.onSharedPreferenceChanged(MozillaSettingsFragment.kt:4)
at android.app.SharedPreferencesImpl$EditorImpl.notifyListeners(SharedPreferencesImpl.java:637)
at android.app.SharedPreferencesImpl$EditorImpl.lambda$notifyListeners$0$SharedPreferencesImpl$EditorImpl(SharedPreferencesImpl.java:643)
at android.app.-$$Lambda$SharedPreferencesImpl$EditorImpl$3CAjkhzA131V3V-sLfP2uy0FWZ0.run(Unknown Source:4)
at android.os.Handler.handleCallback(Handler.java:938)
at android.os.Handler.dispatchMessage(Handler.java:99)
at android.os.Looper.loop(Looper.java:263)
at android.app.ActivityThread.main(ActivityThread.java:8258)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:612)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1006)
```

lobontiumira:
Verified as fixed on the latest Focus Nightly 100.0a1 (build 360680510 with GV 100.0a1-20220308100756), and on Beta 99.0.0-beta.1 with the following devices:
* Xiaomi mi4i (Android 5.0.2)
* Samsung Galaxy Note 8 (Android 9)
[...]

PR #6572, opened 2022-03-04
For #6522 Focus Nightly crashes when accessing Your Rights page

iorgamgabriel:
I made a check sharedPreference Key to be different than  the key that is used to show Cfr for Erase Tabs . 
The value  for the key "number_of_opened_tabs" is int and it will crash the app at this line   ProTips.tipsSettingChanged.record(ProTips.TipsSettingChangedExtra(sharedPreferences.all[key] as Boolean)).

### Pull Request checklist

3. click the `Artifacts` row,

### DIFF
--- a/app/src/main/java/org/mozilla/focus/settings/MozillaSettingsFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/settings/MozillaSettingsFragment.kt
@@ -88,10 +88,12 @@ class MozillaSettingsFragment :
     }
 
     override fun onSharedPreferenceChanged(sharedPreferences: SharedPreferences, key: String) {
-        TelemetryWrapper.settingsEvent(key, sharedPreferences.all[key].toString())
-        ProTips.tipsSettingChanged.record(
-            ProTips.TipsSettingChangedExtra(sharedPreferences.all[key] as Boolean)
-        )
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/focus/settings/MozillaSettingsFragment.kt
```diff
@@ -88,10 +88,12 @@ class MozillaSettingsFragment :
     }
 
     override fun onSharedPreferenceChanged(sharedPreferences: SharedPreferences, key: String) {
-        TelemetryWrapper.settingsEvent(key, sharedPreferences.all[key].toString())
-        ProTips.tipsSettingChanged.record(
-            ProTips.TipsSettingChangedExtra(sharedPreferences.all[key] as Boolean)
-        )
+        if (isAdded && key != requireContext().getString(R.string.number_of_opened_tabs)) {
```
ionutbedregeanu, line 91: Instead of `key != requireContext().getString(R.string.number_of_opened_tabs)` I think we should use `key == requireContext().getString(R.string.pref_key_homescreen_tips)` since is the only preference which can be changed from this screen and has as a value a boolean

iorgamgabriel: Done

ionutbedregeanu (approved):

Issue #6528, opened 2022-02-28
[Experiment] The treatment branch of the Android A/B v97 experiment allows multiple tabs on the first run


heres-maria:
**[Affected Versions]:**
- Firefox Focus 97.2.0 (Build #360591303) GV: 97.0.1-20220216172458
- Firefox Focus 98.0 (Build #360591110) GV: 98.0.20220224185616

**[Affected Platforms]:**
- Android 9 (Samsung Galaxy S8)
- Android 10 (Oneplus 6) 

**[Prerequisites]:**
- Have a Focus for Android build with the following modifications:
     The following lines added to local.properties:
`nimbus.remote-settings.url = https://settings.stage.mozaws.net`
`debuggable = true`
A new “.nimbus” file (in the same location as local.properties) containing the following line:
`https://settings.stage.mozaws.net`
- Have the build installed on an Android device <7 day ago.

**[Steps to reproduce]:**
1. Open the browser from prerequisites to enroll in treatment-a of the [Stage Copy Focus Android A/B v97 Experiment - New Users](https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus/stage-copy-focus-android-ab-v97-experiment-new-users) experiment.
2. Navigate to a page.
3. Close and reopen the browser.
4. Press one of the links and tap “Open link in private tab”.

**[Expected result]:**
-  The “Open link in private tab” option is not displayed and it is not possible to have multiple tabs open.

**[Actual result]:**
-  The “Open link in private tab” option is displayed and it opens the link in a new tab.

**[Notes]:**
- The issue is not reproducible if step 2 is removed.
- The issue is also reproducible for the [Production experiment in Preview](https://experimenter.services.mozilla.com/nimbus/focus-android-ab-v97-experiment-new-users-copy), using Firefox Focus 98 built as Release (with the collection URL used being `https://firefox.settings.services.mozilla.com`).
- The issue is also reproducible for the Existing Users experiment.
- Attached is a [recording of the issue](https://drive.google.com/file/d/1o3XRH0PWQRPrXRoZ7i9S2l0LJr-_vfX-/view?usp=sharing).

heres-maria:
After further investigation, it seems the issue is related to whether the user navigates to a webpage between the startup they are enrolled in the study and the startup when the experiment is applied to the build. I have updated the STR and notes accordingly.

ionutbedregeanu:
[...]

PR #6575, opened 2022-03-04
For #6528 - Re-initialize current experiment variables when Nimbus updates

Mugurell:
This continues the work from
https://github.com/mozilla-mobile/focus-android/pull/6444 and will ensure that
after a new Nimbus configuration is downloaded and internally the experimental
features that depend on Nimbus will be updated to use the new values.

https://user-images.githubusercontent.com/11428869/156792379-4c895319-f35e-418b-8027-05005136c79d.mp4

### Pull Request checklist

3. click the `Artifacts` row,


app/src/main/java/org/mozilla/focus/Components.kt
```diff
@@ -209,9 +209,7 @@ class Components(
         )
     }
 
-    val experimentalFeatures by lazy {
-        ExperimentalFeatures(experiments)
-    }
+    var experimentalFeatures = ExperimentalFeatures { experiments }
```
Mugurell, line 212: Tried avoiding setting this as `var` and also keeping the lazy initialization.
Looked into using something like a mutable lazy but it didn't really work.
I'd say it's important to have this patch qa-tested and integrated in the codebase in preparation of the live experiment and figuring a better solution for the overall flow and configuration can come later to be used also on Fenix.

ionutbedregeanu (approved):

ionutbedregeanu:
I tested your last changes and everything seems ok 👍

jonalmeida (approved):
Ha, good catch - the `lazy` was caching the local value and not updating to the remote value.

This simplification works well for now, although I do wish in the future we could make experiments live in the `AppStore` and therefore we can observe the updates that way (or just query it directly). It would make things more streamlined instead of having multiple mechanisms for each component, but I've digressed.. 🙂

Mugurell:
> Ha, good catch - the `lazy` was caching the local value and not updating to the remote value.
> 
> This simplification works well for now, although I do wish in the future we could make experiments live in the `AppStore` and therefore we can observe the updates that way (or just query it directly). It would make things more streamlined instead of having multiple mechanisms for each component, but I've digressed.. 🙂

Totally agree with this.
Even if we're not using a lazy anymore it might still be possible for components to hold a reference to old experiments values.
Exposing the experiments and variables through an observable stream would mean instant effects. And this is a flow that would work great with Compose.

travis79 (approved):
So, it appears that the `lazy` instantiation of the `ExperimentFeatures` was causing some delay and removing it causes the fetch/apply to occur sooner on startup. I don't see any issues with this approach from a Nimbus perspective, looks good to me!

ionutbedregeanu:
@Mergifyio backport releases_v99.0

ionutbedregeanu:
@Mergifyio backport releases_v99.0

ionutbedregeanu:
@Mergifyio backport releases_v99.0

ionutbedregeanu:
@Mergifyio backport releases_v98.0

ionutbedregeanu:
@Mergifyio backport releases_v99.0

Issue #6542, opened 2022-03-01
[Experiment] Links set to open in new tabs / windows temporarily open in the background if clicked with tabs disabled


heres-maria:
**[Affected Versions]:**
- Firefox Focus 97.2.0 (Build #360591303) GV: 97.0.1-20220216172458
- Firefox Focus 98.0 (Build #360591110) GV: 98.0.20220224185616

**[Affected Platforms]:**
- Android 9 (Samsung Galaxy S8)
-  Android 8 (HUAWEI P20 lite)

**[Prerequisites]:**
- Have a Focus for Android build with the following modifications:
     The following lines added to local.properties:
`nimbus.remote-settings.url = https://settings.stage.mozaws.net/`
`debuggable = true`
A new “.nimbus” file (in the same location as local.properties) containing the following line:
`https://settings.stage.mozaws.net`
- Have a Firefox Nightly for desktop profile open on “about:debugging”.
- Have  “Enable USB devices” enabled on “about:debugging”.

**[Steps to reproduce]:**
1. Open the browser from prerequisites to enroll in [Stage Copy Focus Android A/B v97 Experiment - New Users](https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus/stage-copy-focus-android-ab-v97-experiment-new-users) in the treatment branch (the branch you have been enrolled in can be found looking at logcat logs at startup with the "nimbus" filter set).
2. Close and open the app again (to avoid #6528).
3. Go to “Settings” > “Advanced” and enable “Remote debugging via USB/Wi-Fi”.
4. On the desktop “about:debugging” page, tap the “Connect” button of the Focus app option.
5. Tap the Focus option to view open tabs.
6. Navigate to https://www.rapidtables.com/web/html/link/html-link-new-window.html.
7. Tap the two “Open page in new window” links.
8. Observe the “Tabs” section of the “about:debugging” window.

**[Expected result]:**
-  The page opens in the current tab.
- The “Tabs” section in “about:debugging” displays only one tab.

**[Actual result]:**
-  The page does not seemingly open at all.
- The “Tabs” section in “about:debugging” displays multiple tabs.

**[Notes]:**
- The additional tabs are removed from the “Tabs” section if the user then navigates to another page.
- The issue was also reproducible for links on this page: https://www.freecodecamp.org/news/how-to-use-html-to-open-link-in-new-tab/
- Attached is a recording of the issue.

[...]

Issue #6546, opened 2022-03-01
[Experiment] [Intermittent] The “Report Site Issue…” link either opens multiple tabs in the background or does not work if tabs are disabled


heres-maria:
**[Affected Versions]:**
- Firefox Focus 97.2.0 (Build #360591303) GV: 97.0.1-20220216172458
- Firefox Focus 98.0 (Build #360591110) GV: 98.0.20220224185616 built as Release

**[Affected Platforms]:**
- Android 9 (Samsung Galaxy S8)
- Android 8 (HUAWEI P20 lite)

**[Prerequisites]:**
- Have a Focus for Android build with the following modifications:
     The following lines added to local.properties:
`nimbus.remote-settings.url = https://settings.stage.mozaws.net/`
`debuggable = true`
A new “.nimbus” file (in the same location as local.properties) containing the following line:
`https://settings.stage.mozaws.net`
- Have a Firefox Nightly for desktop profile open on “about:debugging”.
- Have  “Enable USB devices” enabled.

**[Steps to reproduce]:**
1. Open the browser from prerequisites to enroll in [Stage Copy Focus Android A/B v97 Experiment - New Users](https://stage.experimenter.nonprod.dataops.mozgcp.net/nimbus/stage-copy-focus-android-ab-v97-experiment-new-users) in the treatment branch (the branch you have been enrolled in can be found looking at logcat logs at startup with the `nimbus` filter set).
2. Close and open the app again (to avoid #6528).
3. Go to “Settings” > “Advanced” and enable “Remote debugging via USB/Wi-Fi”.
4. On the desktop “about:debugging” page, click the “Connect” button of the Focus app option.
5. Navigate to a page.
6. Tap “Report Site Issue…” 
7. Observe the “Tabs” section of the “about:debugging” window.
8. Repeat steps 5 - 7 a few times.

**[Expected result]:**
-  The page opens in the current tab.
- The “Tabs” section in “about:debugging” displays only one tab.

**[Actual result]:**
-  [Intermittent] The page does not open at all.
-  [Intermittent] The page opens seemingly in the same tab, but the “Tabs” section in “about:debugging” displays multiple tabs.

**[Notes]:**
- The UI of the Focus app only displays one tab.
- Restarting the app will make the page open if stuck in the case when it does not open.
- Attached is a [recording of the issue](https://drive.google.com/file/d/1Ja_KQ087WT_Dno35xNRUPyWJw_0padLd/view?usp=sharing).

[...]

PR #6587, opened 2022-03-07
For #6542 - Open WindowRequest url in current tab if multitab is disabled

Mugurell:
The AC flow for the `window.open` Gecko api is:
- UpdateWindowRequestAction - which adds the WindowRequest to the current tab
- AddTabAction which will open a new tab - first with "about:blank" then in
which the requested window url will be loaded.
AddTabAction for "about:blank" is currently dropped if not supporting multitab.
So in scenario from a users' standpoint the app won't do anything.
- ConsumeWindowRequestAction which will cleanup the WindowRequest from the
current tab's state after the app acted upon that request.
Knowing that this flow is finished Focus will here load the url of the
requested window in the current tab while a bit sooner opening a new tab was
avoided.

https://user-images.githubusercontent.com/11428869/157041898-4b387b3a-d992-450d-bb1a-d9f9564a9f98.mp4

### Pull Request checklist

3. click the `Artifacts` row,


app/src/main/java/org/mozilla/focus/tabs/MergeTabsMiddleware.kt
```diff
@@ -31,6 +33,25 @@ class MergeTabsMiddleware(
         next: (BrowserAction) -> Unit,
         action: BrowserAction
     ) {
+        if (!hasTabs && action is ConsumeWindowRequestAction) {
+            val isRequestToCloseWindow =
+                context.state.privateTabs.first().content.windowRequest?.type == WindowRequest.Type.CLOSE
```
Mugurell, line 38: Tested on https://jsfiddle.net/da5z3egp/ and saw that while on desktop the window will be closed there is no such callback coming from GV for us.
Not sure if/how this request would get to use but the app should be ready to handle it.
In the above video I used a delayed `this.context.components.tabsUseCases.removeAllTabs()` to see how the app would act in the event such a `CLOSE` request would come to the app after a `OPEN` request was acted upon.

csadilek: Here's an alternative that should work fine: https://jsfiddle.net/8dp4huag/1/

We should file a follow-up to investigate why the close call doesn't work in that `setTimeout`, but that's not a blocker for this.

Mugurell:
> @Mugurell The flow here makes sense to me, but tabs added by the `WindowFeature` will also come with a new`engineSession` that has to be used. Same as we discussed in [#6546 (comment)](https://github.com/mozilla-mobile/focus-android/issues/6546#issuecomment-1061125171) and fixed by #6599.
> 
> So we should test this change on top of it to make sure we're not leaking engine sessions and everything works correctly.

Tabs are indeed leaking in this `window.open` scenario.
If allowing to try loading "about:blank" [here](https://github.com/mozilla-mobile/focus-android/blob/66c8d2f27492f8799e0201d0ceeb167cda85e3ea/app/src/main/java/org/mozilla/focus/tabs/MergeTabsMiddleware.kt#L109) and as such pass [this check](https://github.com/mozilla-mobile/focus-android/blob/66c8d2f27492f8799e0201d0ceeb167cda85e3ea/app/src/main/java/org/mozilla/focus/tabs/MergeTabsMiddleware.kt#L49) and get to execute the newly updated code from https://github.com/mozilla-mobile/focus-android/pull/6599 then I get into an infinite loop of tabs being closed and recreated - seemingly from the AC [`consumeWindowRequest` method](https://github.com/mozilla-mobile/android-components/blob/7475b34e7a18cf77222520fa593b529fdf80f39a/components/feature/tabs/src/main/java/mozilla/components/feature/tabs/WindowFeature.kt#L60).

csadilek (commented):
Ok, I think this can be rebased and landed now :). The `AddTabAction` should be processed first and merge the tabs correctly (using the new engine session), then the `ConsumeWindowRequestAction` should trigger loading the url.

The back button will not work as this is essentially a new window. With multiple tabs, the back button would go to the parent session, but since we have to close the parent session (single tab browser) there's nowhere to go back to.

csadilek:
> Tabs are indeed leaking in this window.open scenario.

@Mugurell I've rebased this on top of main and pushed to https://github.com/csadilek/focus-android/commit/604f3dee0a49ddf1d4f274f4af62b19aea6d30c3, which all works fine for me! I don't see any session leaks or an infinite recursion either. Can you double-check, maybe something went wrong in your local branch?

Mugurell:
> > Tabs are indeed leaking in this window.open scenario.
> 
> @Mugurell I've rebased this on top of main and pushed to [csadilek@604f3de](https://github.com/csadilek/focus-android/commit/604f3dee0a49ddf1d4f274f4af62b19aea6d30c3), which all works fine for me! I don't see any session leaks or an infinite recursion either. Can you double-check, maybe something went wrong in your local branch?

Was still seeing leaked EngineSessions in `about:debugging`.
Updated now the patch to merge supporting "open window requests" in the newly updated support for merging tabs.

ionutbedregeanu (approved):

Mugurell:
Tested and all scenarios seems good now.
Landing for Maria to also get the chance to test this in preparation for uplift.

ionutbedregeanu:
@Mergifyio backport releases_v99.0

ionutbedregeanu:
@Mergifyio backport releases_v99.0

ionutbedregeanu:
@Mergifyio backport releases_v99.0

ionutbedregeanu:
@Mergifyio backport releases_v99.0

ionutbedregeanu:
@Mergifyio backport releases_v98.0

PR #6700, opened 2022-03-18
Bootstrap Taskgraph from pypi in decision task

ahal:
Releng has tweaked the recommended way of bootstrapping Taskgraph from a
clone + checkout pinned revision, to Pypi and a requirements.txt.

This provides a few benefits:
1. Since Taskgraph uses SemVer, updating will provide a hint if there
    will be backwards incompatible changes.
2. Using version numbers is nicer than revisions (e.g, Taskgraph's
    CHANGELOG will be useful).
3. Provides a mechanism to pin other non-Taskgraph dependencies.

The reason I'm submitting this now is I'd like to migrate Taskgraph to a
new repository. By moving consumers over to Pypi first, the migration
can happen seamlessly without breaking anything.

Jira: RELENG-809


hneiva (approved):
LGTM 👍

ahal:
The failing task is failing on the main branch as well.

PR #6766, opened 2022-03-28
For #6531 improve coverage for download file types UI tests

AndiAJ:
For #6531 improve coverage for download file types UI tests

✅  `downloadAndOpenPDFFileTest` - Successfully passed 40x on Firebase
✅  `downloadAndOpenWEBMFileTest` - Successfully passed 40x on Firebase
✅  `verifyDownloadedFileOnStorageTest` - Successfully passed 40x on Firebase

### Pull Request checklist

3. click the `Artifacts` row,

### DIFF
--- a/app/src/androidTest/java/org/mozilla/focus/activity/DownloadFileTest.kt
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/DownloadFileTest.kt
@@ -161,4 +161,27 @@ class DownloadFileTest {
             assertNativeAppOpens(GOOGLE_DRIVE)
         }
     }
+
+    @SmokeTest
+    @Test
+    fun downloadAndOpenWebmFileTest() {
+        downloadFileName = "videoSample.webm"
+
+        searchScreen {
+        }.loadPage(downloadTestPage) {
+            progressBar.waitUntilGone(waitingTime)
+            clickLinkMatchingText(downloadFileName)
+        }
+        // If permission dialog appears on devices with API<30, grant it
+        if (permAllowBtn.waitForExists(waitingTime)) {
+            permAllowBtn.click()
+        }
+        downloadRobot {
+            verifyDownloadDialog(downloadFileName)
+            clickDownloadButton()
+            verifyDownloadConfirmationMessage(downloadFileName)
+            openDownloadedFile()
+            assertNativeAppOpens(GOOGLE_PHOTOS)
+        }
+    }
 }
--- a/app/src/androidTest/java/org/mozilla/focus/activity/DownloadFileTest.kt
+++ b/app/src/androidTest/java/org/mozilla/focus/activity/DownloadFileTest.kt
@@ -20,10 +20,10 @@ import org.mozilla.focus.helpers.MockWebServerHelper
 import org.mozilla.focus.helpers.RetryTestRule
 import org.mozilla.focus.helpers.StringsHelper.GOOGLE_PHOTOS
 import org.mozilla.focus.helpers.TestAssetHelper.getImageTestAsset
-import org.mozilla.focus.helpers.TestHelper
 import org.mozilla.focus.helpers.TestHelper.assertNativeAppOpens
 import org.mozilla.focus.helpers.TestHelper.getTargetContext
 import org.mozilla.focus.helpers.TestHelper.mDevice
+import org.mozilla.focus.helpers.TestHelper.permAllowBtn
 import org.mozilla.focus.helpers.TestHelper.verifySnackBarText
 import org.mozilla.focus.helpers.TestHelper.waitingTime
 import org.mozilla.focus.testAnnotations.SmokeTest
@@ -33,6 +33,8 @@ import java.io.IOException
 class DownloadFileTest {
     private lateinit var webServer: MockWebServer
     private val featureSettingsHelper = FeatureSettingsHelper()
+    private val downloadTestPage = "https://storage.googleapis.com/mobile_test_assets/test_app/downloads.html"
+    private var downloadFileName: String = ""
 
     @get:Rule
     var mActivityTestRule = MainActivityIntentsTestRule(showFirstRun = false)
@@ -58,7 +60,7 @@ class DownloadFileTest {
         } catch (e: IOException) {
             throw AssertionError("Could not stop web server", e)
         }
-        deleteFileUsingDisplayName(getTargetContext.applicationContext, "download.jpg")
+        deleteFileUsingDisplayName(getTargetContext.applicationContext, downloadFileName)
         featureSettingsHelper.resetAllFeatureFlags()
     }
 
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/org/mozilla/focus/activity/DownloadFileTest.kt
```diff
@@ -135,4 +140,78 @@ class DownloadFileTest {
...
+    @Test
+    fun downloadAndOpenPDFFileTest() {
+        downloadFileName = "washington.pdf"
+
+        searchScreen {
+        }.loadPage(downloadTestPage) {
+            progressBar.waitUntilGone(waitingTime)
+        }
+        browserScreen {
+            clickLinkMatchingText(downloadFileName)
```
sv-ohorvath, line 154: nit: Could this go in the same brackets as L151?

app/src/androidTest/java/org/mozilla/focus/activity/DownloadFileTest.kt
```diff
@@ -116,7 +121,7 @@ class DownloadFileTest {
     @Test
     fun openDownloadFileTest() {
```
sv-ohorvath, line 122: Should we maybe rename this as downloadAndOpenJpgFileTest to match with the others?

app/src/androidTest/java/org/mozilla/focus/activity/DownloadFileTest.kt
```diff
@@ -135,4 +140,78 @@ class DownloadFileTest {
...
+            verifyDownloadConfirmationMessage(downloadFileName)
+            openDownloadedFile()
+            assertNativeAppOpens(GOOGLE_PHOTOS)
+        }
+    }
+
+    @SmokeTest
+    @Test
+    fun verifyDownloadedFileOnStorageTest() {
+        downloadFileName = "executable.exe"
```
sv-ohorvath, line 197: An exe file is not really important on Android, I would use a more commonly used type of file: apk, mp3, xls, etc

app/src/androidTest/java/org/mozilla/focus/activity/DownloadFileTest.kt
AndiAJ: Done!

app/src/androidTest/java/org/mozilla/focus/activity/DownloadFileTest.kt
AndiAJ: Done!

app/src/androidTest/java/org/mozilla/focus/activity/DownloadFileTest.kt
AndiAJ: Done!

app/src/androidTest/java/org/mozilla/focus/helpers/TestHelper.kt
```diff
@@ -145,6 +147,17 @@ object TestHelper {
     fun verifyTranslatedTextExists(text: String) =
         assertTrue(mDevice.findObject(UiSelector().text(text)).waitForExists(waitingTime))
 
+    @Suppress("Deprecation")
+    fun verifyDownloadedFileOnStorage(fileName: String) {
+        runBlocking {
+            val downloadedFile = File(
+                Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS),
+                fileName
```
AaronMT, line 155: I don't think we should use this in Android 10+ even with a deprecation warning, it's not best practice. Can we try `getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS)` ? I think we still need the runtime grants of `READ_EXTERNAL_STORAGE` There's also [1] talking about `MANAGE_EXTERNAL_STORAGE` 

[1] https://developer.android.com/training/data-storage/manage-all-files

Mind trying this out?

AaronMT (changes requested):

app/src/androidTest/java/org/mozilla/focus/helpers/TestHelper.kt
AndiAJ: Hi @AaronMT from my understanding `getExternalFilesDir` actually returns the path to files from our application
https://stackoverflow.com/a/10124040 so it doesn't really help in our case.

@Mugurell was kind enough to help me out with a different approach similar to what we have on AC.

I've re-ran the the test 40x on Firebase and successfully passed. ✅ 

Please take a look and share your thoughts. 😊

Mugurell: MediaStore can be used on API 30+ - https://developer.android.com/training/data-storage/shared/media and the code to use it in these tests closely follows [the code used in AC for downloads](https://github.com/mozilla-mobile/android-components/blob/6dc07542819cfb29b4bf2a77f2eb8df37f1d54ea/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/AbstractFetchDownloadService.kt#L963-L1007).

For API <=29 `getExternalStoragePublicDirectory` should be safe to use following a check for `SDK_INT`. It is deprecated starting with API 29. 
(And will be undeprecated in the next Android version. Already undeprecated in Android 12L - https://developer.android.com/sdk/api_diff/32/changes. Google..).

AaronMT: If you don't mind, @Mugurell, do you have the time to review this. Everything else looks OK to me, I'm familiar with the code that And wrote in `queryDownloadMediaStore`

AndiAJ: Hi @AaronMT , as previously mentioned @Mugurell helped me out with this.
But sure, please go ahead and leave a review. 
Thanks!

Mugurell (approved):
r+ from me for checking the existence of the downloaded file.
It's the approach Google recommends starting with Android 10 and the scoped storage restrictions and although with a bit more code it allows for a cleaner and a more secure file access with the apps not having to know the full path of it's files but being able to access their own without restrictions.
I understand this was tested locally also with just calling `verifyDownloadedFileOnStorage` with a different `filename` case in which the test failed as expected so it seems to be a reliable method.

app/src/androidTest/java/org/mozilla/focus/helpers/TestHelper.kt
```diff
@@ -162,6 +171,70 @@ object TestHelper {
...
+
+        assertTrue(fileExists)
+    }
+
+    /**
+     * Check the "Downloads" public directory for [fileName] and returns an URI to it.
+     * May be `null` if there is no file with that name in "Downloads".
+     */
+    @TargetApi(Build.VERSION_CODES.Q)
+    @VisibleForTesting
```
Mugurell, line 193: This line is not needed.
It would allow to relax it's visibility just for when it is being tested - we used this for methods that were to be unit tested. Here it's not the case.

AndiAJ: Sure, consider it done!
Many thanks! 🙏🏽

Issue #6545, opened 2022-03-01
Add unit tests for erase cfr in BrowserToolbarIntegrationTest


ionutbedregeanu:
At this moment only one test is written for erase cfr popup in BrowserToolbarIntegrationTest.
Since one of our goals is to improve the coverage of unit tests in Focus I added this ticket as a reminder to increase the coverage of BrowserToolbarIntegrationTest by adding another tests for erase cfr popup

Mugurell:
Thank you for opening this Ionut!
As a drive-by I wanted to propose also adding tests for the newly added `CFRMiddleware` also to know that we have this bigger feature tested and safer from future regressions.

ionutbedregeanu:
This ticket was merged, but can't be tested by QA, so I will close it

PR #6839, opened 2022-04-06
Add unit tests for CfrMiddleware.kt

ionutbedregeanu:
For #6545
Add unit tests just for CfrMiddleware.kt because erase cfr tests fromBrowserToolbarIntegrationTest were updated in a previous pr

### Pull Request checklist

3. click the `Artifacts` row,

### DIFF
--- a/app/src/main/java/org/mozilla/focus/utils/Features.kt
+++ b/app/src/main/java/org/mozilla/focus/utils/Features.kt
@@ -16,7 +16,7 @@ object Features {
     /**
      * Erase tabs CFR
      */
-    var IS_ERASE_CFR_ENABLED = AppConstants.isDevOrNightlyBuild
+    val IS_ERASE_CFR_ENABLED = AppConstants.isDevOrNightlyBuild
 
     /**
      * Replace the current [FirstrunFragment.kt] which has multiple cards with
--- a/app/src/test/java/org/mozilla/focus/cfr/CfrMiddlewareTest.kt
+++ b/app/src/test/java/org/mozilla/focus/cfr/CfrMiddlewareTest.kt
@@ -1,0 +1,154 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package org.mozilla.focus.cfr
+
+import mozilla.components.browser.state.action.ContentAction
+import mozilla.components.browser.state.action.TabListAction
+import mozilla.components.browser.state.state.BrowserState
+import mozilla.components.browser.state.state.SecurityInfoState
+import mozilla.components.browser.state.state.TabSessionState
+import mozilla.components.browser.state.state.createTab
+import mozilla.components.browser.state.store.BrowserStore
+import mozilla.components.support.test.ext.joinBlocking
+import mozilla.components.support.test.libstate.ext.waitUntilIdle
+import mozilla.components.support.test.robolectric.testContext
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mock
+import org.mockito.Mockito.doReturn
+import org.mockito.MockitoAnnotations
+import org.mozilla.focus.Components
+import org.mozilla.focus.state.AppState
+import org.mozilla.focus.state.AppStore
+import org.mozilla.focus.state.Screen
+import org.mozilla.focus.utils.Features
+import org.mozilla.focus.utils.Settings
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class CfrMiddlewareTest {
+    private lateinit var cfrMiddleware: CfrMiddleware
+    private lateinit var browserStore: BrowserStore
+    private lateinit var appStore: AppStore
+    private lateinit var settings: Settings
+
+    @Mock
+    private lateinit var components: Components
+
+    @Before
+    fun setUp() {
+        MockitoAnnotations.openMocks(this)
+
+        cfrMiddleware = CfrMiddleware(components)
+        browserStore = BrowserStore(
+            initialState = BrowserState(),
+            middleware = listOf(cfrMiddleware)
+        )
+        appStore = AppStore(
+            AppState(
+                screen = Screen.Home,
+                showTrackingProtectionCfr = false,
+            )
+        )
+        settings = Settings(testContext)
+
+        doReturn(appStore).`when`(components).appStore
+        doReturn(settings).`when`(components).settings
+    }
+
+    @Test
+    fun `GIVEN erase cfr is enabled and tracking protection cfr is not displayed WHEN AddTabAction is intercepted THEN the numberOfTabsOpened is increased`() {
+        if (Features.IS_ERASE_CFR_ENABLED) {
+            browserStore.dispatch(TabListAction.AddTabAction(createTab())).joinBlocking()
+
+            assertEquals(1, components.settings.numberOfTabsOpened)
+        }
+    }
+
+    @Test
+    fun `GIVEN erase cfr is enabled and tracking protection cfr is not displayed WHEN AddTabAction is intercepted for the third time THEN showEraseTabsCfr is changed to true`() {
+        if (Features.IS_ERASE_CFR_ENABLED) {
+            browserStore.dispatch(TabListAction.AddTabAction(createTab(tabId = 1))).joinBlocking()
+            browserStore.dispatch(TabListAction.AddTabAction(createTab(tabId = 2))).joinBlocking()
+            browserStore.dispatch(TabListAction.AddTabAction(createTab(tabId = 3))).joinBlocking()
+            appStore.waitUntilIdle()
+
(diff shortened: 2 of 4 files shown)


app/src/test/java/org/mozilla/focus/cfr/CfrMiddlewareTest.kt
```diff
@@ -0,0 +1,141 @@
...
+import org.mockito.Mockito.verify
+import org.mockito.MockitoAnnotations
+import org.mozilla.focus.Components
+import org.mozilla.focus.state.AppAction
+import org.mozilla.focus.state.AppState
+import org.mozilla.focus.state.AppStore
+import org.mozilla.focus.utils.Settings
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
```
Mugurell, line 31: With almost everything mocked and no interactions with Android components I don't think we need to use this more expensive runner.

app/src/test/java/org/mozilla/focus/cfr/CfrMiddlewareTest.kt
```diff
@@ -0,0 +1,141 @@
...
+        whenever(components.appStore).thenReturn(appStore)
+        whenever(components.appStore.state).thenReturn(appState)
+        whenever(middlewareContext.state).thenReturn(browserState)
+        whenever(components.settings).thenReturn(settings)
+
+        cfrMiddleware = CfrMiddleware(components)
+    }
+
+    @Test
+    fun `GIVEN a new tab is added WHEN erase cfr is enabled and tracking protection cfr is not displayed THEN the numberOfTabsOpened is increased and checked`() {
```
Mugurell, line 70: Here and below `given` and `when` seems to be missmatched.
- `Given` refers to the initial context, all preconditions
- `When` refers to the event that triggers the outcome we want to test
- `Then` describes the expected outcome that we want to test.

https://cucumber.io/docs/gherkin/reference/
https://cucumber.io/blog/bdd/cucumber-anti-patterns-part-two/#:~:text=No%20clear%20separation%20between%20Given%2C%20When%2C%20and%20Then

Mugurell (commented):
LGTM but this way of testing middlewares is different than what we use on [AC](https://github.com/mozilla-mobile/android-components/blob/71e124a5dbafffb800168ea615fdbc54d7398f50/components/support/locale/src/test/java/mozilla/components/support/locale/LocaleMiddlewareTest.kt) and [Fenix](https://github.com/mozilla-mobile/fenix/blob/26713a58c83bb79ff260034062e2895cb929bc3a/app/src/test/java/org/mozilla/fenix/home/recentvisits/HistoryMetadataMiddlewareTest.kt) where we use more like integration tests with actually observing the state updates after actions pass through middlewares.

Since this is the first middleware test on Focus I think it's important to give a clear direction on what the approach should be for the future.
Asking @csadilek for this.

app/src/test/java/org/mozilla/focus/cfr/CfrMiddlewareTest.kt
```diff
@@ -0,0 +1,122 @@
...
+    private lateinit var appStore: AppStore
+    private lateinit var settings: Settings
+
+    @Mock
+    private lateinit var components: Components
+
+    @Before
+    fun setUp() {
+        MockitoAnnotations.openMocks(this)
+        cfrMiddleware = spy(CfrMiddleware(components))
```
csadilek, line 45: If possible, I would strongly encourage not using spies here. There should be another way e.g. worst case by separating the `shouldShowCfrForTrackingProtection` call from the middleware, but looks like there's a simpler way?

The check is based on the action itself which is under test, and the component settings and state which can be set in the test as well so that `shouldShowCfrForTrackingProtection` either return true or false, as needed.

csadilek (commented):
Thank you, this is looking great now. I have one comment below re: spies, we should try and improve.

app/src/main/java/org/mozilla/focus/utils/Features.kt
```diff
@@ -16,7 +16,7 @@ object Features {
     /**
      * Erase tabs CFR
      */
-    var IS_ERASE_CFR_ENABLED = AppConstants.isDevOrNightlyBuild
+    val IS_ERASE_CFR_ENABLED = AppConstants.isDevOrNightlyBuild
```
Mugurell, line 19: Just wanted to commend this change.
Great that a feature depending on the build type is not a `var` anymore.

app/src/test/java/org/mozilla/focus/cfr/CfrMiddlewareTest.kt
```diff
@@ -0,0 +1,152 @@
...
+        )
+        settings = Settings(testContext)
+
+        doReturn(appStore).`when`(components).appStore
+        doReturn(settings).`when`(components).settings
+    }
+
+    @Test
+    fun `GIVEN erase cfr is enabled and tracking protection cfr is not displayed WHEN AddTabAction is intercepted THEN the numberOfTabsOpened is increased`() {
+        if (AppConstants.isDevOrNightlyBuild) {
```
csadilek, line 64: Can you check for `IS_ERASE_CFR_ENABLED` here instead? Because then when this feature is ultimately enabled in release the tests will just run/work automatically and not require a change.

app/src/test/java/org/mozilla/focus/cfr/CfrMiddlewareTest.kt
```diff
@@ -0,0 +1,152 @@
...
+
+    @Test
+    fun `GIVEN erase cfr is enabled and tracking protection cfr is not displayed WHEN AddTabAction is intercepted for the third time THEN showEraseTabsCfr is changed to true`() {
+        if (AppConstants.isDevOrNightlyBuild) {
+            browserStore.dispatch(TabListAction.AddTabAction(createTab(tabId = 1))).joinBlocking()
+            browserStore.dispatch(TabListAction.AddTabAction(createTab(tabId = 2))).joinBlocking()
+            browserStore.dispatch(TabListAction.AddTabAction(createTab(tabId = 3))).joinBlocking()
+            appStore.waitUntilIdle()
+
+            assertEquals(true, appStore.state.showEraseTabsCfr)
```
csadilek, line 79: nit: use assertTrue

app/src/test/java/org/mozilla/focus/cfr/CfrMiddlewareTest.kt
```diff
@@ -0,0 +1,152 @@
...
+                SecurityInfoState(
+                    secure = true,
+                    host = "test.org",
+                    issuer = "Test"
+                )
+            )
+            browserStore.dispatch(updateSecurityInfoAction).joinBlocking()
+            appStore.waitUntilIdle()
+
+            assertEquals(true, appStore.state.showTrackingProtectionCfr)
```
csadilek, line 97: nit: use assertTrue

app/src/test/java/org/mozilla/focus/cfr/CfrMiddlewareTest.kt
```diff
@@ -0,0 +1,152 @@
...
+                    host = "test.org",
+                    issuer = "Test"
+                )
+            )
+
+            browserStore.dispatch(TabListAction.AddTabAction(insecureTab)).joinBlocking()
+            browserStore.dispatch(updateSecurityInfoAction).joinBlocking()
+            appStore.waitUntilIdle()
+
+            assertEquals(false, appStore.state.showTrackingProtectionCfr)
```
csadilek, line 118: nit: assert false.

app/src/test/java/org/mozilla/focus/cfr/CfrMiddlewareTest.kt
```diff
@@ -0,0 +1,152 @@
...
+                    secure = true,
+                    host = "test.org",
+                    issuer = "Test"
+                )
+            )
+            browserStore.dispatch(TabListAction.AddTabAction(mozillaTab)).joinBlocking()
+            browserStore.dispatch(updateSecurityInfoAction).joinBlocking()
+            appStore.waitUntilIdle()
+
+            assertEquals(false, appStore.state.showTrackingProtectionCfr)
```
csadilek, line 138: nit: assertFalse

csadilek (commented):
Nice! Almost there :). Had a nit and a question below re: the feature flag.

app/src/test/java/org/mozilla/focus/cfr/CfrMiddlewareTest.kt
```diff
@@ -0,0 +1,154 @@
...
+        )
+        settings = Settings(testContext)
+
+        doReturn(appStore).`when`(components).appStore
+        doReturn(settings).`when`(components).settings
+    }
+
+    @Test
+    fun `GIVEN erase cfr is enabled and tracking protection cfr is not displayed WHEN AddTabAction is intercepted THEN the numberOfTabsOpened is increased`() {
+        if (Features.IS_ERASE_CFR_ENABLED) {
```
csadilek, line 66: If you want to cover all ground, you could also add a test that verifies that nothing happens when the feature flag is off.

csadilek (approved):
👍

app/src/test/java/org/mozilla/focus/cfr/CfrMiddlewareTest.kt
ionutbedregeanu: Hmm, this implies to override the current config build for this test, so I will let it as it is for now

Issue #6166, opened 2022-01-05
Fingerprint auth does not protect the home screen (shortcuts)


jonalmeida:
## Steps to reproduce
1. Add a few shortcuts to Focus.
2. Enable fingerprint auth.
3. Swipe away Focus and re-open the app.

### Expected behaviour
- The user sees the fingerprint auth. (UX, is this a correct expectation?)

### Actual behaviour
- There is no auth needed to see your shortcuts.

### Device information

* Android device: All devices
* Focus version: All nightly versions.

amedyne:
@jeffreygee Thoughts on this?

jonalmeida:
Clarification: the reason this is a new issue, is because previously we never had any personal or identifiable data on the Focus home screen. With the addition of Shortcuts,  a user can have information on the home screen that they might not want to be accessible when they have the security auth feature turned on.

jeffreygee:
@amedyne @jonalmeida I'm thinking if we're going to be more privacy/security-focused, this tool should be more flexible. If I was turning this feature on as a user, I would expect that any entry points (either opening the app for the first time or returning to the app) would trigger this. That way it's not page or shortcut based - it's more of a general security feature. Thoughts on this? Happy to discuss

jonalmeida:
That's the direction I was headed towards as well. 👍 

i.e. if the feature is enabled, any visit to the app should require auth.

iorgamgabriel:
So I made the changes . If the user has at least one Shortcut , the lockscreen will appear every time the user enters in the app , doesn't matter the last visited screen .

iorgamgabriel:
<img src="https://user-images.githubusercontent.com/8118371/162430818-46307f9c-eaca-4786-aaf7-35e7559c4076.jpg" width="300" height="600" />

iorgamgabriel:
I think the text "Your fingerprint can unlock Focus  if a URL is already opened in the app. Stealth will be turned on." should be changed . @amedyne @jeffreygee

lobontiumira:
This issue is not fixed yet. Tested on the latest Focus Nightly build from 4/13 with Oppo Find X3 Lite (Android 11), and Samsung Galaxy Note 8 (Android 9).
The STR are:
1. Set a fingerprint auth on the device.
2. Enable the "Use fingerprint to unlock app" on Focus.
3. Open a page, and set is as a shortcut.
4. Close Focus, by swiping away from the recent apps.
5. Re-open Focus.

[...]

PR #6873, opened 2022-04-14
for #6166 Fingerprint auth  protects the home screen (shortcuts)

iorgamgabriel:
I fixed the bug when the user closes the app and removed it from background the LockScreen doesn't appear .
When the user clicks on "New Session" button all the shortcuts will be deleted. 

### Pull Request checklist

3. click the `Artifacts` row,

### DIFF
--- a/app/src/main/java/org/mozilla/focus/biometrics/BiometricAuthenticationDialogFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/biometrics/BiometricAuthenticationDialogFragment.kt
@@ -20,6 +20,7 @@ import org.mozilla.focus.R
 import org.mozilla.focus.databinding.BiometricPromptDialogContentBinding
 import org.mozilla.focus.ext.requireComponents
 import org.mozilla.focus.state.AppAction
+import org.mozilla.focus.topsites.TopSitesIntegration
 
 @RequiresApi(api = Build.VERSION_CODES.M)
 @Suppress("TooManyFunctions")
@@ -27,10 +28,12 @@ class BiometricAuthenticationDialogFragment : AppCompatDialogFragment(), Lifecyc
     private var handler: BiometricAuthenticationHandler? = null
     private var _binding: BiometricPromptDialogContentBinding? = null
     private val binding get() = _binding!!
+    private val topSitesIntegration by lazy { TopSitesIntegration(requireComponents.topSitesStorage) }
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setStyle(DialogFragment.STYLE_NORMAL, R.style.DialogStyle)
+        lifecycle.addObserver(topSitesIntegration)
     }
 
     override fun onCancel(dialog: DialogInterface) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/mozilla/focus/biometrics/BiometricAuthenticationDialogFragment.kt
```diff
@@ -86,16 +92,35 @@ class BiometricAuthenticationDialogFragment : AppCompatDialogFragment(), Lifecyc
...
     }
 
     private fun biometricNewSessionButtonClicked() {
         requireComponents.tabsUseCases.removePrivateTabs()
+        deleteAllTopSites()
+        requireComponents.appStore.dispatch(AppAction.ShowHomeScreen)
         dismiss()
     }
 
+    private fun deleteAllTopSites() {
```
ionutbedregeanu, line 112: I think we should create a separate integration file were to move this kind of logic, it will not be a big step into separation of concerns, but it will be an improvement

ionutbedregeanu:
I reproduced a bug.
Pixel 3a Android 11:
1. Enable Use fingerprint to unlock app
2. Navigate to wikipedia and add it as shortcut
3. Send app in background and open it again
4. Press fingerprint sensor with an unrecognisable finger
5. Press New Session
6. The shortcut is still visible

app/src/main/java/org/mozilla/focus/biometrics/BiometricAuthenticationDialogFragment.kt
iorgamgabriel: done

iorgamgabriel:
I made the changes with LifecycleAwareFeature

ionutbedregeanu (approved):

ionutbedregeanu:
> I reproduced a bug. Pixel 3a Android 11:
> 
> 1. Enable Use fingerprint to unlock app
> 2. Navigate to wikipedia and add it as shortcut
> 3. Send app in background and open it again
> 4. Press fingerprint sensor with an unrecognisable finger
> 5. Press New Session
> 6. The shortcut is still visible

This is an intermitent bug and after a discussion with Gabriel we decide to postpone it until the new biometric component is implemented

app/src/main/java/org/mozilla/focus/biometrics/BiometricAuthenticationDialogFragment.kt
```diff
@@ -93,6 +95,8 @@ class BiometricAuthenticationDialogFragment : AppCompatDialogFragment(), Lifecyc
 
     private fun biometricNewSessionButtonClicked() {
         requireComponents.tabsUseCases.removePrivateTabs()
+        requireComponents.appStore.dispatch(AppAction.ShowHomeScreen)
+        topSitesIntegration.deleteAllTopSites()
```
jonalmeida, line 99: @iorgamgabriel When the 'New session' button is clicked, we remove the users top sites? Those are persistent through a session. Is that really what we want?

You posted this in the linked issue, but it's not clear to me if @jeffreygee acknowledged it.
> When the users clicks on "New Session" button from Lockscreen all the shortcuts will be deleted.

Raising this because, as a user, I would not want and have expected this to happen to my shortcuts.

jonalmeida: If it is intentional, maybe we need to add some copy that tells the user that turning on fingerprint to unlock can cause you to lose your shorcuts?

iorgamgabriel: @jonalmeida If the lockscreen appears if you have one shortcut I assume that this is a security issue . So if the user can't login in the app the shortcut will be deleted. I documented in the task . https://github.com/mozilla-mobile/focus-android/issues/6166#issuecomment-1092780991

Issue #6792, opened 2022-03-30
[custom tab] Some elements from the toolbar are white


lobontiumira:
## Steps to reproduce
1. Make Focus Nightly your default browser.
2. From Gmail open a link on a custom tab.
3. Observe the back arrow, and share icon.

### Expected behavior
All the elements have the same colour.

### Actual behavior
The back arrow, and share icon are white, no matter which theme the device is set on.

### Device information
* Android devices: Oppo Find X3 Lite (Android 11), Google Pixel (Android 10)
* Not reproducible on Lenovo tablet M10 (Android 10)
* Focus version: Focus Nightly from 3/30
* Not reproducible on RC 99.1.0
<img src = https://user-images.githubusercontent.com/40694158/160876826-ccf0ba64-6421-422f-9f9e-39d0afc9a511.jpg  width="40%" />

ionutbedregeanu:
Should be fixed after https://github.com/mozilla-mobile/focus-android/pull/6751 merge

lobontiumira:
Reproducible on Beta 100.0.0-beta.1 with Lenovo Tab M10 (Android 10).

lobontiumira:
Reproducible on Beta 100.0.0-beta.2 with Oppo Find X3 Lite (Android 11):

![Screenshot_3](https://user-images.githubusercontent.com/40694158/162428707-6762acdb-5c63-4372-8ac5-716c71b2ff36.jpg)

ionutbedregeanu:
@jeffreygee @amedyne What color should the toolbar have in custom tab?

jeffreygee:
@ionutbedregeanu @amedyne - all icons in the toolbar on light mode should be #291D4F

iorgamgabriel:
Focus supports any color  to the toolbar that an external app sends in the intent when it opens a Custom Tab . Now we should set the color of the icons from the toolbar to be in contrast with the toolbar background color . I think this is the bug . 
We should support something like Google Chrome https://developer.chrome.com/docs/android/custom-tabs/integration-guide/#opening-a-custom-tab .
@jeffreygee  do you agree?

iorgamgabriel:
No , we are setting the color of the toolbar based on what the external app sends . One of the most important aspects of Custom Tabs should be  the ability for an external app to change the color of the toolbar  to be consistent with the app's theme. . We should programmatic set the color of the icons based on the toolbar background color. @jeffreygee

jeffreygee:
[...]

PR #6987, opened 2022-05-09
for #6792 [custom tab] Fixed Toolbar icons color. 

iorgamgabriel:
Fixed Toolbar icons colors and  base on toolbar color set the same color for StatusBar and NavigationBar

 

 

### Pull Request checklist

3. click the `Artifacts` row,

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
@@ -160,9 +160,12 @@ class BrowserToolbarIntegration(
                 toolbar,
                 sessionId = customTabId,
                 useCases = customTabsUseCases,
+                closeListener = { fragment.closeCustomTab() },
+                updateToolbarBackground = true,
+                forceActionButtonTinting = false,
                 menuBuilder = menu.menuBuilder,
                 menuItemIndex = menu.menuBuilder.items.size - 1,
-                closeListener = { fragment.closeCustomTab() }
+                window = fragment.activity?.window,
             )
         }
 


app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -364,10 +370,38 @@ class BrowserFragment :
...
         )
-
+        setColorsForCustomTab(customTabConfig)
         setSitePermissions(view)
     }
 
+    /**
+     * Base on toolbar color set the same color for StatusBar and NavigationBar ,if the tab is custom.
+     */
+    private fun setColorsForCustomTab(customTabConfig: CustomTabConfig?) {
```
jonalmeida, line 380: Why do we need to do this? Does passing the `window` in the constructor of the custom tabs feature not work properly?

In general, when we try to fix bugs, we should consider:
1. Does it affect all our other browsers?
2. Do we want to change the behaviour of just _this particular_ browser or all our browsers?

Accordingly, we should decide if a change should be upstreamed to Android Components or if the feature can be written in a re-usable way so that our other browsers can also "compose" them into their app.

jonalmeida:
Slightly related: I see that this commit (and another before that is already merged) has the commit author set as our bot "MickeyMoz" with Sebastian's email address. Could you please update the commit to reflect that the change is made by you? 🙂 

I think the git commands are:
```
git config --global user.email <your email here>
```

app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
iorgamgabriel: I made the changes and I opened another pull request because of the commit author 
https://github.com/mozilla-mobile/focus-android/pull/6996

iorgamgabriel: I will closed this one

iorgamgabriel: @jonalmeida

PR #7027, opened 2022-05-16
DO NOT LAND - Test older GV with newest AC

Mugurell:
Latest try to update the AC version had many ui test failing - https://github.com/mozilla-mobile/focus-android/pull/7026/checks?check_run_id=6447251975
Let's try the same but with a slightly older GV version.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -35,6 +35,14 @@ android {
         multiDexEnabled true
     }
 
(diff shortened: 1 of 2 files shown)


rvandermeulen:
The regressing change was backed out from mozilla-central and post-backout #7034 has now successfully landed.

Issue #7052, opened 2022-05-19
Fix focus memory leak TipManager.listOfTips 


iorgamgabriel:
All memory leaks are reported by LeakCanary.  LeakCanary is a memory leak detection library for Android.

<img src="https://user-images.githubusercontent.com/8118371/168751023-25c30487-7b41-4016-ab54-26a816c46426.png" width="300" height="600" />

* Android device: all devices 
* Focus version: all version

PR #7055, opened 2022-05-19
for #7052 Fix focus memory leak TipManager.listOfTips

iorgamgabriel:
All memory leaks are reported by LeakCanary. LeakCanary is a memory leak detection library for Android.
Tests, Screenshots, Accessibility not needed .

### Pull Request checklist

3. click the `Artifacts` row,


app/src/main/java/org/mozilla/focus/fragment/UrlInputFragment.kt
```diff
@@ -231,7 +231,7 @@ class UrlInputFragment :
 
     private fun updateTipsLabel() {
         val context = context ?: return
-        val tips = TipManager.getAvailableTips(context)
+        val tips = TipManager().getAvailableTips(context)
```
jonalmeida, line 234: I understand why you've chosen to make this an instance instead of the original singleton (so that we don't leak the context). However, similar to my comment above, I think we would regress in performance if we're doing this work every time we focus on the `UrlInputFragment`.

Why not create a lazy singleton that takes the context as a parameter so you can call `getAvailableTips` safely?


```kotlin
// In Components.kt
val tipManager by lazy { TipManager(context) }

// In UrlInputFragment.kt
val tips = context.components.tipManager.getAvailableTips()
```

Does that work?

iorgamgabriel: It's working . I made the changes . Thank you!

jonalmeida: Excellent, I'm glad it worked out!

jonalmeida (approved):
📝 💧 ("memo"-ry leak)

😄

Issue #7057, opened 2022-05-19
The text for FingerPrint subtitle should be change 


iorgamgabriel:
The text for FingerPrint subtitle from Settings -> Privacy & Security should be change .
Now the text is  "Your fingerprint can unlock Focus if a URL is already opened in the app. Stealth will be turned on."
Here it was the initial request https://github.com/mozilla-mobile/focus-android/issues/6166 

<img src="https://user-images.githubusercontent.com/8118371/162430818-46307f9c-eaca-4786-aaf7-35e7559c4076.jpg" width="300" height="600" /> 

* Android device: all devices 
* Focus version: all versions

amedyne:
Tagging @jeffreygee for input.

lobontiumira:
Verified as fixed on the latest Focus Nightly 102.0a1 build from 5/26 with Google Pixel (Android 10), and Sony Xperia Z5 Premium (Android 7.1.1).

![Screenshot_20220715-065410](https://user-images.githubusercontent.com/40694158/170488856-069c6841-9f44-436f-9c0e-d310b210f707.png)

PR #7106, opened 2022-05-25
for #7057 Update the fingerprint settings summary

iorgamgabriel:
### Pull Request checklist

3. click the `Artifacts` row,

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -299,9 +299,11 @@
     <string name="preference_security_biometric">Use fingerprint to unlock app</string>
     <!-- Preference Summary: Biometrics (iris scan, fingerprint, or facial scan) can be used to unlock Focus if the app had a URL
         open when the app was last run to ensure that only the phone's owner can view a previous session.
-        %1$s will be replaced by the app name (e.g. Firefox Focus) -->
-    <string name="preference_security_biometric_summary">Your fingerprint can unlock %1$s if a URL is already opened in the app.
+        %s will be replaced by the app name (e.g. Firefox Focus)
+        Shortcuts is a feature that allows the user to pin his favorite websites to Focus home screen so he can quickly access them-->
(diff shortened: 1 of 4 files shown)


Mugurell (changes requested):
Whenever the value of an existing string changes we need to actually add a new string (same name as before just with a monotonically incremented suffix, in this case `preference_security_biometric_summary2`) for localizers to get this new string for translating.

A system for removing the old strings similar to Fenix - https://github.com/mozilla-mobile/fenix/wiki/Removing-strings should probably be considered for Focus also.

nit: Commit title should use the imperative mood - https://github.com/google/eng-practices/blob/master/review/developer/cl-descriptions.md#first-line-firstline. In this case maybe `Update the fingerpring settings summary` would offer just a bit more context.

app/src/main/res/values/strings.xml
```diff
@@ -300,8 +300,9 @@
     <!-- Preference Summary: Biometrics (iris scan, fingerprint, or facial scan) can be used to unlock Focus if the app had a URL
         open when the app was last run to ensure that only the phone's owner can view a previous session.
         %1$s will be replaced by the app name (e.g. Firefox Focus) -->
-    <string name="preference_security_biometric_summary">Your fingerprint can unlock %1$s if a URL is already opened in the app.
+    <string name="preference_security_biometric_summary" moz:removedIn="102" tools:ignore="UnusedResources">Your fingerprint can unlock %1$s if a URL is already opened in the app.
     Stealth will be turned on.</string>
+    <string name="preference_security_biometric_summary2">Unlock using fingerprint if you’ve added Shortcuts or when a website is already open in %1$s.</string>
```
Mugurell, line 305: Looking at the screenshot posted by Jeff there is no punctuation mark at the end of these strings.
Maybe we should check with Jeff on the general approach to be used but in the meantime since all strings have a dot a the end I think it's okay to use one here also.

Looking at just the string it may not be clear what "Shortcuts" refer to. Can you update the string comment to also offer more details about the "Shortcuts" feature?

nit: There is no need for the `%1` positional argument if there is only one argument this string takes. `%s` should be enough.

Mugurell (approved):
LGTM!

Issue #6641, opened 2022-03-11
[L10N RTL] Beta build version is not displayed fully on RTL languages


lobontiumira:
## Steps to reproduce
1. Set the device to any RTL language.
2. In Focus Beta, go to Settings - Mozilla - About, and observe the build version.

### Expected behavior
The Beta build version is displayed fully.

### Actual behavior
The Beta build version is not displayed fully.

### Device information
* Android device: Samsung Galaxy Note 8 (Android 9)
* Focus version: Beta 99.0.0-beta.1

![Screenshot_1](https://user-images.githubusercontent.com/40694158/157879443-d1348f23-dabe-4fd0-a8f3-3581554b8968.jpg)

lobontiumira:
This issue is still reproducible on Beta 100.0.0-beta.3.

lobontiumira:
Reproducible on Beta 101.0.0-beta.1 with Xiaomi Mi8 (Android 10).

lobontiumira:
Reproducible on Beta 102.0.0-beta.2 with Samsung Galaxy Tab A6 (Android 5.1.1).

lobontiumira:
Reproducible on Beta 103.0.0-beta.1 with Lenovo tablet (Android 10).

Mugurell:
The versions header has been modified in the meantime and the offending dash disappeared.
The l10n team helped confirm in https://github.com/mozilla-mobile/focus-android/pull/7358#issuecomment-1190702461 that the current state is fine.

<image src="https://user-images.githubusercontent.com/11428869/180142504-30304d9f-cbac-4396-ba4b-c17b0f5a5ea0.png" width="33%" />

@lobontiumira Do you agree on closing this ?

ItielMaN:
Hmm, how does this currently look like in beta? If the "beta" part still appears on the left, the whole thing should be LTR (while retaining the same text alignment.

lobontiumira:
@ItielMaN this is how it looks now in Beta 103.0.0-beta.5:

![Untitled](https://user-images.githubusercontent.com/40694158/180194852-0ed031bf-92fa-4150-8782-385eaefeb55a.png)

ItielMaN:
If this would look the same without the dash (as @Mugurell mentioned), i.e. if the number is on the right and the beta (which I assume is not localizable) appears on the left, it should be set to LTR whilw being aligned to the right. Injecting LRM before the version number would also work.

Mugurell:
[...]

PR #7358, opened 2022-07-13
Fix About screen header text alignment

ionutbedregeanu:
For #6641
In Unicode many characters have a specific directionality, which lets the system know
it has to be written, say LTR.
The paragraph usually uses the direction of its first character.
#### LTR

No unit tests required
### Pull Request checklist

3. click the `Artifacts` row,

### DIFF
--- a/app/src/main/java/org/mozilla/focus/fragment/about/AboutFragment.kt
+++ b/app/src/main/java/org/mozilla/focus/fragment/about/AboutFragment.kt
@@ -111,9 +111,11 @@ class AboutFragment : BaseSettingsLikeFragment() {
         val packageInfo = requireContext().packageManager.getPackageInfo(requireContext().packageName, 0)
         val versionCode = PackageInfoCompat.getLongVersionCode(packageInfo).toString()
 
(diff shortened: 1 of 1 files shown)


Mugurell:
I would've expected somewhat of a mirrored text and based on the screenshots this is not achieved.
As in the LTR text
`104.0 (Build #361941655  🦎 ...`
to be written in RTL as
`... 🦎 361941655# Build) 104.0`

From the time the ticket was filed the versions for AC and AS were added to this about screen and need to be updated to support RTL also. As in I expect they should say something like
`93.5.0 :AS`
(Not sure if in this case because the formatting for the version follows a certain schema: `major.minor.patch version` the digits should also have reverse position to respect the schema to rather say `0.5.93 :AS`)

@Delphine Do you know if someone from l10n can help us with what is the expected result in RTL?
@ionutbedregeanu Not sure if you've looked into this already but one other direction could be using using [BidiFormatter](https://developer.android.com/reference/android/text/BidiFormatter).

Mugurell:
@ItielMaN Thank you for the explanation. It really helps!

With Ionut in PTO I'll close this PR based on the above.

PR #7395, opened 2022-07-20
For #7388 - Add GitHub actions to tag milestones, link issues and reopen issues for QA

### LINKED ISSUE #7388: Add GitHub actions to tag milestones, link issues and reopen issues for QA

gabrielluong:
Fixes #7388

### Pull Request checklist

3. click the `Artifacts` row,

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -6,9 +6,18 @@
 - [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
 - [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.
 
+### QA
+<!-- Before submitting the PR, please address each item -->
+- [x] **QA Needed**
+
 ### To download an APK when reviewing a PR:
 1. click on `Show All Checks`,
 2. click `Details` next to `build-focus-debug` or `build-klar-debug` for changes targeting Klar,
 2. click `View task in Taskcluster`,
 3. click the `Artifacts` row,
 4. click to download any of the apks listed here which use an appropriate name for each CPU architecture.
+
+### GitHub Automation
+<!-- Do not add anything below this line -->
+
+Used by GitHub Actions.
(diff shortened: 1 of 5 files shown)


.github/PULL_REQUEST_TEMPLATE
```diff
@@ -6,9 +6,18 @@
 - [ ] **Screenshots**: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
 - [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.
 
+### QA
+<!-- Before submitting the PR, please address each item -->
+- [ ] **QA Needed**
```
mcarare, line 11: Should this be checked by default and only unchecked when necessary?
I think the number of issues that need QA is significantly higher than the ones that do not need QA.

And as for the result of forgetting to change the state of this, IMO it would be preferable to tag issues for QA when it is not needed, rather than not tagging ones that do.

sv-ohorvath: I agree with mcarare here: as a QA, I can always remove the label and close an issue if I find nothing that needs testing. On the other hand, I wouldn't want to find unknown changes too late in a cycle to have the time to test. If you agree, can this be applied to Fenix too?

gabrielluong: @sv-ohorvath I am totally okay with going with QA's preference. Before I proceed, I wanted to double check if this decision was okay with the rest of the QA team. The implication could mean a larger number of issues are tagged for QA needed (not necessarily a bad thing), but potentially issues that shouldn't need QA as well. Overall, I would just be happy to see that all issues are being QA validated.

sv-ohorvath: Yes, I have the same concern. Ideally, people would un-check this option if QA is not needed if we can spread that information as clearly as possible. If what Mihai said is true: "the number of issues that need QA is significantly higher than the ones that do not need QA", I hope that won't be a big issue.

gabrielluong: Ok, let me know when an executive decision is made. I won't necessarily block on this decision in the meantime since it's an easy change if we need more time to get the OK from the rest of the QA team.

sv-ohorvath: @gabrielluong I talked to the team and everyone is okay with having to double-check a few slipped issues that don't need QA (hopefully a minimal number) rather than missing one that does (a higher risk). So, as far as we are concerned, I'd say go ahead with checking the item by default if there are not any other concerns. Thank you for doing this, btw!

mcarare (approved):

PR #7468, opened 2022-08-08
Fixes #7467: Update ktlint to 0.46.1

### LINKED ISSUE #7467: Update code quality tools
This should include updating ktlint and detekt, and also fixing issues and removing entries from baselines (if any).

mcarare:
### Pull Request checklist

### QA

- [x] **QA Needed**

3. click the `Artifacts` row,

### GitHub Automation
Fixes #7467

### DIFF
--- a/app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
+++ b/app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
@@ -54,12 +54,12 @@ class BrowserToolbarIntegration(
     private val eraseActionListener: () -> Unit,
     private val tabCounterListener: () -> Unit,
     private val customTabId: String? = null,
-    inTesting: Boolean = false
+    inTesting: Boolean = false,
 ) : LifecycleAwareFeature {
     private val presenter = ToolbarPresenter(
         toolbar,
         store,
-        customTabId
+        customTabId,
     )
 
     @VisibleForTesting
@@ -76,7 +76,7 @@ class BrowserToolbarIntegration(
     private val eraseAction = BrowserToolbar.Button(
         imageDrawable = AppCompatResources.getDrawable(
             toolbar.context,
-            R.drawable.mozac_ic_delete
+            R.drawable.mozac_ic_delete,
         )!!,
         contentDescription = toolbar.context.getString(R.string.content_description_erase),
         iconTintColorResource = R.color.primaryText,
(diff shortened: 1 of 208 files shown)


app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
```diff
@@ -165,7 +165,7 @@ class BrowserToolbarIntegration(
                 menuItemIndex = menu.menuBuilder.items.size - 1,
                 closeListener = { fragment.closeCustomTab() },
                 updateToolbarBackground = true,
-                forceActionButtonTinting = false,
```
jonalmeida, line 168: I think the team wanted to keep trailing commas around. We probably want to enable these [two rules](https://pinterest.github.io/ktlint/rules/standard/#trailing-comma-on-call-site)?

jonalmeida (approved):
Aside from the comment below, I'm fine if we want to land this sooner than later. Is there any reason to wait?

app/src/main/java/org/mozilla/focus/browser/integration/BrowserToolbarIntegration.kt
mcarare: Unfortunately, running ktlint auto format would mean either removing all existing trailing commas or adding trailing commas everywhere. The alternative would be manually modifying a serious amount of files.

I searched for a style guide or a decision in this issue but did not find one. IMO we should add trailing commas everywhere :)

jonalmeida: We could use a baseline file for ktlint but using ktlintFormat to update them all once isn't so bad either.

mcarare: I added a baseline file, I think it is a better option than having a commit with 80 + modified files.
Also tried for AC, there would have been 200+ modified files. IMO we should stick to the same approach in all 3 repos.
And maybe improve the existing issues only when that code is touched. Thoughts?

jonalmeida: > And maybe improve the existing issues only when that code is touched. Thoughts?

Yeah, I'm fine with that. 

In general, I'm usually more flexible with Focus because it's a smaller project (code base) so we can try out experiments like this with ease.

Issue #7732, opened 2022-09-25
(Crash) java.lang.NullPointerException (org.mozilla.geckoview.GeckoResult$UncaughtException: at org.mozilla.geckoview.GeckoResult.dispatchLocked(GeckoResult.java))


cadeyrn:
### Steps to reproduce

Unfortunately no reliable steps to reproduce. :( But it suddenly happened two times after bringing back Firefox Focus from the background to the foreground. I am not sure if Firefox Focus already crashed before pressing the "Unlock" button or after that. It doesn't always reproduce so I can't check that.

### Expected behavior

No crash. :) 

### Actual behavior

Crash report: https://crash-stats.mozilla.org/report/index/4b611361-8354-4bae-8f77-57f210220925

### Device information

* Android device: OnePlus 10 Pro, Android 12
* Focus version: Firefox Klar 105.1.0

mcarare:
STR:
0. Have at least one tab opened and fingerprint unlock enabled.
1. Open Exceptions list.
2. Turn the screen off / background the app.
3. Turn the screen on / go back to the app.
4. Crash.

lobontiumira:
Verified as fixed on the latest Focus Nightly build 107.0a1 from 10/7 with Google Pixel (Android 10), and Oppo Reno 6 (Android 12).

PR #7758, opened 2022-09-27
For #7732: Run refresh block only if fragment is still attached.

mcarare:
NPE is thrown because binding is null when the refresh block is executed if the lock fragment has to be displayed and thus `onDestroy` for `ExceptionsListFragment` has been called. 

### Pull Request checklist

### QA

- [x] **QA Needed**

3. click the `Artifacts` row,

### GitHub Automation

Used by GitHub Actions.


app/src/main/java/org/mozilla/focus/exceptions/ExceptionsListFragment.kt
```diff
@@ -158,12 +158,15 @@ open class ExceptionsListFragment : BaseSettingsLikeFragment(), CoroutineScope {
         showToolbar(getString(R.string.preference_exceptions))
 
         (binding.exceptionList.adapter as DomainListAdapter).refresh(requireActivity()) {
```
iorgamgabriel, line 160: You can check with "isAdded" if the fragment is attached.

mcarare: We already use this pattern in numerous places in Fenix, I think we should stick to it.

iorgamgabriel (approved):

PR #7796, opened 2022-10-03
Remove extraneous coroutines_version

rvandermeulen:
We use the one set by Dependencies.kt, so remove this one to avoid confusion.


jonalmeida (approved):
✨

PR #7803, opened 2022-10-05
Update some Gradle dependencies

rvandermeulen:
Similar to https://github.com/mozilla-mobile/android-components/pull/12888. Probably nothing super critical in here, but might as well.


mcarare (approved):
I went to all the release notes for the upgrades and LGTM.

I had a question about Sentry Self-hosted Compatibility, but the release notes state "If you are using [sentry.io](https://sentry.io/), no action is needed.", so I guess no additional changes are required by the upgrade.

PR #7841, opened 2022-10-10
Bug 1802577 - Target Android 13 (SDK 33)

### LINKED ISSUE #7840: Target Android 13 (SDK 33)
Latest SDK apps can target is 33: https://developer.android.com/about/versions/13

See also https://github.com/mozilla-mobile/focus-android/issues/7400.

mcarare:
### Pull Request checklist

### QA

- [x] **QA Needed**

3. click the `Artifacts` row,

### GitHub Automation
Fixes #7840


app/src/main/AndroidManifest.xml
```diff
@@ -35,6 +35,12 @@
     <!-- Needed for Google Play policy https://support.google.com/googleplay/android-developer/answer/6048248 -->
     <uses-permission android:name="com.google.android.gms.permission.AD_ID"/>
 
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
+
+    <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
+    <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
+    <uses-permission android:name="android.permission.READ_MEDIA_AUDIO" />
```
jonalmeida, line 42: What are these permissions needed for? Downloads? 

Accordingly, I wonder if we can do better to ask for some runtime permission when it's needed. [This image](https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions) on the Android 13 behaviour changes suggests that we will show a permission dialog to accept these permissions. Did you see that in your testing?

mcarare: Currently, we can download files with or without these permissions, without requesting any runtime permissions. 

What I do not get from the documentation versus the testing is that no permission is asked when opening downloaded files (via the open button in the download snack bar) but the files are opened in other apps. The same is when declaring these permissions or not.

This is true for media files, but the permission request shows up when I tried opening a downloaded docx file. 

From what I understand from the documentation, and testing, we might not need this at all in Focus. ( is there any scenario where we want to open files created by other apps?)

Mugurell: A scenario for opening files owned by other apps is when users try to upload files to various websites.

The scoped storage framework supports the following two scenarios:
- for accessing other app's audio/image/video files we need to use the MediaStore and we need the above permissions - https://developer.android.com/training/data-storage/shared/media#access-other-apps-files
- for accessing documents and other file types we should use the system's file picker in which case we don't need any permission - https://developer.android.com/training/data-storage/shared/documents-files

mcarare: Thank you @Mugurell. That totally slipped my mind.
I remember from testing that I could not trigger file selection with or without the permissions in the manifest when targeting 13 and I left the matter for more investigations after we also compile with 13 in Fenix.

ohall-m: > LE: The above was observed only on an Android 13 device. Additional changes might be needed here: https://github.com/mozilla-mobile/firefox-android/blob/a7eb0fe9d057982cb5eb1ad670e8195afe7fba1a/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/MimeType.kt#L72

Just chiming in with a test case - I think I reproduced what you were mentioning about `MimeType.kt` when testing selecting a file using the [MDN file input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file) page on a Pixel 4a Android 13 real device. File selection options didn't open and also had [line 51](https://github.com/mozilla-mobile/firefox-android/blob/a7eb0fe9d057982cb5eb1ad670e8195afe7fba1a/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/MimeType.kt#L51) in the stack trace. (Saw slightly different behavior in an emulator.)

cpeterso:
The Focus bug for targeting Android 13 has moved from #7840 to Bugzilla https://bugzilla.mozilla.org/show_bug.cgi?id=1802577.

gabrielluong:
Hey @mcarare, we're tidying the opened PRs in this repo since this project moved to a new repository. It is now developed and maintained at: https://github.com/mozilla-mobile/firefox-android. Please reopen a new PR there if the issue is still relevant. Thank you!

PR #7875, opened 2022-10-18
Use a consistent mockwebserver version

rvandermeulen:
I'm cleaning up some dependencies and noticed that we're currently specifying two different versions of mockwebserver. Hopefully this is green because it'll make my life easier later if it is.


rvandermeulen:
I could roll this change into the bigger cleanup patch I'm doing, but it might be nicer to land it separately first in the interest of the "no dependency changes" guarantee to that work.

rvandermeulen:
Turns out the declaration in `service-telemetry/build.gradle` isn't actually used at all, so we can just drop it outright. And I'll do that in the other PR.

PR #7883, opened 2022-10-20
Revert espresso-contrib back to 3.1.0-alpha4

rvandermeulen:
#7876 moved a lot of things around but wasn't supposed to change any actual dependency versions, with one exception: it changed `espresso-contrib` from 3.1.0-alpha4 to 3.4.0 like the other espresso libs already were. While CI on the pull request was green, it turns out that the non-debug UI tests (which don't run on PRs) are now permafailing:
```
RunTests
java.nio.BufferUnderflowException
	at java.base/java.nio.Buffer.nextGetIndex(Buffer.java:651)
	at java.base/java.nio.HeapByteBuffer.getInt(HeapByteBuffer.java:402)
	at com.linkedin.dex.spec.AnnotationOffItem.<init>(AnnotationOffItem.kt:13)
	at com.linkedin.dex.spec.AnnotationSetItem$Companion.create(AnnotationSetItem.kt:18)
	at com.linkedin.dex.parser.AnnotationUtilsKt.checkIfAnnotationIsInherited(AnnotationUtils.kt:89)
	at com.linkedin.dex.parser.AnnotationUtilsKt.getTestAnnotation(AnnotationUtils.kt:77)
	at com.linkedin.dex.parser.AnnotationUtilsKt.getClassAnnotationValues(AnnotationUtils.kt:46)
	at ftl.run.platform.android.CreateAndroidTestContextKt.isParametrizedClass(CreateAndroidTestContext.kt:210)
	at ftl.run.platform.android.CreateAndroidTestContextKt.getParametrizedClasses(CreateAndroidTestContext.kt:200)
	at ftl.run.platform.android.CreateAndroidTestContextKt.getFlankTestMethods(CreateAndroidTestContext.kt:153)
	at ftl.run.platform.android.CreateAndroidTestContextKt.calculateShardsNormally(CreateAndroidTestContext.kt:86)
	at ftl.run.platform.android.CreateAndroidTestContextKt.calculateShards(CreateAndroidTestContext.kt:112)
	at ftl.run.platform.android.CreateAndroidTestContextKt.access$calculateShards(CreateAndroidTestContext.kt:1)
	at ftl.run.platform.android.CreateAndroidTestContextKt$setupShards$2$1$1.invokeSuspend(CreateAndroidTestContext.kt:49)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
[...]

PR #7884, opened 2022-10-20
Update AndroidX testing libraries to the latest releases

rvandermeulen:
Was reverted back to 3.1.0-alpha4 in #7883. This is aiming to get it back to 3.4.0 again.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -333,11 +333,13 @@ dependencies {
     androidTestImplementation Dependencies.fastlane
     implementation Dependencies.falcon // Required by fastlane
 
-    androidTestImplementation 'androidx.test.espresso:espresso-contrib:3.1.0-alpha4', {
-        exclude group: 'com.android.support', module: 'support-annotations'
-        exclude group: 'com.android.support', module: 'support-v4'
-        exclude group: 'com.android.support', module: 'design'
-        exclude group: 'com.android.support', module: 'recyclerview-v7'
+    androidTestImplementation Dependencies.espresso_contrib, {
+        exclude module: 'appcompat-v7'
+        exclude module: 'support-v4'
+        exclude module: 'support-annotations'
+        exclude module: 'recyclerview-v7'
+        exclude module: 'design'
+        exclude module: 'espresso-core'
     }
     testImplementation Dependencies.androidx_test_core
     testImplementation Dependencies.androidx_test_runner
--- a/buildSrc/src/main/java/Dependencies.kt
+++ b/buildSrc/src/main/java/Dependencies.kt
@@ -47,10 +47,10 @@ object Versions {
     }
 
     object Testing {
-        const val androidx_core = "1.4.0"
-        const val androidx_espresso = "3.4.0"
-        const val androidx_ext_junit = "1.1.3"
-        const val androidx_orchestrator = "1.4.1"
+        const val androidx_core = "1.5.0"
+        const val androidx_espresso = "3.5.0"
+        const val androidx_ext_junit = "1.1.4"
+        const val androidx_orchestrator = "1.4.2"
         const val androidx_uiautomator = "2.2.0"
         const val falcon = "2.2.0"
         const val fastlane = "2.1.1"


rvandermeulen:
OK, espresso-contrib 3.5.0 apparently fixes whatever issue is at play here. From what I can tell, this is a crash in the dex-test-parsing library (coming in from Flank I believe?). There's some upstream bugs that look possibly related (one specifically called out fixing a crash in Espresso 3.4.0), but it also doesn't appear that upstream Flank is staying up to date with that dependency. Anyway, if the current setup isn't actively hurting anything, we can probably just live with this until Espresso 3.5.0 ships.

rvandermeulen:
Updating all of espresso to 3.5.0-beta01 is consistently hitting debug UI test failures in CustomTabTest
```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.mozilla.focus.activity.robots.BrowserRobot.verifyPageURL(BrowserRobot.kt:71)
	at org.mozilla.focus.activity.CustomTabTest$openCustomTabInFocusTest$5.invoke(CustomTabTest.kt:119)
	at org.mozilla.focus.activity.CustomTabTest$openCustomTabInFocusTest$5.invoke(CustomTabTest.kt:118)
	at org.mozilla.focus.activity.robots.CustomTabRobot$Transition.clickOpenInFocusButton(CustomTabRobot.kt:101)
	at org.mozilla.focus.activity.CustomTabTest.openCustomTabInFocusTest(CustomTabTest.kt:118)
    
```

rvandermeulen:
@AaronMT Trying to update these libraries, I'm seeing a consistent CustomTabTest failure. Given that it's debug-only, is that pointing to a timing issue maybe?

AaronMT:
Don't think so as there's a visible behavior difference watching the videos and how custom tabs intents are working here when they return to the browser, as it's supposed to load the custom tab : [Failure](https://storage.cloud.google.com/focus_android_test_artifacts/2022-10-29_01-19-00.579125_QGKv/matrix_0/Pixel2.arm-30-en_US-portrait-shard_2/video.mp4?authuser=1) (browser home-screen, empty URL bar focused) / [Pass](https://storage.cloud.google.com/focus_android_test_artifacts/2022-10-29_02-09-24.390280_rUhA/matrix_0/Pixel2.arm-30-en_US-portrait-shard_2/video.mp4?authuser=1). 

possibly some newer library differences in the intent building here [createCustomTabIntent](https://github.com/mozilla-mobile/focus-android/blob/main/app/src/androidTest/java/org/mozilla/focus/helpers/TestHelper.kt#L275) although I'm not familiar with this code, and it's also marked deprecated suppressed (maybe @mcarare has an idea?)

mcarare:
It seems that the page crashes (or at least the browser tab is not opened) when pressing the "open in focus" button.
The deprecation annotation is related to the way we show the share button and the way we set the toolbar color. I will add a patch to update that to the new API but is not related to the test failure. 
The recent changes to the file, related to retrieving the package name also do not influence the test results.

Will look more into what is happening here.

mcarare:
The failure occurs when upgrading `androidx.test:core` from 1.5.0-alpha01 to 1.5.0-alpha02.
In the release notes, there is no concrete code change, but it mentions:

> API changes
> 
> Modify ActivityScenario#launch to remove use of Bootstrap Activity API. This change should improve performance and stability of ActivityScenario#launch. ActivityScenario#getResult will now enforce that it can only be used with ActivityScenario#launchActivityForResult

sv-ohorvath:
Thanks, @mcarare! I agree with your solution. For my part, feel free to remove the first steps before launching the custom tab activity, to avoid any more trouble.

> It seems like launching the new activity (when pressing "open in focus" button) clashes with the initial `launchActivity` with null params, used to open the initial browser page.
> 
> I suggest 2 paths to fix the tests :
> 
> * investigate the changes that make the interaction problematic
> * eliminate the overhead and keep the code that actually checks just the "open in focus" feature:
> 
> ```
>     fun openCustomTabInFocusTest() {
>         val customTabPage = getGenericTabAsset(webServer, 1)
> 
>         launchActivity<IntentReceiverActivity>(createCustomTabIntent(customTabPage.url))
>         customTab {
>             progressBar.waitUntilGone(waitingTime)
>             verifyPageURL(customTabPage.url)
>             openCustomTabMenu()
>         }.clickOpenInFocusButton {
>             verifyPageURL(customTabPage.url)
>         }
>     }
> ```
> 
> cc: @sv-ohorvath ,@AndiAJ

rvandermeulen:
@mcarare Looks good now, thanks! I'll keep an eye out for the final releases of those AndroidX test libraries and update the PR when they're ready.

rvandermeulen:
Opt and Debug UI tests are green with the final releases. I think this is ready to go!

jonalmeida (approved):
Rubber stamping to get this landed. Thanks all!

PR #7895, opened 2022-10-21
Update linters

rvandermeulen:
Update ktlint to 0.47.1 and detekt to 1.21.0


rvandermeulen:
The `UndocumentedPublicClass` and `UndocumentedPublicFunction` checks caused hundreds of errors that seem pretty low-value. I guess we could just use a baseline for them otherwise, but I'd question how much we care about these in general.

I disabled `UtilityClassWithPublicConstructor` for now mainly just to get a green run so I could see what the rest of CI looks like afterwards. Not sure what we want to do about these - change/suppress/otherwise?
```
UtilityClassWithPublicConstructor - [The class AppReviewUtils only contains utility functions. Consider defining it as an object at app/src/main/java/org/mozilla/focus/appreview/AppReviewUtils.kt:23:1
UtilityClassWithPublicConstructor - [The class IconGenerator only contains utility functions. Consider defining it as an object at app/src/main/java/org/mozilla/focus/shortcut/IconGenerator.kt:20:1
```

app/src/main/java/org/mozilla/focus/exceptions/ExceptionsListFragment.kt
```diff
@@ -73,7 +73,7 @@ open class ExceptionsListFragment : BaseSettingsLikeFragment(), CoroutineScope {
                 return true
             }
 
-            override fun onSwiped(viewHolder: RecyclerView.ViewHolder, direction: Int) {}
```
mcarare, line 76: This might actually need adding a dummy implementation ( maybe a single-line comment) because non-abstract methods need a body.

rvandermeulen: Thanks for the heads-up! I'm planning to come back to this PR after https://github.com/mozilla-mobile/reference-browser/pull/2056 lands.

gabrielluong:
Hey @rvandermeulen , I am interested in our ktlint and detekt process so I can help take over your work to completion.

rvandermeulen:
Ah, I'd been sitting on this one until the [PR in r-b](https://github.com/mozilla-mobile/reference-browser/pull/2056) landed as I wanted to follow the same process for this one (and eventually Fenix/AC as well). If you want to take it, feel free though!

rvandermeulen:
I also hadn't decided whether I wanted to handle the detekt 1.22 update in that PR still vs. getting it in a follow-up (though I was sorta leaning towards the latter)

gabrielluong:
@rvandermeulen I upgraded our detekt the last time and manually compared the base configuration detekt.yml with our detekt.yml to align the changes. If it's okay with you, I can at least take over the detekt portion.

rvandermeulen:
Go for it. Probably best to start a new PR at this point.

PR #7902, opened 2022-10-24
Update to Android-Components 108.0.20221023143208.

badboy:
Supersedes #7899 

---

The underlying exception is:

```
java.lang.IllegalStateException: WorkManager is not initialized properly.  You have explicitly disabled WorkManagerInitializer in your manifest, have not manually called WorkManager#initialize at this point, and your Application does not implement Configuration.Provider.
```

This didn't show in previous versions because it happens in some side-thread I guess (or we caught and dropped it, I'm not 100% sure).
Recent UniFFI changes surface that as a Rust panic (I'll work on a followup to make this easier to debug).

I am not sure at all why it worked before in practice.
For tests it probably didn't matter, because it failed in tests that don't test Glean.

This adds a new dependency, I'm not sure if that needs additional approval from Focus folks?
It fixes the test run for me locally.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -619,6 +619,11 @@ if (gradle.hasProperty('localProperties.autoPublish.application-services.dir')) 
     apply from: "../${appServicesSrcDir}/build-scripts/substitute-local-appservices.gradle"
 }
 
+if (gradle.hasProperty('localProperties.autoPublish.glean.dir')) {
+    ext.gleanSrcDir = gradle."localProperties.autoPublish.glean.dir"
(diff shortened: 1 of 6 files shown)


badboy:
So a release build sends pings just fine, so WorkManager is running there without issues.

badboy:
```
[task 2022-10-24T13:47:21.266Z] /builds/worker/checkouts/vcs/app: Error: Remove androidx.work.WorkManagerInitializer from your AndroidManifest.xml when using on-demand initialization. [RemoveWorkManagerInitializer from work-runtime-2.7.1]
```

uhm ... I don't know anymore what's going on

rocketsroger (approved):
🚢

PR #7919, opened 2022-10-25
Don't use password inputType on AddToHomeScreenSiteTitle

rocka:
Setting `inputType` of an EditText to `Password` or its variants (`VisiblePassword`) would completely disable the ablility to switch input method (language) other than English in many IMEs, including Gboard. So it's not possible to type CJK characters or emoji into the EditText, even though you can type them somewhere else then copy/paste them.

|current state|expected state|
|:-|:-|
|Screenshot_20221025-212719|Screenshot_20221025-213823|
|can't switch to other language, it is locked to "English"|able to switch to other language|

"the line below the text" mentioned in comments is actually [composing text](https://developer.android.com/reference/android/view/inputmethod/InputConnection#setComposingText(java.lang.CharSequence,%20int)), which means "unfinished" inputing state and IME may provide auto correction to replace the composing text span. If you really want to get rid of the auto correction behavior of IME, [InputType#TYPE_TEXT_FLAG_NO_SUGGESTIONS](https://developer.android.com/reference/android/text/InputType#TYPE_TEXT_FLAG_NO_SUGGESTIONS) should be used instead. Even though some IMEs won't follow that flag, it's their choice, and I consider which would be better than not allowing user to switch to other language at all.

### Pull Request checklist

### QA

- [x] **QA Needed**

3. click the `Artifacts` row,

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -75,8 +75,7 @@
 
     <style name="AddToHomeScreenSiteTitle" parent="Widget.AppCompat.EditText">
         <item name="android:textSize">20sp</item>
-        <!-- Setting textVisiblePassword as inputType removes the line below the text-->
-        <item name="android:inputType">text|textVisiblePassword</item>
+        <item name="android:inputType">text|textNoSuggestions</item>
         <item name="android:textColor">@color/primaryText</item>
         <item name="android:textCursorDrawable">@null</item>
     </style>


mcarare:
@rocka Thank you for taking the time to submit the patch!

However, this patch regresses the issue `textVisiblePassword` tried to fix, adding the text underline for both Gboard and SwiftKeyKeyboard.  

Also, you can already select another language:
-  with Gboard, by long-pressing the space key (the one with the selected language name).
- with SwiftKey Keyboard by swiping left or right on the space key

Given the reasons above, I am closing this PR.

rocka:
it's not the case, setting inputType to `textVisiblePassword` completely takes away the ability to switch languages, no matter how you change it.

https://user-images.githubusercontent.com/13914967/198232450-e8613ac1-fbbb-4e7e-8c50-e0458e714b85.mp4

mcarare:
@rocka Did you try to long press the space bar? I do not see that in the video you posted.

rocka:
@mcarare I've already tried. Could you please try it yourself? I tried all approaches I can think of, only to find that the EditText is `VisiblePassword` and does not allow switching language.

mcarare:
@rocka of course, I have tried it, on both Gboard and SwiftKey. Otherwise, I wouldn't have mentioned it.

rocka:
https://user-images.githubusercontent.com/13914967/198238294-e3a21a7f-c509-448b-a413-d8b5c4071c14.mp4

mcarare:
https://user-images.githubusercontent.com/48995920/198240137-13e9b2b8-7b0b-4cec-ae6c-ae5eec7723b8.mp4

The Google Keyboard version on the device is 12.2.05.469624536.

rocka:
Also I'm curious why you don't like the "underline" of composing text. It's a state of unfinished imputing text, and IME can provide spell check or auto correction to replace it's content. And text field in web pages can have "underline" in them too. If you really don't like spell check, it's better to disable it in input method, rather than disable it for all users.

WhiredPlanck:
@mcarare I think you may need to switch to a Chinese, Korean or Japanese or more language keyboard, this issue may just affect specified languages.

rocka:
you should try some more complex language which can have more character than the keyboard can display, such as Chinese or Japanese.

lilydjwg:
I haven't tried Focus but I can see that, keys on the keyboard work literally, and can't be used to compose Chinese or Japanese characters as they need to be composed rather than pressed. The underline indicates the composing code to reach the desired characters (it's also used to spellcheck, but I suppose spellcheck can be disabled separately?).

mcarare:
I am putting this on my to-do list to try another solution that allows special characters input while preserving the design in the specs.

PR #8026, opened 2022-11-14
Update various AndroidX libraries to the latest releases

rvandermeulen:
This updates most of our AndroidX dependencies to their latest releases (and a bonus Material Components update while I'm in there). The only one I've omitted for now is Preference as there's some actual changes needed in our code to account for the 1.2.0 release, so I'll handle that in its own PR.

Changelogs:
* [AppCompat](https://developer.android.com/jetpack/androidx/releases/appcompat#1.5.1)
* [Browser](https://developer.android.com/jetpack/androidx/releases/browser#1.4.0)
* [Compose Foundation](https://developer.android.com/jetpack/androidx/releases/compose-foundation#1.3.1) (looks like that's the only Compose library that actually changed from 1.3.0 to 1.3.1)
* [Fragment](https://developer.android.com/jetpack/androidx/releases/fragment#1.5.4)
* [Material Components](https://github.com/material-components/material-components-android/releases) - There's...been a lot of since our last update.
* [RecyclerView](https://developer.android.com/jetpack/androidx/releases/recyclerview#recyclerview-1.2.1)
* [Transition](https://developer.android.com/jetpack/androidx/releases/transition#1.4.1)


jonalmeida:
cc: @mcarare @iorgamgabriel for awareness or if you have strong opinions here.

mcarare:
I've been reluctant in the past to update the Material library because it might change some default values in the widgets we use. Maybe we should do it separately?

As a side note, maybe we should do upgrades that are NOT dot releases in their own PR.

rvandermeulen:
Per discussion with @mcarare on Slack, I'll split out the Material Components and Appcompat updates to their own PRs and leave the others in this one.

mcarare (approved):
Looked into the release notes ( TY for adding them) and LGTM!

PR #8095, opened 2022-12-01
Bug 1803448 - Generate a Debug Version Name

gabrielluong:
@ionutbedregeanu @iorgamgabriel We won't have to manually set the version after this change.

### Pull Request checklist

### QA

- [ ] **QA Needed**

3. click the `Artifacts` row,

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -26,7 +26,8 @@ android {
         targetSdkVersion 32
         versionCode 11 // This versionCode is "frozen" for local builds. For "release" builds we
                        // override this with a generated versionCode at build time.
-        versionName "106.0"
+        // The versionName is dynamically overridden for all the build variants at build time.
+        versionName Config.generateDebugVersionName()
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         testInstrumentationRunnerArguments clearPackageData: 'true'
 
--- a/buildSrc/src/main/java/Config.kt
+++ b/buildSrc/src/main/java/Config.kt
@@ -3,8 +3,22 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import org.gradle.api.Project
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
 
 object Config {
+
+    @JvmStatic
+    private fun generateDebugVersionName(): String {
+        val today = Date()
+        // Append the year (2 digits) and week in year (2 digits). This will make it easier to distinguish versions and
+        // identify ancient versions when debugging issues. However this will still keep the same version number during
+        // the week so that we do not end up with a lot of versions in tools like Sentry. As an extra this matches the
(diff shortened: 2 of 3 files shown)


csadilek (approved):
👍 

It seems better in general to give our debug builds a different version e.g., for Sentry logs/crashes and telemetry. This aligns it with Fenix.

For Nightly|Beta|Release we already override this property anyway, and it's kind of strange to allow it to be out-of-date :). It's currently 106.

cc @jonalmeida as well, in case we're missing anything.

buildSrc/src/main/java/Config.kt
```diff
@@ -3,8 +3,22 @@
...
 object Config {
+
+    @JvmStatic
+    private fun generateDebugVersionName(): String {
+        val today = Date()
+        // Append the year (2 digits) and week in year (2 digits). This will make it easier to distinguish versions and
+        // identify ancient versions when debugging issues. However this will still keep the same version number during
+        // the week so that we do not end up with a lot of versions in tools like Sentry. As an extra this matches the
+        // sections we use in the changelog (weeks).
+        return SimpleDateFormat("1.0.yyww", Locale.US).format(today)
```
jonalmeida, line 19: From personal experience, having the yy-mm-dd (minutes/seconds are optional) for a debug build is helpful to avoid user error when you think you're running a new debug but somehow it wasn't deployed to your device; week number is hard to grok.

Same would go for Fenix.

jonalmeida, line 19: Maybe 0.0 since was a really version we released at one point?

```suggestion
        return SimpleDateFormat("0.0.yyww", Locale.US).format(today)
```

app/build.gradle
```diff
@@ -26,7 +26,7 @@ android {
         targetSdkVersion 32
         versionCode 11 // This versionCode is "frozen" for local builds. For "release" builds we
                        // override this with a generated versionCode at build time.
-        versionName "106.0"
+        versionName Config.generateDebugVersionName()
```
jonalmeida, line 29: If you can, a similar-ish comment for the `versionCode` would be good to add here even though it wasn't because of your changes - took me a few seconds to realize this is an overridden property as well (without having looked at Christian's comments above).

jonalmeida (approved):
Just left some opinions, this is fine to land as-is too.

buildSrc/src/main/java/Config.kt
gabrielluong: I filed a followup for this https://bugzilla.mozilla.org/show_bug.cgi?id=1803686

gabrielluong: Same as above. Filed as followup https://bugzilla.mozilla.org/show_bug.cgi?id=1803686

app/build.gradle
gabrielluong: Added a comment

